PR #1, opened 2015-06-09
Addon docs

akx:


suutari-ai:
Can you rewrap the rst-text at 72-79 column or so? That would make it easier to read in text form (especially in the diffs).

doc/index.rst
```diff
@@ -22,6 +22,7 @@ Developer documentation
    provides
    templates
    taxes
+   addons
```
suutari-ai: Addons and provides are somewhat connected. IMHO addons should be added before provides into that list. What do you think?

akx:
Done, @suutari-ai

suutari-ai:
Very nice doc! Thanks @akx! :D

suutari-ai:
retest this please

PR #4, opened 2015-06-10
Fixes for early initialization

akx:


shoop/core/taxing/_tax_summary.py
```diff
@@ -16,7 +16,14 @@
...
     @classmethod
-    def from_line_taxes(cls, line_taxes, untaxed=TaxlessPrice(0)):
+    def from_line_taxes(cls, line_taxes, untaxed=None):
+        """
+        :type line_taxes: Iterable[LineTax]
+        :type untaxed: TaxlessPrice
+        :rtype: TaxSummary
+        """
+        if untaxed is None:
+            untaxed = TaxlessPrice(0)
```
suutari-ai: This should probably be fixed in the Price-object initialization, which uses the settings. It was a bad idea to use settings.SHOOP_HOME_CURRENCY in the constructor. The constructor should just set the self._currency to None if one is not given and the Price class should have a property getter for currency, which will return settings.SHOOP_HOME_CURRENCY when self._currency is None and self._currency otherwise. What do you think?

akx: That works for me. If you make a PR out of that, feel free to close this one :)

suutari-ai: I created PR #10 for that, but it does not have your commit aa52149ee3397c1ef33692834a77970121c3c097. Maybe just drop the other commit from this PR?

akx:
@suutari-ai: Rebased :)

PR #7, opened 2015-06-10
Fix copyright misinformation (SHOOP-978)

suutari-ai:
There was one more vendored file that was incorrectly classified as
being ours: djangodocs.py.  Remove the incorrect copyright header from
it.  Also make ensure_license_headers.py script to ignore vendored files
in the future and fix JS header format.


akx:
Good stuff. Cheers :)

PR #9, opened 2015-06-10
README: Add "Chat" section

suutari-ai:
Promote the Gitter chat room with a badge.

### DIFF
--- a/README.md
+++ b/README.md
@@ -29,6 +29,13 @@ published under other AGPLv3-compatible licenses.  For these, please
 refer to VENDOR-LICENSES.md file in the source code tree or the licenses
 included within each package.
 
+## Chat
+
+We have a Gitter chat room for Shoop.  Come chat with us!
+[![Join chat][join-chat-img]](https://gitter.im/shoopio/shoop)
+
+[join-chat-img]: https://badges.gitter.im/Join%20Chat.svg
+
 ## Getting started with Shoop development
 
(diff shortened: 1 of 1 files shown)


akx:
Malformed commit message. Also, "chat room".

suutari-ai:
Fixed typo "root" -> "room" in the commit message. What do you mean by malformed?

akx:
@suutari-ai: Title line not in imperative case.

> Readme: Promote the Gitter chat room with a badge.

would be enough imho :)

suutari-ai:
It's a title (or headline). It should NOT end with a period, and it doesn't have to be a sentence. (Compare to newspaper headlines etc.) Most important function for the title is to easily identify bunch of commits (e.g. when reading the commit logs, shortlogs, "git gui blame" output, etc.). IMHO shortness matters there very much. That's why I prefer forms like "Foo: Bar" over "Add Bar to Foo" for the title and then use more words in the commit message body to explain it in more details.

README.md
```diff
@@ -29,6 +29,13 @@ published under other AGPLv3-compatible licenses.  For these, please
 refer to VENDOR-LICENSES.md file in the source code tree or the licenses
 included within each package.
 
+## Chat
+
+We have a chat room for Shoop at Gitter.   Welcome to chat with us!
```
akx: > We have a Gitter chat room for Shoop. Come chat with us!

suutari-ai:
Any better now?

PR #13, opened 2015-06-15
Persistent baskets

akx:


suutari-ai:
Added some comments to commit 09fb50f. (Commenting here, because if the commit gets rewritten, those comments will be lost, if the commit id is not linked from the PR directly.)

shoop/front/models/stored_basket.py
```diff
@@ -0,0 +1,42 @@
...
+
+
+class StoredBasket(models.Model):
+    # A combination of the PK and key is used to retrieve a basket for session situations.
+    key = models.CharField(max_length=32, default=generate_key)
+    owner_contact = models.ForeignKey("shoop.Contact", blank=True, null=True)
+    owner_user = models.ForeignKey(settings.AUTH_USER_MODEL, blank=True, null=True)
+
+    created_on = models.DateTimeField(auto_now_add=True, db_index=True, editable=False)
+    updated_on = models.DateTimeField(auto_now=True, db_index=True, editable=False)
```
suutari-ai: Why would we want created_on and updated_on be indexed?

shoop/front/models/stored_basket.py
```diff
@@ -0,0 +1,42 @@
...
+    persistent = models.BooleanField(db_index=True, default=False)
+    deleted = models.BooleanField(db_index=True, default=False)
+    finished = models.BooleanField(db_index=True, default=False)
+    title = models.CharField(max_length=64, blank=True)
+    data = PickledObjectField(compress=True)
+
+    # For statistics etc., as `data` is opaque:
+    taxless_total = MoneyField(default=0)
+    taxful_total = MoneyField(default=0)
```
suutari-ai: IMHO we should just store the total_price and a flag that tells if it's taxful or taxless, since toggling taxfulness of a price can be costly operation (literally, since that needs a request to tax module and e.g. Avalara charges per transaction).

shoop/front/models/stored_basket.py
```diff
@@ -0,0 +1,42 @@
...
+    product_count = models.IntegerField(default=0)
```
suutari-ai: Would it be bad idea to store -- in addition to price and product_count -- a many-to-many foreign key to the actual products in the basket? That would allow searching for baskets that have a certain product etc.

suutari-ai: That also would have a side effect of preventing deletion of those products, which IMHO could actually be a good thing.

shoop/front/models/stored_basket.py
akx: Because I foresee searching for baskets in a given time period to be a fairly common thing to do.

shoop/front/models/stored_basket.py
akx: I have the feeling that's not going to be an acceptable solution in places like our very own Finland where tax calculations are constant. Merchants will likely want (they have wanted this before) to see both taxless and taxful price statistics in pretty graphs for orders; I don't see why they might not want that for (abandoned) baskets too.

Besides, we can probably cache the taxation information, can't we?

shoop/front/basket/storage.py
```diff
@@ -0,0 +1,129 @@
...
+        The actual semantics of what finalization does are up to each backend.
+
+        :type basket: shoop.front.basket.objects.BaseBasket
+        """
+        self.delete()
+
+
+class DirectSessionBasketStorage(BasketStorage):
+    def __init__(self):
+        if settings.SESSION_SERIALIZER == "django.contrib.sessions.serializers.JSONSerializer":  # pragma: no cover
```
akx: Re @suutari-ai 's comments on JSON storage:

> Why don't we just make it JSON-serializable? I.e. convert the datetimes, Decimals, etc. to strings in OrderSource.to_dict (or whatever will be the serializing method) and back to correct types in OrderSource.from_dict (or whatever)?
> IMHO it would be a big pro, if we wouldn't have to use pickle for storing the baskets, since pickling between Python-versions and/or Django-versions might be problematic.

Yeah, I do agree, but this was the easier way fttb.

I'll see what I can do... Thing is we can't really know the schema of the basket data beforehand for customizations, so we'll have to do with some sort of [tagged JSON serialization](https://github.com/lynaghk/json-tagged-literals/#usage).

shoop/front/models/stored_basket.py
akx: I was thinking about that, yeah... Maybe it is a good idea.

shoop/front/models/stored_basket.py
suutari-ai: OK, sounds reasonable.

shoop/front/basket/storage.py
akx: @suutari-ai: the newest commit replaces pickling (in the database anyway) with tagged JSON. What do you think?

shoop/front/models/stored_basket.py
suutari-ai: If merchant has to pay a transaction fee (once or possibly several times, because the basket changes while customer adds products to it) for each abandoned basket, he probably doesn't care so much about having both taxful and taxless prices for those. For ordered baskets, i.e. orders, it's totally different of course, but that's not our concern here.

When calculating taxless from/to taxful is free and cheap in terms of I/O bandwith and CPU time, then it would be wise to cache both prices, but only then. We could add a method to TaxModule API for querying if the tax-requests should be avoided or not? (like `should_avoid_requests()` which by default returns True and in default_tax would return False)

Maybe we should add both taxless and taxful fields as you suggested, but make those fields nullable and then if TaxModule tells us to avoid requests only one of those will have non-NULL value. What do you think?

shoop/front/models/stored_basket.py
```diff
@@ -0,0 +1,41 @@
...
+ID_KEYSPACE = "0123456789abcdefghijklmnopqrstuvwxyz"
+
+
+def generate_key():
+    return "".join(random.choice(ID_KEYSPACE) for x in range(32))
+
+
+class StoredBasket(models.Model):
+    # A combination of the PK and key is used to retrieve a basket for session situations.
+    key = models.CharField(max_length=32, default=generate_key)
```
suutari-ai: Just checking: If the random source is good, even max_length=19 (> 98 bits) would be enough to to store 1000 baskets per second for 100 years (total 3.1536e12  baskets) with collision probability less than 0.1% (see https://en.wikipedia.org/wiki/Birthday_problem#Probability_table), so we should be safe then... But it is a big if.

I'm just not sure if using `random.choice` will result in good enough randomness (since it uses a PRNG) or should we use `os.urandom` or something similar. What's the purpose of basket key anyway? Preventing access to someone else's basket? I'm worried that it's possible to determine the state of the PRNG from a few basket keys and then all the following basket keys are predetermined with that state info...

suutari-ai: It seems that my concerns about predetermination of basket keys aren't so far fetched after all. Here's a quote from http://www.quadibloc.com/crypto/co4814.htm

> Despite the fact that the Mersenne Twister is an extremely good pseudo-random number generator, it is not cryptographically secure by itself for a very simple reason. It is possible to determine all future states of the generator from the state the generator has at any given time, and either 624 32-bit outputs, or 19,937 one-bit outputs are sufficient to provide that state. Using a cryptographically-secure hash function, such as SHA-1, on the output of the Mersenne Twister has been recommended as one way of obtaining a keystream useful in cryptography.

suutari-ai: Seems that Django folks have thought this already in `django.utils.crypto.get_random_string`. See https://github.com/django/django/commit/1525874238fd705ec17a066291935a9316bd3044.

akx: Using `get_random_string` now. :)

package.json
```diff
@@ -6,7 +6,8 @@
     "doc": "doc"
   },
   "scripts": {
-    "build": "python setup.py build_resources"
+    "build": "python build-all.py",
+    "dev:test": "py.test -vvv --cov shoop --cov-report html shoop_tests"
```
suutari-ai: There's no build-all.py anymore.

suutari-ai:
Cool stuff! :D

PR #14, opened 2015-06-16
Init default shop as active

Pikkupomo:
This will enable creation of `ShopProduct` with `ShopProductFormPart` in `shoop/admin/modules/products/views/edit.py:56` when shoop was initialized with management command `shoop_init`.

Add status parameter `status=ShopStatus.ENABLED` when creating default shop.


akx:
Good catch. Thanks!

PR #19, opened 2015-06-20
Fix getting started links

jaywink:
Currently getting_started links to getting_started_dev and the other way around. Github doesn't know how to render the links, these versions work according to the preview at least.


suutari-ai:
Yep, those work with Github, but don't work with the generated documentation. We're targeting to get the docs generated for readthedocs.org or pythonhosted.org so can't accept a patch that breaks the generated docs. Sorry.

PR #21, opened 2015-06-21
Add missing dependency django-registration-redux

jaywink:


akx:
This is already mentioned in the `extras` for `'shoop.front.apps.registration'` (it's not a strict requirement to use the Shoop Core, nor the default frontend either if you don't need registration).

suutari-ai:
That's actually very odd that we define only that dep as optional. We probably should (1) move all the deps to the same list or (2) move all the optional ones to extras or (3) create separate Python packages for Shoop Core, Shoop Front, Shoop Admin, etc.

akx:
@suutari-ai -- it's odd only because it's the first one to have had that sort of dependency. `shoop.api` (once it lands) will be another.

jaywink:
I think #28 handles this so closing :)

PR #22, opened 2015-06-22
Persist GET parameters in checkout phase redirect

Pikkupomo:
This is because some custom phases might need `GET` params to operate.


shoop/front/views/checkout.py
```diff
@@ -5,6 +5,9 @@
 #
 # This source code is licensed under the AGPLv3 license found in the
 # LICENSE file in the root directory of this source tree.
+from six.moves import urllib
+
+from django.core.urlresolvers import reverse
```
suutari-ai: Move the six import last (alphabetical order), please.

akx:
Looks good to me.

PR #24, opened 2015-06-23
Better admin login flow (SHOOP-801)

akx:
Features some other incidental improvements too, as usual.

### DIFF
--- a/shoop/utils/excs.py
+++ b/shoop/utils/excs.py
@@ -12,6 +12,30 @@ class Problem(Exception):
 
     message = property(lambda self: self.args[0] if self.args else None)
 
+    def __init__(self, message, title=None):
+        super(Problem, self).__init__(message)
+        self.title = title
+        self.links = []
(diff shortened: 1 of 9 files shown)


shoop/utils/excs.py
```diff
@@ -12,6 +12,19 @@ class Problem(Exception):
...
+    def __init__(self, message, title=None):
+        super(Problem, self).__init__(message)
+        self.title = title
+        self.links = []
+
+    def with_link(self, url, title):
+        # This API is designed after `Exception.with_traceback()`,
+        # so you can fluently chain this in a `raise` statement:
+        #
+        # > raise Problem("Oops").with_link("...", "...")
```
suutari-ai: Rather use a real docstring so this will end up in the API documentation too. Will need a title then though, and the code example should be formatted as `>>> raise Problem...`.

akx:
retest this please

PR #25, opened 2015-06-23
Make order reprs not die when billing address is missing.

akx:


suutari-ai:
Not really necessary for this, but maybe we should derive more/most/all of our models from ShoopModel or ShoopTranslatedModel?

suutari-ai:
I mean ShoopModel or TranslatableShoopModel, see https://github.com/shoopio/shoop/blob/master/shoop/core/models/_base.py and e.g. https://github.com/shoopio/shoop/blob/master/shoop/core/models/taxes.py .

PR #26, opened 2015-06-23
Telemetry Sender

akx:
Implements an **opt-in-by-default** telemetry sender.

The receiving side is <del>still missing</del>currently in PR review, so this doesn't do much except slow down loading of the admin dashboard for a few seconds every 72 hours.

### DIFF
--- a/shoop/admin/__init__.py
+++ b/shoop/admin/__init__.py
@@ -16,6 +16,7 @@ class ShoopAdminAppConfig(AppConfig):
     required_installed_apps = ["bootstrap3"]
     provides = {
         "admin_module": [
+            "shoop.admin.modules.system:SystemModule",
             "shoop.admin.modules.products:ProductModule",
             "shoop.admin.modules.product_types:ProductTypeModule",
             "shoop.admin.modules.media:MediaModule",
@@ -26,7 +27,7 @@ class ShoopAdminAppConfig(AppConfig):
             "shoop.admin.modules.users:UserModule",
             "shoop.admin.modules.methods:MethodModule",
             "shoop.admin.modules.attributes:AttributeModule",
-            "shoop.admin.modules.demo:DemoModule"
+            "shoop.admin.modules.demo:DemoModule",
         ]
     }
 
--- a/shoop/admin/modules/system/views/telemetry.py
+++ b/shoop/admin/modules/system/views/telemetry.py
@@ -1,0 +1,38 @@
+# -*- coding: utf-8 -*-
+# This file is part of Shoop.
+#
+# Copyright (c) 2012-2015, Shoop Ltd. All rights reserved.
+#
+# This source code is licensed under the AGPLv3 license found in the
+# LICENSE file in the root directory of this source tree.
+
+from django.http.response import HttpResponseRedirect, HttpResponse
+from django.utils.translation import ugettext_lazy as _
+from django.views.generic.base import TemplateView
+from shoop.core import telemetry
+
+
+class TelemetryView(TemplateView):
+    template_name = "shoop/admin/system/telemetry.jinja"
+
+    def get_context_data(self, **kwargs):
+        context = super(TelemetryView, self).get_context_data(**kwargs)
+        context.update({
+            "opt_in": not telemetry.is_opt_out(),
+            "is_grace": telemetry.is_in_grace_period(),
+            "last_submission_time": telemetry.get_last_submission_time(),
+            "submission_data": telemetry.get_telemetry_data(request=self.request, indent=2),
+            "title": _("Telemetry")
+        })
+        return context
+
+    def get(self, request, *args, **kwargs):
+        if "last" in request.GET:
+            return HttpResponse(telemetry.get_last_submission_data(), content_type="text/plain; charset=UTF-8")
+        return super(TelemetryView, self).get(request, *args, **kwargs)
+
+    def post(self, request, *args, **kwargs):
+        opt = request.POST.get("opt")
+        if opt:
+            telemetry.set_opt_out(opt == "out")
+        return HttpResponseRedirect(request.path)
--- a/shoop/admin/templates/shoop/admin/system/telemetry.jinja
+++ b/shoop/admin/templates/shoop/admin/system/telemetry.jinja
@@ -1,0 +1,57 @@
+{% extends "shoop/admin/base.jinja" %}
+{% block content %}
+    <div class="container-fluid">
+        <div class="row">
+            <div class="col-md-4">
+                <h2>About Shoop Telemetry</h2>
+                <p>Shoop will occasionally send telemetry data to a Shoop.io server.</p>
+                <p>
+                    The data contains an unique installation key,
+                    as well as the hostname of the installation (as sent by the visiting browser; currently
+                    <code>{{ request.get_host() }}</code>).
+                </p>
+                <p>
+                    No business data or other customer information is ever sent.
+                    Shoop.io only uses the telemetry data for statistical purposes and for insights
+                    about the software's installation base.
+                </p>
+                <h2>Last Telemetry</h2>
+                {% if is_grace and opt_in %}
+                    <div class="alert alert-info">
+                        Telemetry submission is currently suspended;
+                        data will only be sent 24 hours after installation if still opted in at that point.
+                    </div>
+                {% elif last_submission_time %}
+                    Telemetry data was last submitted at {{ last_submission_time|datetime }}
+                    ({{ last_submission_time|timesince }} ago).<br>
+                    <a href="?last">See the data that was submitted.</a>
+                {% else %}
+                    Telemetry data has not yet been submitted.
+                {% endif %}
+            </div>
+            <div class="col-md-4">
+                <h2>Opt-in/opt-out</h2>
+                <form action="{{ request.path }}" method="post">
+                    {% csrf_token %}
+                    {% if opt_in %}
+                        <div class="alert alert-success">
+                            You are currently <b>opted in</b> to send telemetry data to Shoop.io.<br>
+                            <b>Thank you</b> for your valuable contribution. &hearts;<br><br>
+                            <button type="submit" value="out" name="opt" class="btn btn-default btn-block">Opt out</button>
+                        </div>
+                    {% else %}
+                        <div class="alert alert-danger">
+                            You are currently <b>opted out</b> of Shoop telemetry.<br><br>
+                            <button type="submit" value="in" name="opt" class="btn btn-default btn-block">Opt in</button>
+                        </div>
+                    {% endif %}
+                </form>
+            </div>
+            <div class="col-md-4">
+                <h2>Telemetry Data</h2>
+                <p>Here is an example of the data that would be submitted right now.</p>
+                <pre><code>{{ submission_data }}</code></pre>
+            </div>
+        </div>
+    </div>
+{% endblock %}
--- a/shoop/admin/views/dashboard.py
+++ b/shoop/admin/views/dashboard.py
@@ -8,6 +8,7 @@
 from django.views.generic.base import TemplateView
 from shoop.admin.dashboard import get_activity
 from shoop.admin.module_registry import get_modules
+from shoop.core.telemetry import try_send_telemetry
 
 
 class DashboardView(TemplateView):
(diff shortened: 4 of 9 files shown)


Pikkupomo:
If I'm reading this correctly (and might be I'm not :)) it seems that user is opted in by default, which means that atleast one telemetry is being sent without consent. If so, I think it's not good. This could be resolved by directing first login to admin to `TelemetryView` (thus skipping dashboard which sends the telemetry).

If we want to send 1 telemetry anyways, we should then (in dashboard) say that `Shoop is collecting telemetry from installation, see <link> for more information and a possibility to opt-out if you want`. Or by directing there after first access to dashboard. 

Currently `the feeling` is that its happening behind the curtains without user ever noticing it.  I think we should push this more towards these 2 facts:
* User understands that this is happening
* We have his `permission` to do so

Otherwise looks good :+1:

akx:
You're absolutely correct: one telemetry payload is sent unless the user knows to navigate to `/sa/system/telemetry/` before visiting the dashboard to opt out of the feature.

I'm not sure I agree with the procedure either, though I reasoned with myself that it isn't _that_ bad, considering it contains no business information.  But yeah, you're right, this is definitely happening very quietly, even sneakily, behind the curtains.

I also have the feeling giving users the choice to opt-out by default would have more users opting out than leaving it be, and I got the impression from @tomiap that we really do want to gather these statistics anyway.

For the record, here's an example payload (as displayed on the Telemetry information page):

```json
{
  "apps": [
    "django.contrib.admin",
    "django.contrib.auth",
    "django.contrib.contenttypes",
    "django.contrib.sessions",
    "django.contrib.messages",
    "django.contrib.staticfiles",
    "django_jinja",
    "filer",
    "easy_thumbnails",
    "shoop.core",
    "shoop.simple_pricing",
    "shoop.simple_supplier",
    "shoop.default_tax",
    "shoop.front",
    "shoop.front.apps.registration",
    "registration",
    "shoop.front.apps.auth",
    "shoop.front.apps.customer_information",
    "shoop.front.apps.personal_order_history",
    "shoop.front.apps.simple_order_notification",
    "shoop.front.apps.simple_search",
    "shoop.admin",
    "shoop.addons",
    "shoop.testing",
    "bootstrap3",
    "shoop.notify",
    "shoop.simple_cms"
  ],
  "host": "127.0.0.1:8000",
  "key": "icbTWzYGQb9UT4OteMGioIAsjrUabx2GbuHlnldRwwSq6wjo",
  "machine": "AMD64",
  "platform": "Windows-8-6.2.9200",
  "python_version": "3.4.2 (v3.4.2:ab2c023a9432, Oct  6 2014, 22:16:31) [MSC v.1600 64 bit (AMD64)]",
  "shoop_version": "1.0.0"
}
```

So yeah -- this really is a policy issue. I'm happy to rework the code to be opt-in by default, or something else altogether, but I do need PO input on that.

akx:
As discussed in meatspace, I added a 24 hour grace period during which telemetry payloads will never be sent.

shoop/core/telemetry.py
```diff
@@ -0,0 +1,196 @@
...
+from shoop.core.models.persistent_cache import PersistentCacheEntry
+import json
+import platform
+import requests
+import shoop
+import sys
+
+OPT_OUT_KWARGS = dict(module="telemetry", key="opt_out")
+INSTALLATION_KEY_KWARGS = dict(module="telemetry", key="installation_key")
+LAST_DATA_KWARGS = dict(module="telemetry", key="last_data")
```
suutari-ai: Why you're using `dict(foo=bar)` rather than the normal `{'foo': bar}` construct here?

suutari-ai: Ah, I see... It seems more logical now that I looked the code that uses them... These are OK.

shoop/core/telemetry.py
```diff
@@ -0,0 +1,196 @@
...
+                raise TelemetryNotSent("Trying to resend too soon", "age")
+
+    data = get_telemetry_data(request)
+    try:
+        resp = requests.post(url=settings.SHOOP_TELEMETRY_URL, data=data)
+    except Exception as exc:
+        data = {
+            "data": data,
+            "error": force_text(exc),
+        }
```
suutari-ai: The exceptions are handled nicely here, but I'm still a bit worried, since it could block for a long time if the server is unreachable for some reason (no net, no DNS-server, outgoing traffic is firewalled, our end not listening, etc.).

Can we easily make this async, e.g. by doing the stuff from lines 160 to 173 in a multiprocessing.Pool.apply_async or something similar? That would of course change the way response is reported back, because it cannot be returned as normal return value then. We could provide a future to it though, so caller could call something like
```python
telemetry_result = send_telemetry(request)
try:
    result = telemetry_result.get(timeout=1)
except multiprocessing.TimeoutError:
    result = None
...
```

shoop/admin/views/dashboard.py
```diff
@@ -22,3 +23,7 @@ def get_context_data(self, **kwargs):
             blocks.extend(module.get_dashboard_blocks(request=self.request))
         context["activity"] = get_activity(request=self.request)
         return context
+
+    def get(self, request, *args, **kwargs):
+        send_telemetry(request)
+        return super(DashboardView, self).get(request, *args, **kwargs)
```
suutari-ai: I would rather see this code here being more explicit about what's going to happend, e.g.
```python
    if telemetry.should_send(request):
        telemetry.send(request)
    return super(DashboardView, self).get(request, *args, **kwargs)
```
That would make it more clear that we don't actually send telemetry for each request, and reader can see the `telemetry.should_send` function to see when it actually happens if he cares. (Feel free to ignore renaming the `telemetry.send_telemetry` to just `telemetry.send`... just a personal taste..)

shoop/admin/__init__.py
```diff
@@ -26,7 +26,8 @@ class ShoopAdminAppConfig(AppConfig):
             "shoop.admin.modules.users:UserModule",
             "shoop.admin.modules.methods:MethodModule",
             "shoop.admin.modules.attributes:AttributeModule",
-            "shoop.admin.modules.demo:DemoModule"
+            "shoop.admin.modules.demo:DemoModule",
+            "shoop.admin.modules.system:SystemModule",
```
suutari-ai: I would put SystemModule as first module in the list.

shoop_tests/core/test_telemetry.py
```diff
@@ -0,0 +1,91 @@
+# -*- coding: utf-8 -*-
+from django.test.utils import override_settings
```
suutari-ai: License header is missing.

shoop/admin/modules/system/views/telemetry.py
```diff
@@ -0,0 +1,38 @@
...
+    template_name = "shoop/admin/system/telemetry.jinja"
+
+    def get_context_data(self, **kwargs):
+        context = super(TelemetryView, self).get_context_data(**kwargs)
+        context.update({
+            "opt_in": not is_opt_out(),
+            "is_grace": is_in_grace_period(),
+            "last_submission_time": get_last_submission_time(),
+            "submission_data": get_telemetry_data(request=self.request, indent=2),
+            "title": "Telemetry"
```
suutari-ai: What is this title here for? (Why it's not translated?)

shoop/admin/modules/system/views/telemetry.py
```diff
@@ -0,0 +1,38 @@
...
+        })
+        return context
+
+    def get(self, request, *args, **kwargs):
+        if "last" in request.REQUEST:
```
suutari-ai: Isn't request.REQUEST deprecated?

shoop/admin/modules/system/views/telemetry.py
```diff
@@ -0,0 +1,38 @@
...
+# This file is part of Shoop.
+#
+# Copyright (c) 2012-2015, Shoop Ltd. All rights reserved.
+#
+# This source code is licensed under the AGPLv3 license found in the
+# LICENSE file in the root directory of this source tree.
+
+from django.http.response import HttpResponseRedirect, HttpResponse
+from django.views.generic.base import TemplateView
+from shoop.core.telemetry import is_opt_out, get_last_submission_time, get_telemetry_data, set_opt_out, \
```
suutari-ai: Don't use backslashes (`\`)... I would consider importing this as `from shoop.core import telemetry`.

shoop/admin/templates/shoop/admin/system/telemetry.jinja
```diff
@@ -0,0 +1,56 @@
+{% extends "shoop/admin/base.jinja" %}
+{% block content %}
+    <div class="container-fluid">
+        <div class="row">
+            <div class="col-md-4">
+                <h2>About Shoop Telemetry</h2>
+                <p>Shoop will occasionally send telemetry data to a Shoop.io server.</p>
+                <p>
+                    The data contains an unique installation key,
+                    as well as the hostname of the installation (as sent by the HTTP client).
```
suutari-ai: I don't understand "HTTP client" in this context and I bet many Shoop administrator will feel the same.

shoop/admin/modules/system/views/telemetry.py
akx: Sassafrassin' Django devs keep deprecating things from under my feet, darnit.

shoop/admin/views/dashboard.py
akx: Not having a separate public "should send" function is a conscious choice.

I don't want it to ever be possible for telemetry to be accidentally sent.

shoop/core/telemetry.py
akx: I don't think multiprocessing would work in an uWSGI context.
Ditto with threads being disabled in some environments/configs.

I wouldn't honestly worry about it too much. Regardless, I've added `timeout=5`, which will help a bit.

akx:
retest this please

shoop/core/telemetry.py
suutari-ai: OK then, probably the timeout will save us from most of the problems.

shoop/admin/views/dashboard.py
suutari-ai: Still, it makes the code less readable. At least the function should be named "send_if_supposed" or something like that then...

akx: The docstring makes it pretty clear, doesn't it...

suutari-ai: No, it doesn't. The code says "send telemetry", which does not sound conditional in any way. Just renaming the function would make the code much more explicit.

shoop/admin/templates/shoop/admin/system/telemetry.jinja
```diff
@@ -0,0 +1,57 @@
...
+{% block content %}
+    <div class="container-fluid">
+        <div class="row">
+            <div class="col-md-4">
+                <h2>About Shoop Telemetry</h2>
+                <p>Shoop will occasionally send telemetry data to a Shoop.io server.</p>
+                <p>
+                    The data contains an unique installation key,
+                    as well as the hostname of the installation (as sent by the visiting browser; currently
+                    <code>{{ request.get_host() }}</code>).
```
suutari-ai: Ah, very nice improvement to the "HTTP client". Much better, IMHO. Thanks.

shoop/core/telemetry.py
```diff
@@ -0,0 +1,196 @@
...
+        data = {
+            "data": data,
+            "response": force_text(resp.content, errors="replace"),
+            "status": resp.status_code,
+        }
+    save_telemetry_submission(data)
+    return data
+
+
+def send_telemetry(request=None, max_age_hours=72, raise_on_error=False):
```
suutari-ai: max_age_hours should probably be a setting rather than argument here.

shoop/core/telemetry.py
```diff
@@ -0,0 +1,196 @@
...
+    :type raise_on_error: bool
+    :return: Sent data (possibly with response information) or False if not sent.
+    :rtype: dict|bool
+    """
+    try:
+        return _send_telemetry(request=request, max_age_hours=max_age_hours)
+    except TelemetryNotSent:
+        if raise_on_error:
+            raise
+        return False  # pragma: no cover
```
suutari-ai: Please don't add "no cover" pragmas for code that can easily be tested.

suutari-ai:
Btw, commit message title should not end with period.

akx:
Renamed to `try_send_telemetry`.

PR #27, opened 2015-06-25
Requirement upgrade

akx:


akx:
Those pesky newlines...

akx:
retest this please

akx:
retest this please

suutari-ai:
Yeah, probably we should keep these updated, since we're pinning to exact versions. Though probably we should stop pinning to exact versions, at least for the Python packages listed in setup.py, and instead just specify the minimum (and maximum) versions with `>=` (and `<=`).

Anyway, thanks for the update!

akx:
You're right -- though I think we shouldn't necessarily loosen up _all_ versions. Something like Django, though, sure.

suutari-ai:
I think it would be better to pin the exact versions in the project that uses Shoop (with requirements.txt or setup.py) and let Shoop just specify the the min/max ranges for each requirement so that if the project needs a specific version for some reason it would be less likely that the version would conflict with Shoop requirements. Therefore I'd say we probably want to loosen up all versions, or do you see any exception: any package that we want to pin to specific version?

PR #29, opened 2015-06-26
Documentation!

akx:


akx:
So yeah, let it be noted `SHOOP_HOME_LOCALE` was not used anywhere, so it's gone.

PR #30, opened 2015-06-26
shoop_setup_utils: Fix find_packages (SHOOP-994)

suutari-ai:
Both versions of shoop_setup_utils.find_packages function always
returned None, rather than the correct result of package list, because
return statements were missing.  Fix that.


suutari-ai:
Can't understand why this passed the tests last time though...

akx:
Huh, wow. :D

PR #31, opened 2015-06-29
Harmonize 'New' buttons in admin

akx:
Refs SHOOP-686

### DIFF
--- a/shoop/admin/modules/products/__init__.py
+++ b/shoop/admin/modules/products/__init__.py
@@ -87,14 +87,14 @@ class ProductModule(AdminModule):
             url = reverse("shoop_admin:product.new")
             if " " in query:
                 yield SearchResult(
-                    text="Create product called '%s'" % query,
+                    text=_("Create Product Called \"%s\"") % query,
                     url=manipulate_query_string(url, name=query),
                     is_action=True
                 )
             else:
                 if query.lower() not in skus_seen:
                     yield SearchResult(
-                        text="Create product with SKU '%s'" % query,
+                        text=_("Create Product with SKU \"%s\"") % query,
                         url=manipulate_query_string(url, sku=query),
                         is_action=True
                     )
--- a/shoop/admin/toolbar.py
+++ b/shoop/admin/toolbar.py
@@ -14,6 +14,7 @@ from django.utils.encoding import force_text
 from django.utils.html import conditional_escape, format_html
 from django.utils.safestring import mark_safe
 from django.utils.translation import ugettext as _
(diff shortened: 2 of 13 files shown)


akx:
Wait, never mind just yet...

akx:
Erm, waiiit...

akx:
There, that's more like it!

shoop/admin/modules/products/__init__.py
```diff
@@ -87,14 +87,14 @@ def get_search_results(self, request, query):
             url = reverse("shoop_admin:product.new")
             if " " in query:
                 yield SearchResult(
-                    text="Create product called '%s'" % query,
+                    text="New product called '%s'" % query,
```
suutari-ai: Other buttons have the entity in title case, also using double quotes seems more appropriate than single quotes and this probably should be translated, i.e. the text here should be `_('New Product called "%s"')`.

shoop/admin/modules/products/__init__.py
```diff
@@ -87,14 +87,14 @@ def get_search_results(self, request, query):
...
                     url=manipulate_query_string(url, name=query),
                     is_action=True
                 )
             else:
                 if query.lower() not in skus_seen:
                     yield SearchResult(
-                        text="Create product with SKU '%s'" % query,
+                        text="New product with SKU '%s'" % query,
```
suutari-ai: same here

shoop/admin/toolbar.py
```diff
@@ -109,10 +110,28 @@ class NewActionButton(URLActionButton):
...
+        :rtype: shoop.admin.toolbar.NewActionButton|None
+        """
+
+        if "url" not in kwargs:
+            try:
+                url = get_model_url(model, kind="new")
+            except NoModelUrl:
+                return None
+            kwargs["url"] = url
+        kwargs.setdefault("text", _("New %(model)s") % {"model": model._meta.verbose_name})
```
suutari-ai: Hmm... this won't get the same case, since verbose name is in lower case and other buttons are in title case. But unconditionally applying title case to this would result in wrong case for non-English languages, e.g. "Uusi Tuote" doesn't look very good.

akx:
I decided the search result actions should still be "Create", because it's a proper verb.
They're translatable, now, though, and in Title Case All The Way.

shoop/admin/toolbar.py
suutari-ai: This is very minor thing though and this PR improves the consistency quite a bit so I won't care about this minor detail.

PR #32, opened 2015-06-30
Docker, Docker everywhere

akx:
This PR adds a basic Dockerfile that can be used to boot up an "evaluation" instance of Shoop.

### DIFF
--- a/Dockerfile
+++ b/Dockerfile
@@ -1,0 +1,15 @@
+FROM node:0.12.5
+MAINTAINER Aarni Koskela <aarni.koskela@shoop.io>
+EXPOSE 8080
+ADD . /var/www/shoop/working_copy
+
+RUN apt-get update && \
+    apt-get install -y --no-install-recommends python3-minimal python3-virtualenv python3-pip python3-dev python3-pil && \
+    apt-get clean && \
+    rm -rf /var/lib/apt/lists/*
+RUN python3 -m virtualenv -p /usr/bin/python3 --system-site-packages /var/www/shoop/venv && \
+    /var/www/shoop/venv/bin/pip install -U pip setuptools
+RUN /var/www/shoop/venv/bin/pip install /var/www/shoop/working_copy
+RUN /var/www/shoop/venv/bin/python -m shoop_workbench migrate
+RUN /var/www/shoop/venv/bin/python -m shoop_workbench shoop_populate_mock --with-superuser=admin
+CMD /var/www/shoop/venv/bin/python -m shoop_workbench runserver 0:8080
--- a/.dockerignore
+++ b/.dockerignore
@@ -1,0 +1,7 @@
+*.pyc
+*.pyd
+*.pyo
+*__pycache__*
+*node_modules*
+.dockerignore
+Dockerfile
--- a/shoop_setup_utils/commands.py
+++ b/shoop_setup_utils/commands.py
@@ -19,7 +19,7 @@ class BuildCommand(distutils.command.build.build):
         my_cmds = [
             BuildProductionResourcesCommand.command_name,
         ]
-        return super_cmds + my_cmds
+        return my_cmds + super_cmds
 
 
 class BuildResourcesCommand(distutils.core.Command):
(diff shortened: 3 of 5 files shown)


JsseL:
PR or commit messages doesn't say much, but looks like we get somekind of docker support for Shoop. That sounds nice.

suutari-ai:
> Build our resources before build_py (that copies files)

Please add better explanation to this commit message to explain why the order has to be changed. There is some stuff in parenthesis that does not belong to title line, but rather you should have a paragraph or two explaining what used to happen and how this commit changes it. We used quite a lot of time to debug this and commit message is a good place to log the (relevant) things we found out while debugging it.

akx:
Fixed (to some degree), @suutari-ai

Dockerfile
```diff
@@ -0,0 +1,11 @@
+FROM node:0.12.5
+RUN apt-get update && apt-get install -y --no-install-recommends python3-minimal python3-virtualenv python3-pip python3-dev python3-pil
+RUN python3 -m virtualenv -p /usr/bin/python3 --system-site-packages /var/www/shoop/venv
```
suutari-ai: Why system-site-packages?

Also, why not use "python3 -m venv" rather than "virtualenv" ?

Dockerfile
```diff
@@ -0,0 +1,11 @@
...
+RUN /var/www/shoop/venv/bin/pip install -U pip
+ADD . /var/www/shoop/working_copy
+RUN /var/www/shoop/venv/bin/pip install /var/www/shoop/working_copy
+RUN SHOOP_WORKBENCH_DISABLE_MIGRATIONS=1 /var/www/shoop/venv/bin/python -m shoop_workbench migrate
```
suutari-ai: Disabling migrations for the docker image is not a good idea, since if someone wants to develop using this image, he might want to make new migrations and testing them will need the base migrations being applied.

Optimizing the time it takes to build the docker image is also irrelevant, since it can be done in Docker Hub and users can just fetch the result.

Dockerfile
```diff
@@ -0,0 +1,11 @@
...
+RUN /var/www/shoop/venv/bin/python -m shoop_workbench shoop_populate_mock --with-superuser=admin
+RUN apt-get clean && rm -rf /var/lib/apt/lists/* /tmp/* /var/tmp/*
```
suutari-ai: Move the apt-get clean && rm -rf /var/lib/apt/lists/* part to same RUN command that does the "apt-get update" and "apt-get install". That'll make the intermediate image smaller.

Also move the rm -rf /tmp/* /var/tmp/* to same RUN command that leaves the cruft in /tmp and /var/tmp. Is it pip? Btw. are you sure that something really leaves cruft to both of these?

Dockerfile
akx: > Why system-site-packages?

`python3-pil`

suutari-ai:
@akx, your commit message:
> Build our resources before original build tasks
> 
> It turns out that the `build_py` command actually also copies data files
> into the build directory.  This naturally means we'll have to do the
> Node.js based builds before `build_py` is allowed to copy those artifacts
> further.
>
> Refs SHOOP-994

Exactly what I meant. Thanks! :+1:

Dockerfile
suutari-ai: OK then. I usually let PIL compile itself, but maybe installing it from deb packages is better idea for the Docker image.

akx: Also, `venv` doesn't seem to work:

> Command '['/var/www/shoop/venv/bin/python3', '-Im', 'ensurepip', '--upgrade', '--default-pip']' returned non-zero exit status 1

and I don't really have the inclination to figure out why when `virtualenv` does the trick :)

suutari-ai: Let's use virtualenv then. Dunno what's the base image of node:0.12.5, but if it's Ubuntu, then probably the reason is this bug: https://bugs.launchpad.net/ubuntu/+source/python3.4/+bug/1290847

Dockerfile
akx: What would be the workflow of someone developing using the image anyway? Besides, they could just use `migrate --fake` to fix the migration history table; using `DISABLE_MIGRATIONS` just makes Django use the faster logic that Just Creates Tables.

suutari-ai:
Good work, @akx. Thanks!

PR #35, opened 2015-07-02
Clean up and speed up admin statics build

akx:
* I noticed the NPM `glob` module has an ignore mechanism, so why not use that.
* As discussed on Slack, clean installations would kvetch about Bower files being missing before running `gulp bower`. Now, the regular tasks that use Bower files aren't even being installed if we're doing `gulp bower`, so that won't happen. (In addition, there's lots less code to load in this case, so speed-ups too! Win win!)
* Code style, man! JSCS'd the heck out of those bits.


suutari-ai:
Very nice! Thanks @akx!

PR #37, opened 2015-07-02
Unicode, literally (and licenses too~)

akx:


_misc/ensure_unicode_literals.py
```diff
@@ -42,24 +42,29 @@ def process_file(path):
...
+    need_encoding_comment = True
+    first_non_comment_line_index = 0
+    for line_index, line in enumerate(source_lines):
+        if not line.strip():
+            continue
+        if encoding_comment_regexp.match(line):
+            need_encoding_comment = False
+        if not line.startswith("#"):
+            first_non_comment_line_index = line_index
+            break
```
suutari-ai: IMHO we should only install encoding comment headers to files that are not all ASCII.

akx: Done and done.

suutari-ai: i.e. `any(x > 127 for x in six.iterbytes(open(FILENAME, 'rb').read()))`

suutari-ai:
Cool bits!

PR #40, opened 2015-07-02
workbench: Port execfile usage to Python 3

suutari-ai:
There is no execfile in Python 3, so convert the code to use open,
compile and exec directly.  Now it should work with Python 3 too.

SHOOP-994 / SHOOP-1027

### DIFF
--- a/shoop_workbench/settings/__init__.py
+++ b/shoop_workbench/settings/__init__.py
@@ -27,8 +27,12 @@ def configure(setup):
 
     # Load local settings from file
     if local_settings_file:
-        local_settings_ns = {}
-        execfile(local_settings_file, local_settings_ns)
+        local_settings_ns = {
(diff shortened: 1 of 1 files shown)


akx:
Considering shoop_workbench might be the template many people use for their own projects, it's starting to turn (needlessly?) complicated.

suutari-ai:
I agree. I thought that we should move the LOCAL_SETTINGS_FILE environment stuff to Shoop Core, but didn't want to do that yet.

Anyway, if someone wants to create a pip installable project, there has to be some way to read settings from outside the installation dir, it can be environment variables, ini-file, or some fancy key-value storage like Redis etc. Wouldn't be too bad if had support for one/some of those...

suutari-ai:
I mean built-in support...

PR #42, opened 2015-07-03
Prepare for 1.1 release (SHOOP-987)

suutari-ai:
Update change log and add release instructions.

### DIFF
--- a/ChangeLog.rst
+++ b/ChangeLog.rst
@@ -10,29 +10,53 @@ Unreleased
   appropriate version header and this help text will be removed.
 
 - Improve "Getting Started with Shoop Development" documentation
+
 - Add a basic REST API for reading/writing products and reading orders
+
 - Use the database to store shopping baskets by default
+
 - Implement pluggable shopping basket storage backends
+
 - Implement basic contact group admin
+
 - Add telemetry (usage statistics) system
+
 - Add Dockerfile
+
 - Improve admin login flow
+
 - Document settings; make documentation builds available on ReadTheDocs
+
 - Make release packaging much more robust
+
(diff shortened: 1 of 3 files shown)


setup.py
```diff
@@ -15,6 +15,28 @@
 LONG_DESCRIPTION_FILE = os.path.join(TOPDIR, 'README.rst')
 VERSION_FILE = os.path.join(TOPDIR, 'shoop', '_version.py')
 
+# Release instructions
```
akx: I don't think these belong in a comment in setup.py. Can you make a Real, Actual Document out of these -- maybe .rst in `doc/`, even?

ChangeLog.rst
```diff
@@ -0,0 +1,62 @@
+Shoop Change Log
```
akx: I guess the changelog could live in `doc/` too? Would make it easily browseable via RTFD once it unbreaks.

suutari-ai: It's kind of "standard" that ChangeLog is in repository root. Maybe we could symlink it to doc though, but let's check that before the next release. OK?

setup.py
suutari-ai: You only need these when releasing. It's just cruft for those who don't do releases.

On the other hand, when releasing, the first thing you try to accomplish, is to update the version number, which you'll find in setup.py. That's why it's the most visible place for instructions like this. I can move them to doc and add a "link" though, but I'm a bit scared that then nobody follows the "link".

akx:
Let's fix those minor issues later.

suutari-ai:
Thanks, I'll create an issue to our backlog.

PR #47, opened 2015-07-30
Add possibility to set primary image for product

jaywink:
Product edit allows choosing one of the linked ProductMedia active and public image types as primary for particular product.

While this is not shop based, it does fix problem with default front templates that product images are not visible unless one is set as active.


juhakujala:
Closing, since this is being implemented in [PR #57](https://github.com/shoopio/shoop/pull/57)

PR #48, opened 2015-07-31
Fix showing contact detail at admin

tulimaki:
Order.orderer is PersonContact so it can't be filtered with CompanyContact object.

Showing current addresses fail since None is not iterable. This happens when contact has only one address set.


akx:
Looks reasonable. Thanks!

PR #49, opened 2015-08-03
Admin: Product choice widget and cross-sell management

akx:
This PR adds a product choice widget that works like the pre-existing media choice widget.

I figured the best way to let users pick from possibly millions of products is to use the same Picotable they already use to browse through their selection.

In addition, to test the functionality I added a simple cross-sell (related products) management view.

Refs SHOOP-31
Refs SHOOP-1038


suutari-ai:
I like the idea to use Product List View as a product selector. Great work!

PR #56, opened 2015-08-06
Add link to admin panel in frontend

Pikkupomo:
Refs SHOOP-1053

### DIFF
--- a/shoop/front/templates/shoop/front/includes/_navigation.jinja
+++ b/shoop/front/templates/shoop/front/includes/_navigation.jinja
@@ -119,9 +119,10 @@
                             </li>
                             {% endif %}
                             {% if user.is_superuser %}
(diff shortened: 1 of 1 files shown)


juhakujala:
:+1:

akx:
This will break if the `shoop.admin` app isn't installed.

suutari-ai:
Yeah, shoop.admin should add such link through UI extension mechanisms which are not yet implemented (nor designed). Until that, we can just assume that shoop.admin is always installed.

akx:
@suutari-ai: for now we should just follow what the other conditionally enabled links do, i.e. wrap the link in

```jinja
{% if shoop.urls.has_url("shoop_admin:dashboard") %}
```

suutari-ai:
That's fine too, but not required IMHO. The ifs are going away anyway.

PR #62, opened 2015-08-12
Related products helper: rename, cache, correct order

akx:
There's more plumbing than actual fixing in here, really... :)

### DIFF
--- a/shoop/front/template_helpers/product.py
+++ b/shoop/front/template_helpers/product.py
@@ -5,6 +5,7 @@
 #
 # This source code is licensed under the AGPLv3 license found in the
 # LICENSE file in the root directory of this source tree.
+from django.utils.deprecation import warn_about_renamed_method
 from jinja2.utils import contextfunction
 
 from shoop.core.models import AttributeVisibility, Product, ProductAttribute, ProductCrossSell, ProductCrossSellType
@@ -17,21 +18,11 @@ def get_visible_attributes(product):
     )
 
 
-# Deprecated, see `get_product_cross_sells()`
+# Deprecated, see `get_related_products()`
 @contextfunction
+@warn_about_renamed_method("", "get_products_bought_with", "get_related_products", DeprecationWarning)
 def get_products_bought_with(context, product, count=5):
-    related_product_cross_sells = (
-        ProductCrossSell.objects
-        .filter(product1=product, type=ProductCrossSellType.COMPUTED)
-        .order_by("-weight")[:(count * 4)])
-    products = []
-    for cross_sell in related_product_cross_sells:
-        product2 = cross_sell.product2
-        if product2.is_visible_to_user(context["request"].user) and product2.is_list_visible():
-            products.append(product2)
-        if len(products) >= count:
-            break
-    return products
+    return get_related_products(context, product, "computed", count)
 
 
 @contextfunction
@@ -44,7 +35,13 @@ def is_visible(context, product):
 
 
 @contextfunction
+@warn_about_renamed_method("", "get_product_cross_sells", "get_related_products", DeprecationWarning)
 def get_product_cross_sells(context, product, relation_type="related", count=4):
+    return get_related_products(context, product, relation_type, count)
+
+
+@contextfunction
+def get_related_products(context, product, relation_type="related", count=4):
     request = context["request"]
     rtype = ProductCrossSellType.RELATED
     if relation_type == "computed":
--- a/shoop/front/template_helpers/product.py
+++ b/shoop/front/template_helpers/product.py
@@ -8,7 +8,10 @@
 from django.utils.deprecation import warn_about_renamed_method
 from jinja2.utils import contextfunction
 
+from shoop.core import cache
 from shoop.core.models import AttributeVisibility, Product, ProductAttribute, ProductCrossSell, ProductCrossSellType
+from shoop.utils.enums import map_enum
+from shoop.utils.models import get_in_id_order
 
 
 def get_visible_attributes(product):
@@ -42,27 +45,40 @@ def get_product_cross_sells(context, product, relation_type="related", count=4):
 
 @contextfunction
 def get_related_products(context, product, relation_type="related", count=4):
+    """
+    Get at most `count` visible products related (by type `relation_type`) to the given `product`.
+
+    :param context: The rendering context. This parameter is passed implicitly by Jinja2.
+    :type context: jinja2.runtime.Context
+    :param product: Source product
+    :type product: shoop.core.models.Product
+    :param relation_type: Relation type (string or ProductCrossSellType value)
+    :type relation_type: str|ProductCrossSellType
+    :param count: Number of products to return.
+    :type count: int
+    :return: List of at most `count` products.
+    :rtype: list[shoop.core.models.Product]
+    :raises ValueError: ValueError may be raised if the `relation_type` passed isn't valid.
+    """
+    relation_type = map_enum(ProductCrossSellType, relation_type)
+    id_query_count = max(count * 4, 64)  # IDs to query for caching
+    cache_key = "related_product_ids:%d_%d_%d" % (product.pk, relation_type.value, id_query_count)
+    related_product_ids = cache.get(cache_key)
+    if related_product_ids is None:
+        related_product_ids = list(
+            ProductCrossSell.objects
+            .filter(product1=product, type=relation_type)
+            .order_by("weight")[:id_query_count].values_list("product2_id", flat=True)
+        )
+        cache.set(cache_key, related_product_ids)
+
+    if not related_product_ids:
+        return []
+
     request = context["request"]
-    rtype = ProductCrossSellType.RELATED
-    if relation_type == "computed":
-        rtype = ProductCrossSellType.COMPUTED
-    elif relation_type == "recommended":
-        rtype = ProductCrossSellType.RECOMMENDED
 
-    related_product_ids = list((
-        ProductCrossSell.objects
-        .filter(product1=product, type=rtype)
-        .order_by("weight")[:(count * 4)]).values_list("product2_id", flat=True)
+    return get_in_id_order(
+        Product.objects.list_visible(shop=request.shop, customer=request.customer),
+        related_product_ids,
+        count
     )
-
-    related_products = list(
-        Product.objects
-        .filter(id__in=related_product_ids)
-        .list_visible(shop=request.shop, customer=request.customer)
-    )
-
-    # Order related products by weight. Related product ids is in weight order.
-    # If same related product is linked twice to product then lowest weight stands.
-    related_products.sort(key=lambda prod: list(related_product_ids).index(prod.id))
-
-    return related_products[:count]
--- a/shoop/themes/classic_gray/templates/classic_gray/shoop/front/macros.jinja
+++ b/shoop/themes/classic_gray/templates/classic_gray/shoop/front/macros.jinja
@@ -34,8 +34,8 @@
     </div>
 {% endmacro %}
 
-{% macro render_cross_sell_products(product, relation_type="", title="") %}
-    {%- set cross_sell_products = shoop.product.get_product_cross_sells(product, relation_type) %}
+{% macro render_cross_sell_products(product, relation_type="related", title="") %}
+    {%- set cross_sell_products = shoop.product.get_related_products(product, relation_type) %}
     {% if cross_sell_products %}
         <hr>
         {% if title %}<h3>{{ title }}</h3>{% endif %}
--- a/shoop/utils/enums.py
+++ b/shoop/utils/enums.py
@@ -1,0 +1,65 @@
+# -*- coding: utf-8 -*-
+from __future__ import absolute_import
+from numbers import Number
+
+import six
+
+RAISE = object()
+
+
+def map_enum(enum_cls, value, default=RAISE, allow_magic_numbers=False):
+    """
+    Attempt to cast the given `value` into a value of the given `enum_cls`.
+
+    Tries by value first, then case-insensitively by name.
+
+    If casts fail, the `default` is returned (unless it is the default sentinel
+    object `RAISE`, in which case a `ValueError` is raised).
+
+    This is useful for places where it doesn't make sense to export the actual
+    enumeration classes (such as into templates, etc.) while still avoiding
+    "magic constants" (i.e. the actual enum values).  For the aforementioned
+    reason of being used in templates, it is purposefully slightly more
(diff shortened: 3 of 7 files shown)


suutari-ai:
This "fix" is quite BIG change to Shoop Core. Do we really need to introduce a whole new caching system for this? Have you discussed this implementation through with someone?

akx:
No, I haven't. Now's as good a time as any, hey?

Also, it's not a very big change, and it's entirely opt-in for things that care about caches.

So far, only two modules in the current master Shoop use `django.core.cache` at all (as you can see from the diff).

The new `shoop.core.cache` module is only a wrapper for `django.core.cache` with better semantics regarding [versioned caches](https://andy.wordpress.com/2008/03/01/cache-with-versioned-tags/).

Using the module should make it very easy to invalidate some cached items (without wholesale clearing the entire cache) -- for instance, when related products are changed, `bump_version("related_products")` will make subsequent attempts of retrieving related products re-cache things.

akx:
(Also, considering there haven't been too many other core devs available today, I figure asking for forgiveness rather than permission wasn't a bad policy here...)

akx:
(Double also: There's really more docstrings than code in the `shoop.core.cache` module.)

suutari-ai:
I bet the MySQL query-bug can be fixed with a lot smaller diff than this and IMHO that fix should be done in a separate PR. When the fix is merged, we should talk about this caching etc. (There were some parts that I'd like to comment about, but don't have time at the moment.)

akx:
Well, sure, I'll write the temporary fix, but I'd still like to keep this PR open :)

Pikkupomo:
now that versions change etc, we should check this thru, correct @suutari-ai?

akx:
This is rebased on d56dd24954e2921e3d4934a0e56387c84fbf4192 now.

akx:
Rebased on top of 4ffb38aa542346c7886db3d3d8a0cd725026570f

akx:
Assigning @suutari-ai because

> (There were some parts that I'd like to comment about, but don't have time at the moment.)

shoop/front/template_helpers/product.py
```diff
@@ -44,28 +38,46 @@ def is_visible(context, product):
...
+    :param product: Source product
+    :type product: shoop.core.models.Product
+    :param relation_type: Relation type (string or ProductCrossSellType value)
+    :type relation_type: str|ProductCrossSellType
+    :param count: Number of products to return.
+    :type count: int
+    :return: List of at most `count` products.
+    :rtype: list[shoop.core.models.Product]
+    """
+    relation_type = map_enum(ProductCrossSellType, relation_type, default=ProductCrossSellType.RELATED)
```
suutari-ai: What's the default for here? IMHO it should just raise if API user gives unknown value.

shoop/utils/enums.py
```diff
@@ -0,0 +1,36 @@
...
+    """
+    try:
+        return enum_cls(value)
+    except ValueError:
+        pass
+    members = getattr(enum_cls, "_map_enum_cache_", None)
+    if members is None:
+        members = dict((k.lower(), v) for (k, v) in enum_cls.__members__.items())
+        enum_cls._map_enum_cache_ = members
+    return members.get(six.text_type(value).lower(), default)
```
suutari-ai: `default` is not good keyword for a fallback value here, since it's just a way to hide ValueError of unknown enum value.

Is it really needed? I think fallbacking to None is more harmful than raising ValueError.

shoop/utils/models.py
```diff
@@ -11,6 +11,7 @@
 
 from django.db import models
 from django_countries.fields import Country
+from jinja2.filters import do_batch
```
suutari-ai: Using jinja2 filter for this does not feel right. How about using `shoop.utils.iterables.batch` instead?

shoop_tests/utils/test_enum.py
```diff
@@ -0,0 +1,10 @@
+# -*- coding: utf-8 -*-
+from shoop.core.models.products import ProductMode
+from shoop.utils.enums import map_enum
+
+
+def test_map_enum():
+    assert map_enum(ProductMode, 0) is ProductMode.NORMAL
```
suutari-ai: What's the point of mapping int to enum with this function? Doesn't it produce unreadable code (because of magic numbers)?

shoop_tests/utils/test_enum.py
```diff
@@ -0,0 +1,10 @@
...
+    assert map_enum(ProductMode, "normal") is ProductMode.NORMAL
+    assert map_enum(ProductMode, "Normal") is ProductMode.NORMAL
+    assert map_enum(ProductMode, "shower") is None
```
suutari-ai: This last one should raise ValueError (as said in previous comments).

shoop/utils/models.py
```diff
@@ -49,3 +50,27 @@ def build_or_query(over_fields, term, operator=""):
     def add_term(query_q, field):
         return (query_q | models.Q(**{("%s%s" % (field, operator)): term}))
     return reduce(add_term, over_fields, models.Q())
+
+
+def get_in_id_order(queryset, id_order, count):
+    """
+    Returns the first `count` identified objects of the given queryset in the given order.
```
suutari-ai: Naming of this function is somehow misleading...

shoop/utils/models.py
akx: Sure. The reason it wasn't used in the first place was that, well, it didn't exist back then. (It was only added in 2576caba9614da5008b1b23754722b8c547da65f.)

shoop_tests/utils/test_enum.py
akx: The point is that it is _able_ to do that mapping (which I foresee being useful for, say, import/export situations), but user code probably shouldn't use magic numbers.

shoop/front/template_helpers/product.py
suutari-ai: It would be more useful (for the template developer too) to raise an exception rather than returning possibly incorrect set of products (if e.g. the template developer typoes "related" as "reltaed"). Don't you think?

shoop_tests/utils/test_enum.py
suutari-ai: That should be another function IMHO, since it's entirely different use case and having the possibility map from int here allows template developers to write code with magic numbers and we shouldn't make that easy.

shoop/front/template_helpers/product.py
akx: That's an excellent point I didn't think about.

akx:
@suutari-ai:

I taught `map_enum` to not map numbers by default and removed the funny default from `get_related_products` (and added a nicer error message).

shoop/utils/enums.py
```diff
@@ -0,0 +1,64 @@
...
+
+    :param enum_cls: Enum class
+    :type enum_cls: enum.Enum
+    :param value: value or name
+    :type value: object|str
+    :param default: Default value to return in case of trouble
+    :type default:
+    :param allow_magic_numbers: Allow mapping via "magic numbers". If this is False and
+                                the value is a number, the default `enum_cls(value)` logic
+                                is not attempted.
```
suutari-ai: This allow_magic_numbers stuff does not belong to same function IMHO. A separate function would be much clearer, since it's a separate functionality. `map_enum(MyEnum, 42, allow_magic_numbers=True)` vs `map_enum_int(MyEnum, 42)`

Having the int mapping in separate function would also make this one much clearer, since the value would always be string.

shoop/utils/models.py
```diff
@@ -11,6 +11,7 @@
 
 from django.db import models
 from django_countries.fields import Country
+from shoop.utils.iterables import batch
```
suutari-ai: grouping

shoop/utils/enums.py
```diff
@@ -0,0 +1,64 @@
...
+    # The mapping is stored as a "private" attribute of the enum class.
+
+    members = getattr(enum_cls, "_map_enum_cache_", None)
+    if members is None:
+        members = dict((k.lower(), v) for (k, v) in enum_cls.__members__.items())
+        enum_cls._map_enum_cache_ = members
+
+    # Try looking up the value in the mapping.
+
+    retval = members.get(six.text_type(value).lower(), default)
```
suutari-ai: I don't like touching private parts of another class (the `_map_enum_cache_` member).

Though... Thinking about this members cache... The only reason it's needed at all is because we can't be sure that if the Enum's members are lower or uppercase (or worst, mixed case). How about just trying it like this:
```python
enum_val = enum_cls.__members__.get(value.upper())
if enum_val is None:
    enum_val = enum_cls.__members__.get(value.lower())
    if enum_val is None:
        for (k, v) in enum_cls.__members__.items():
            if k.lower() == value.lower():
                enum_val = k
                break
```
This would avoid building a cache and would be faster most of the time, since the first or the second will very likely catch all cases. It would be slower for cases where the enum members are mixed case, but that should be very rare.

shoop/utils/enums.py
akx: The separate function is then just, uh,

```python
def map_enum_int(enum_cls, value):
    return enum_cls(value)
```

I kind of like that `map_enum`, when told so, actively disallows magic numbers.

shoop/front/template_helpers/product.py
```diff
@@ -44,28 +39,49 @@ def is_visible(context, product):
...
+    :type relation_type: str|ProductCrossSellType
+    :param count: Number of products to return.
+    :type count: int
+    :return: List of at most `count` products.
+    :rtype: list[shoop.core.models.Product]
+    """
+    try:
+        relation_type = map_enum(ProductCrossSellType, relation_type)
+    except ValueError as ve:
+        six.raise_from(ValueError("%r is not a valid cross-sell type" % relation_type), ve)
```
suutari-ai: `The "Can't map value 'foo' to any member of enum ProductCrossSellType"` error message should be fine, so rather let the exception propagate that mess with it.

shoop/utils/enums.py
akx: That looks more complicated though.

My reasoning for building the cache is that -- due to Murphy's law of data imports -- we end up going into the slow path, say, 10,000 times, when some 3rd party has decided that "Related" is a nicer spelling than "RELATED" or "related".

OTOH, maybe it'd be better to just decorate this method with `lru_cache`..?

shoop/front/template_helpers/product.py
```diff
@@ -44,28 +39,49 @@ def is_visible(context, product):
...
 
-    # Order related products by weight. Related product ids is in weight order.
-    # If same related product is linked twice to product then lowest weight stands.
-    related_products.sort(key=lambda prod: list(related_product_ids).index(prod.id))
+    :param context: The rendering context. This parameter is passed implicitly by Jinja2.
+    :type context: jinja2.runtime.Context
+    :param product: Source product
+    :type product: shoop.core.models.Product
+    :param relation_type: Relation type (string or ProductCrossSellType value)
+    :type relation_type: str|ProductCrossSellType
```
suutari-ai: Why relation_type is not always a string? I don't think there even is any easy way to get the ProductCrossSellType enum value in the template, nor does it make much sense to use that instead of string.

shoop/themes/classic_gray/templates/classic_gray/shoop/front/macros.jinja
```diff
@@ -35,7 +35,7 @@
 {% endmacro %}
 
 {% macro render_cross_sell_products(product, relation_type="", title="") %}
-    {%- set cross_sell_products = shoop.product.get_product_cross_sells(product, relation_type) %}
+    {%- set cross_sell_products = shoop.product.get_related_products(product, relation_type) %}
```
suutari-ai: The default parameter for render_cross_sell_products macro does not seem correct anymore. Shouldn't it be "related"?

shoop/front/template_helpers/product.py
akx: Because this method _does_ accept the actual enum value by virtue of `map_enum`.

shoop/themes/classic_gray/templates/classic_gray/shoop/front/macros.jinja
akx: Fixed in a future revision, thanks.

shoop/front/template_helpers/product.py
akx: Sure.

shoop/utils/enums.py
suutari-ai: You're right it looks more complicated. I think the decoration approach would be nicer.

shoop/front/template_helpers/product.py
suutari-ai: API doc is a contract, not a specification of it's current implementation. It could probably also support any object with pk as product too. ;)

akx:
Huh... One would think `from __future__ import absolute_import` was enough to make that `from numbers import Number` work.

shoop/front/template_helpers/product.py
akx: Well, in any case I think it's sane to allow the enum value itself. For instance to support the case

```jinja
{% set relation_type = x.y.z.some_function_from_a_third_party(product) %}
{% for product in get_related_products(product, relation_type=relation_type) %}
...
```

suutari-ai: OK, that's a possible use case (not very like IMHO though), but fine. :)

suutari-ai:
How about just dropping the number test from map_enum and let it always do its job by comparing the value as a string? I.e. remove the part that tries to call the enum constructor with the value. (Also adding the lru_cache while at it.. to make the code cleaner.) That would eliminate the "from numbers import Number" problem. Simple is better than complex.

suutari-ai:
This is what I'm after:

```python
from django.utils.lru_cache import lru_cache


RAISE = object()


def map_enum(enum_cls, string, default=RAISE):
    """
    Cast string to value of given enum class.

    ...

    :type enum_cls: enum.EnumMeta
    :type string: str
    :type default: Any
    :rtype: enum.Enum|Any
    """
    value = _map_enum_cached(enum_cls, string)
    if value is None and default is RAISE:
        raise ValueError("Cannot map %r as value of %s" % (string, enum_cls))
    return value if (value is not None) else default


@lru_cache
def _map_enum_cached(enum_cls, string):
    for (key, value) in enum_cls.__members__.items():
        if key.lower() == string.lower():
            return value
    return None
```

Quite simple? Isn't it? (Probably broken too, since it's not tested ;) )

Pikkupomo:
Closing this, last comment is months ago. If you want some bits or pieces of this, please make own pr's of those. @akx @suutari-ai

PR #63, opened 2015-08-12
Fix simple variation form (and add tests)

akx:
Refs SHOOP-1075


suutari-ai:
This is cool! Thanks @akx!

PR #65, opened 2015-08-12
Admin: Show payment details in order view

akx:


shoop/admin/templates/shoop/admin/orders/detail.jinja
```diff
@@ -126,6 +126,32 @@
                             </div>
                         </div>
                     {% endcall %}
+
+                    {% set payments=order.payments.all() %}
+                    {% if payments %}
+                        {% call content_block(_("Payments"), "fa-dollar") %}
+                            <table class="table table-striped table-bordered">
```
juhakujala: Since we are using table here, could this be done the same way as the tab "order contents" where the table is hidden in mobile and list-group is used instead?

akx: This table is a fair bit simpler than the order content one (4 columns, most of which will have short contents).

juhakujala: Well if you are sure that it will fit the 320px screen, then it's fine by me.

suutari-ai: @akx: Have you checked this with a such small display size?

akx:
Rebased and added mobile listgroup.

image

PR #67, opened 2015-08-13
Import cleaning (direct model imports; isort)

akx:


shoop/core/utils/reference.py
```diff
@@ -9,7 +9,7 @@
 from django.utils.encoding import force_text
 
 from shoop.utils.importing import load
-from shoop.core.models.counters import Counter, CounterType
+from shoop.core.models.counters import Counter, CounterType  # must be direct import
```
suutari-ai: Why this must be direct import?

shoop/front/apps/simple_search/views.py
```diff
@@ -6,14 +6,14 @@
...
 from django.utils.encoding import force_bytes
 from django.views.generic import ListView
-import hashlib
-from shoop.core.models.products import Product
-from django.core.cache import cache
-from shoop.front.utils.views import cache_product_things
+from shoop.core.models import Product
 from shoop.front.utils.product_sorting import PRODUCT_SORT_CHOICES, sort_products
+from shoop.front.utils.views import cache_product_things
+import hashlib
```
suutari-ai: While fixing import statement ordering make it conform PEP-8 groping too, i.e. first stdlib imports, then third party imports, then local imports (= shoop imports in our case), each group in alphabetical order and groups separated by a blank line.

suutari-ai:
@akx: Hey, thanks. This is good stuff. :) Can you reply to the couple inline comments above and rebase this, so I may merge?

shoop/api/factories.py
```diff
@@ -5,19 +17,10 @@
...
 from shoop.api._vendor.parler_rest.fields import TranslatedFieldsField
 from shoop.api._vendor.parler_rest.serializers import TranslatableModelSerializer
-from rest_framework import serializers, viewsets, permissions
 
-
-def serializer_factory(model, serializer_class=None, attrs=None, meta=None):
-    attrs = attrs or {}
-    meta = meta or {}
-    if not serializer_class:
-        if issubclass(model, TranslatableModel):
```
suutari-ai: WTF happened here?

shoop/core/fields/tagged_json.py
```diff
@@ -5,6 +5,20 @@
...
+from django.core.exceptions import ImproperlyConfigured
+from django.utils import six
+from jsonfield.encoder import JSONEncoder
+from six import text_type
+
+from shoop.utils.importing import load
+from shoop.utils.iterables import first
+
 """
 "Tagged JSON" encoder/decoder.
```
suutari-ai: This is not correct. The module docstring should go above the imports. Quote from PEP-8:
> Imports are always put at the top of the file, just after any module comments and docstrings, and before module globals and constants.

suutari-ai: Maybe the reason why isort does this is because the unicode_literals import is too soon?

akx:
There, I think that's about it :)

shoop/core/fields/tagged_json.py
suutari-ai: This is interesting... If the imports go before the module docstring, it won't be treated as module docstring at all. Also the type of module docstring in Python 2 is unicode even when the unicode_literals import follows the docstring.
```
$ cat a.py 
"""
Docstring.
"""

import sys

print(sys.version, type(__doc__))
$ cat b.py 
import sys

"""
Docstring.
"""

print(sys.version, type(__doc__))
$ python a.py
('2.7.6 (default, Jun 22 2015, 17:58:13) \n[GCC 4.8.2]', <type 'unicode'>)
$ python b.py
('2.7.6 (default, Jun 22 2015, 17:58:13) \n[GCC 4.8.2]', <type 'NoneType'>)
$ python3 a.py
3.4.0 (default, Jun 19 2015, 14:20:21) 
[GCC 4.8.2] <class 'str'>
$ python3 b.py
3.4.0 (default, Jun 19 2015, 14:20:21) 
[GCC 4.8.2] <class 'NoneType'>
```

akx:
retest this please

akx:
retest this please

akx:
Well *that's* more like it.

akx:
(Self-)merging this per Slack discussion now that tests pass.

PR #68, opened 2015-08-13
Fix `get_product_cross_sells` to work on MySQL

akx:
Refs SHOOP-1073

Will be superseded by #62


suutari-ai:
Yep, I thought so, that the fix will be small! :D

PR #69, opened 2015-08-13
Fix variable variation form regression (and an important root cause)

akx:
Form regression introduced in 43a88d2092a525ac1785588443d2789c095403c5, refs SHOOP-1075.

More importantly: this adds a brand-new migration to fix the slightly broken `InternalIdentifierField` implementation. Refs SHOOP-1092.


akx:
retest this please

akx:
That's a good boy, Jenkins!

PR #70, opened 2015-08-14
Product image fixes

Pikkupomo:
Refs SHOOP-1089


akx:
The commit messages could be better.


* Fix issue with tabs -> "Ensure unique IDs for product media tab sets"
* Fix issue with thumbnails -> "Fix check for image type files for thumbnailing" (or something)
* Fix issue with primary images -> "Use .prop() instead of .attr() to check primary image checkboxes"

akx:
Thanks. :)

PR #75, opened 2015-08-24
Rework `SimplePricing` and pricing in general

Pikkupomo:
Add `PriceInfo` to simplify price handling
Add `prices_include_tax` into `Shop`

The idea behind this change is that `ShopProduct` determines the default price
of the product. Templates should only use `PriceInfo` the pricing module
offers thru `Product`. `SimplePricing` adds Customer Group Specific pricing on
top of that.

`SimplePricing` continues to return lowest possible price for `ShopProduct`

NOTE: The changes in `Product.get_base_price` will cause backwards incompability.

Refs SHOOP-514

### DIFF
--- a/shoop/core/models/product_shops.py
+++ b/shoop/core/models/product_shops.py
@@ -14,7 +14,7 @@ from django.utils.translation import ugettext_lazy as _
 from enumfields import EnumIntegerField
 
 from shoop.core.excs import ProductNotOrderableProblem, ProductNotVisibleProblem
-from shoop.core.fields import QuantityField, UnsavedForeignKey
+from shoop.core.fields import MoneyField, QuantityField, UnsavedForeignKey
 from shoop.core.models.product_media import ProductMediaKind
 from shoop.core.models.products import ProductVisibility, StockBehavior
 from shoop.core.signals import get_orderability_errors, get_visibility_errors
@@ -55,6 +55,10 @@ class ShopProduct(models.Model):
     shop_primary_image = models.ForeignKey(
         "ProductMedia", null=True, blank=True, related_name="primary_image_for_shop_products", on_delete=models.SET_NULL
     )
+
+    # the default price of this product in the shop, taxfulness is determined in
+    # `Shop.prices_include_tax`
+    default_price = MoneyField(verbose_name=_("Default price"), null=True, blank=True)
 
     class Meta:
         unique_together = (("shop", "product",),)
--- a/shoop/core/models/products.py
+++ b/shoop/core/models/products.py
@@ -258,20 +258,43 @@ class Product(AttributableMixin, TranslatableModel):
             for c in self.variation_children.all()
         )[0]
 
-    def get_price(self, context, quantity=1):
+    def get_price_info(self, context, quantity=1):
         """
+        returns a `PriceInfo` object for product
+
+        Returned `PriceInfo` object contains calculated `price` and `base_price`.
+        The calculation of prices is handled in the current pricing module.
+
         :type context: shoop.core.contexts.PriceTaxContext
-        :rtype: shoop.core.pricing.Price
+        :rtype: shoop.core.pricing.PriceInfo
         """
         from shoop.core.pricing import get_pricing_module
         module = get_pricing_module()
         pricing_context = module.get_context(context)
-        return module.get_price(pricing_context, product_id=self.pk, quantity=quantity)
+        return module.get_price_info(pricing_context, product=self, quantity=quantity)
 
-    def get_base_price(self):
-        from shoop.core.pricing import get_pricing_module
-        module = get_pricing_module()
-        return module.get_base_price(product_id=self.pk)
+    def get_price(self, context, quantity=1):
+        """
+        Returns `price` of the product calculated in the current pricing module.
+
+        :type context: shoop.core.contexts.PriceTaxContext
+        :rtype: shoop.core.pricing.Price
+        """
+        return self.get_price_info(context, quantity).price
+
+    def get_base_price(self, context):
+        """
+        Returns `base_price` of the product calculated in the current pricing module.
+
+        `base_price` defaults to `price` but the pricing module can decide what the value is.
+
+        For example in `SimplePricing` this value is `ShopProduct.default_price` or calculated `price`
+        based on customer groups.
+
+        :type context: shoop.core.contexts.PriceTaxContext
+        :rtype: shoop.core.pricing.Price
+        """
+        return self.get_price_info(context, quantity=1).base_price
 
     def get_taxed_price(self, context, quantity=1):
         """
--- a/shoop/core/models/products.py
+++ b/shoop/core/models/products.py
@@ -197,8 +197,8 @@ class Product(AttributableMixin, TranslatableModel):
     gross_weight = MeasurementField(unit="g", verbose_name=_('gross weight (g)'))
 
     # Misc.
-    purchase_price = MoneyField(verbose_name=_('purchase price'))
-    suggested_retail_price = MoneyField(verbose_name=_('suggested retail price'))
+    purchase_price = MoneyField(verbose_name=_('purchase price'), blank=True, null=True)
+    suggested_retail_price = MoneyField(verbose_name=_('suggested retail price'), blank=True, null=True)
     manufacturer = models.ForeignKey("Manufacturer", blank=True, null=True, verbose_name=_('manufacturer'))
     primary_image = models.ForeignKey(
         "ProductMedia", null=True, blank=True,
--- a/shoop/core/pricing/__init__.py
+++ b/shoop/core/pricing/__init__.py
@@ -5,7 +5,6 @@
 #
 # This source code is licensed under the AGPLv3 license found in the
 # LICENSE file in the root directory of this source tree.
-from __future__ import unicode_literals
 """
 Shoop modular product pricing functionality.
 
@@ -27,19 +26,25 @@ for more advanced uses, when you do not have access to an HTTP request,
 
 After you have acquired the module and a context, you can calculate
 prices for a product with the module's
-:func:`~PricingModule.get_price` method.
+:func:`~PricingModule.get_price_info` method.
 (:class:`~shoop.core.models.products.Product` objects contain the
-convenience method
-:func:`~shoop.core.models.products.Product.get_price` which
-do these steps for you.)
+convenience methods
+:func:`~shoop.core.models.products.Product.get_price`,
+:func:`~shoop.core.models.products.Product.get_base_price`,
+and :func:`~shoop.core.models.products.Product.get_base_price`
+which do these steps for you.)
 
 If you have multiple products, it will likely be more efficient --
 depending on the implementation of the module -- to use the
-:func:`~PricingModule.get_prices` method.
+:func:`~PricingModule.get_price_infos` method.
 
 TODO: document the concepts of base price and the pricing steps API.
 TODO: caching.
 """
+
+from __future__ import unicode_literals
+import abc
+import six
 
 import hashlib
 from shoop.apps.provides import load_module
@@ -48,9 +53,11 @@ from django.utils.encoding import force_bytes
 from django.utils.timezone import now
 
 from .price import Price, TaxfulPrice, TaxlessPrice
+from .price_info import PriceInfo
 
 __all__ = [
     "Price",
+    "PriceInfo",
     "PricingContext",
     "PricingModule",
     "TaxfulPrice",
@@ -90,10 +97,9 @@ class PricingContext(object):
     cache_key = property(get_cache_key)
 
 
-class PricingModule(object):
+class PricingModule(six.with_metaclass(abc.ABCMeta)):
     identifier = None
     name = None
-
     pricing_context_class = PricingContext
 
     def get_context(self, context):
@@ -116,26 +122,14 @@ class PricingModule(object):
     def get_context_from_data(self, **context_data):
         return self.pricing_context_class(**context_data)
 
-    def get_price(self, context, product_id, quantity=1):
+    @abc.abstractmethod
+    def get_price_info(self, context, product, quantity=1):
         """
-        Get context-specific Price for the given product and quantity.
-
-        May return TaxlessPrice or TaxfulPrice.
-
-        :type context: PricingContext
-        :rtype: Price
+        :param product: `Product` object or id of `Product`
+        :type product: shoop.core.models.Product|int
+        :rtype: PriceInfo
         """
-        return TaxlessPrice(0)
-
-    def get_base_price(self, product_id):
-        """
-        Get base Price for the given product.
-
-        May return TaxlessPrice or TaxfulPrice.
-
-        :rtype: Price
-        """
-        return TaxlessPrice(0)
+        pass
 
     def get_pricing_steps(self, context, product_id):
         """
@@ -155,15 +149,15 @@ class PricingModule(object):
         """
         return [(0, TaxlessPrice(0))]
 
-    def get_prices(self, context, product_ids, quantity=1):
+    def get_price_infos(self, context, products, quantity=1):
+        """
+        :param products: a list of `Product`s or id's
+        :type products:  Iterable[shoop.core.models.Product|int]
+        :rtype: dict[int,PriceInfo]
+        """
+        product_ids = [getattr(x, "pk", x) for x in products]
         return {
-            product_id: self.get_price(context, product_id=product_id, quantity=quantity)
-            for product_id in product_ids
-        }
-
-    def get_base_prices(self, product_ids):
-        return {
-            product_id: self.get_base_price(product_id=product_id)
+            product_id: self.get_price_info(context=context, product=product_id, quantity=quantity)
             for product_id in product_ids
         }
 
--- a/shoop/core/pricing/price_info.py
+++ b/shoop/core/pricing/price_info.py
@@ -1,0 +1,46 @@
+# -*- coding: utf-8 -*-
+# This file is part of Shoop.
+#
+# Copyright (c) 2012-2015, Shoop Ltd. All rights reserved.
+#
+# This source code is licensed under the AGPLv3 license found in the
+# LICENSE file in the root directory of this source tree.
+
+from shoop.core.pricing import Price
+
+
+class PriceInfo(object):
+    def __init__(self, price, base_price, expires_on=None):
+        """
+        Initialize PriceInfo with two given Price objects.
+
+        Given prices can be taxful or taxless, but their types must match.
+
+        :param price: Price of the product that is calculated in the pricing module
+        :type price: shoop.core.pricing.Price
+        :param base_price: Base price of the product, discounts are calculated based on this.
+        :type base_price: shoop.core.pricing.Price
+        """
+
+        assert isinstance(price, Price)
+        assert isinstance(base_price, Price)
+        assert price._units_match(base_price)
+        self.price = price
+        self.base_price = base_price
+        self.expires_on = expires_on
+
+    @property
+    def includes_tax(self):
+        return self.price.includes_tax
+
+    @property
+    def discount_amount(self):
+        return (self.base_price - self.price)
+
+    @property
+    def discount_percentage(self):
+        return (1 - (self.price / self.base_price)) * 100
+
+    @property
+    def is_discounted(self):
+        return (self.price < self.base_price)
--- a/shoop/core/shortcuts/__init__.py
+++ b/shoop/core/shortcuts/__init__.py
@@ -18,7 +18,6 @@ def update_order_line_from_product(request, order_line, product, quantity=1, sup
         raise Exception("set_from_product may not be used on saved lines")
 
     if not product:  # pragma: no cover
-
         raise Exception("set_from_product may not be used without product")
 
     # TODO: (TAX) Taxes in update_order_line_from_product
--- a/shoop/simple_pricing/admin_form_part.py
+++ b/shoop/simple_pricing/admin_form_part.py
@@ -26,8 +26,8 @@ class SimplePricingForm(forms.Form):
             self._build_fields()
 
     def _build_fields(self):
-        self.shops = [None] + list(Shop.objects.all())
-        self.groups = [None] + list(ContactGroup.objects.filter(
+        self.shops = list(Shop.objects.all())
+        self.groups = list(ContactGroup.objects.filter(
             Q(show_pricing=True) |
             Q(
                 id__in=SimpleProductPrice.objects.filter(product=self.product)
@@ -40,6 +40,7 @@ class SimplePricingForm(forms.Form):
             in SimpleProductPrice.objects.filter(product=self.product)
             .values_list("shop_id", "group_id", "price")
         )
+
         for group in self.groups:
             for shop in self.shops:
                 shop_group_id_tuple = self._get_id_tuple(shop, group)
--- a/shoop/simple_pricing/migrations/0002_remove_simpleproductprice_includes_tax.py
+++ b/shoop/simple_pricing/migrations/0002_remove_simpleproductprice_includes_tax.py
@@ -1,0 +1,18 @@
+# -*- coding: utf-8 -*-
+from __future__ import unicode_literals
+
+from django.db import models, migrations
+
+
+class Migration(migrations.Migration):
+
+    dependencies = [
+        ('simple_pricing', '0001_initial'),
+    ]
+
+    operations = [
+        migrations.RemoveField(
+            model_name='simpleproductprice',
+            name='includes_tax',
+        ),
+    ]
--- a/shoop/simple_pricing/migrations/0003_remove_groupless_prices.py
+++ b/shoop/simple_pricing/migrations/0003_remove_groupless_prices.py
@@ -1,0 +1,34 @@
+# -*- coding: utf-8 -*-
+from __future__ import unicode_literals
+
+from django.db import models, migrations
+from django.db.models import Q
+
+
+def remove_groupless_prices(apps, schema_editor):
+    model = apps.get_model("simple_pricing", "SimpleProductPrice")
+    objs = model.objects.filter(Q(group_id__isnull=True) | Q(shop_id__isnull=True))
+    if objs.count():
+        print("** Removing %d groupless or shopless SimpleProductPrices **" % objs.count())
+        objs.delete()
+
+class Migration(migrations.Migration):
+
+    dependencies = [
+        ('simple_pricing', '0002_remove_simpleproductprice_includes_tax'),
+    ]
+
+    operations = [
+        migrations.RunPython(remove_groupless_prices, migrations.RunPython.noop),
+
+        migrations.AlterField(
+            model_name='simpleproductprice',
+            name='group',
+            field=models.ForeignKey(to='shoop.ContactGroup'),
+        ),
+        migrations.AlterField(
+            model_name='simpleproductprice',
+            name='shop',
+            field=models.ForeignKey(to='shoop.Shop'),
+        ),
+    ]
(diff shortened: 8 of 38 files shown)


shoop/core/models/products.py
```diff
@@ -258,20 +258,29 @@ def get_cheapest_child_price(self, context, quantity=1):
             for c in self.variation_children.all()
         )[0]
 
-    def get_price(self, context, quantity=1):
+    def get_price_info(self, context, quantity=1):
```
akx: Could we get proper docstrings for `get_price_info`, `get_price` and `get_base_price`?

I mean, I'm not sure about the semantics of the last one, at least :)

shoop/core/pricing/__init__.py
```diff
@@ -90,10 +95,10 @@ def get_cache_key(self):
     cache_key = property(get_cache_key)
 
 
+@six.add_metaclass(abc.ABCMeta)
```
akx: I think we've used the `six.with_metaclass()` base class in the past...

shoop/core/pricing/price_info.py
```diff
@@ -0,0 +1,46 @@
...
+
+
+class PriceInfo(object):
+    def __init__(self, price, base_price, expires_on=None):
+        """
+        Initialize PriceInfo with two given Price objects.
+
+        Given prices can be taxful or taxless, but they must match.
+
+        :param price:
```
akx: Undescribed `price`.

shoop/core/pricing/price_info.py
```diff
@@ -0,0 +1,46 @@
...
+        :type price: shoop.core.pricing.Price
+        :param base_price:
```
akx: Undescribed `base_price`

shoop/simple_pricing/admin_form_part.py
```diff
@@ -38,8 +38,9 @@ def _build_fields(self):
             ((shop_id or 0, group_id or 0), price)
             for (shop_id, group_id, price)
             in SimpleProductPrice.objects.filter(product=self.product)
-            .values_list("shop_id", "group_id", "price")
+            .values_list("shop_id", "group_id", "price") if group_id
```
akx: Shouldn't this filtering be done in a `.filter()`/`.exclude()` clause?

Also, should a migration perhaps clean groupless `SimpleProductPrice`s out so this won't be necessary?

shoop/simple_pricing/admin_form_part.py
```diff
@@ -65,6 +66,7 @@ def _process_single_save(self, shop, group):
         name = self._get_field_name(shop_group_id_tuple)
         value = self.cleaned_data.get(name)
         clear = (value is None or value < 0)
+        print(clear)
```
akx: Stray print.

shoop/simple_pricing/admin_form_part.py
```diff
@@ -78,6 +80,7 @@ def save(self):
             #  but short-circuits out if `has_changed()` returns false.
             #  That, in kind, would cause `self.cleaned_data.get(name)` in `_process_single_save`
             #  to return Nones, clearing out all prices. Oops.)
+            print("No changes")
```
akx: Stray print.

shoop_tests/core/test_order_creator.py
```diff
@@ -57,7 +58,10 @@ def test_order_creator(admin_user):
         require_verification=True,
     ))
 
-    creator = OrderCreator(request=None)
+    request = RequestFactory().get("/")
```
akx:
```python
request = RequestFactory().get("/")
request.shop = get_default_shop()
```

is copypasted into ALL THE PLACES. Any way we could refactor it? (FWIW, there's `apply_request_middleware`, which will install all the usual attributes into a request...)

shoop_tests/front/test_basket.py
```diff
@@ -28,9 +28,14 @@ def test_basket(rf, storage):
...
         product = create_product(printable_gibberish(), shop=shop, supplier=supplier)
-        SimpleProductPrice.objects.get_or_create(
-            shop=shop, product=product, defaults={"price": 50, "includes_tax": False}
-        )
+
+        shop_product = product.get_shop_instance(shop)
+        shop_product.default_price = 50
+        shop_product.save()
+        #
+        # SimpleProductPrice.objects.get_or_create(
```
akx: Dead code?

shoop_tests/front/test_basket.py
```diff
@@ -28,9 +28,14 @@ def test_basket(rf, storage):
     products_and_quantities = []
     for quantity in quantities:
         product = create_product(printable_gibberish(), shop=shop, supplier=supplier)
```
akx: Maybe add `default_price` as a kwarg to `create_product`?

shoop_tests/simple_pricing/test_form.py
```diff
@@ -57,23 +54,31 @@ def test_no_changes_into_form():
 def test_change_shop_price():
     product = _get_test_product()
     shop = get_default_shop()
+    group = get_default_customer_group()
+
+    # cg = get_default_customer_group()
```
akx: Dead code.

shoop_tests/simple_pricing/test_simple_pricing.py
```diff
@@ -15,37 +16,96 @@
...
     request = rf.get("/")
     request.shop = shop
-    product = create_product("Just-A-Product-Too")
-    SimpleProductPrice.objects.create(product=product, shop=None, price=250)
-    SimpleProductPrice.objects.create(product=product, shop=shop, price=199)
+    request.customer = customer
+
+    product = create_product("Just-A-Product-Too", shop)
+
+    price_cls = TaxfulPrice if shop.prices_include_tax else TaxlessPrice
```
akx: Could you add parentheses around the rvalue?

shoop_tests/simple_pricing/test_simple_pricing.py
```diff
@@ -15,37 +16,96 @@
...
 
 
 @pytest.mark.django_db
-def test_set_taxful_price_works():
-    product = create_product("Anuva-Product")
-    spp = SimpleProductPrice(product=product, shop=None)
-    spp.price = 250
-    spp.includes_tax = True
+def test_set_taxful_price_works(rf):
+    shop = get_default_shop()
```
akx: This setup fragment is duplicated... Refactor?

shoop_tests/simple_pricing/test_simple_pricing.py
```diff
@@ -15,37 +16,96 @@
...
-    assert spp.price == 250
-    assert spp.includes_tax == True
+
     spm = SimplePricingModule()
-    pp = spm.get_price(spm.get_context_from_data(), product.pk, quantity=1)
+    pricing_context = spm.get_context_from_request(request)
+    price_info = product.get_price_info(pricing_context, quantity=1)
+
+    assert price_info.price == TaxfulPrice(250)
+    assert price_info.includes_tax == True
```
akx: No need to compare to True.

shoop/core/pricing/__init__.py
```diff
@@ -116,26 +120,9 @@ def get_context_from_request(self, request):
...
-        """
-        Get base Price for the given product.
-
-        May return TaxlessPrice or TaxfulPrice.
-
-        :rtype: Price
-        """
-        return TaxlessPrice(0)
+    @abc.abstractmethod
+    def get_price_info(self, context, product, quantity=1):
```
akx: So yeah, as mentioned, this method should accept a product ID as product as well.

shoop/core/pricing/__init__.py
```diff
@@ -155,16 +142,16 @@ def get_pricing_steps(self, context, product_id):
         """
         return [(0, TaxlessPrice(0))]
 
-    def get_prices(self, context, product_ids, quantity=1):
+    def get_prices(self, context, products, quantity=1):
```
akx: Ditto here; this should also be able to accept a list of product IDs or products.

Pikkupomo:
retest this please

shoop/core/pricing/__init__.py
```diff
@@ -157,13 +148,13 @@ def get_pricing_steps(self, context, product_id):
 
     def get_prices(self, context, product_ids, quantity=1):
```
akx: This API should also accept a list of IDs or a list of products (though the retval should still be as it is, a dict from ID to prices.

shoop/testing/factories.py
```diff
@@ -548,8 +557,12 @@ def create_random_order(customer=None, products=(), completion_probability=0):
     if not customer:
         raise ValueError("No valid contacts")
 
-    request = RequestFactory().get("/")
-    request.customer = customer
+    shop = get_default_shop()
+
+    request = apply_request_middleware(RequestFactory().get("/"),
+                                       shop=get_default_shop(),
```
akx: You shouldn't need to pass `shop` here -- the ShopMiddleware should always use the default shop...

shoop/core/models/products.py
```diff
@@ -258,20 +258,41 @@ def get_cheapest_child_price(self, context, quantity=1):
...
         """
+        Returned `PriceInfo` object contains calculated `price` and `base_price`.
```
akx: This docstring is missing the title line.

shoop/core/pricing/__init__.py
```diff
@@ -6,6 +6,9 @@
 # This source code is licensed under the AGPLv3 license found in the
 # LICENSE file in the root directory of this source tree.
 from __future__ import unicode_literals
+import abc
+import six
+
 """
 Shoop modular product pricing functionality.
```
akx: Is this module docstring up-to-date?

shoop/core/pricing/__init__.py
```diff
@@ -116,26 +120,13 @@ def get_context_from_request(self, request):
...
-        :type context: PricingContext
-        :rtype: Price
+    @abc.abstractmethod
+    def get_price_info(self, context, product, quantity=1):
         """
-        return TaxlessPrice(0)
-
-    def get_base_price(self, product_id):
+        :param product: `Product` object or id of `Product`
+        :type product: `Product` or int
```
akx: Use PyCharm-compatible type annotations:

```shoop.core.models.Product|int```

File: shoop/core/pricing/__init__.py [hunk 22]
```diff
@@ -155,15 +146,18 @@ def get_pricing_steps(self, context, product_id):
...
 
-    def get_prices(self, context, product_ids, quantity=1):
-        return {
-            product_id: self.get_price(context, product_id=product_id, quantity=quantity)
-            for product_id in product_ids
-        }
+    def get_price_infos(self, context, products, quantity=1):
+        """
+        :param products: a list of `Product`s or id's
+        :type products:  list
```
akx: Same here for the type annotation.

shoop/core/pricing/__init__.py
```diff
@@ -155,15 +146,18 @@ def get_pricing_steps(self, context, product_id):
...
+        """
+        if not all(isinstance(item, six.integer_types) for item in products):
```
akx: Hmm. If `products` is a generator, this will cause its (partial) exhaustion, meaning `product_ids` may end up being an empty iterable, which will be a fairly obscure bug to solve.

Pinging dr. @suutari-ai :)

shoop/core/pricing/price_info.py
```diff
@@ -0,0 +1,46 @@
...
+
+from shoop.core.pricing import Price
+
+
+class PriceInfo(object):
+    def __init__(self, price, base_price, expires_on=None):
+        """
+        Initialize PriceInfo with two given Price objects.
+
+        Given prices can be taxful or taxless, but they must match.
```
akx: > but they must match.

should maybe be reworded as

> but they must be of the same type.

or

> but their types must match.

shoop/simple_pricing/admin_form_part.py
```diff
@@ -38,8 +38,9 @@ def _build_fields(self):
             ((shop_id or 0, group_id or 0), price)
             for (shop_id, group_id, price)
             in SimpleProductPrice.objects.filter(product=self.product)
-            .values_list("shop_id", "group_id", "price")
+            .values_list("shop_id", "group_id", "price").exclude(group_id__isnull=True)
```
akx: I like it when `.filter()` and `.exclude()` are grouped logically together and "output options" such as `values_list` comes last.

shoop/simple_pricing/migrations/0003_remove_groupless_prices.py
```diff
@@ -0,0 +1,21 @@
...
+from __future__ import unicode_literals
+
+from django.db import models, migrations
+
+def remove_groupless_prices(apps, schema_editor):
+    model = apps.get_model("simple_pricing", "SimpleProductPrice")
+    objs = model.objects.filter(group_id__isnull=True)
+    if objs.count():
+        print("** Removing %d groupless SimpleProductPrices **" % objs.count())
+        model.objects.filter(group_id__isnull=True).delete()
```
akx: Just use `objs.delete()` :)

shoop/simple_pricing/migrations/0002_remove_simpleproductprice_includes_tax.py
```diff
@@ -0,0 +1,18 @@
+# -*- coding: utf-8 -*-
+from __future__ import unicode_literals
+
+from django.db import models, migrations
+
+
+class Migration(migrations.Migration):
+
+    dependencies = [
+        ('simple_pricing', '0001_initial'),
```
akx: I wonder if this should actually depend on the core migration adding the per-shop taxation field.

Then we could have a data migration that toggles the per-shop taxation field based on the current taxness of the prices for each shop. (Ambiguous cases, where a shop has both taxful and taxless prices, would warrant a warning.) Otherwise we're just, well, dropping taxfulness information here.

shoop/simple_pricing/module.py
```diff
@@ -29,50 +31,52 @@ def get_context_from_request(self, request):
         customer = getattr(request, "customer", None)
         shop = getattr(request, "shop", None)
 
+        assert shop is not None
```
akx: This assertion isn't really necessary, is it?

shoop/simple_pricing/module.py
```diff
@@ -29,50 +31,52 @@ def get_context_from_request(self, request):
...
+
         if not customer or customer.is_anonymous:
             customer_group_ids = []
         else:
             customer_group_ids = sorted(customer.groups.all().values_list("id", flat=True))
 
         return self.pricing_context_class(
-            shop_id=(shop.pk if shop else None),
+            shop=(shop if shop else None),
```
akx: Since `shop` is always required, this ternary operator shouldn't be required anyway.

shoop/simple_pricing/module.py
```diff
@@ -29,50 +31,52 @@ def get_context_from_request(self, request):
...
+            price = default_price
 
+        base_price = price
 
-def _create_price(price, includes_tax):
-    if includes_tax:
-        return TaxfulPrice(price)
-    else:
-        return TaxlessPrice(price)
+        price_cls = TaxfulPrice if includes_tax else TaxlessPrice
```
akx: Can you add parentheses around the rvalue?

shoop/core/pricing/__init__.py
suutari-ai: Yeah, it needs updating. Also: Move it just below the copyright notice, before the imports. That's the correct location for module docstring.

shoop/core/pricing/__init__.py
suutari-ai: I.e.
```
:type products: Iterable[Product|int]
```

shoop/core/pricing/__init__.py
suutari-ai: I'm not a doctor :P

Anyway, you could do something like this
```python
product_ids = [getattr(x, "pk", x) for x in products]
```

And that would allow even mixing Products and product ids to the same iterable.

shoop_tests/simple_pricing/test_simple_pricing.py
```diff
@@ -6,46 +6,138 @@
...
+
+
+@pytest.mark.django_db
+def test_shop_specific_cheapest_price_1(rf):
+    request, shop, group = initialize_test(rf, False)
+
+    product = create_product("Just-A-Product", shop, default_price=200)
+
+    # determine which is the taxfulness
+    price_cls = TaxfulPrice if settings.SHOOP_DEFAULT_PRICES_INCLUDE_TAX else TaxlessPrice
```
suutari-ai: Shouldn't this use the Shop.prices_include_tax instead? (and the setting be removed)

shoop/core/settings.py
```diff
@@ -136,3 +136,6 @@
 
 #: The submission URL for Shoop's telemetry (statistics) system
 SHOOP_TELEMETRY_URL = "https://telemetry.shoop.io/collect/"
+
+#: Defines whether the default prices for products are assumed to be taxful or taxless.
+SHOOP_DEFAULT_PRICES_INCLUDE_TAX = False
```
suutari-ai: This should not be needed, since now you're adding the `Shop.prices_include_tax`.

Don't add this setting.

shoop/simple_pricing/module.py
suutari-ai: @akx: Do you mean that it should rather use just
```python
shop = request.shop
```
?

Though, I would still add an assert, but maybe something like
```python
assert isinstance(shop, shoop.core.models.Shop)
```
would be more useful...

shoop/simple_pricing/models.py
```diff
@@ -18,7 +18,6 @@ class SimpleProductPrice(models.Model):
     shop = models.ForeignKey("shoop.Shop", blank=True, null=True, db_index=True)
     group = models.ForeignKey("shoop.ContactGroup", blank=True, null=True, db_index=True)
```
suutari-ai: Shouldn't shop and group be changed to blank=False, null=False now, since shopless prices are not supported anymore and the groupless price is stored to ShopProduct.default_price?

That would also allow removing a few checks of group is null (like `exclude(group__isnull=True)`) in some other places.

shoop/simple_pricing/admin_form_part.py
suutari-ai: This exclude is unneeded after you make group_id non-nullable in SimpleProductPrice.

shoop/core/pricing/__init__.py
akx: That seems good.

shoop/simple_pricing/models.py
akx: Probably a good idea. (The migration will naturally need to get rid of the currently-null-or-blank SPPs before it edits the schema.)

shoop/simple_pricing/migrations/0003_remove_groupless_prices.py
Pikkupomo: ups!

shoop/core/pricing/apps.py
```diff
@@ -0,0 +1,19 @@
...
+# This file is part of Shoop.
+#
+# Copyright (c) 2012-2015, Shoop Ltd. All rights reserved.
+#
+# This source code is licensed under the AGPLv3 license found in the
+# LICENSE file in the root directory of this source tree.
+from shoop.apps import AppConfig
+
+
+class DefaultPricingAppConfig(AppConfig):
```
akx: I don't think we need a separate app just to provide `default_pricing`... Just move the provide to core's provides?

shoop/core/pricing/__init__.py
Pikkupomo: updated it

Pikkupomo:
retest this please

shoop/testing/factories.py
```diff
@@ -393,7 +394,7 @@ def get_completed_order_status():
     return OrderStatus.objects.get_default_complete()
 
 
-def create_product(sku, shop=None, supplier=None):
+def create_product(sku, shop=None, supplier=None, default_price=None):
```
akx: If this is `default_price=0`, you can just put it in the `ShopProduct.objects.create()` call (as 0 is the default anyway)

shoop/core/models/product_shops.py
```diff
@@ -114,7 +118,7 @@ def get_visibility_errors(self, customer):
 
     # TODO: Refactor get_orderability_errors, it's too complex
     def get_orderability_errors(  # noqa (C901)
-            self, supplier, quantity, customer, ignore_minimum=False):
+                                  self, supplier, quantity, customer, ignore_minimum=False):
```
suutari-ai: unneeded change

shoop/core/models/product_shops.py
```diff
@@ -171,7 +175,7 @@ def get_orderability_errors(  # noqa (C901)
                             'for example %(smaller_p)s %(unit)s.') % {
                     "package_size": purchase_multiple,
                     "smaller_p": smaller_p,
-                    "unit": self.product.sales_unit,
+                    "unit": self.product.sales_unit
```
suutari-ai: unneeded change

shoop/core/models/product_shops.py
```diff
@@ -179,19 +183,19 @@ def get_orderability_errors(  # noqa (C901)
                     "package_size": purchase_multiple,
                     "smaller_p": smaller_p,
                     "larger_p": larger_p,
-                    "unit": self.product.sales_unit,
+                    "unit": self.product.sales_unit
```
suutari-ai: unneeded change (and the final comma is good to be there)

shoop/core/models/product_shops.py
```diff
@@ -179,19 +183,19 @@ def get_orderability_errors(  # noqa (C901)
...
                 }
             yield ValidationError(message, code="invalid_purchase_multiple")
 
         for receiver, response in get_orderability_errors.send(
-            ShopProduct, shop_product=self, customer=customer, supplier=supplier, quantity=quantity
+                ShopProduct, shop_product=self, customer=customer, supplier=supplier, quantity=quantity
```
suutari-ai: unneeded change (and the indentation should be 4, not 8)

shoop/core/models/product_shops.py
```diff
@@ -179,19 +183,19 @@ def get_orderability_errors(  # noqa (C901)
...
         ):
             for error in response:
                 yield error
 
     def raise_if_not_orderable(self, supplier, customer, quantity, ignore_minimum=False):
         for message in self.get_orderability_errors(
-            supplier=supplier, quantity=quantity, customer=customer, ignore_minimum=ignore_minimum
+                supplier=supplier, quantity=quantity, customer=customer, ignore_minimum=ignore_minimum
```
suutari-ai: unneeded change (and the indentation should be 4, not 8)

akx:
Maybe title the first commit

> Remove default from `MoneyField`

instead of "Rework" ...?

shoop/core/shortcuts/__init__.py
```diff
@@ -18,7 +18,6 @@ def update_order_line_from_product(request, order_line, product, quantity=1, sup
         raise Exception("set_from_product may not be used on saved lines")
 
     if not product:  # pragma: no cover
-
```
akx: ?

shoop/simple_pricing/migrations/0002_remove_simpleproductprice_includes_tax.py
akx: :arrow_up:

shoop/testing/factories.py
```diff
@@ -411,8 +412,11 @@ def create_product(sku, shop=None, supplier=None):
     product.save()
     if shop:
         sp = ShopProduct.objects.create(product=product, shop=shop)
+        if default_price:
```
akx: This should probably be `if default_price is not None:`, so users could set zero here

akx: Come to think of it, you can just simply put this in the `.create()` call. If it's `None`, it's NULL; if it's some other value, it just works.

shoop_tests/core/test_order_creator.py
```diff
@@ -76,13 +80,16 @@ def test_order_creator_supplierless_product_line_conversion_should_fail(admin_us
         quantity=1,
         unit_price=TaxlessPrice(10),
     ))
-    creator = OrderCreator(request=None)
+
+    request = apply_request_middleware(RequestFactory().get("/"))
```
akx: In top-level test functions, you can use [the `rf` fixture](https://pytest-django.readthedocs.org/en/latest/helpers.html#rf-requestfactory) instead of having to import `RequestFactory` yourself:

```python
def test_order_creator(admin_user):
    request = blah(RequestFactory().blah)
```
-->
```python
def test_order_creator(admin_user, rf):
    request = blah(rf.blah)
```

shoop_workbench/settings/base_settings.py
```diff
@@ -30,7 +30,7 @@
     'filer',
     'easy_thumbnails',
     'shoop.core',
-    'shoop.simple_pricing',
+    'shoop.core.pricing',
```
akx: Still not an app anymore!

Pikkupomo: for some very strange reason this has re-appeared here, not sure why :) Will remove \o7

shoop/core/pricing/__init__.py
```diff
@@ -155,15 +149,15 @@ def get_pricing_steps(self, context, product_id):
...
-            product_id: self.get_price(context, product_id=product_id, quantity=quantity)
-            for product_id in product_ids
-        }
-
-    def get_base_prices(self, product_ids):
+    def get_price_infos(self, context, products, quantity=1):
+        """
+        :param products: a list of `Product`s or id's
+        :type products:  iterable[shoop.core.models.Product|int]
+        :rtype: iterable[PriceInfo]
```
suutari-ai: It's `Iterable` with a capital "I" and rtype should be `dict[int,PriceInfo]`.

shoop/simple_pricing/migrations/0002_remove_simpleproductprice_includes_tax.py
suutari-ai: Good idea, but if no such data migration is created, this does not matter. The SimpleProductPrice.includes_tax was not visible in the UI so I don't think many have set it manually. Therefore we could as well use the same default value for Shop.prices_include_tax and even if that turns out to be wrong for some case, it's usually not a big deal to manually fix couple booleans from the Admin (you're not having millions of Shops in a single Shoop instance -- hopefully).

suutari-ai: I.e. no need for changes to this IMHO. I don't think adding a data migration has much value.

shoop/simple_pricing/models.py
```diff
@@ -39,5 +38,5 @@ def __repr__(self):
             self.shop_id,
             self.group_id,
             self.price,
-            self.includes_tax,
+            self.shop.prices_include_tax
```
suutari-ai: Hmm.. including properties of the shop to repr does not seem appropriate.

Pikkupomo: ye, just wanted to keep the old functionality, removed.

shoop_tests/core/test_default_pricing.py
```diff
@@ -0,0 +1,72 @@
+# -*- coding: utf-8 -*-
```
akx: This file is added in the wrong commit.

shoop_tests/core/test_basic_order.py
```diff
@@ -69,9 +70,13 @@ def create_order(request, creator, customer, product):
...
 @pytest.mark.django_db
 def test_basic_order(rf, admin_user):
+    shop = get_default_shop()
+    group = get_default_customer_group()
+
     request = rf.get('/')
+    request.shop = shop
     product = get_default_product()
-    SimpleProductPrice.objects.create(product=product, group=None, price=Decimal("100.00"), includes_tax=False)
+    #SimpleProductPrice.objects.create(product=product, group=group, shop=shop, price=Decimal("100.00"))
```
akx: Why the comment-out?

akx:
10 days and 92 comments later, I think we're about ready to merge this, after Monsieur Jenkins gives his green light.

PR #76, opened 2015-08-24
Prepare for Shoop 1.2 release (SHOOP-1107)

suutari-ai:
* Update Change Log.

* Add couple missing files to admin/generated_resources.txt, which were
  found out while checking that it builds fine as a wheel.


tulimaki:
:+1:

PR #78, opened 2015-08-25
Admin: Add basic Supplier views

tulimaki:
Refs SHOOP-1059

### DIFF
--- a/shoop/admin/modules/suppliers/__init__.py
+++ b/shoop/admin/modules/suppliers/__init__.py
@@ -1,0 +1,40 @@
+# -*- coding: utf-8 -*-
+# This file is part of Shoop.
+#
+# Copyright (c) 2012-2015, Shoop Ltd. All rights reserved.
+#
+# This source code is licensed under the AGPLv3 license found in the
+# LICENSE file in the root directory of this source tree.
+from __future__ import unicode_literals
+
+from django.utils.translation import ugettext_lazy as _
+
+from shoop.admin.base import AdminModule, MenuEntry
+from shoop.admin.utils.urls import derive_model_url, get_edit_and_list_urls
+from shoop.core.models import Supplier
+
+
+class SupplierModule(AdminModule):
+    name = _("Suppliers")
+    breadcrumbs_menu_entry = MenuEntry(name, url="shoop_admin:suppliers.list")
+
+    def get_urls(self):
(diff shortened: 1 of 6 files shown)


shoop/admin/modules/suppliers/__init__.py
```diff
@@ -0,0 +1,40 @@
...
+            view_template="shoop.admin.modules.suppliers.views.Supplier%sView",
+            name_template="suppliers.%s"
+        )
+
+    def get_menu_entries(self, request):
+        category = _("Products")
+        return [
+            MenuEntry(
+                text=_("Suppliers"),
+                icon="fa fa-truk",
```
akx: `truck`, I assume? :smile:

shoop/admin/modules/suppliers/views/edit.py
```diff
@@ -0,0 +1,34 @@
...
+from shoop.core.models import Supplier
+
+
+class SupplierForm(forms.ModelForm):
+    class Meta:
+        model = Supplier
+        exclude = ("module_data",)
+        widgets = {
+            "module_identifier": forms.Select(
+                choices=model.get_module_choices(
```
akx: Mmmmm... `get_module_choices` can't necessarily be safely called at import time (it may `load()` something with circular dependencies).

Move the choices assignment to `get_form()` in the view instead, please :) It doesn't need to be anything more complex than

```python
def get_form(self, form_class=None):
    form = super(SupplierEditView, self).get_form(form_class=form_class)
    choices = self.model.get_module_choices(...)
    form["module_identifier"].choices = form["module_identifier"].widget.choices = choices
    return form
````

File: shoop/admin/modules/suppliers/views/list.py [hunk 3]
```diff
@@ -0,0 +1,35 @@
...
+            _(u"Name"),
+            sort_field="name",
+            display="name",
+            filter_config=TextFilter(
+                filter_field="name",
+                placeholder=_("Filter by name...")
+            )
+        ),
+        Column("type", _(u"Type")),
+        Column("module_identifier", _(u"Module identifier"), display="get_module_display", sortable=False)
```
akx: The title of this column could just be `Module`, as you're actually showing the module name anyway.

akx:
Good stuff! Thank you!  :bow:

PR #80, opened 2015-08-26
workbench: Allow overriding couple settings from env

suutari-ai:
Allow overriding BASE_DIR and SHOOP_ENABLED_ADDONS_FILE from the
environment to make it easier to customize deployment of shoop_workbench
for QA environments and such.


akx:
Looks harmless AND useful! Good combo.

PR #82, opened 2015-08-26
Front: Add ordering for cross sells template helper

tulimaki:
Add ordering by weight to product cross sells template helper.
Use same ordering as in admin where lowest weight is shown first.

Refs SHOOP-1064


akx:
This duplicates work done in #62... :\

tulimaki:
Since https://github.com/shoopio/shoop/pull/62 contains better solution for this

suutari-ai:
I'd rather see this bug fixed separately and the cache thing being discussed by itself. IMHO this fix is good.

PR #83, opened 2015-08-27
Translated fields to tabs

tulimaki:
Add more tools to render language dependent content
Use these tools to render translated fields for attributes, product attributes, contact groups, methods and shops correctly.

Refs SHOOP-1049


akx:
Nice and clean! Thank you!

PR #84, opened 2015-08-27
Xtheme Core

akx:
This is an initial version of the Shoop extensible theme engine.

The grand goal is to support "widget"/"portlet" style flexible editing of frontend themes while still allowing creative control and (a semblance of) familiarity for theme authors (the designery and developery types).

Some of the content in this PR is most definitely still in flux and may/will change in future PRs, but I figured there's more than enough to review as-is...


akx:
retest this please

Pikkupomo:
This looks ok in my opinion. when @suutari-ai has time, he could recheck this beast.

tulimaki:
What "Make DocCov saner"-commit has to do with xthemes? If we have to deny this pr we don't get that fix?

akx:
@tulimaki: It doesn't, other than that I needed a better-behaving DocCov when documenting Xtheme.

If the Xtheme-related things are considered no go, I can cut the general improvements (the three first commits here) into a separate PR. I always try to reorganize general improvement commits like that to be at the start of any given PR to make that easier. There're similar general improvements at the start of the forthcoming xtheme-editor PR.

akx:
@tulimaki: I solemnly swear <del>that I am up to no good</del> that I'll improve the docstrings in further commits I have the time for it. Good catches though :)

akx:
https://media.giphy.com/media/hTh9bSbUPWMWk/giphy.gif

PR #85, opened 2015-08-27
Core: Add non-ASCII support for supplier name

tulimaki:
Based on test test_module_interface_for_scandinavian_letters add force ascii to shoop.core.modules.interface load. Also based on same test change supplier model to inherit ShoopModel instead of models.Model.

Since shoop.core.utils.text.force_ascii is already implemented in shoop.utils.text this is removed. Migrate it usages to shoop.utils.text.force_ascii.

Refs SHOOP-1059


akx:
Hmm.

I think it'd be better to:

* remove `shoop.core.utils.text.force_ascii` (migrate its usages to `shoop.utils.text.force_ascii`, which already exists)
* use `shoop.utils.text.force_ascii` in `load()` (where a possibly-unicode string may be passed in as `context_explanation`
* add a test case for this failure ;)

shoop/core/models/suppliers.py
```diff
@@ -33,7 +34,7 @@ class Supplier(ModuleInterface, models.Model):
     module_data = JSONField(blank=True, null=True)
 
     def __str__(self):
-        return self.name
+        return ascii_safe_repr(self.name)
```
suutari-ai: I think this is wrong place for the fix and the fix is wrong.

Since the error you're trying to fix happens when ModuleInterface._load_module tries to do repr for the module, IMHO the correct fix is to use ascii_safe_repr there (i.e. in shoop/core/modules/interface.py).

Also str != repr: Don't make string presentation of the module be ascii only, since that is shown in Admin UIs etc. Instead, if you want to make the repr behave nicely, derive the Supplier from ShoopModel rather than models.Model. (`from ._base import ShoopModel`)

suutari-ai: Btw. this is also the reasons tests fail:
```
>           assert str(supplier) == DEFAULT_NAME
E           assert "u'Default'" == 'Default'
```

shoop/core/modules/interface.py
```diff
@@ -31,10 +31,10 @@ def _load_module(self):
         if module_identifier:
             impls = self.get_module_implementation_map()
             if module_identifier not in impls:
-                raise ModuleNotFound("Invalid module identifier %r in %r" % (module_identifier, self))
+                raise ModuleNotFound("Invalid module identifier %r in %r" % (module_identifier, force_ascii(self)))
```
akx: This should now be

```python
raise ModuleNotFound("Invalid module identifier %r in %s" % (module_identifier, force_ascii(repr(self))))
```

to retain the old behavior. (Otherwise this would be `force_ascii()`ing the `str()` representation of the object, not the `repr`).

shoop/core/modules/interface.py
```diff
@@ -31,10 +31,10 @@ def _load_module(self):
...
             spec = impls[module_identifier]
 
-        cls = load(spec, context_explanation="Loading module for %r" % self)
+        cls = load(spec, context_explanation="Loading module for %r" % force_ascii(self))
```
akx: Same here :)

shoop_tests/core/test_module_interface.py
```diff
@@ -0,0 +1,26 @@
...
+from __future__ import unicode_literals
+import pytest
+
+from shoop.core.models import Supplier
+from shoop.testing.factories import get_default_supplier
+
+
+@pytest.mark.django_db
+def test_module_interface_for_scandinavian_letters(rf):
+    supplier = get_default_supplier()
```
akx: This test shouldn't modify the default supplier. (It's fine to create a new supplier for this test -- and if you want, you can even `.delete()` it at the end of this test to clean things up)

shoop/core/modules/interface.py
```diff
@@ -31,10 +32,12 @@ def _load_module(self):
         if module_identifier:
             impls = self.get_module_implementation_map()
             if module_identifier not in impls:
-                raise ModuleNotFound("Invalid module identifier %r in %r" % (module_identifier, self))
+                raise ModuleNotFound(
+                    "Invalid module identifier %r in %r" % (module_identifier, force_ascii(repr(self)))
```
akx: This needs to be `"... %r in %s"`, otherwise you're getting double repring.

shoop/core/modules/interface.py
```diff
@@ -31,10 +32,12 @@ def _load_module(self):
...
+                )
             spec = impls[module_identifier]
 
-        cls = load(spec, context_explanation="Loading module for %r" % self)
+        cls = load(spec, context_explanation="Loading module for %r" % force_ascii(repr(self)))
```
akx: This needs to be `"for %s"`, otherwise you're getting double repring.

akx:
Nice! :clap:

PR #86, opened 2015-09-01
Change empty label for admin supplier module edit

tulimaki:
Refs SHOOP-1059


akx:
Hmm. `instance.module.name` should never be falsy. If it is, it means

* the instance is somehow using `BaseSupplierModule` (which returns `0` for all stock counts and is thus sort of useless)
* the instance is using another supplier module that has no proper `name = ...` attribute set (which is a bug on the module programmer's behalf).

tulimaki:
Changed empty label for editview, since empty label selection means that module_identifier is empty and there is no module to handle stocks etc.

PR #88, opened 2015-09-08
Fix issue with price not being saved

Pikkupomo:
Refs SHOOP-514

### DIFF
--- a/shoop/simple_pricing/admin_form_part.py
+++ b/shoop/simple_pricing/admin_form_part.py
@@ -69,9 +69,12 @@ class SimplePricingForm(forms.Form):
         if clear:
             SimpleProductPrice.objects.filter(product=self.product, group=group, shop=shop).delete()
         else:
-            spp, _ = SimpleProductPrice.objects.get_or_create(product=self.product, group=group, shop=shop)
-            spp.price = value
-            spp.save()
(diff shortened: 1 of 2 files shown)


akx:
Could we also get a test case that proves this bug was fixed? :)

shoop/simple_pricing/admin_form_part.py
```diff
@@ -69,7 +69,7 @@ def _process_single_save(self, shop, group):
         if clear:
             SimpleProductPrice.objects.filter(product=self.product, group=group, shop=shop).delete()
         else:
-            spp, _ = SimpleProductPrice.objects.get_or_create(product=self.product, group=group, shop=shop)
+            spp, _ = SimpleProductPrice.objects.get_or_create(product=self.product, group=group, shop=shop, price=value)
             spp.price = value
             spp.save()
```
suutari-ai: Is this really correct? Why the price is assigned twice? Shouldn't the get_or_create just specify the key fields (or if it specifies data fields too, then with the `default` dictionary)? And since the price is set and saved below, this seems redundant anyway.

akx: Good points. :) This commit would probably actually break things more than fix them -- which makes it curious that Jenkins thinks everything is hunky dory.

PR #89, opened 2015-09-08
Xtheme: Fix too-early import of `xtheme.models`

akx:


suutari-ai:
Missing refs.

akx:
Ref added.

PR #92, opened 2015-09-09
Admin: Optionally disable creating shops

tulimaki:
Refs SHOOP-1118

### DIFF
--- a/shoop/admin/modules/shops/views/edit.py
+++ b/shoop/admin/modules/shops/views/edit.py
@@ -5,13 +5,16 @@
 #
 # This source code is licensed under the AGPLv3 license found in the
 # LICENSE file in the root directory of this source tree.
-
 from __future__ import unicode_literals
 
+from django.conf import settings
 from django.forms.models import modelform_factory
+from django.utils.translation import ugettext_lazy as _
 
 from shoop.admin.utils.views import CreateOrUpdateView
+from shoop.admin.toolbar import get_default_edit_toolbar
 from shoop.core.models import Shop
+from shoop.utils.excs import Problem
 from shoop.utils.multilanguage_model_form import MultiLanguageModelForm
 
 
@@ -20,3 +23,16 @@ class ShopEditView(CreateOrUpdateView):
     form_class = modelform_factory(Shop, MultiLanguageModelForm, exclude=("owner", "options"))
     template_name = "shoop/admin/shops/edit.jinja"
     context_object_name = "shop"
+
+    def get_object(self, queryset=None):
+        obj = super(ShopEditView, self).get_object(queryset)
+        if not settings.SHOOP_ENABLE_MULTIPLE_SHOPS:
+            if not obj.pk and Shop.objects.count() >= 1:
+                # This seems like a new object, but there already is (at least) one shop...
+                raise Problem(_(u"Only one shop permitted."))
+
+        return obj
+
+    def get_toolbar(self):
+        save_form_id = self.get_save_form_id()
+        return get_default_edit_toolbar(self, save_form_id, with_split_save=settings.SHOOP_ENABLE_MULTIPLE_SHOPS)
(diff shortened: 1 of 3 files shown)


shoop/admin/modules/shops/views/list.py
```diff
@@ -25,3 +27,6 @@ class ShopListView(PicotableListView):
         Column("identifier", _(u"Identifier")),
         Column("status", _(u"Status"), filter_config=ChoicesFilter(choices=ShopStatus.choices)),
     ]
+
+    def get_toolbar(self):
+        return super(ShopListView, self).get_toolbar() if settings.SHOOP_ENABLE_MULTIPLE_SHOPS else Toolbar([])
```
suutari-ai: Use real if rather than ternary-operator for this, i.e.:
```python
if settings.SHOOP_ENABLE_MULTIPLE_SHOPS:
    super(ShopListView, self).get_toolbar()
else:
    return Toolbar([])
```
because IMHO that is easier to read.

shoop/admin/modules/shops/views/edit.py
```diff
@@ -20,3 +24,14 @@ class ShopEditView(CreateOrUpdateView):
     form_class = modelform_factory(Shop, MultiLanguageModelForm, exclude=("owner", "options"))
     template_name = "shoop/admin/shops/edit.jinja"
     context_object_name = "shop"
+
+    def get_object(self, queryset=None):
+        if not settings.SHOOP_ENABLE_MULTIPLE_SHOPS and (Shop.objects.count() > 1 or not self.kwargs["pk"]):
+            raise Problem(_(u"Only one shop permitted."))
```
suutari-ai: I would split this if to two separate iffs from the "and" to ease the readability.

akx: And maybe move it after the super call for more readability:

```python
def get_object(self, queryset=None):
    obj = super(ShopEditView, self).get_object(queryset)
    if not settings.SHOOP_ENABLE_MULTIPLE_SHOPS:
        if not obj.pk and Shop.objects.count() >= 1:
            # This seems like a new object, but there already is (at least) one shop...
            raise Problem(_(u"Only one shop permitted."))
    return obj
```

shoop/admin/modules/shops/views/edit.py
```diff
@@ -20,3 +24,14 @@ class ShopEditView(CreateOrUpdateView):
...
+
+        return super(ShopEditView, self).get_object(queryset)
+
+    def get_toolbar(self):
+        save_form_id = self.get_save_form_id()
+        if save_form_id:
```
akx: In this case, you don't need this `if`; you'll _always_ know this view has a save_form_id and thus should have a toolbar.

shoop/admin/modules/shops/views/list.py
akx: Though with the `return` missing from @suutari-ai 's example added ;)

shoop_tests/admin/test_multishop_setting.py
```diff
@@ -0,0 +1,34 @@
...
+def test_multishop_edit_view(rf, admin_user):
+    get_default_shop()
+
+    request = apply_request_middleware(rf.get("/"), user=admin_user)
+    view = ShopEditView(request=request, kwargs={"pk": None}) 
+
+    with override_settings(SHOOP_ENABLE_MULTIPLE_SHOPS=False):
+        raised_problem = False
+        try:
+            view.get_object()  # Now view object should throw Problem
```
akx: Use the `pytest.raises` context manager :)

```python
with pytest.raises(Problem):
    view.get_object()   # Now view object should throw Problem
```

The context manager will deal with failing the test if the given exception is raised within the block.

akx:
Yay! Looking good. :eyes: 

:+1:

shoop/admin/modules/shops/views/list.py
suutari-ai: oops :)

PR #93, opened 2015-09-09
Add `Product Media` tab to product editing

Pikkupomo:
Check commits for more information.

Refs SHOOP-1045

### DIFF
--- a/shoop/admin/modules/products/views/forms.py
+++ b/shoop/admin/modules/products/views/forms.py
@@ -153,7 +153,7 @@ class ProductAttributesForm(forms.Form):
         self.product.clear_attribute_cache()
 
 
-class ProductMediaForm(MultiLanguageModelForm):
+class BaseProductMediaForm(MultiLanguageModelForm):
     class Meta:
         model = ProductMedia
         fields = (
@@ -172,7 +172,7 @@ class ProductMediaForm(MultiLanguageModelForm):
         self.product = kwargs.pop("product")
         self.allowed_media_kinds = kwargs.pop("allowed_media_kinds")
         default_shop = kwargs.pop("default_shop")
-        super(ProductMediaForm, self).__init__(**kwargs)
+        super(BaseProductMediaForm, self).__init__(**kwargs)
 
         self.fields["file"].widget = MediaChoiceWidget()  # Filer misimplemented the field; we need to do this manually.
         self.fields["file"].required = True
@@ -186,7 +186,18 @@ class ProductMediaForm(MultiLanguageModelForm):
             self.fields["shops"].initial = [default_shop]
 
         self.file_url = self.instance.url
-        self.thumbnail = None
+
+        if self.instance.pk and self.instance.file:
+            if isinstance(self.instance.file, Image):
+                thumbnail = self.instance.easy_thumbnails_thumbnailer.get_thumbnail({
+                    'size': (64, 64),
+                    'crop': True,
+                    'upscale': True,
+                })
+                self.file_url = self.instance.url
+                self.thumbnail = thumbnail.url or None
+        else:
+            self.thumbnail = None
 
     def pre_master_save(self, instance):
         instance.product = self.product
@@ -225,24 +236,34 @@ class BaseProductMediaFormSet(BaseModelFormSet):
         return self.form_class(**kwargs)
 
 
+class ProductMediaForm(BaseProductMediaForm):
+
+    def __init__(self, **kwargs):
+        super(ProductMediaForm, self).__init__(**kwargs)
+        self.fields["file"].required = False
+
+    def clean_external_url(self):
+        external_url = self.cleaned_data.get("external_url")
+
+        if "DELETE" in self.changed_data:
+            return external_url
+
+        file = self.cleaned_data.get("file")
+        if external_url and file:
+            raise ValidationError(_("Use only URL or file, not both"))
+        return external_url
+
+
 class ProductMediaFormSet(BaseProductMediaFormSet):
     form_class = ProductMediaForm
 
 
-class ProductImageMediaForm(ProductMediaForm):
+class ProductImageMediaForm(BaseProductMediaForm):
     is_primary = forms.BooleanField(required=False)
 
     def __init__(self, **kwargs):
         super(ProductImageMediaForm, self).__init__(**kwargs)
         if self.instance.pk and self.instance.file:
-            if isinstance(self.instance.file, Image):
-                thumbnail = self.instance.easy_thumbnails_thumbnailer.get_thumbnail({
-                    'size': (64, 64),
-                    'crop': True,
-                    'upscale': True,
-                })
-                self.file_url = self.instance.url
-                self.thumbnail = thumbnail.url or None
             if self.product.primary_image_id == self.instance.pk:
                 self.fields["is_primary"].initial = True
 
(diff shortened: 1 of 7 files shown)


shoop/admin/modules/products/views/forms.py
```diff
@@ -263,3 +284,23 @@ def save(self, commit=True):
 class ProductImageMediaFormSet(ProductMediaFormSet):
     allowed_media_kinds = [ProductMediaKind.IMAGE]
     form_class = ProductImageMediaForm
+
+    def save(self, commit=True):
+        """
+        Add first saved image as  primary image for this product by default if none is selected as such
```
suutari-ai: The docstring is misleading, since it does not have a title and therefore reader gets the idea that the "adding first saved image..." is the only thing that is being done, which is not true.

Add a docstring title and wrap it to 79 chars/line (as our style guide sais about docstring formatting).

e.g.
```python
def save(self, commit=True):
    """
    Save the form.

    In addition add the first saved image as primary image for the
    product if none is selected as such.
    """
```

shoop/admin/modules/products/views/forms.py
```diff
@@ -263,3 +284,23 @@ def save(self, commit=True):
...
+        super(ProductImageMediaFormSet, self).save(commit)
+
+        if not self.forms:
+            return
+
+        has_primary = False
+        for form in self.forms:
+            if form.cleaned_data.get("is_primary"):
+                has_primary = True
+                break
```
suutari-ai: Too much detail for such a simple test. You could replace the lines 294-301 (from the `if not self.forms` to this line) with just
```python
has_primary = any(form.cleaned_data.get("is_primary") for form in self.forms)
```
Or if the `self.forms` could be None use `for form in (self.forms or [])` in that.

suutari-ai: Though then the test below would need to be amended to "if self.forms and not has_primary", so the 294-295 if reasonable.

suutari-ai:
Looks good to me except for the couple comments. May be merged (or I will merge) when those are fixed.

shoop/admin/modules/products/views/forms.py
```diff
@@ -186,7 +186,18 @@ def __init__(self, **kwargs):
             self.fields["shops"].initial = [default_shop]
 
         self.file_url = self.instance.url
-        self.thumbnail = None
+
+        if self.instance.pk and self.instance.file:
+            if isinstance(self.instance.file, Image):
+                thumbnail = self.instance.easy_thumbnails_thumbnailer.get_thumbnail({
```
akx: This should be wrapped in some sort of `try: except`; `get_thumbnail` can and will throw when it just can't deal with things.

shoop/admin/modules/products/views/forms.py
```diff
@@ -225,24 +236,34 @@ def form(self, **kwargs):
...
+class ProductMediaForm(BaseProductMediaForm):
+
+    def __init__(self, **kwargs):
+        super(ProductMediaForm, self).__init__(**kwargs)
+        self.fields["file"].required = False
+
+    def clean_external_url(self):
+        external_url = self.cleaned_data.get("external_url")
+
+        if "DELETE" in self.changed_data:
```
akx: What's the point of this? Not quite obvious from the code.

shoop/admin/modules/products/views/forms.py
Pikkupomo: @akx has always said that there is no point documenting the obvious. I thought that mentioning that `save` actually saves the form goes into that category :)

Agree though that the title is missing and was thinking that when I made the docstring. Will fix :)

shoop/admin/modules/products/views/forms.py
```diff
@@ -186,7 +186,20 @@ def __init__(self, **kwargs):
...
+            try:
+                thumbnail = self.instance.easy_thumbnails_thumbnailer.get_thumbnail({
+                    'size': (64, 64),
+                    'crop': True,
+                    'upscale': True,
+                })
+                self.file_url = self.instance.url
+                self.thumbnail = thumbnail.url
+            except Exception:
+                self.thumbnail = None
```
suutari-ai: The file_url is set in two places, the former one in the if is not needed.

Move the try-except to a free helper function to make this code cleaner
```python
class BaseProductMediaForm(...):
    ...
    def __init__(...):
        self.thumbnail = _try_get_thumbnail_url(self.instance)
...

def _try_get_thumbnail_url(media):
    """
    Get thumbnail URL of product media object if it is an image.

    :type media: ProductMedia  ## Is this correct?
    """
    if media.pk and isinstance(media.file, Image):
        try:
            return media.easy_thumbnails_thumbnailer.get_thumbnail(thumbnail_params)
        except WhatExactExceptionTheThumbnailerWillThrow:  # Don't use bare Exception
            pass
    return None
```

suutari-ai: Though, now that I think about it... maybe the _try_get_thumbnail_url free function should be a instance method of ProductMedia?

shoop/core/models/product_media.py
```diff
@@ -40,7 +40,10 @@ class ProductMedia(TranslatableModel):
         ProductMediaKind, db_index=True, default=ProductMediaKind.GENERIC_FILE, verbose_name=_('kind')
     )
     file = FilerFileField(blank=True, null=True, verbose_name=_('file'))
-    external_url = models.URLField(blank=True, null=True, verbose_name=u'URL')
+    external_url = models.URLField(
+        blank=True, null=True, verbose_name=u'URL',
+        help_text=_("Enter URL to external file. If this field is filled, the selected media doesn't apply.")
+    )
```
suutari-ai: This help text and the validation that both url and file are not set is very good improvement. Good work! :)

shoop/admin/modules/products/views/forms.py
```diff
@@ -186,7 +186,15 @@ def __init__(self, **kwargs):
...
-        self.thumbnail = None
+
+        if self.instance and self.instance.file:
+            try:
+                thumbnail = self.instance.get_thumbnail()
+                self.thumbnail = thumbnail.url
+            except Exception:
+                self.thumbnail = None  # no thumbnail available for whatever reason
+        else:
+            self.thumbnail = None
```
suutari-ai: 1. All this iffing and try-execpting should be in the ProductMedia.get_thumbnail, since it is ProductMedia's own business to know when it can't be thumbnailed or how the thumbnailing can fail (and return None in that case).
2. The catch is still too broad. Find out the exact exceptions easy_thumbnails_thumbnailer.get_thumbnail will throw when it fails and catch only those, not every Exception.
3. You're using self.instance.url above so the `if sefl.instance` here is probably not needed (or it should be done in the `self.file_url` too?).
Also,

Pikkupomo: 1. Originally this was my intention, somehow @akx turned my head to make this kind of solution. It's probably not what @akx was after, but he said `ProductMedia.get_thumbnail` should raise when errors occur, now it does. 
2. Agree catch is too broad, and I need to find out what are the exact Exceptions. It has been catched this way in prior implementations inside Shoop.
3. `file_url` is the url to original file, thumbnail url is for the thumbnail, if that was something you were after here, not sure I completely understand :)

akx:
Looks spiffy to me!

:shipit:

PR #96, opened 2015-09-10
Add missing migrations

tulimaki:
Core: Add migration to drop default values from moneyfields
Notify: Add migration to make notification identifier not unique
Simple pricing: Add migration to drop default from price


akx:
LGTM, but requesting another pair of :eyes: too. @suutari-ai ?

suutari-ai:
Yeah, it's good. Merged and tested this. Thanks @tulimaki!

PR #98, opened 2015-09-10
Xtheme: add "extra view" dispatcher

akx:
Refs SHOOP-1177


shoop/xtheme/views/extra.py
```diff
@@ -0,0 +1,69 @@
...
+
+def _get_view_by_name(theme, view_name):
+    view = getattr(theme, "views", {}).get(view_name)
+    if isinstance(view, six.string_types):  # Handle loading views from specs
+        try:
+            view = load(view)
+        except ImproperlyConfigured:
+            if settings.DEBUG:
+                raise
+            return None
```
suutari-ai: Why hide this error if not debug?

akx: It'll end up being a 404, which feels better for production.

suutari-ai: Ah, I see.. it's a parameter passed in from the URL...

suutari-ai: Wait.. no.. it's not.. it's already seeked from the theme's `views` dict (which btw. should be `{}` by default to avoid getattr here).

Can't we just let the theme have the views as non-lazy (i.e. not as strings)?

suutari-ai: If theme's views dict has the views as objects rather than strings, it makes it easier to navigate from the theme class to the views by go-to-definition feature of IDE, since that works better for objects than for strings. That's also where Django is heading with its urls stuff: https://docs.djangoproject.com/en/dev/releases/1.8/#django-conf-urls-patterns

suutari-ai: And if you're woried about the lazyness vs. import ordering when the views contains objects instead of strings, maybe using a method rather than attribute for this, i.e. `views = Theme.get_views()` (would return a dict from view_name to view object).

akx: I'm thinking about this from the viewpoint of making things easier for theme devs who may not be all that Python savvy. In that sense it's nicer to also allow load-by-spec.

suutari-ai: The `get_views` is an advanced feature anyway; you'd have to write the view too and that means, you'd know something about Python.

No matter how this is implemented, the only way to make it accessible for average theme designer, is to have good documentation that tells how to create such custom views. It's just copy-paste from the documentation being it a dict of strings or a method returning a dict of objects with couple import statements.

akx: I'd just also kinda love that `Theme` objects remain "declarative" for all usual use cases.

Aside from possibly harder navigation, what downsides do you see for load-by-spec?

suutari-ai: * It's making the `_get_view_by_name` more complicated than it could be.
* It's hiding dependencies from all code analysis / refactoring tools.
* No autocomplete by IDEs. (as said in the django link I provided earlier)

Yeah, declarative is cool, but since it already is in a *.py file, why not take advantage of the full power of the Python language rather than parsing a class content like it was some sort of XML or something?

You could btw. also provide a Theme base class that does make things easier for dummies:
```python
class XmlDeclaredTheme(Theme):
    def get_views(self):
        return {
            spec: self._load_view_by_spec(spec)
            for (view_name, spec) in self._get_views_from_xml_file()
        }
```
and then just instruct the dummies to dump their view stuff to the XML file.

akx: Fine, fine. I'll make it `get_views()`. :)

akx:
I went one further, @suutari-ai, and made it `get_view` for ultimate power!

suutari-ai:
Very nice! Thanks!

PR #102, opened 2015-09-11
Add new fields to shop core

tulimaki:
Core: Add logo and translatable public name for Shop
Front: Use shop logo and public name in templates

Refs SHOOP-1178


akx:
Looks nice. Thanks!

PR #109, opened 2015-09-14
Add `ImageChoiceWidget`

Pikkupomo:
Also some minor plumbing to allow testing this feature without ui part.

Refs SHOOP-1278

### DIFF
--- a/shoop/admin/modules/media/views.py
+++ b/shoop/admin/modules/media/views.py
@@ -16,7 +16,9 @@ from django.utils.encoding import force_text
 from django.utils.translation import ugettext as _, ugettext_lazy as _l
 from django.views.generic import TemplateView
 from filer.models import File, Folder
+from filer.models.imagemodels import Image
 from mptt.templatetags.mptt_tags import cache_tree_children
+
 from shoop.admin.modules.media.utils import delete_folder
 from shoop.utils.excs import Problem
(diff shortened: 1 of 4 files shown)


akx:
Can you rebase? Also, this seems to add a widget _and_ some extra functionality in the media browser view?

Pikkupomo:
rebased and added comment about the extra functionality to commit message

shoop/admin/modules/media/views.py
```diff
@@ -82,6 +84,8 @@ def get_context_data(self, **kwargs):
         return context
 
     def get(self, request, *args, **kwargs):
+        self.filter = request.REQUEST.get("filter")
```
akx: You should change line 82 to also use `self.filter`, now that we have that.

shoop/admin/modules/products/views/forms.py
```diff
@@ -268,6 +268,8 @@ class ProductImageMediaForm(BaseProductMediaForm):
 
     def __init__(self, **kwargs):
         super(ProductImageMediaForm, self).__init__(**kwargs)
+        self.fields["file"].widget = ImageChoiceWidget()  # Filer misimplemented the field; we need to do this manually.
```
akx: The comment is unnecessary (and not really correct here imho)

akx:
Could you separate the additions into separate commits?

PR #110, opened 2015-09-14
Media fixes

Pikkupomo:
Fix issue with saving images
Fix issue with product images and medias mixing

Refs SHOOP-1345

### DIFF
--- a/shoop/admin/modules/products/views/forms.py
+++ b/shoop/admin/modules/products/views/forms.py
@@ -241,7 +241,6 @@ class BaseProductMediaFormSet(BaseModelFormSet):
 
 
 class ProductMediaForm(BaseProductMediaForm):
-
     def __init__(self, **kwargs):
         super(ProductMediaForm, self).__init__(**kwargs)
         self.fields["file"].required = False
@@ -300,8 +299,10 @@ class ProductImageMediaFormSet(ProductMediaFormSet):
         super(ProductImageMediaFormSet, self).save(commit)
 
         has_primary = any(form.cleaned_data.get("is_primary") for form in (self.forms or []))
+        eligible_forms = [form for form in (self.forms or []) if
+                          (form.cleaned_data.get("file") and not form.cleaned_data.get("DELETE"))]
 
-        if self.forms and not has_primary:
+        if eligible_forms and not has_primary:
             # make first form be the primary image as well
             form_instance = self.forms[0]
             form_instance.product.primary_image = form_instance.instance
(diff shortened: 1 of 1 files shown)


shoop/admin/modules/products/views/forms.py
```diff
@@ -183,6 +184,13 @@ def __init__(self, **kwargs):
             self.fields["kind"].initial = self.allowed_media_kinds[0]
             self.fields["kind"].widget = forms.HiddenInput()
 
+        if self.disallowed_media_kinds:
+            choices = []
+            for (i, choice) in self.fields["kind"].choices:
```
akx: Uh, so what's happening in this loop?

akx:
I don't understand what the issue is ("error when saving image with URL"?) and how this fixes it.

The second commit is equally confusing (and ref-less)...

Pikkupomo:
>I don't understand what the issue is ("error when saving image with URL"?) and how this fixes it.

The original ticket description was misleading, it has been fixed

>The second commit is equally confusing (and ref-less)...

It should't confuse you anymore... I hope :)

shoop/admin/modules/products/views/forms.py
```diff
@@ -178,10 +178,19 @@ def __init__(self, **kwargs):
         self.fields["file"].widget = MediaChoiceWidget()  # Filer misimplemented the field; we need to do this manually.
         self.fields["file"].required = True
 
-        if len(self.allowed_media_kinds) == 1:
-            # only one media kind given, no point showing the dropdown
+        if self.allowed_media_kinds:
+            if len(self.allowed_media_kinds) == 1:
+                # only one media kind given, no point showing the dropdown
+                self.fields["kind"].widget = forms.HiddenInput()
+            else:
```
akx: You should filter the choices even if the input is hidden; `choices` affects the validation as well.

shoop/admin/modules/products/views/forms.py
```diff
@@ -178,10 +178,19 @@ def __init__(self, **kwargs):
...
+                # multiple media kinds allowed, filter the choices list to reflect the `self.allowed_media_kinds`
+                allowed_kinds_values = set(v.value for v in self.allowed_media_kinds)
+                self.fields["kind"].choices = [
+                    (value, choice)
+                    for value, choice in self.fields["kind"].choices
+                    if value in allowed_kinds_values
+                ]
```
akx: Does this end up filtering the actual select widget as well?

akx:
:shipit:

PR #114, opened 2015-09-14
Add logo and name for classic gray

tulimaki:
Classic gray theme: Add missing bootstrap-select to bower.json
Classic gray theme: Use shop logo and public name in templates

Refs SHOOP-1178


akx:
Seems legit.

PR #116, opened 2015-09-14
Xtheme: Correct URL pattern order

akx:
(I am a dummy, veritably.)

Augments d0ca6e9d8e21e9150
Refs SHOOP-1177


juhakujala:
:+1:

PR #118, opened 2015-09-14
Product preview and add with ajax

tulimaki:
Classic gray theme: Add products to basket with AJAX
Classic gray theme: Enable product previews

Refs SHOOP-1210, SHOOP-1205, SHOOP-1206, SHOOP-1207, SHOOP-1208, SHOOP-1209

### DIFF
--- a/shoop/themes/classic_gray/__init__.py
+++ b/shoop/themes/classic_gray/__init__.py
@@ -15,6 +15,10 @@ class ClassicGrayTheme(Theme):
     author = "Juha Kujala"
     template_dir = "classic_gray/"
 
+    def get_view(self, view_name):
+        import shoop.themes.classic_gray.views as views
+        return getattr(views, view_name, None)
+
 
 class ClassicGrayThemeAppConfig(AppConfig):
     name = "shoop.themes.classic_gray"
--- a/shoop/themes/classic_gray/templates/classic_gray/shoop/front/base.jinja
+++ b/shoop/themes/classic_gray/templates/classic_gray/shoop/front/base.jinja
@@ -52,10 +52,13 @@
         </div>
     </footer>
 
+    
+
     {# Include JavaScript libraries #}
     <script src="{{ STATIC_URL }}classic_gray/js/classic-gray.js"></script>
-
     {# Include all Extra JavaScript #}
     {% block extrajs %}{% endblock %}
+    {# Enable adding products to basket with ajax #}
+    {% include "classic_gray/add_product_to_basket.jinja" %}
 </body>
 </html>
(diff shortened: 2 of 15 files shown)


shoop/themes/classic_gray/__init__.py
```diff
@@ -4,8 +4,9 @@
 #
 # This source code is licensed under the AGPLv3 license found in the
 # LICENSE file in the root directory of this source tree.
-
 from shoop.apps import AppConfig
+from shoop.themes.classic_gray.views.product_preview import product_quick_view
```
akx: You can't safely import views (that eventually do any model imports) here.

Move these imports to `get_view`?

You can even do

```python
def get_view(self, view_name):
    import shoop.themes.classic_gray.views as views
    return getattr(views, view_name, None)
```

and in `views/__init__.py` import the views with the xtheme extra view names:

```python
from .product_preview import product_quick_view as preview
from .basket import basket_partial
```

:sunglasses:

shoop/themes/classic_gray/templates/classic_gray/shoop/front/base.jinja
```diff
@@ -52,10 +52,33 @@
...
+        function showPreview(productId) {
+            var modal_select = "#product-" + productId + "-modal";
+            var product_modal = $(modal_select);
+            if (product_modal.length) {
+                product_modal.modal("show");
+                return;
+            }
+
+            $.ajax({
+                url: '{{ url("shoop:xtheme_extra_view", view="preview") }}?id=' + productId,
```
akx: You can rely on this being `/xtheme/preview`, no need for reversing.

shoop/themes/classic_gray/__init__.py
```diff
@@ -13,8 +12,18 @@ class ClassicGrayTheme(Theme):
     identifier = "shoop.themes.classic_gray"
     name = "Shoop Classic Gray Theme"
     author = "Juha Kujala"
+
     template_dir = "classic_gray/"
 
+    views = {
```
akx: You can get rid of this dict now :)

shoop/themes/classic_gray/views/__init__.py
```diff
@@ -0,0 +1,2 @@
+from .product_preview import product_quick_view as preview
+from .basket import basket_partial
```
akx: Ah, and to appease the vengeful spirit that is flake8, you'll also have to add

```python

__all__ = ["preview", "basket_partial"]

```

here.

akx:
Spanky!

:+1: :shipit:

PR #135, opened 2015-09-17
Pagination for category view

tulimaki:
Front: Add pagination option for CategoryView
Classic gray theme: Use pagination provided by shoop.front

Refs SHOOP-1419

### DIFF
--- a/shoop/front/utils/views.py
+++ b/shoop/front/utils/views.py
@@ -7,6 +7,8 @@
 # LICENSE file in the root directory of this source tree.
 from __future__ import with_statement
 
+from django.conf import settings
+from django.core.paginator import Paginator
 from django.utils.translation import get_language
 
 from shoop.core.models import Product, ProductAttribute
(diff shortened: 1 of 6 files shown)


shoop/themes/classic_gray/templates/classic_gray/shoop/front/product/category.jinja
```diff
@@ -88,6 +81,8 @@
 
 {% block extrajs %}
     <script>
+        // Absolute url for products reload
+        var ABSOLUTE_URL = "{{ request.path }}";
```
akx: No need for a separate variable. Use [`location.pathname`](https://developer.mozilla.org/en-US/docs/Web/API/Location) instead.

shoop/front/utils/views.py
```diff
@@ -24,3 +26,30 @@ def cache_product_things(request, products, language=None, attribute_identifiers
...
     return products
+
+
+def get_pagination_variables(request, products):
+    """
+    Used to update pagination variables to category view context
+    """
+    context = {}
+
+    limit = getattr(settings, "SHOOP_PRODUCTLIST_PRODUCT_COUNT", None)
```
Pikkupomo: Some sane defaults would be nice since you are not defining this on the `settings` by default.

tulimaki: But maybe we don't like to paginate for every theme. So like this pagination is optional for CategoryView.

shoop/front/utils/views.py
```diff
@@ -24,3 +26,30 @@ def cache_product_things(request, products, language=None, attribute_identifiers
...
+    if not limit:
+        return context
+    if len(products):
+        context["paginator"] = paginator = Paginator(products, (limit if limit else len(products)))
```
Pikkupomo: This will always use `limit` since you are exiting previously if no limit. Therefore the `len(products)` is useless check.

shoop/front/utils/views.py
```diff
@@ -24,3 +26,30 @@ def cache_product_things(request, products, language=None, attribute_identifiers
...
+                context["products"] = page.object_list
+            except:
+                pass
+        else:
+            context["is_paginated"] = False
+    else:
+        context["is_paginated"] = False
+        context["paginator"] = None
+
+    return context
```
Pikkupomo: how about something like
```python
limit = limit if limit else (len(products) if products else 0)

if not limit:	
	context["is_paginated"] = False
	context["paginator"] = None
	return context

context["paginator"] = paginator = Paginator(products, limit)
context["is_paginated"] = (paginator.num_pages > 1)
try:
    context["page"] = page = paginator.page(request.GET.get("page") or 1)
    context["products"] = page.object_list
except:
	pass

return context
```

PR #139, opened 2015-09-17
Fix PriceInfo usage with non-one quantities (SHOOP-1462)

suutari-ai:
Also set the correct discount amounts for SourceLine, BasketLine and
OrderLine when they are created.

Refs SHOOP-975 / SHOOP-1462

### DIFF
--- a/shoop/core/models/shops.py
+++ b/shoop/core/models/shops.py
@@ -16,6 +16,7 @@ from jsonfield import JSONField
 from parler.models import TranslatableModel, TranslatedFields
 
 from shoop.core.fields import InternalIdentifierField
+from shoop.core.pricing import TaxfulPrice, TaxlessPrice
 
 
 class ShopStatus(Enum):
@@ -40,3 +41,18 @@ class Shop(TranslatableModel):
 
     def __str__(self):
         return self.safe_translation_getter("name", default="Shop %d" % self.pk)
+
+    def create_price(self, value):
+        """
+        Create a price with given value and settings of this shop.
+
+        Takes the ``prices_include_tax`` setting into account, and in
+        future might also do the same for a currency setting.
+
+        :type value: decimal.Decimal
+        :rtype: shoop.core.pricing.Price
+        """
+        if self.prices_include_tax:
+            return TaxfulPrice(value)
+        else:
+            return TaxlessPrice(value)
--- a/shoop/core/pricing/__init__.py
+++ b/shoop/core/pricing/__init__.py
@@ -161,7 +161,8 @@ class PricingModule(six.with_metaclass(abc.ABCMeta)):
             for product_id in product_ids
         }
 
-    def get_pricing_steps_for_products(self, context, product_ids):
+    def get_pricing_steps_for_products(self, context, products):
+        product_ids = [getattr(x, "pk", x) for x in products]
         return {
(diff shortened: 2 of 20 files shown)


shoop/core/pricing/__init__.py
```diff
@@ -161,7 +161,8 @@ def get_price_infos(self, context, products, quantity=1):
             for product_id in product_ids
         }
 
-    def get_pricing_steps_for_products(self, context, product_ids):
+    def get_pricing_steps_for_products(self, context, products):
```
tulimaki: Missing docstring.

suutari-ai: Yeah, was already missing. Feel free to add that. ;)

tulimaki:
Now products get_price function has quantity 1 by default. Should we make sure that quantity is at least minimum purchase amount allowed? Would make it more simple to get from templates etc...

shoop/core/models/shops.py
```diff
@@ -40,3 +41,15 @@ class Shop(TranslatableModel):
 
     def __str__(self):
         return self.safe_translation_getter("name", default="Shop %d" % self.pk)
+
+    def create_price(self, value):
+        """
+        Create a price with given value and settings of this shop.
+
+        Takes the ``prices_include_tax`` setting into account, and in
+        future might also do the same for a currency setting.
```
akx: I assume `value` is just an unadorned decimal?

(`:type:` annotations would be nice.)

suutari-ai:
> Now products get_price function has quantity 1 by default. Should we make sure that quantity is at least minimum purchase amount allowed? Would make it more simple to get from templates etc...

Are you proposing something like this?
```python
def get_price_info(self, context, quantity=None):
    from shoop.core.pricing import get_pricing_module
    module = get_pricing_module()
    pricing_context = module.get_context(context)
    if quantity is None:
        shop_product = self.get_shop_instance(pricing_context.shop)
        quantity = shop_product.minimum_purchase_quantity
    return module.get_price_info(pricing_context, product=self, quantity=quantity)
```

Seems somewhat logical, but how about the `Product.get_cheapest_child_price`, what should it do, since the minimum_purchase_quantity could be different for each child?

shoop/core/models/shops.py
suutari-ai: Right. Yep, I'll add the type annotations.

tulimaki:
Yes something like that. I guess cheapest child price should return price that you can purchase some children cheapest. It think it should always force to minimum_purchase_quantity.

suutari-ai:
> Yes something like that. I guess cheapest child price should return price that you can purchase some children cheapest. It think it should always force to minimum_purchase_quantity.

Force to minimum_purchase_quantity of which product? (the parent or one of the children?)

Also, quantity=1 as default is used in many other places too, e.g. in `PricingModule.get_price_info`, and `PricingModule.get_price_infos` and some others. Maybe changing the default from 1 to minimum_purchase_quantity should be another issue?

(As a side note: `PricingModule` methods should probably not have a default value for the quantity)

shoop/core/models/shops.py
suutari-ai: Added.

tulimaki:
@suutari-ai I think we should talk about this in some meeting. And it should be different task since it effects so many places.

tulimaki:
great work

PR #140, opened 2015-09-17
Admin: Show errors more clearly

tulimaki:
Admin: Activate first sidebar list item with errors
Admin: Make first language tab with errors work better

Refs SHOOP-1398, SHOOP-1396

### DIFF
--- a/shoop/admin/static_src/base/js/content-blocks.js
+++ b/shoop/admin/static_src/base/js/content-blocks.js
@@ -21,4 +21,6 @@ $(function() {
             $(this).find(".block-title").addClass("mobile-error-indicator");
         }
     });
+    // Activate first sidebar-list-item with errors
+    $("a.sidebar-list-item.errors").first().trigger("click");
 });
--- a/shoop/admin/static_src/base/js/form-utils.js
+++ b/shoop/admin/static_src/base/js/form-utils.js
@@ -56,9 +56,10 @@ $(function() {
             }
             const tabHasErrors = ($tabPane.find(".has-error").length > 0);
             if(tabHasErrors) {
-                $tab.find("a").append($(" <div class=error-indicator><i class=\"fa fa-exclamation-circle\"></i></div>"));
+                const $tabLink = $tab.find("a");
+                $tabLink.append($(" <div class=error-indicator><i class=\"fa fa-exclamation-circle\"></i></div>"));
                 if (!firstTabWithErrorsOpened) {
-                    $tab.tab("show");
+                    $tabLink.tab("show");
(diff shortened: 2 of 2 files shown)


shoop/admin/static_src/base/js/content-blocks.js
```diff
@@ -21,4 +21,9 @@ $(function() {
             $(this).find(".block-title").addClass("mobile-error-indicator");
         }
     });
+    // Activate first sidebar-list-item with errors
+    var $firstSidebarListItemWithErrors = $("a.sidebar-list-item.errors").first();
+    if ($firstSidebarListItemWithErrors.length > 0) {
```
akx: I don't think you actually even need to check for the length -- [the trigger would act on a collection of 0 element, i.e. not much at all](https://github.com/jquery/jquery/blob/master/src/event.js#L912-L916).

shoop/admin/static_src/base/js/content-blocks.js
```diff
@@ -21,4 +21,9 @@ $(function() {
             $(this).find(".block-title").addClass("mobile-error-indicator");
         }
     });
+    // Activate first sidebar-list-item with errors
+    var $firstSidebarListItemWithErrors = $("a.sidebar-list-item.errors").first();
```
akx: This could be const-bound :)

tulimaki:
@akx also added one commit to make language-tabs errors work better.

shoop/admin/static_src/base/js/form-utils.js
```diff
@@ -56,9 +56,10 @@ $(function() {
             }
             const tabHasErrors = ($tabPane.find(".has-error").length > 0);
             if(tabHasErrors) {
-                $tab.find("a").append($(" <div class=error-indicator><i class=\"fa fa-exclamation-circle\"></i></div>"));
+                const $tab_link = $tab.find("a");
```
akx: Use camelCase in JavaScript :)

shoop/admin/static_src/base/js/content-blocks.js
```diff
@@ -21,4 +21,9 @@ $(function() {
             $(this).find(".block-title").addClass("mobile-error-indicator");
         }
     });
+    // Activate first sidebar-list-item with errors
+    const $firstSidebarListItemWithErrors = $("a.sidebar-list-item.errors").first();
+    if ($firstSidebarListItemWithErrors) {
```
akx: No need for this `if` altogether, I meant (`$(selector)` never returns a falsy value).,

Just

```javascript
$("a.sidebar-list-item.errors").first().trigger("click");
```

should be plenty.

PR #141, opened 2015-09-18
Classic gray theme: Reload product prices with AJAX in quantity change

tulimaki:
* Create custom product_prices view for classic gray theme
* Call product_prices view with AJAX, with product id, quantity and
optional child id parameters
* Render correct prices at _detail_order_section template based on
previous parameters

Refs SHOOP-1356

### DIFF
--- a/shoop/themes/classic_gray/static_src/js/custom.js
+++ b/shoop/themes/classic_gray/static_src/js/custom.js
@@ -56,6 +56,29 @@ function reloadProducts() {
     $("#ajax_content").load(location.pathname + filterString);
 }
 
+function updateProductPrice(productId) {
+    var $quantity = $("#product-quantity");
+    if (!$quantity.is(":valid")) {
+        return;
+    }
+    var data = {
+        id: productId,
+        quantity: $quantity.val()
+    };
+    var $selectedChild = $("#product-variations");
+    if ($selectedChild.length > 0) {
+        data.child = $selectedChild.val();
+    }
+    $.ajax({
+        url: "/xtheme/product_price",
+        method: "GET",
+        data: data,
+        success: function(html) {
+            $("#product-price-section").html(html);
+        }
+    });
+}
+
 $(function() {
     $("#search-modal").on("show.bs.modal", function() {
         setTimeout(function(){
--- a/shoop/themes/classic_gray/templates/classic_gray/shoop/front/product/_detail_order_section.jinja
+++ b/shoop/themes/classic_gray/templates/classic_gray/shoop/front/product/_detail_order_section.jinja
@@ -10,6 +10,10 @@
     </script>
 {% endif %}
 
+{% if not quantity %}
+    {% set quantity = shop_product.rounded_minimum_purchase_quantity %}
+{% endif %}
+
 {% macro quantity_box() %}
     <div class="form-group amount">
         <label for="product-quantity">{% trans %}Quantity{% endtrans %}</label>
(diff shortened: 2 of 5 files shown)


shoop/themes/classic_gray/templates/classic_gray/shoop/front/product/_detail_order_section.jinja
```diff
@@ -9,6 +9,8 @@
         });
     </script>
 {% endif %}
+{% set quantity = product.sales_unit.round(request.GET.get("quantity") or 0) or shop_product.rounded_minimum_purchase_quantity %}
```
akx: This is starting to be a little too much logic in the template for my liking...

shoop/themes/classic_gray/templates/classic_gray/shoop/front/product/_detail_order_section.jinja
```diff
@@ -20,8 +22,9 @@
             name="quantity"
             id="product-quantity"
             step="{{ shop_product.quantity_step }}"
-            value="{{ shop_product.rounded_minimum_purchase_quantity }}"
-            min="{{ shop_product.rounded_minimum_purchase_quantity }}">
+            value="{{ quantity }}"
+            min="{{ shop_product.rounded_minimum_purchase_quantity }}"
+            onchange="javascript:updateProductPrice('{{ product.pk }}')">
```
akx: `on*` attributes just take JavaScript code, not `javascript:` URLs

shoop/themes/classic_gray/static_src/js/custom.js
```diff
@@ -29,6 +29,28 @@ function getProductListViewMode() {
     return "grid";
 }
 
+function updateProductPrice(productId) {
+    const $quantity = $("#product-quantity");
+    if (!$quantity.is(":valid")) {
+        return
+    }
+    var quantityParam = '&quantity=' + $quantity.val();
```
RauliL: Shouldn't this be URL escaped?

shoop/themes/classic_gray/static_src/js/custom.js
```diff
@@ -29,6 +29,28 @@ function getProductListViewMode() {
...
+
+    var childParam = "";
+    const $selectedChild = $("#product-variations");
+    if ($selectedChild.length > 0) {
+        var childParam = "&child=" + $selectedChild.val();
```
RauliL: No need for `var` here.

shoop/themes/classic_gray/static_src/js/custom.js
```diff
@@ -29,6 +29,28 @@ function getProductListViewMode() {
...
+    }
+
+    $.ajax({
+        url: '/xtheme/product_price?id=' + productId + quantityParam + childParam,
```
RauliL: Instead of constructing the URL with string concatenation, you could use `data` property here.

RauliL: With something like this:
```JavaScript
var data = {
    id: productId,
    quantity: $quantity.val()
};
...
if ($selectedChild.length > 0) {
    data.child = $selectedChild.val();
}
...
$.ajax({
    url: "/xtheme/product_price",
    method: "GET",
    data: data,
    success: $("#product-price-section").html
});
```

shoop/themes/classic_gray/static_src/js/custom.js
```diff
@@ -29,6 +29,29 @@ function getProductListViewMode() {
     return "grid";
 }
 
+function updateProductPrice(productId) {
+    const $quantity = $("#product-quantity");
```
akx: If `custom.js` is not transpiled with Babel (I don't know about the current gulpfile), we can't use `const` safely cross-browser.

shoop/themes/classic_gray/static_src/js/custom.js
```diff
@@ -29,6 +29,29 @@ function getProductListViewMode() {
...
+    if (!$quantity.is(":valid")) {
+        return;
+    }
+    var data = {
```
akx: If we can use const binds, this could also be one (you're never reassigning `data`, just editing it)

PR #142, opened 2015-09-18
Admin: Make folder clicks work at media browser

tulimaki:
Refs SHOOP-1354

### DIFF
--- a/shoop/admin/modules/media/static_src/media/browser/components/folderClick.js
+++ b/shoop/admin/modules/media/static_src/media/browser/components/folderClick.js
@@ -1,0 +1,15 @@
+/**
+ * This file is part of Shoop.
+ *
+ * Copyright (c) 2012-2015, Shoop Ltd. All rights reserved.
+ *
+ * This source code is licensed under the AGPLv3 license found in the
+ * LICENSE file in the root directory of this source tree.
+ */
+export default function(ctrl, folder) {
+    return function clickFolder(event) {
+        ctrl.setFolder(folder.id);
+        event.preventDefault();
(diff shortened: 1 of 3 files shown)


shoop/admin/modules/media/static_src/media/browser/components/folderClick.js
```diff
@@ -0,0 +1,19 @@
...
+const _ = require("lodash");
+
+export default function(ctrl, folder) {
+    function clickFolder(event, folderId) {
+        ctrl.setFolder(folderId);
+        event.preventDefault();
+        return false;
+    }
+
+    return _.partialRight(clickFolder, folder.id);
```
akx: You don't need to use `_.partialRight` since you're already making `clickFolder` an inner function that has `folder` in its closure:

```javascript
export default function(ctrl, folder) {
    return function clickFolder(event) {
        ctrl.setFolder(folder.id);
        event.preventDefault();
        return false;
    }
}
```

shoop/admin/modules/media/static_src/media/browser/components/folderClick.js
```diff
@@ -0,0 +1,17 @@
+/**
+ * This file is part of Shoop.
+ *
+ * Copyright (c) 2012-2015, Shoop Ltd. All rights reserved.
+ *
+ * This source code is licensed under the AGPLv3 license found in the
+ * LICENSE file in the root directory of this source tree.
+ */
+const _ = require("lodash");
```
akx: Unnecessary `require` now :)

PR #152, opened 2015-09-23
Maintenance mode (SHOOP-1153)

tulimaki:
Core: Add missing migration for product media external_url
Core: Enable maintenance mode

Refs SHOOP-1153

### DIFF
--- a/shoop/core/migrations/0008_maintenance_mode.py
+++ b/shoop/core/migrations/0008_maintenance_mode.py
@@ -1,0 +1,24 @@
+# -*- coding: utf-8 -*-
+from __future__ import unicode_literals
+
+from django.db import models, migrations
+
+
+class Migration(migrations.Migration):
+
+    dependencies = [
+        ('shoop', '0007_product_media'),
+    ]
+
+    operations = [
+        migrations.AddField(
+            model_name='shop',
+            name='maintenance_mode',
+            field=models.BooleanField(default=False, verbose_name='maintenance mode'),
+        ),
+        migrations.AddField(
+            model_name='shoptranslation',
+            name='maintenance_message',
+            field=models.CharField(max_length=300, blank=True),
+        ),
+    ]
--- a/shoop/core/models/shops.py
+++ b/shoop/core/models/shops.py
@@ -32,11 +32,13 @@ class Shop(TranslatableModel):
     owner = models.ForeignKey("Contact", blank=True, null=True)
     options = JSONField(blank=True, null=True)
     prices_include_tax = models.BooleanField(default=True)
-    logo = FilerImageField(verbose_name=_('logo'), blank=True, null=True)
+    logo = FilerImageField(verbose_name=_("logo"), blank=True, null=True)
+    maintenance_mode = models.BooleanField(verbose_name=_("maintenance mode"), default=False)
 
     translations = TranslatedFields(
         name=models.CharField(max_length=64),
-        public_name=models.CharField(max_length=64)
+        public_name=models.CharField(max_length=64),
+        maintenance_message=models.CharField(max_length=300, blank=True)
     )
 
     def __str__(self):
--- a/shoop/front/middleware.py
+++ b/shoop/front/middleware.py
@@ -9,6 +9,8 @@ from django.conf import settings
 from django.contrib.auth.models import AnonymousUser
 from django.contrib.auth.signals import user_logged_in, user_logged_out
 from django.core.exceptions import ImproperlyConfigured
+from django.http import HttpResponse
+from django.template import loader
 from django.utils import timezone
 
 from shoop.core.middleware import ExceptionMiddleware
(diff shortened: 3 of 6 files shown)


suutari-ai:
@tulimaki:
> Refs SHOOP-1134

Fix this. It should be SHOOP-1153 probably.

shoop/front/middleware.py
```diff
@@ -88,6 +90,11 @@ def _set_timezone(self, request):
     def process_response(self, request, response):
         if hasattr(request, "basket") and request.basket.dirty:
             request.basket.save()
+
+        is_superuser = request.user and request.user.is_superuser
```
akx: You'll want to do this during `process_request`, otherwise the pages do all their work and then end up showing a "not available" page :\

akx: Also, you shouldn't need to check for the truthiness of `request.user` in the first place; if no one is logged in, it's an `AnonymousUser`

akx:
Could we get tests for this feature?

shoop/front/middleware.py
```diff
@@ -60,6 +62,8 @@ class ShoopFrontMiddleware(object):
 
     def process_request(self, request):
         self._set_shop(request)
+        if request.shop.maintenance_mode and not request.user.is_superuser:
+            return HttpResponse(loader.render_to_string("shoop/front/maintenance.jinja", request=request))
```
suutari-ai: Put this to own method to allow that functionality be replaced by subclassing.
```python
...
self._set_shop(request)
maintenance_response = self._get_maintenance_response(request)
if maintenance_response:
    return maintenance_response
self._set_person(request)
...
```

shoop/core/migrations/0008_maintenance_mode.py
```diff
@@ -0,0 +1,25 @@
...
+
+    dependencies = [
+        ('shoop', '0007_product_media'),
+    ]
+
+    operations = [
+        migrations.AddField(
+            model_name='shop',
+            name='maintenance_mode',
+            field=models.BooleanField(default=True, verbose_name='maintenance mode'),
```
suutari-ai: This being default=True breaks all tests that need the front middleware to set the login.

shoop/front/middleware.py
akx: :+1: Great idea.

akx: On that note, it should be done after all of the other `_set`s, to ensure the request is in a decent state.

shoop/core/models/shops.py
```diff
@@ -32,11 +32,13 @@ class Shop(TranslatableModel):
     owner = models.ForeignKey("Contact", blank=True, null=True)
     options = JSONField(blank=True, null=True)
     prices_include_tax = models.BooleanField(default=True)
-    logo = FilerImageField(verbose_name=_('logo'), blank=True, null=True)
+    logo = FilerImageField(verbose_name=_("logo"), blank=True, null=True)
+    maintenance_mode = models.BooleanField(verbose_name=_("maintenance mode"), default=True)
```
akx: The default should definitely be `False`, as @suutari-ai alludes.

shoop/core/models/shops.py
```diff
@@ -32,11 +32,13 @@ class Shop(TranslatableModel):
...
 
     translations = TranslatedFields(
         name=models.CharField(max_length=64),
-        public_name=models.CharField(max_length=64)
+        public_name=models.CharField(max_length=64),
+        maintenance_message=models.CharField(max_length=300)
```
akx: `maintenance_message` should be `blank=True`.

shoop/front/middleware.py
```diff
@@ -85,6 +91,10 @@ def _set_timezone(self, request):
             timezone.activate(request.person.timezone)
             # TODO: Fallback to request.shop.timezone (and add such field)
 
+    def _get_maintenance_response(self, request):
+        if request.shop.maintenance_mode and not request.user.is_superuser:
+            return HttpResponse(loader.render_to_string("shoop/front/maintenance.jinja", request=request))
```
akx: Come to think of it, it might be a good idea to return a non-200 error code from here, to avoid, say, search engines indexing the maintenance page as regular content. 503 ("Service Unavailable") sounds like a good fit.

This will also simplify the test case; you can just look at the status code there.

shoop_tests/front/test_middleware.py
```diff
@@ -100,3 +100,25 @@ def test_intra_request_user_changing(rf, regular_user):
...
+
+@pytest.mark.django_db
+def test_maintenance_mode(rf, regular_user, admin_user):
+    shop = shoop.core.models.Shop.objects.create(
+        name="test shop",
+        identifier="test",
+        status=shoop.core.models.ShopStatus.ENABLED,
+        maintenance_mode=True
+    )
+    mw = ShoopFrontMiddleware()
```
akx: Considering `ShoopFrontMiddleware` does `request.shop = Shop.objects.first()`, I don't see how the freshly created shop would end up there...

Maybe just modify `get_default_shop()` and return it to normal after the test?

tulimaki: Well without shop create here test fails because there is no shop.

akx: `get_default_shop()` get-or-creates that shop for you.

akx:
:+1: :lollipop:

PR #153, opened 2015-09-23
Fix shoop.core.migrations.0006

suutari-ai:
Do not import Shop models directly from shoop.core.models, but use the
apps.get_model instead.  This is needed, because importing Shop
directly will use the newest version of the model, not the version of
the previous migration level and that makes it impossible e.g. to add
fields to Shop.

Refs SHOOP-1153 / SHOOP-1502


akx:
:+1: Will merge as soon as Jenkins says okay.

PR #154, opened 2015-09-23
Classic gray theme: Add pagination for category view

tulimaki:
* Add get_pagination_variables context function under
front.template_helpers.category
* Render product list based on pagination variables
* Reload category page with AJAX

Refs SHOOP-1419

### DIFF
--- a/shoop/themes/classic_gray/templates/classic_gray/shoop/front/macros.jinja
+++ b/shoop/themes/classic_gray/templates/classic_gray/shoop/front/macros.jinja
@@ -233,3 +233,44 @@
     {% endif %}
 
 {% endmacro %}
+
+{% macro pagination(page, paginator) %}
+    <script>
+        window.PAGE_NUMBER = "{{ page.number }}";
+    </script>
+    <nav>
+        <ul class="pagination text-center">
+            {% if page.has_previous() %}
+                <li id="previous_page">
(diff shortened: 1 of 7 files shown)


shoop/front/template_helpers/category.py
```diff
@@ -22,3 +23,31 @@ def get_manufacturers(context):
...
+        variables["paginator"] = None
+        variables["is_paginated"] = False
+        return variables
+
+    variables["paginator"] = paginator = Paginator(products, limit)
+    variables["is_paginated"] = (paginator.num_pages > 1)
+    try:
+        variables["page"] = page = paginator.page(context["request"].GET.get("page") or 1)
+        variables["products"] = page.object_list
+    except:
```
akx: I don't like the look of this bare `try:except:` :)

shoop/themes/classic_gray/templates/classic_gray/shoop/front/macros.jinja
```diff
@@ -233,3 +233,48 @@
     {% endif %}
 
 {% endmacro %}
+
+{% macro pagination(page, paginator) %}
+    <script>
+        {% if page %}
```
akx: Why _wouldn't_ we have a `page` if we're paginating something?

akx:
So yeah, as discussed in meatspace, this could probably be made more generic and simplified a little more :)

akx:
Nice! Thank you!

PR #156, opened 2015-09-23
Tests: Make test_user_detail_contact_seed not fail randomly

suutari-ai:
Refs https://github.com/shoopio/shoop/pull/155#issuecomment-142689539

### DIFF
--- a/shoop_tests/admin/test_user_module.py
+++ b/shoop_tests/admin/test_user_module.py
@@ -41,6 +41,13 @@ def test_user_detail_works_at_all(rf, admin_user):
 def test_user_detail_contact_seed(rf):
     get_default_shop()
     contact = create_random_person()
+
+    # Using random data for name and email would need escaping when
+    # checking if it is rendered, therefore use something very basic instead
+    contact.name = "Matti Perustyyppi"
+    contact.email = "matti.perustyyppi@perus.fi"
+    contact.save()
+
     view_func = UserDetailView.as_view()
     # Check that fields populate . . .
     request = apply_request_middleware(rf.get("/", {"contact_id": contact.pk}))


shoop_tests/admin/test_user_module.py
```diff
@@ -41,6 +41,13 @@ def test_user_detail_works_at_all(rf, admin_user):
 def test_user_detail_contact_seed(rf):
     get_default_shop()
     contact = create_random_person()
+
+    # Using random data for name and email would need escaping when
+    # checking if it is rendered, therefore use something very basic instead
```
akx: Why not just un-escape the content?

suutari-ai: Because there is no point in having random data like this in the tests. What are we trying to test here anyway?

akx: I kind of feel it's better to have random data that could discover failures in the code by chance...

Pikkupomo: I have to agree with @akx in this. Random data is better because this wont reveal if person with name `趙 雅萍` causes some strange behavior or breaks something.

suutari-ai: Then we should test with name "趙 雅萍" and all the others that might be problematic and not fail randomly.

Pikkupomo: Perhaps, but name like that is delivered from faker.

suutari-ai:
I hate that my coding is interrupted by randomly failing test. If you feel that we should have random data in the tests, please implement this test correctly then. Until this is fixed properly, I'm just gonna use my fix. Sorry!

akx:
You could've just disabled this test locally, maybe.

suutari-ai:
Or just remove it, since it's invalid.

PR #157, opened 2015-09-24
Front middleware: Avoid maintenance mode in admin

tulimaki:
Refs SHOOP-1418

### DIFF
--- a/shoop/front/middleware.py
+++ b/shoop/front/middleware.py
@@ -67,10 +67,6 @@ class ShoopFrontMiddleware(object):
         self._set_basket(request)
         self._set_timezone(request)
 
-        maintenance_response = self._get_maintenance_response(request)
-        if maintenance_response:
-            return maintenance_response
-
     def _set_shop(self, request):
         # TODO: Not the best logic :)
         request.shop = Shop.objects.first()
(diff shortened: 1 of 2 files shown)


shoop/front/middleware.py
```diff
@@ -92,6 +93,9 @@ def _set_timezone(self, request):
             # TODO: Fallback to request.shop.timezone (and add such field)
 
     def _get_maintenance_response(self, request):
+        if resolve(request.path).app_name == "shoop_admin":
```
akx: What if resolution fails? Also, I don't like the implication that URL resolution will be done twice per request...

shoop/front/middleware.py
```diff
@@ -91,15 +87,22 @@ def _set_timezone(self, request):
...
             # TODO: Fallback to request.shop.timezone (and add such field)
 
-    def _get_maintenance_response(self, request):
-        if request.shop.maintenance_mode and not request.user.is_superuser:
-            return HttpResponse(loader.render_to_string("shoop/front/maintenance.jinja", request=request), status=503)
-
     def process_response(self, request, response):
         if hasattr(request, "basket") and request.basket.dirty:
             request.basket.save()
+        maintenance_response = self._get_maintenance_response(request)
```
akx: Now we're back to the actual view doing its work before we override the response with another. :\

akx: Could we use [process_view](https://docs.djangoproject.com/en/1.8/topics/http/middleware/#process-view) instead?  Then we'd have `resolver_match` available, and we could even introspect the `view` argument (I assume the point here is to allow logging in to maintenance mode shops) to see if it "smells like" a login view.

PR #160, opened 2015-09-24
template_helpers: Fix get_pagination_variables

suutari-ai:
The GET parameter is a string and should be converted to int.

Refs SHOOP-1419


akx:
Good catch. Thanks!

PR #161, opened 2015-09-24
Customer URL now requires login

Pikkupomo:
Front Auth: Ensure `next` is transferred through login flow
Front: view `/customer/` now requires login
`LOGIN_URL` now defined in workbench settings

Refs SHOOP-1514


akx:
@suutari-ai: That's just testing Django's `login_required` decorator though.

suutari-ai:
Still, it is a bug, and that's what test cases are good at. See #163.

PR #162, opened 2015-09-24
Ensure user is logged in after activating account

Pikkupomo:
No ref


suutari-ai:
> No ref

This is actually https://project.anders.fi/browse/SHOOP-614

PR #163, opened 2015-09-24
Add test case for accessing customer view unauthorized

suutari-ai:
Refs SHOOP-1514

### DIFF
--- a/shoop_tests/front/test_customer_information.py
+++ b/shoop_tests/front/test_customer_information.py
@@ -6,8 +6,10 @@
 # This source code is licensed under the AGPLv3 license found in the
 # LICENSE file in the root directory of this source tree.
 from __future__ import unicode_literals
+from django.conf import settings
 from django.contrib.auth import get_user_model
 from django.core.urlresolvers import reverse
+from django.shortcuts import resolve_url
 import pytest
 from shoop.testing.factories import get_default_shop
 from shoop.testing.soup_utils import extract_form_fields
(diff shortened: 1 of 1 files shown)


shoop_tests/front/test_customer_information.py
```diff
@@ -52,3 +52,15 @@ def test_new_user_information_edit():
...
     assert get_user_model().objects.get(pk=user.pk).email == new_email
+
+
+@pytest.mark.django_db
+def test_user_information_unavailable_if_not_logged_in():
+    get_default_shop()  # Front middleware needs a Shop to exists
+    client = SmartClient()
+    customer_edit_url = reverse("shoop:customer_edit")
+    login_path = reverse("shoop:login")
+    login_path = (login_path[:-1] if login_path.endswith("/") else login_path)
```
akx: Huh, wait, why is this necessary?

shoop_tests/front/test_customer_information.py
```diff
@@ -52,3 +52,15 @@ def test_new_user_information_edit():
...
     assert response.status_code == 302
     assert get_user_model().objects.get(pk=user.pk).email == new_email
+
+
+@pytest.mark.django_db
+def test_user_information_unavailable_if_not_logged_in():
+    get_default_shop()  # Front middleware needs a Shop to exists
+    client = SmartClient()
+    customer_edit_url = reverse("shoop:customer_edit")
+    login_path = reverse("shoop:login")
```
akx: This should probably be `resolve_url(settings.LOGIN_URL)`, actually ([that's what Django does](https://github.com/django/django/blob/1.8.4/django/contrib/auth/decorators.py#L24))

suutari-ai:
@akx: Thanks for the comments. Your approach got the test much cleaner too. :)

Also removed some temporary variables for brevity.

PR #165, opened 2015-09-28
Maintenance mode: add maintenance mode exempt decorator

tulimaki:
Refs SHOOP-1518

### DIFF
--- a/shoop/core/utils/maintenance.py
+++ b/shoop/core/utils/maintenance.py
@@ -1,0 +1,17 @@
+# -*- coding: utf-8 -*-
+# This file is part of Shoop.
+#
+# Copyright (c) 2012-2015, Shoop Ltd. All rights reserved.
+#
+# This source code is licensed under the AGPLv3 license found in the
+# LICENSE file in the root directory of this source tree.
+
+def maintenance_mode_exempt(view_func):
+    """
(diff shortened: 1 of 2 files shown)


shoop/core/utils/maintenance.py
```diff
@@ -0,0 +1,16 @@
...
+
+
+def maintenance_mode_exempt(view_func):
+    """
+    Make view ignore shop maintenance mode
+
+    :param view_func: view attached to this decorator
+    :return: view wrapped with maintenance_mode_exempt attribute
+    """
+    def wrapped_view(*args, **kwargs):
```
akx: I don't think you need to wrap the view... Just do

```python
view_func.maintenance_mode_exempt = True
return view_func
```

?

shoop/front/middleware.py
```diff
@@ -126,7 +126,7 @@ def process_view(self, request, view_func, *view_args, **view_kwargs):
             return maintenance_response
 
     def _get_maintenance_response(self, request, view_func):
-        if "login" in view_func.__name__:
```
akx: I think the `login` heuristic would be good to have there, for third-party apps that likely use a view named "login", in addition to the new flag.

akx:
Looks brilliant, thanks!

PR #166, opened 2015-09-28
Various fixes

Pikkupomo:
Front: Show message if simple product not sellable.
Admin: Require minimum purchase quantity be over 0 
Admin: Require Sales Unit
Admin: Round Minimum Purchase Price based on the chosen Sales Unit

Refs SHOOP-1527

### DIFF
--- a/shoop/admin/modules/products/views/edit.py
+++ b/shoop/admin/modules/products/views/edit.py
@@ -83,6 +83,14 @@ class ShopProductFormPart(FormPart):
                 continue
             if not shop_product_form.instance.pk:
                 shop_product_form.instance.product = self.object
+
+            original_quantity = shop_product_form.instance.minimum_purchase_quantity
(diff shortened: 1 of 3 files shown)


shoop/admin/modules/products/views/edit.py
```diff
@@ -83,6 +83,14 @@ def form_valid(self, form):
                 continue
             if not shop_product_form.instance.pk:
                 shop_product_form.instance.product = self.object
+
+            original_quantity = shop_product_form.instance.minimum_purchase_quantity
```
akx: So this change deprecates the rounded_minimum_purchase_quantity property, yeah?

Should the PR mark it as Deprecated.

Pikkupomo: Maybe not, because of various "importers" or such that can create products outside Admin UI

akx: Fair enough.

PR #175, opened 2015-09-29
Front middleware: Let shoop admin through when in maintenance mode

tulimaki:
Refs SHOOP-1539

### DIFF
--- a/shoop/front/middleware.py
+++ b/shoop/front/middleware.py
@@ -126,8 +126,13 @@ class ShoopFrontMiddleware(object):
             return maintenance_response
 
     def _get_maintenance_response(self, request, view_func):
-        if "login" in view_func.__name__ or getattr(view_func, "maintenance_mode_exempt", False):
-            return
+        if getattr(view_func, "maintenance_mode_exempt", False):
+            return None
+        if "login" in view_func.__name__:
+            return None
+        resolver_match = getattr(request, "resolver_match", None)
+        if resolver_match and resolver_match.app_name == "shoop_admin":
+            return None
 
         if request.shop.maintenance_mode and not request.user.is_superuser:
             return HttpResponse(loader.render_to_string("shoop/front/maintenance.jinja", request=request), status=503)


shoop/front/middleware.py
```diff
@@ -126,7 +126,11 @@ def process_view(self, request, view_func, *view_args, **view_kwargs):
             return maintenance_response
 
     def _get_maintenance_response(self, request, view_func):
-        if "login" in view_func.__name__ or getattr(view_func, "maintenance_mode_exempt", False):
+        if getattr(view_func, "maintenance_mode_exempt", False):
+            return
```
suutari-ai: Explicit `return None` would be better than just `return`.

shoop/front/middleware.py
```diff
@@ -126,7 +126,11 @@ def process_view(self, request, view_func, *view_args, **view_kwargs):
...
+        if "login" in view_func.__name__:
+            return
```
suutari-ai: Same here... and change the third one too.

shoop/front/middleware.py
```diff
@@ -126,7 +126,11 @@ def process_view(self, request, view_func, *view_args, **view_kwargs):
...
+        if getattr(request, "resolver_match") and getattr(request.resolver_match, "app_name") == "shoop_admin":
```
suutari-ai: Why not just `if request.resolver_match.app_name == "shoop_admin"`? (Since you're using the two-param getattr, i.e. without default value. That'll raise AttributeError if there is no such attribute.)

suutari-ai: Or would `request.url_name == "shoop_admin"` be more correct? (dunno about that stuff, but just guessing)

tulimaki: I think there might be case that request.resolver_match is None. In that case it returns False in the first if statement and second getattr is never done. url_name is also under resolver_match.

shoop/front/middleware.py
```diff
@@ -126,8 +126,12 @@ def process_view(self, request, view_func, *view_args, **view_kwargs):
             return maintenance_response
 
     def _get_maintenance_response(self, request, view_func):
-        if "login" in view_func.__name__ or getattr(view_func, "maintenance_mode_exempt", False):
-            return
+        if getattr(view_func, "maintenance_mode_exempt", False):
+            return None
+        if "login" in view_func.__name__:
+            return None
+        if getattr(request, "resolver_match") and getattr(request.resolver_match, "app_name") == "shoop_admin":
```
akx: Maybe make this

```python
resolver_match = getattr(request, "resolver_match", None)
if resolver_match and resolver_match.app_name == "shoop_admin":
```

We can probably assume that if we have a `resolver_match`, it will have `app_name`.

PR #178, opened 2015-09-29
Classic gray theme: Fix selector issue with valid in price update

tulimaki:
Check that quantity input actually exist before validating it.

Refs SHOOP-1544


akx:
Looks good. Merging before tests pass, because I know Javascript isn't tested by Jenkins at the moment.

PR #184, opened 2015-09-30
Admin: Time interval attribute now renders as `DecimalField`

Pikkupomo:
Time interval field had `IntegerField` as a formfield type. This was not
intented as the value was actually considered as decimal.

Refs SHOOP-1550

### DIFF
--- a/shoop_tests/admin/test_product_attributes_form.py
+++ b/shoop_tests/admin/test_product_attributes_form.py
@@ -7,14 +7,15 @@
 # LICENSE file in the root directory of this source tree.
 from __future__ import unicode_literals
 import datetime
+from decimal import Decimal
+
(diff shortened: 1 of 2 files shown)


akx:
Why was this change made? Test case?

shoop_tests/admin/test_product_attributes_form.py
```diff
@@ -27,6 +27,7 @@ def test_product_attributes_form():
     product.set_attribute_value("bogomips", 6400)
     product.set_attribute_value("genre", "Kauhu", "fi")
     product.set_attribute_value("genre", "Horror", "en")
+    product.set_attribute_value("time_to_finish", 12)
```
akx: Maybe set it to a decimal value and ensure it remains the same? :)

PR #186, opened 2015-09-30
Classic Gray: Fix issue with variation children prices

Pikkupomo:
Fix issue where product has variation children but none have price available
(no products attached to the variation).

No ref


tulimaki:
:lollipop:

PR #193, opened 2015-10-01
Front: Fix issue with variation children being listed for admin user

Pikkupomo:
Customer with `is_all_seeing` `True` could see variation children in product
listings and this caused multiple problems and was not intented.

Refs SHOOP-1596

### DIFF
--- a/shoop/core/models/products.py
+++ b/shoop/core/models/products.py
@@ -117,7 +117,8 @@ class ProductQuerySet(TranslatableQuerySet):
         root = (self.language(language) if language else self)
 
         if customer and customer.is_all_seeing:
-            qs = root.all().exclude(deleted=True).filter(shop_products__shop=shop)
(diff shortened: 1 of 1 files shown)


shoop/core/models/products.py
```diff
@@ -117,7 +117,8 @@ def list_visible(self, shop, customer, language=None):
         root = (self.language(language) if language else self)
 
         if customer and customer.is_all_seeing:
-            qs = root.all().exclude(deleted=True).filter(shop_products__shop=shop)
+            exclude_q = Q(Q(deleted=True) | Q(mode=ProductMode.VARIATION_CHILD))
```
akx: This can be written as just

```python
exclude_q = Q(deleted=True) | Q(mode=ProductMode.VARIATION_CHILD)
```

PR #198, opened 2015-10-02
Admin: Fix issue where visibility errors caused an error

Pikkupomo:
Admin has no customer, therefore the visibility cannot be checked the same way
as in frontend.

SHOOP-1604

### DIFF
--- a/shoop/core/models/product_shops.py
+++ b/shoop/core/models/product_shops.py
@@ -90,6 +90,7 @@ class ShopProduct(MoneyPropped, models.Model):
     def get_visibility_errors(self, customer):
         if self.product.deleted:
             yield ValidationError(_('This product has been deleted.'), code="product_deleted")
+
         if customer and customer.is_all_seeing:  # None of the further conditions matter for omniscient customers.
             return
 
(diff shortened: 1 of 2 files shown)


shoop/core/models/product_shops.py
```diff
@@ -103,7 +103,7 @@ def get_visibility_errors(self, customer):
                 _('The Product is invisible to users not logged in.'),
                 code="product_not_visible_to_anonymous")
 
-        if self.visibility_limit == ProductVisibility.VISIBLE_TO_GROUPS:
+        if customer and self.visibility_limit == ProductVisibility.VISIBLE_TO_GROUPS:
```
akx: Hmm... This should probably use `is_logged_in`, because `AnonymousUser`s won't have `groups` either.

Also, can you add a test case for this fix?

shoop/core/models/product_shops.py
```diff
@@ -103,7 +104,7 @@ def get_visibility_errors(self, customer):
                 _('The Product is invisible to users not logged in.'),
                 code="product_not_visible_to_anonymous")
 
-        if self.visibility_limit == ProductVisibility.VISIBLE_TO_GROUPS:
+        if customer and customer.is_logged_in and self.visibility_limit == ProductVisibility.VISIBLE_TO_GROUPS:
```
akx: Uh... Customers don't have an `is_logged_in` attribute.

Pikkupomo: ok ill try to wake up sorry :)

shoop_tests/admin/test_product_module.py
```diff
@@ -34,12 +35,15 @@ def test_product_module_search(rf, admin_user):
 
 @pytest.mark.django_db
 def test_product_edit_view_works_at_all(rf, admin_user):
-    get_default_shop()
+    shop = get_default_shop()
+    default_product = get_default_product()
```
akx: As mentioned in :meat_on_bone: :space_invader: :

If you're going to mutate the visibility limit, it'd be better to create a new product for testing

PR #201, opened 2015-10-02
Front: Fix issue with variation children visible in search results

Pikkupomo:
Ref SHOOP-1606


akx:
Could you reword the commit message title to something like

> Simple Search: Don't show variation children in search results

PR #206, opened 2015-10-04
Prevent `Shop` being deleted when image was deleted

Pikkupomo:
`Shop` and `Category` were both missing `on_delete=models.SET_NULL`. This caused them to be deleted also when attached image was deleted in media browser.

Fixes #205

PR #211, opened 2015-10-05
Saner sanity tools

akx:
Unifies some bits of the sanity tools living in `_misc` -- no functional changes as far as I can tell -- and adds a `ForeignKey` checking tool for #206 / SHOOP-1617.


_misc/sanity_utils.py
```diff
@@ -0,0 +1,151 @@
...
+    :param ignored_patterns: fnmatch file patterns to ignore
+    :type ignored_patterns: Iterable[str]
+    :param ignored_dirs: fnmatch directory patterns to ignore
+    :type ignored_dirs: Iterable[str]
+    :param ignored_path_regexps: Path regexps to ignore
+    :type ignored_path_regexps: Iterable[str]
+    :param allowed_extensions: Extensions (really filename suffixes) to ignore (optional)
+    :type allowed_extensions: Iterable[str]|None
+    :return: Iterable of file paths
+    :rtype: Iterable[str]
```
suutari-ai: Wrap at 72 (or 79) as per style guide.

_misc/sanity_utils.py
```diff
@@ -0,0 +1,151 @@
...
+    """
+    if generated_resources is None:
+        generated_resources = set()
+    if isinstance(roots, six.string_types):
+        roots = [roots]
+    if isinstance(allowed_extensions, six.string_types):
+        allowed_extensions = set([allowed_extensions])
```
suutari-ai: Don't use six, since that's not available at our Jenkins host while doing sanity checks.

akx: Already hone.

_misc/sanity_utils.py
```diff
@@ -0,0 +1,159 @@
+import sys
+import fnmatch
+import os
+import posixpath
+import re
+from ast import iter_fields, AST, Attribute, Name
+
+if sys.version_info[0] == 3:
+    string_types = (str,)
```
suutari-ai: Maybe this should be ">= 3" for Python 4 compatibility. :)

akx: Linear curve fitting over Python's major versions and their release dates seems to say we're hitting Python 4 around 1st June 2020, so I don't think we need to worry just yet.

PR #223, opened 2015-10-12
Admin: Transfer sku properly to product

Pikkupomo:
Fix issue with `sku` and `name` `GET` params not being handled in product
view when adding products through search.

Refs SHOOP-1062


akx:
Can you see that the same functionality for names also works correctly?

akx:
Missing test case.

PR #226, opened 2015-10-14
Fix Eslint complaints

akx:
Refs SHOOP-1297

### DIFF
--- a/shoop/notify/static_src/script-editor.js
+++ b/shoop/notify/static_src/script-editor.js
@@ -78,7 +78,7 @@ function Controller() {
         }
     };
     ctrl.setStepItemEditorState = function(state) {
-        if(!!state) {
+        if(state) {
             document.getElementById("step-item-wrapper").style.display = "block";
         } else {
             document.getElementById("step-item-wrapper").style.display = "none";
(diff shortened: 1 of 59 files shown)


shoop/notify/static_src/script-editor.js
```diff
@@ -78,7 +78,7 @@ function Controller() {
         }
     };
     ctrl.setStepItemEditorState = function(state) {
-        if(!!state) {
+        if(state) {
```
suutari-ai: Doesn't eslint care whether it is `if(state)` or `if (state)`? I see both styles being used and IMHO we should use only one (the `if (state)` one if you ask me).

akx: Oh, that's more of a code style thing (which JSCS handles). Eslint deals with semantics, not form. (And with JavaScript, missing or extraneous semicola are a semantic thing, thanks to ASI.)

akx: I'll do a JSCS pass in the same PR.

akx:
@suutari-ai: Updated jscs and the rule file, then ran `jscs --fix` (and hand-fixed the couple overlong lines).

suutari-ai:
@akx Good clean-up! Thanks!

PR #227, opened 2015-10-15
Add autoexpiring versioned cache API (SHOOP-1702)

akx:
This is the versioned cache API as introduced in #62, rebased, extracted and slightly updated too.

### DIFF
--- a/shoop/core/cache/impl.py
+++ b/shoop/core/cache/impl.py
@@ -1,0 +1,151 @@
+# -*- coding: utf-8 -*-
+# This file is part of Shoop.
+#
+# Copyright (c) 2012-2015, Shoop Ltd. All rights reserved.
+#
+# This source code is licensed under the AGPLv3 license found in the
+# LICENSE file in the root directory of this source tree.
+from __future__ import unicode_literals
+import threading
+import time
+
+from django.conf import settings
+from django.core.cache import caches
+from django.core.signals import request_finished
+from django.utils.encoding import force_str
+
+DEFAULT_CACHE_DURATIONS = {
+    # Add default durations for various namespaces here (in seconds)
+}
+_versions = threading.local()
+
+
+def _clear_versions_for_request(**kwargs):
+    _versions.__dict__.clear()
+
+
+request_finished.connect(_clear_versions_for_request, dispatch_uid="shoop.core.cache._clear_versions_for_request")
+
+
+def _get_cache_key_namespace(cache_key):
+    """
+    Split the given cache key by the first colon to get a namespace.
+
+    :param cache_key: Cache key string
+    :type cache_key: str
+    :return: Cache namespace string
+    :rtype: str
+    """
+    return force_str(cache_key).split(str(":"), 1)[0]
+
+
+def get_cache_duration(cache_key):
+    """
+    Determine a cache duration for the given cache key.
+
+    :param cache_key: Cache key string
+    :type cache_key: str
+    :return: Timeout seconds
+    :rtype: int
+    """
+    namespace = _get_cache_key_namespace(cache_key)
+    duration = settings.SHOOP_CACHE_DURATIONS.get(namespace)
+    if duration is None:
+        duration = DEFAULT_CACHE_DURATIONS.get(namespace, settings.SHOOP_DEFAULT_CACHE_DURATION)
+    return duration
+
+
+class VersionedCache(object):
+    def __init__(self, using):
+        """
+        :param using: Cache alias
+        :type using: str
+        """
+        self.using = using
+        self._cache = caches[self.using]
+
+    def bump_version(self, cache_key):
+        """
+        Bump up the cache version for the given cache key/namespace.
+
+        :param cache_key: Cache key or namespace
+        :type cache_key: str
+        """
+        namespace = _get_cache_key_namespace(cache_key)
+        version = str(time.time())
+        self.set(str("_version:") + namespace, version)
+        setattr(_versions, namespace, version)
+
+    def get_version(self, cache_key):
+        """
+        Get the cache version (or None) for the given cache key/namespace.
+
+        The cache version is stored in thread-local storage for
+        the current request, so unless bumped in-request,
+        all `get`s within a single request should get coherently versioned
+        data from the cache.
+
+        :param cache_key: Cache key or namespace
+        :type cache_key: str
+        :return: Version ID or none
+        :rtype: str|None
+        """
+        namespace = _get_cache_key_namespace(cache_key)
+        if not hasattr(_versions, namespace):
+            version = self._cache.get(str("_version:") + namespace)
+            setattr(_versions, namespace, version)
+        else:
+            version = getattr(_versions, namespace, None)
+        return version
+
+    def set(self, key, value, timeout=None, version=None):
+        """
+        Set the value for key `key` in the cache.
+
+        Unlike `django.core.caches[using].set()`, this also derives
+        timeout and versioning information from the key (and cached
+        version data) if the key begins with a colon-separated namespace,
+        such as `foo:bar`.
+
+        :param key: Cache key
+        :type key: str
+        :param value: Value to cache
+        :type value: object
+        :param timeout: Timeout seconds or None (for auto-determination)
+        :type timeout: int|None
+        :param version: Version string or None (for auto-determination)
+        :type version: str|None
+        :param using: Cache alias
+        :type using: str
+        """
+        if timeout is None:
+            timeout = get_cache_duration(key)
+        if version is None:
+            version = self.get_version(key)
+        self._cache.set(key, value, timeout=timeout, version=version)
+
+    def get(self, key, version=None, default=None):
+        """
+        Get the value for key `key` in the cache.
+
+        Unlike `django.core.caches[using].get()`, versioning information
+        can be auto-derived from the key (and cached version data) if the
+        key begins with a colon-separated namespace, such as `foo:bar`.
+
+        :param key: Cache key
+        :type key: str
+        :param version: Version string or None (for auto-determination)
+        :type version: str|None
+        :param default: Default value, if the key is not found
+        :type default: object
+        :param using: Cache alias
+        :type using: str
+        :return: cached value
+        :rtype: object
+        """
+        if version is None:
+            version = self.get_version(key)
+        return self._cache.get(key, default=default, version=version)
+
+    def clear(self):
+        self._cache.clear()
--- a/shoop/core/settings.py
+++ b/shoop/core/settings.py
@@ -136,3 +136,10 @@ SHOOP_TELEMETRY_ENABLED = True
 
 #: The submission URL for Shoop's telemetry (statistics) system
 SHOOP_TELEMETRY_URL = "https://telemetry.shoop.io/collect/"
+
+#: Default cache duration for various caches in seconds
+SHOOP_DEFAULT_CACHE_DURATION = 60 * 30
+
+#: Overrides for default cache durations by key namespace.
+#: These override possible defaults in `shoop.core.cache.impl.DEFAULT_CACHE_DURATIONS`.
+SHOOP_CACHE_DURATIONS = {}
--- a/shoop/core/cache/__init__.py
+++ b/shoop/core/cache/__init__.py
@@ -1,0 +1,38 @@
+# -*- coding: utf-8 -*-
+# This file is part of Shoop.
+#
+# Copyright (c) 2012-2015, Shoop Ltd. All rights reserved.
+#
+# This source code is licensed under the AGPLv3 license found in the
+# LICENSE file in the root directory of this source tree.
+"""
+Utilities for versioned caching and automatic timeout determination.
+
+Versioning works by way of namespaces. Namespaces are the first
+colon-separated part of cache keys.
+
+For instance, the cache keys ``price:10``, ``price:20``, and ``price``
+all belong to the ``price`` namespace and can be invalidated with
+one ``bump_version("price")`` call.
+
+The versions themselves are stored within the cache, within the
+``_version`` namespace.  (As an implementation detail, this allows one
+to invalidate _all_ versioned keys by bumping the version of
+``_version``. Very meta!)
+"""
+
+from .impl import VersionedCache
+
+__all__ = [
+    "bump_version",
+    "clear",
+    "get",
+    "set",
+    "VersionedCache",
+]
+
+_default_cache = VersionedCache(using="default")
+get = _default_cache.get
+set = _default_cache.set
+bump_version = _default_cache.bump_version
+clear = _default_cache.clear
--- a/shoop/front/apps/simple_search/views.py
+++ b/shoop/front/apps/simple_search/views.py
@@ -10,12 +10,12 @@ from __future__ import unicode_literals
 import hashlib
 
 from django import forms
-from django.core.cache import cache
 from django.db.models import Q
 from django.utils.encoding import force_bytes
 from django.views.generic import ListView
 
-from shoop.core.models import Product
+from shoop.core import cache
+from shoop.core.models.products import Product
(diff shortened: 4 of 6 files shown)


suutari-ai:
Please check the commit messages: Add refs and reasonings, Remove final period from titles.

shoop/core/cache.py
```diff
@@ -0,0 +1,161 @@
...
+Versioning works by way of "namespaces". Namespaces are the first colon-separated part
+of cache keys.
+
+For instance, the cache keys `price:10`, `price:20`, and `price` all belong to the
+`price` namespace and can be invalidated with one `bump_version("price")` call.
+
+The versions themselves are stored within the cache, within the `_version` namespace.
+(As an implementation detail, this allows one to invalidate _all_ versioned keys by
+bumping the version of `_version`. Very meta!)
+"""
```
suutari-ai: * Missing license header.
* No need for encoding header.
* Docstrings should be wrapped at col 79 (or I would say preferably around 72 or so, but that's not in our style guide so cannot demand that). -> While wrapping, make the title shorter so it will fit to one line.

shoop/core/cache.py
```diff
@@ -0,0 +1,161 @@
...
+import time
+from django.conf import settings
+from django.utils.encoding import force_text
+from django.core.cache import caches
+from django.core.signals import request_finished
+from threading import local
+
+DEFAULT_CACHE_DURATIONS = {}
+_versions = local()
```
suutari-ai: Function named local just feel very stupid like that. Why not just use it as threading.local() instead?

shoop/core/cache.py
```diff
@@ -0,0 +1,161 @@
...
+The versions themselves are stored within the cache, within the `_version` namespace.
+(As an implementation detail, this allows one to invalidate _all_ versioned keys by
+bumping the version of `_version`. Very meta!)
+"""
+import time
+from django.conf import settings
+from django.utils.encoding import force_text
+from django.core.cache import caches
+from django.core.signals import request_finished
+from threading import local
```
suutari-ai: Group the imports correctly (as isort would do).

shoop/core/cache.py
```diff
@@ -0,0 +1,161 @@
...
+
+def get_cache_duration(cache_key):
+    """
+    Get a suitable cache duration for the given cache key (by the default duration and namespace mapping).
+
+    :param cache_key: Cache key string
+    :type cache_key: str
+    :return: Timeout seconds
+    :rtype: int
+    """
```
suutari-ai: Wrapping & don't include all stuff to the title but make it short title + longer body instead.

shoop/core/cache.py
```diff
@@ -0,0 +1,161 @@
...
+
+        The cache version is stored in thread-local storage for the current request, so unless
+        bumped in-request, all `get`s within a single request should get coherently versioned
+        data from the cache.
+
+        :param cache_key: Cache key or namespace
+        :type cache_key: str
+        :return: Version ID or none
+        :rtype: str|none
+        """
```
suutari-ai: wrapping

shoop/core/settings.py
```diff
@@ -136,3 +136,9 @@
 
 #: The submission URL for Shoop's telemetry (statistics) system
 SHOOP_TELEMETRY_URL = "https://telemetry.shoop.io/collect/"
+
+#: Default cache duration for various caches in seconds
+SHOOP_DEFAULT_CACHE_DURATION = 60 * 30
```
suutari-ai: I would make this datetime.timedelta rather than int, since it's a duration.

shoop/core/cache.py
```diff
@@ -0,0 +1,161 @@
...
+        self._cache.clear()
+
+
+_default_cache = VersionedCache("default")
+get = _default_cache.get
+set = _default_cache.set
+bump_version = _default_cache.bump_version
+clear = _default_cache.clear
+
+# TODO: Add API for PersistentCacheEntry?
```
suutari-ai: This TODO could be removed IMHO, because I don't think we'll need any PersistentCacheEntry, since the Dynamic Configuration could be used for that.

shoop/core/cache.py
```diff
@@ -0,0 +1,161 @@
...
+bumping the version of `_version`. Very meta!)
+"""
+import time
+from django.conf import settings
+from django.utils.encoding import force_text
+from django.core.cache import caches
+from django.core.signals import request_finished
+from threading import local
+
+DEFAULT_CACHE_DURATIONS = {}
```
suutari-ai: I don't see why we would like to have this third level of defaults here. Isn't `settings.SHOOP_CACHE_DURATIONS` and `settings.SHOOP_DEFAULT_CACHE_DURATION` enough?

shoop/core/cache.py
```diff
@@ -0,0 +1,161 @@
...
+def _get_cache_key_namespace(cache_key):
+    """
+    Split the given cache key by the first colon to get a namespace.
+
+    :param cache_key: Cache key string
+    :type cache_key: str
+    :return: Cache namespace string
+    :rtype: str
+    """
+    return force_text(cache_key).split(":")[0]
```
suutari-ai: I'd rather `assert isinstance(cache_key, six.text_type)` than force_text it to avoid unneeded UTF-8/unicode conversions. API caller should know that the API expects text rather than binary and use it with str (or on py2: unicode) rather than bytes.

And, `split(":", 1)[0]`, since we don't need to split more than once.

shoop/core/cache.py
```diff
@@ -0,0 +1,161 @@
...
+
+    def bump_version(self, cache_key):
+        """
+        Bump up the cache version for the given cache key/namespace.
+
+        :param cache_key: Cache key or namespace
+        :type cache_key: str
+        """
+        namespace = _get_cache_key_namespace(cache_key)
+        version = str(time.time())
```
suutari-ai: This could also have the `assert isinstance(cache_key, six.text_type)` and the version should be formed with `"%s" % time.time()` with unicode_literals active so it will be unicode on both py2 and py3.

shoop/core/settings.py
akx: The Django cache API uses seconds, this mirrors that.

shoop/core/cache.py
akx: We already have PersistentCacheEntry, but it doesn't have an API.

suutari-ai: Yeah, for the installation key and the "opt-in for telemetry" setting, but those IMHO should be moved to dynamic configuration when that's implemented.

shoop/core/cache.py
suutari-ai: And PersistentCacheEntry should be deprecated...

Can we just rename the PersistentCacheEntry to ConfigurationEntry so that migrations would move the previous installation key to the new configuration API?

suutari-ai:
This would probably make a nicer API if the VersionedCache class would be moved to a separate module and just the default cache and the get/set/bump_version functions would be in `cache/__init__.py`. I.e. something like this:

`shoop/core/cache/__init__.py`:
```
from .implementation import VersionedCache

__all__ = [
    "VersionedCache",
    "bump_version",
    "clear",
    "get",
    "set",
]

_default_cache = VersionedCache("default")
get = _default_cache.get
set = _default_cache.set
bump_version = _default_cache.bump_version
clear = _default_cache.clear
```

That would make it stand out that you normally don't even create a VersionedCache object yourself, but rather use the cache.get/cache.set functions directly.

shoop/core/settings.py
suutari-ai: Using bare ints rather than typed values has the disadvantage that it's harder to spot what value that magical 151200 is (compared to `timedelta(hours=42)`). We don't have to repeat Django's way of using bare numbers for time intervals here, timedelta has `total_seconds()` for passing the number of seconds to Django.

shoop/core/cache.py
suutari-ai: And PersistentCacheEntry should be deprecated...

Can we just rename the PersistentCacheEntry to ConfigurationEntry so that migrations would move the previous installation key to the new configuration API?

shoop/core/cache.py
akx: `force_text()` begins with

```python
    # Handle the common case first for performance reasons.
    if isinstance(s, six.text_type):
        return s
```

anyway, so I don't think that's much of a problem.

shoop/core/cache.py
akx: Sure. I was just mimicking Django here.

image

shoop/core/cache.py
akx: Because if someone were to overwrite `SHOOP_CACHE_DURATIONS` in their settings for their own uses, there would no longer be _any_ default cache durations.

shoop/core/settings.py
akx: Another point: I really don't want people to have to import `datetime` in their `settings` for this.

shoop/core/cache.py
akx: PersistentCacheEntry was used for caching Matkahuolto pick-up point data, etc., so I think it still has its place.

akx:
Replies to the outdated-diff comments:

* Re making `SHOOP_DEFAULT_CACHE_DURATION` a timedelta: I don't want people to have to `import datetime` in the settings for this.
* Re `PersistentCacheEntry`: `PersistentCacheEntry` was used (in A Previous Incarnation) for caching Matkahuolto pick-up point data, etc., so I think it still has its place.
* Re the `DEFAULT_CACHE_DURATIONS` dict: If someone were to overwrite `SHOOP_CACHE_DURATIONS` in their settings for their own uses, there would no longer be any default cache durations.

shoop/core/cache.py
suutari-ai: Can't this be done with something like this?

`my_settings.py`:

```python
import shoop.core.settings

SHOOP_CACHE_DURATIONS = dict(shoop.core.settings.SHOOP_CACHE_DURATIONS,
   some_cache_namespace=60,
   another_key=3600,
)
```

shoop/core/settings.py
suutari-ai: Fine

shoop/core/cache.py
suutari-ai: Hmm.. maybe you're right. It's better to have the defaults somewhere else than in the settings and so the default setting would be just empty dict as you proposed.

Can you add a comment to `SHOOP_CACHE_DURATIONS` or `SHOOP_DEFAULT_CACHE_DURATION` setting that there is also `shoop.core.cache.DEFAULT_CACHE_DURATIONS` so that people configuring the cache durations know why some cache duration does not have `SHOOP_DEFAULT_CACHE_DURATION` as its value even though it's not in the `SHOOP_CACHE_DURATIONS` dict?

shoop/core/cache.py
suutari-ai: It is, since you can code incorrectly with `bytes` and it works otherwise fine, but is unnecessarily slow. Better teach the API user that we expect `str` here and throw `bytes` back to his face (with assertion failure).

suutari-ai:
OK, I rechecked the comments and your replies and checked the newest diff. Add some more comments, but since they get easily buried to outdated diffs, I add them here too as a check list:

* Can you add a comment to SHOOP_CACHE_DURATIONS or SHOOP_DEFAULT_CACHE_DURATION setting that there is also shoop.core.cache.DEFAULT_CACHE_DURATIONS so that people configuring  the cache durations know why some cache duration does not have SHOOP_DEFAULT_CACHE_DURATION as its value even though it's not in the SHOOP_CACHE_DURATIONS dict?
* Change cache key force_texts to asserts. Reasoning: It is problematic to accept `bytes` here, since you can code incorrectly with bytes and it works otherwise fine, but is unnecessarily slow. Better teach the API user that we expect `str` here and throw `bytes` back to his face (with assertion failure).
* Splitting the cache key should use `split(":", 1)[0]` (i.e. no splitting more than once)
* Add `from __future__ import unicode_literals` to `shoop/core/cache/impl.py`.
* Wrapping of the module docstring in `shoop/core/cache/__init__.py` (I suggest you use your editor to do docstring wrapping automatically, since it seems to be forgotten very easily. Emacs and Sublime Text have Alt+Q for this, I bet PyCharm can do it too.)
* The single quoted example strings in `shoop/core/cache/__init__.py` module docstring (e.g. ``` `price` ```) should be quoted with quote-quote (e.g. ``` ``price`` ```), since they are not references, but inline code and single quotes mean a reference for Sphinx, i.e. Sphinx will try to find a symbol named `price` and link it there.

akx:
I still disagree with the bytes thing, because it's exceedingly easy to pass a bytestring in Python 2 and the speed difference will be utterly negligible.

However since I assume Memcached is a popular cache backend and [this is a thing in Django's memcached cache](https://github.com/django/django/blob/master/django/core/cache/backends/memcached.py#L66-L68), I'll change it to `force_str`. Best of both worlds, yeah? :)

shoop/core/cache/impl.py
```diff
@@ -0,0 +1,151 @@
...
+def _get_cache_key_namespace(cache_key):
+    """
+    Split the given cache key by the first colon to get a namespace.
+
+    :param cache_key: Cache key string
+    :type cache_key: str
+    :return: Cache namespace string
+    :rtype: str
+    """
+    return force_str(cache_key).split(":", 1)[0]
```
suutari-ai: On Python 2 the `force_str` has no effect if you split it with unicode string:
```
In [1]: b'foo:bar'.split(u":", 1)
Out[1]: [u'foo', u'bar']
```

This should work for Py2&Py3:
```python
return force_str(cache_key).split(str(":"), 1)[0]
```

shoop/core/cache/impl.py
```diff
@@ -0,0 +1,151 @@
...
+        """
+        Bump up the cache version for the given cache key/namespace.
+
+        :param cache_key: Cache key or namespace
+        :type cache_key: str
+        """
+        namespace = _get_cache_key_namespace(cache_key)
+        version = "%s" % time.time()
+        self.set("_version:%s" % namespace, version)
+        setattr(_versions, namespace, version)
```
suutari-ai: Same here, unicodes and bytes being mixed. This should fix it:
```python
        namespace = _get_cache_key_namespace(cache_key)
        version = str(time.time())
        self.set(str("_version:") + namespace, version)
        setattr(_versions, namespace, version)
```

shoop/core/cache/impl.py
```diff
@@ -0,0 +1,151 @@
...
+        data from the cache.
+
+        :param cache_key: Cache key or namespace
+        :type cache_key: str
+        :return: Version ID or none
+        :rtype: str|none
+        """
+        namespace = _get_cache_key_namespace(cache_key)
+        if not hasattr(_versions, namespace):
+            version = self._cache.get("_version:%s" % namespace)
```
suutari-ai: Change to
```python
version = self._cache.get(str("_version:") + namespace)
```

shoop/core/cache/impl.py
```diff
@@ -0,0 +1,151 @@
...
+
+        The cache version is stored in thread-local storage for
+        the current request, so unless bumped in-request,
+        all `get`s within a single request should get coherently versioned
+        data from the cache.
+
+        :param cache_key: Cache key or namespace
+        :type cache_key: str
+        :return: Version ID or none
+        :rtype: str|none
```
suutari-ai: `str|None`

suutari-ai:
retest this please

suutari-ai:
Re: force_text/assert/force_str. Fine. Since Python memcached expects the cache keys to be differently typed on Py 2 vs Py 3 (bytes vs unicode), I guess we should do the same here. This should be good now that the other unicode/str stuff is in line with that too.

suutari-ai:
IMHO this looks now very polished. Good work @akx!

Will merge after Mr. Jenkins gives a go.

PR #228, opened 2015-10-15
Xtheme: don't crash when trying to revert unsaved configuration

akx:
Refs SHOOP-1646


suutari-ai:
Nice and elegant fix, and with a test. Merging!

PR #230, opened 2015-10-16
Tuning login with username

tulimaki:
Auth: add error message to login when email is not in use
Password recovery: add better error handling

Refs SHOOP-1670 / SHOOP-1703, SHOOP-1704, SHOOP-1706

### DIFF
--- a/shoop/front/apps/auth/forms.py
+++ b/shoop/front/apps/auth/forms.py
@@ -9,7 +9,7 @@ from django import forms
 from django.contrib.auth import get_user_model
 from django.contrib.auth.forms import AuthenticationForm
 from django.contrib.auth.tokens import default_token_generator
-from django.core.exceptions import MultipleObjectsReturned
+from django.core.exceptions import ObjectDoesNotExist, MultipleObjectsReturned
 from django.core.mail import send_mail
 from django.db.models import Q
 from django.template import loader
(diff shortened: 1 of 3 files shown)


shoop/front/apps/auth/forms.py
```diff
@@ -30,6 +30,11 @@ def clean_username(self):
         if '@' in username:
             try:
                 username = user_model.objects.get(email=username).username
+            except ObjectDoesNotExist:
+                raise forms.ValidationError(
+                    _("Email and password didn't match, please try again."),
```
akx: This should be the same error message (and code) that `AuthenticationForm` itself uses when authentication fails. Otherwise an attacker could use the error message to deduce whether or not the email they entered actually exists in the database.

suutari-ai: That would still be vulnerable to timing attacks, since then the password hashing would not be ran. See https://code.djangoproject.com/ticket/20760 for details.

IMHO it's probaly just best to just remove the `if '@' in username` part and always first query for an user by username and then query by email (even if exists with username to make timing attacks harder) and then select the first one that exists. I.e. something like this:
```python
# Note: Always searches by username AND by email from database, to prevent timing attacks
try:
    user_by_name = user_model._default_manager.get_by_natural_key(username)
except user_model.DoesNotExist:
    user_by_name = None
try:
    user_by_email = user_model._default_manager.get(email=email)
except user_model.DoesNotExist:
    user_by_email = None
except MultipleObjectsReturned:
    raise forms.ValidationError(...)  # or should we mask this too for security? That's against the spec though

if not user_by_name and user_by_email:
    return getattr(user_by_email, user_model.USERNAME_FIELD)
return username
```

suutari-ai:
LGTM

PR #232, opened 2015-10-26
Fix tax calc and implement override groups (SHOOP-1622)

suutari-ai:
Fix tax calculations of default_tax to make added and compounded taxes
work correctly.

Implement override groups for TaxRules.

Refs SHOOP-1636/SHOOP-1622


tulimaki:
:+1:

PR #233, opened 2015-10-28
Dynamic configuration

tulimaki:
Cache: Change bump_version to increment one to current version
Core: Add model for dynamic configurations
Add API for dynamic configuration

Refs SHOOP-1719 / SHOOP-1720 / SHOOP-1698

### DIFF
--- a/shoop/configuration.py
+++ b/shoop/configuration.py
@@ -1,0 +1,137 @@
+# -*- coding: utf-8 -*-
+# This file is part of Shoop.
+#
+# Copyright (c) 2012-2015, Shoop Ltd. All rights reserved.
+#
+# This source code is licensed under the AGPLv3 license found in the
+# LICENSE file in the root directory of this source tree.
+"""
+API for Shoop's Dynamic Configuration.
+
+Idea of the Dynamic Configuration is to allow storing configuration
+values similarly as :ref:`Django settings <django-settings-module>`
+allows, but in a more flexible way: Dynamic Configuration can be changed
+with a simple API and there is no need restart the application server
+after changing a value.
+
+Dynamic configuration values are permanent.  Current implementation
+stores the values with `~shoop.core.models.ConfigurationItem` model into
+database, but that may change in the future.
+
+Configuration values are get and set by a key string.  There is a global
+configuration and a shop specific configuration for each shop.  Values
+in shop specific configuration override the values in global
+configuration.
+"""
+from __future__ import unicode_literals
+
+from shoop.core import cache
+from shoop.core.models import ConfigurationItem
+
+
+def set(shop, key, value):
+    """
+    Set configuration item value for a shop or globally.
+
+    If given `shop` is ``None``, the value of given `key` is set
+    globally for all shops.  Otherwise sets a shop specific value which
+    overrides the global value in configuration of the specified shop.
+
+    :param shop: Shop to set value for, or None to set a global value
+    :type shop: shoop.core.models.Shop|None
+    :param key: Name of the key to set
+    :type key: str
+    :param value: Value to set.  Note: Must be JSON serializable.
+    :type value: Any
+    """
+    ConfigurationItem.objects.update_or_create(
+        shop=shop, key=key, defaults={"value": value})
+    if shop:
+        cache.set(_get_cache_key(shop), None)
+    else:
+        cache.bump_version(_SHOP_CONF_NAMESPACE)
+
+
+def get(shop, key, default=None):
+    """
+    Get configuration value by shop and key.
+
+    Global configuration can be accessed with ``shop=None``.
+
+    :param shop: Shop to get configuration value for, or None
+    :type shop: shoop.core.models.Shop|None
+    :param key: Configuration item key
+    :type key: str
+    :param default:
+      Default value returned if no value is set for given key (globally
+      or in given shop).
+    :type default: Any
+    :return: Configuration value or the default value
+    :rtype: Any
+    """
+    return _get_configuration(shop).get(key, default)
+
+
+def _get_configuration(shop):
+    """
+    Get global or shop specific configuration with caching.
+
+    :param shop: Shop to get configuration for, or None
+    :type shop: shoop.core.models.Shop|None
+    :return: Global or shop specific configuration
+    :rtype: dict
+    """
+    configuration = cache.get(_get_cache_key(shop))
+    if configuration is None:
+        configuration = _cache_shop_configuration(shop)
+    return configuration
+
+
+def _cache_shop_configuration(shop):
+    """
+    Cache global or shop specific configuration.
+
+    Global configuration (`shop` is ``None``) is read first, then `shop`
+    based configuration is updated over that.
+
+    :param shop: Shop to cache configuration for, or None
+    :type shop: shoop.core.models.Shop|None
+    :return: Cached configuration
+    :rtype: dict
+    """
+    configuration = {}
+    configuration.update(_get_configuration_from_db(None))
+    if shop:
+        configuration.update(_get_configuration_from_db(shop))
+    cache.set(_get_cache_key(shop), configuration)
+    return configuration
+
+
+def _get_configuration_from_db(shop):
+    """
+    Get global or shop specific configuration from database.
+
+    :param shop: Shop to fetch configuration for, or None
+    :type shop: shoop.core.models.Shop|None
+    :return: Configuration as it was saved in database
+    :rtype: dict
+    """
+    configuration = {}
+    for conf_item in ConfigurationItem.objects.filter(shop=shop):
+        configuration[conf_item.key] = conf_item.value
+    return configuration
+
+
+_SHOP_CONF_NAMESPACE = str("shop_config")
+
+
+def _get_cache_key(shop):
+    """
+    Get global or shop specific cache key.
+
+    :param shop: Shop to get cache key for, or None
+    :type shop: shoop.core.models.Shop|None
+    :return: Global or shop specific cache key
+    :rtype: str
+    """
+    return str("%s:%s") % (_SHOP_CONF_NAMESPACE, shop.pk if shop else 0)
--- a/shoop/core/cache/impl.py
+++ b/shoop/core/cache/impl.py
@@ -8,6 +8,7 @@
 from __future__ import unicode_literals
 import threading
 import time
+import random
 
 from django.conf import settings
 from django.core.cache import caches
@@ -72,7 +73,7 @@ class VersionedCache(object):
         :type cache_key: str
         """
         namespace = _get_cache_key_namespace(cache_key)
-        version = str(time.time())
+        version = str("%s/%s" % (time.time(), random.random()))
         self.set(str("_version:") + namespace, version)
         setattr(_versions, namespace, version)
 
--- a/shoop/core/models/configurations.py
+++ b/shoop/core/models/configurations.py
@@ -1,0 +1,37 @@
+# -*- coding: utf-8 -*-
+# This file is part of Shoop.
+#
+# Copyright (c) 2012-2015, Shoop Ltd. All rights reserved.
+#
+# This source code is licensed under the AGPLv3 license found in the
+# LICENSE file in the root directory of this source tree.
+from __future__ import unicode_literals
+
+from django.db import models
+from django.utils.encoding import python_2_unicode_compatible
+from django.utils.translation import ugettext_lazy as _
+from jsonfield import JSONField
+
+from ._base import ShoopModel
+from .shops import Shop
+
+
+@python_2_unicode_compatible
+class ConfigurationItem(ShoopModel):
+    shop = models.ForeignKey(Shop, related_name="+", null=True, blank=True, on_delete=models.CASCADE)
+    key = models.CharField(verbose_name=_("key"), max_length=100)
+    value = JSONField(verbose_name=_("value"))
+
+    class Meta:
+        unique_together = [("shop", "key")]
+        verbose_name = _("configuration item")
+        verbose_name_plural = _("configuration items")
+
+    def __str__(self):
+        if self.shop:
+            return _("%(key)s for shop %(shop)s") % vars(self)
+        else:
+            return _("%(key)s (global)") % vars(self)
+
+    def __repr__(self):
+        return '<%s "%s" for %r>' % (type(self).__name__, self.key, self.shop)
--- a/shoop_tests/core/test_configurations.py
+++ b/shoop_tests/core/test_configurations.py
@@ -1,0 +1,134 @@
+# -*- coding: utf-8 -*-
+# This file is part of Shoop.
+#
+# Copyright (c) 2012-2015, Shoop Ltd. All rights reserved.
+#
+# This source code is licensed under the AGPLv3 license found in the
+# LICENSE file in the root directory of this source tree.
+import pytest
+
+from shoop import configuration
+from shoop.core import cache
+from shoop.core.models import ConfigurationItem
+from shoop.testing.factories import get_default_shop
+
+
+@pytest.mark.django_db
+def test_simple_set_and_get_with_shop():
+    shop = get_default_shop()
+    configuration.set(shop, "answer", 42)
+    assert configuration.get(shop, "answer") == 42
+
+    assert configuration.get(shop, "non-existing") is None
+    configuration.set(shop, "non-existing", "hello")
+    assert configuration.get(shop, "non-existing") == "hello"
+
+
+@pytest.mark.django_db
+def test_simple_set_and_get_without_shop():
+    configuration.set(None, "answer", 42)
+    assert configuration.get(None, "answer") == 42
+
+    assert configuration.get(None, "non-existing") is None
+    configuration.set(None, "non-existing", "hello")
+    assert configuration.get(None, "non-existing") == "hello"
+
+
+@pytest.mark.django_db
+def test_simple_set_and_get_cascading():
+    shop = get_default_shop()
+    configuration.set(None, "answer", 42)
+    assert configuration.get(None, "answer") == 42
+    assert configuration.get(shop, "answer", 42)
+
+    assert configuration.get(None, "non-existing") is None
+    assert configuration.get(shop, "non-existing") is None
+    configuration.set(shop, "non-existing", "hello")
+    assert configuration.get(None, "non-existing") is None
+    assert configuration.get(shop, "non-existing") == "hello"
+
+    assert configuration.get(None, "foo") is None
+    assert configuration.get(shop, "foo") is None
+    configuration.set(None, "foo", "bar")
+    configuration.set(shop, "foo", "baz")
+    assert configuration.get(None, "foo") == "bar"
+    assert configuration.get(shop, "foo") == "baz"
+
+
+@pytest.mark.django_db
+def test_configuration_gets_saved():
+    configuration.set(None, "x", 1)
+    assert configuration.get(None, "x") == 1
+    configuration.set(None, "x", 2)
+    assert configuration.get(None, "x") == 2
+    configuration.set(None, "x", 3)
+    assert configuration.get(None, "x") == 3
+    conf_item = ConfigurationItem.objects.get(shop=None, key="x")
+    assert conf_item.value == 3
+
+
+@pytest.mark.django_db
+def test_configuration_set_and_get():
+    cache.clear()
+    shop = get_default_shop()
+    test_conf_data = {"data": "test"}
+    configuration.set(shop, "key", test_conf_data)
+
+    # Get the configuration via configuration API
+    assert configuration.get(shop, "key") == test_conf_data
+    # Check that configuration is saved to database
+    assert ConfigurationItem.objects.get(shop=shop, key="key").value == test_conf_data
+
+
+@pytest.mark.django_db
+def test_configuration_update():
+    cache.clear()
+    shop = get_default_shop()
+    configuration.set(shop, "key1", {"data": "test1"})
+    configuration.set(shop, "key2", {"data": "test2"})
+    configuration.set(shop, "key3", {"data": "test3"})
+    assert configuration.get(shop, "key1").get("data") == "test1"
+    assert configuration.get(shop, "key3").get("data") == "test3"
+
+    # Update configuration
+    configuration.set(shop, "key3", {"data": "test_bump"})
+    assert configuration.get(shop, "key3").get("data") == "test_bump"
+
+
+@pytest.mark.django_db
+def test_global_configurations():
+    cache.clear()
+    shop = get_default_shop()
+    configuration.set(None, "key1", {"data": "test1"})
+    configuration.set(shop, "key2", {"data": "test2"})
+
+    # key1 from shop should come from global configuration
+    assert configuration.get(shop, "key1").get("data") == "test1"
+    # key2 shouldn't be in global configurations
+    assert configuration.get(None, "key2") is None
+
+    # Update global configuration
+    configuration.set(None, "key1", {"data": "test_bump"})
+    assert configuration.get(shop, "key1").get("data") == "test_bump"
+
+    # Override shop data for global key1
+    configuration.set(shop, "key1", "test_data")
+    assert configuration.get(shop, "key1") == "test_data"
+
+    # Update shop configuration for global key1
+    configuration.set(shop, "key1", "test_data1")
+    assert configuration.get(shop, "key1") == "test_data1"
+
+
+@pytest.mark.django_db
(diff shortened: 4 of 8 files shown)


shoop/core/cache/impl.py
```diff
@@ -72,7 +71,8 @@ def bump_version(self, cache_key):
         :type cache_key: str
         """
         namespace = _get_cache_key_namespace(cache_key)
-        version = str(time.time())
+        current_version = self.get_version(cache_key) or 1
+        version = current_version + 1
```
suutari-ai: Should the version still be string?

I.e. this would be something like:
 ```python
lastver = self.get_version(cache_key)
lastver_int = int(lastver) if isinstance(lastver, str) and lastver.isdigit() else 1
version = str(lastver_int + 1)
```

File: shoop/core/models/configuration.py [hunk 2]
```diff
@@ -0,0 +1,30 @@
...
+from django.db import models
+from django.utils.encoding import python_2_unicode_compatible
+from django.utils.translation import ugettext_lazy as _
+from jsonfield import JSONField
+
+from _base import ShoopModel
+
+
+@python_2_unicode_compatible
+class Configuration(ShoopModel):
```
suutari-ai: This should be ConfigurationItem IMHO, since Configuration is a collection of ConfigurationItems.

shoop/core/models/configuration.py
```diff
@@ -0,0 +1,30 @@
...
+    shop = models.ForeignKey("Shop", related_name="configurations", null=True, blank=True, on_delete=models.CASCADE)
```
suutari-ai: I don't think we need a related name for this. I would set it to just `"+"` here.

shoop/core/models/configuration.py
```diff
@@ -0,0 +1,30 @@
...
+    key = models.CharField(verbose_name=_("key"), max_length=60)
+    data = JSONField(verbose_name=_("data"))
+
+    class Meta:
+        unique_together = (("shop", "key"),)
+        verbose_name = _('configuration')
+        verbose_name_plural = _('configurations')
```
suutari-ai: _("configuration item") and _("configuration items"), i.e. fix quotes and the naming.

shoop/core/models/configuration.py
```diff
@@ -0,0 +1,30 @@
...
+
+    def __str__(self):
+        return self.key
```
suutari-ai: Not very good str. It should include the shop and the key at least. How about:
```
if self.shop:
    return _("%(conf_item_key)s for shop %(shop)s") % dict(
        conf_item_key=self.key, shop=self.shop)
else:
    return _("%(conf_item_key)s (global)") % dict(conf_item_key=self.key)
```

shoop/configuration.py
```diff
@@ -0,0 +1,102 @@
...
+
+from shoop.core import cache
+from shoop.core.models import Shop, Configuration
+
+
+SHOP_CONFIGURATION_CACHE_PREFIX = "shop_configuration_%s"
+
+
+def get_cache_key(shop):
+    return SHOP_CONFIGURATION_CACHE_PREFIX % shop.pk if shop else ""
```
akx: Add parens around after % (the precedence is kinda ambiguous as it is)

shoop/configuration.py
```diff
@@ -0,0 +1,102 @@
...
+
+
+def set(shop, key, data):
+    """
+    Set shop, key, data combination to configurations.
```
akx: Misformatted docstring (missing empty line after summary). Also I'm not quite sure what the summary is trying to say.

shoop/configuration.py
```diff
@@ -0,0 +1,102 @@
...
+    After configuration create or update bump cache version
+    or invalidate cache for all shops
+
+    :param shop: shop for configuration
```
akx: Missing `:type`s

shoop/configuration.py
```diff
@@ -0,0 +1,102 @@
...
+
+def cache_shop_configurations(shop):
+    """
+    Cache shop configurations
+
+    :param shop: shop which configurations is cached
+    :return: configurations cached
+    """
+    configurations = {}
+    configurations.update(get_configurations_from_db(None))
```
akx: Would be nice to have comments explaining the cascading logic here.

shoop/configuration.py
```diff
@@ -0,0 +1,102 @@
...
+    if shop:
+        configurations.update(get_configurations_from_db(shop))
+    cache.set(get_cache_key(shop), configurations)
+    return configurations
+
+
+def get_configurations_from_db(shop):
```
akx: This should probably be an underscore-prefixed function, as no outside caller should touch it.

shoop/configuration.py
```diff
@@ -0,0 +1,102 @@
...
+    """
+    Get configurations from database
+
+    :param shop: shop which configuration is fetched. Also None allowed
+    :return: all saved configurations saved in database as a dict
+    """
+    query = {"shop": shop} if shop else {"shop__isnull": True}
+    configurations = {}
+    for conf in Configuration.objects.filter(**query):
```
akx: This can be more succinctly said as 
```python
return dict(Configuration.objects.filter(**query).values_list("key", "data"))
```

shoop/configuration.py
```diff
@@ -0,0 +1,102 @@
...
+    """
+    Set shop, key, data combination to configurations.
+    After configuration create or update bump cache version
+    or invalidate cache for all shops
+
+    :param shop: shop for configuration
+    :param key: configuration key
+    :param data: data set to configuration
+    """
+    conf, _ = Configuration.objects.get_or_create(shop=shop, key=key)
```
akx: This will end up saving the object twice, once with no `data` and again with `data`.

That seems kind of wasteful, and since this isn't an atomic operation, you could possibly end up with a race condition of some sort...

shoop/configuration.py
```diff
@@ -0,0 +1,102 @@
...
+    conf.data = data
+    conf.save()
+
+    if shop:
+        cache.bump_version(get_cache_key(shop))
```
akx: Uhm, I think you've misunderstood how the versioning API works. It works for key _prefixes_ (namespaces), not full keys.

You'll really just want to do
```python
cache.set(get_cache_key(shop), None)
```
here, _or_ refactor the code to have a common namespace (`shop_config` sounds good), so the actual cache keys are of the format `shop_config:123` (for shop id 123), after which you can do
```python
if shop:
    cache.set(get_cache_key(shop), None)  # single-shop setting changed
else:
    bump_version("shop_config")  # global setting changed
```
instead of having to iterate through all shops to invalidate the global setting layer.

shoop/configuration.py
```diff
@@ -0,0 +1,102 @@
...
+from __future__ import unicode_literals
+
+from shoop.core import cache
+from shoop.core.models import Shop, Configuration
+
+
+SHOP_CONFIGURATION_CACHE_PREFIX = "shop_configuration_%s"
+
+
+def get_cache_key(shop):
```
akx: This should probably also be a private (underscore-prefixed) function.

shoop/core/cache/impl.py
suutari-ai: Maybe it's better to use int in this case.  Anyway.. the docstring of get_version() should be updated, because now it returns int rather than str.

akx: This is also prone to race conditions, since this is not an atomic operation and multiple readers may be incrementing the version concurrently. Consider:

```
       Request 1       Request 2
       =========       =========
 T |   get_version
 i |    ==> 1
 m |                   get_version
 e |                    ==> 1      <-- oops, getting
   |                                   an old version while
 | |   increment       increment       Request 1 is already
 | |    ==> 2           ==> 2          bumping things...
 V |
   |   set_version
   |    ==> 2
   |                   set_version
   |                    ==> 2      <-- this should've been 3!
```

The correct fix is to either
* use a more high-resolution timer ([since Python 3.3, there're multiple high-res timers available on Unices](https://docs.python.org/3/library/time.html#time.clock_gettime)) if the current timer is the problem (as I understand from the commit message)
* or, more portably, to append a random string (Django's `get_random_string()` maybe?) to the version.

shoop/core/models/configuration.py
akx: I noticed the same. My first gut feeling was `ConfigurationEntry`.

shoop/core/models/configuration.py
akx: :+1:

shoop/core/models/configuration.py
akx:
```python
if self.shop:
    return _("%(key)s for shop %(shop)s") % vars(self)
else:
    return _("%(key)s (global)") % vars(self)
```
would be more concise and pretty much as readable.

shoop_tests/core/test_configurations.py
```diff
@@ -0,0 +1,64 @@
...
+import pytest
+
+from shoop import configuration
+from shoop.core import cache
+from shoop.core.models import Configuration
+from shoop.testing.factories import get_default_shop
+
+
+@pytest.mark.django_db
+def test_configurations_1():
```
akx: Test names should be more descriptive.

shoop_tests/core/test_configurations.py
```diff
@@ -0,0 +1,64 @@
...
+    cache.clear()
+    shop = get_default_shop()
+    test_conf_data = {"data": "test"}
+    configuration.set(shop, "key", test_conf_data)
+
+    assert configuration.get(shop, "key") == test_conf_data
+    assert Configuration.objects.get(shop=shop, key="key").data == test_conf_data
+    assert cache.get(configuration.SHOP_CONFIGURATION_CACHE_PREFIX % shop.pk).get("key") == test_conf_data
```
akx: The fact that the configuration API is backed by the cache shouldn't be very public.

Of course you'll want to test that stuff gets cached (again, more descriptive test names, please?), but is it a good idea that all three tests touch the cache?

shoop/configuration.py
tulimaki: values_list doesn't return objects. So it is not usable here I think.

shoop/core/models/configuration.py
suutari-ai: :+1:

shoop/core/cache/impl.py
suutari-ai: Should we just use Django cache's `incr_version()`? (See https://docs.djangoproject.com/en/1.8/topics/cache/#cache-versioning)

shoop/configuration.py
akx: You don't need the actual `Configuration` objects anyway, you're just accessing two `values` from them.

shoop/core/cache/impl.py
akx: No -- it only deals with single cache keys, not namespaces. (`shoop.core.cache` does use Django's cache versioning as the underlying mechanism.)

shoop/configuration.py
tulimaki: But if data is {"key123": "value"} this returns {"key", '{"key123": "value"}'} instead of {"key", {"key123": "value"}} which is not good imo.

akx: Does it? Okay then, never mind.

shoop/configuration.py
suutari-ai: Yep, and moved to bottom of this file, since public before private is better order.

shoop/configuration.py
suutari-ai: Shouldn't it be `SHOP_CONFIGURATION_CACHE_PREFIX + str(shop.pk)` anyway? (it's a prefix, not a template)

i.e. with parenthesis too:
```python
return (SHOP_CONFIGURATION_CACHE_PREFIX + str(shop.pk)) if shop else ""
```

shoop/configuration.py
```diff
@@ -0,0 +1,102 @@
...
+#
+# This source code is licensed under the AGPLv3 license found in the
+# LICENSE file in the root directory of this source tree.
+from __future__ import unicode_literals
+
+from shoop.core import cache
+from shoop.core.models import Shop, Configuration
+
+
+SHOP_CONFIGURATION_CACHE_PREFIX = "shop_configuration_%s"
```
suutari-ai: Should be
```python
SHOP_CONFIGURATION_CACHE_PREFIX = str("shop_configuration_")
```
I.e. always str (bytes on Py2, unicode on Py3) and really a prefix rather than template.

suutari-ai: And this too should be private... or actually... do we even need a global variable for this? why not just dump that to `_get_cache_key` as is?

shoop/configuration.py
akx:
```python
return SHOP_CONFIGURATION_CACHE_PREFIX % (shop.pk or 0)
```

would also be nice.

shoop/configuration.py
suutari-ai: `data` should be renamed to `value`.

The docstring could be something like
```python
def set(shop, key, value):
    """
    Set configuration item value for a shop or globally.

    If given `shop` is ``None``, the value of given `key` is set
    globally for all shops.  Otherwise sets a shop specific value which
    overrides the global value in configuration of the specified shop.

    :type shop: shoop.core.models.Shop|None
    :param shop: Shop to set value for
    :type key: str
    :param key: Name of the key to set
    :type value: Any
    :param value: Value to set.  Note: Must be JSON serializable.
    """
```

shoop/configuration.py
suutari-ai: Solution is to use get_or_create with `defaults` and if it does not create a new item, then set and save it.
```python
    (conf_item, created) = ConfigurationItem.objects.get_or_create(
        shop=shop, key=key,
        defaults=dict(value=value))
    if not created:
        conf_item.value = value
        conf_item.save()
```

shoop/core/models/configuration.py
```diff
@@ -0,0 +1,30 @@
...
+from jsonfield import JSONField
+
+from _base import ShoopModel
+
+
+@python_2_unicode_compatible
+class Configuration(ShoopModel):
+    shop = models.ForeignKey("Shop", related_name="configurations", null=True, blank=True, on_delete=models.CASCADE)
+    key = models.CharField(verbose_name=_("key"), max_length=60)
+    data = JSONField(verbose_name=_("data"))
```
suutari-ai: Rename data to value.

shoop/configuration.py
```diff
@@ -0,0 +1,102 @@
...
+    if shop:
+        cache.bump_version(get_cache_key(shop))
+    else:
+        invalidate_cache()
+
+
+def invalidate_cache():
+    """
+    Invalidate cache for all configuration keys
+    """
```
suutari-ai: Should this really be public? IMHO it should be private.

shoop/configuration.py
```diff
@@ -0,0 +1,102 @@
...
+    cache.set(get_cache_key(None), None)
+    for shop in Shop.objects.all():
+        cache.set(get_cache_key(shop), None)
+
+
+def get(shop, key, default=None):
+    """
+    Get configuration data based on shop and key.
```
suutari-ai: Get configuration value by shop and key.

shoop/configuration.py
```diff
@@ -0,0 +1,102 @@
...
+
+    :param shop: shop for configuration fetched
+    :param key: configuration key
+    :param default: default data returned if no configuration available
+    :return: configuration data or default
```
suutari-ai: Missing type specifiers (for params and for rtype).

Should mention if shop=None is allowed here or not.

shoop/configuration.py
```diff
@@ -0,0 +1,102 @@
...
+    return setting_cache.get(key, default)
+
+
+def get_configuration_cache(shop):
+    """
+    Get cached configurations for shop
+
+    :param shop: shop object for settings
+    :return: settings for shop
+    """
```
suutari-ai: IMHO this too should be private.

shoop/configuration.py
```diff
@@ -0,0 +1,102 @@
...
+    return configurations
+
+
+def cache_shop_configurations(shop):
+    """
+    Cache shop configurations
+
+    :param shop: shop which configurations is cached
+    :return: configurations cached
+    """
```
suutari-ai: IMHO this too should be private.

shoop/configuration.py
suutari-ai: Cascading logic should be explained in module docstring.

shoop/configuration.py
suutari-ai: :+1:

shoop/configuration.py
akx: There, too, but the implementation should be commented in kind.

shoop/configuration.py
```diff
@@ -0,0 +1,102 @@
...
+
+
+def get_configurations_from_db(shop):
+    """
+    Get configurations from database
+
+    :param shop: shop which configuration is fetched. Also None allowed
+    :return: all saved configurations saved in database as a dict
+    """
+    query = {"shop": shop} if shop else {"shop__isnull": True}
```
suutari-ai: Doesn't `.filter(shop=None)` do the same as `.filter(shop__isnull=True)` ?

If so, you could drop this `query` variable and just use
```python
    for conf_item in ConfigurationItem.objects.filter(shop=shop):
```
below.

shoop/configuration.py
```diff
@@ -0,0 +1,107 @@
...
+
+    After configuration create or update invalidate cache
+
+    :param shop: shop for configuration
+    :type shop: shoop.core.models.Shop
+    :param key: configuration key
+    :type key: str
+    :param data: data set to configuration
+    :type data: JSON serializable object
+    """
```
suutari-ai: I had a better docstring written for this in previous rounds comments. Why not use it?

shoop_tests/core/test_configurations.py
suutari-ai: Exactly. The caching should be as invisible as possible.

shoop/configuration.py
```diff
@@ -0,0 +1,107 @@
...
+
+from shoop.core import cache
+from shoop.core.models import ConfigurationItem
+
+
+SHOP_CONFIGURATION_NAMESPACE = "shop_config"
+
+
+def _get_cache_key(shop):
+    return "%s:%s" % (SHOP_CONFIGURATION_NAMESPACE, shop.pk if shop else "")
```
suutari-ai: Move this and SHOP_CONFIGURATION_NAMESPACE to bottom. Make the global variable private, i.e. `_SHOP_CONFIGURATION_NAMESPACE`.

shoop/configuration.py
```diff
@@ -0,0 +1,107 @@
...
+
+def get(shop, key, default=None):
+    """
+    Get configuration value by shop and key.
+
+    :param shop: shop for configuration fetched
+    :type shop: shoop.core.models.Shop
+    :param key: configuration key
+    :type key: str
+    :param default: default data returned if no configuration available
```
suutari-ai: default value returned if no value is set for given shop and key

shoop/core/cache/impl.py
```diff
@@ -72,7 +73,7 @@ def bump_version(self, cache_key):
         :type cache_key: str
         """
         namespace = _get_cache_key_namespace(cache_key)
-        version = str(time.time())
+        version = str("%s/%s") % (time.time(), random.random())
```
akx: I think the `str()` call should encompass the whole expression:

```python
str("%s/%s" % (time.time(), random.random()))
```

akx:
LGTM (at long last) :lollipop: 

Won't merge before @suutari-ai 's blessing though.

shoop/configuration.py
```diff
@@ -0,0 +1,125 @@
...
+# LICENSE file in the root directory of this source tree.
+"""
+API for shoop dynamic configurations. Values is stored to database in
+`shoop.core.configurations.ConfigurationItem model`. Shop specific
+configurations is cached.
+
+Shop can be None which means that configuration item is global. For shop
+first all global configurations is read and then shop specific
+configuration which may override global configurations.
+"""
```
suutari-ai: * Module docstring should have a title too.
* It's "dynamic configuration", not "dynamic configurations".
* "Values is" -> "Values are"
* Drop the "model" word from the quotes or Sphinx won't find the class: It should be ``` `ConfigurationItem` model ``` rather than ``` `ConfigurationItem model` ```)

Maybe just change the first sentence to title:
```python
API for Shoop's Dynamic Configuration.

Values are stored to database in `ConfigurationItem` model. ...
```

shoop/configuration.py
```diff
@@ -0,0 +1,125 @@
...
+    globally for all shops.  Otherwise sets a shop specific value which
+    overrides the global value in configuration of the specified shop.
+
+    :param shop: Shop to set value for
+    :type shop: shoop.core.models.Shop|None
+    :param key: Name of the key to set
+    :type key: str
+    :param value: Value to set.  Note: Must be JSON serializable.
+    :type value: Any
+    """
```
suutari-ai: I still think the value should be named `value`, not `data`.

suutari-ai: I.e. in the parameter name too.

shoop/configuration.py
```diff
@@ -0,0 +1,125 @@
...
+    conf, _ = ConfigurationItem.objects.update_or_create(shop=shop, key=key, defaults={"value": data})
```
suutari-ai: This does not update the value if it was already set. The `if not created:` block in [my example](https://github.com/shoopio/shoop/pull/233#discussion_r43363876) was there for a reason.

suutari-ai: Add a test case that value gets updated in the database. Even if set twice.

shoop/configuration.py
```diff
@@ -0,0 +1,125 @@
...
+    Get configuration value by `shop` and `key`.
+
+    :param shop: shop for configuration fetched
+    :type shop: shoop.core.models.Shop|None
+    :param key: configuration key
+    :type key: str
+    :param default: default value returned if no value is set for given key (globally or in given shop)
+    :type default: object
+    :return: configuration data or default
+    :rtype: object
```
suutari-ai: `Any` is the correct rtype for function that may return anything.

shoop/configuration.py
```diff
@@ -0,0 +1,125 @@
...
+    """
+    Get configuration value by `shop` and `key`.
+
+    :param shop: shop for configuration fetched
+    :type shop: shoop.core.models.Shop|None
+    :param key: configuration key
+    :type key: str
+    :param default: default value returned if no value is set for given key (globally or in given shop)
+    :type default: object
+    :return: configuration data or default
```
suutari-ai: configuration data -> configuration value

shoop/configuration.py
```diff
@@ -0,0 +1,125 @@
...
+
+def get(shop, key, default=None):
+    """
+    Get configuration value by `shop` and `key`.
+
+    :param shop: shop for configuration fetched
+    :type shop: shoop.core.models.Shop|None
+    :param key: configuration key
+    :type key: str
+    :param default: default value returned if no value is set for given key (globally or in given shop)
```
suutari-ai: wrap this

shoop/configuration.py
suutari-ai: The model is `shoop.core.models.ConfigurationItem`, not `shoop.core.configurations.ConfigurationItem`.

shoop/core/models/configurations.py
```diff
@@ -0,0 +1,33 @@
...
+# This source code is licensed under the AGPLv3 license found in the
+# LICENSE file in the root directory of this source tree.
+from __future__ import unicode_literals
+
+from django.db import models
+from django.utils.encoding import python_2_unicode_compatible
+from django.utils.translation import ugettext_lazy as _
+from jsonfield import JSONField
+
+from _base import ShoopModel
```
suutari-ai: Should be `from ._base` (it's explicit relative import).  This does not even work for Py3 I guess.

suutari-ai:
@tulimaki I'll look into these issues I've listed here and create a few fixup commits for you. Will push them somewhere soon...

shoop/configuration.py
suutari-ai: Sorry, I was wrong about this. Didn't notice that the call was `update_or_create` (thought it was `get_or_create`). Anyway.. the conf is unneeded and so the return value may be totally discarded.

suutari-ai:
@tulimaki Please pull my changes from my [dynamic-configuration branch](https://github.com/suutari-ai/shoop/tree/dynamic-configuration) with
```
git pull git@github.com:suutari-ai/shoop.git dynamic-configuration
```
and rebase the changes in with `git rebase -i --autosquash master`.

PR #240, opened 2015-10-30
Remove PriceTaxContext and related clean-up (SHOOP-1735)

suutari-ai:
The context classes for pricing and taxing are not as interleaved as I
first thought.  There seems to be no reason for PriceTaxContext after
all, so therefore remove it.  Also do much clean-up related to contexts
in general, because that was needed to so that I can better understand
what kind of contexts we need and where/when.

See the individual commit messages too.

Refs SHOOP-1636/SHOOP-1735


tulimaki:
:+1:

PR #241, opened 2015-10-30
Remove Priceful.total_price (SHOOP-1734)

suutari-ai:
The Priceful interface was introduced a while ago, but the refactoring
was not done through.  This step of unifying `price`/`total_price`
usages was left, so do it now for cleaning up our code base.

Also includes couple somewhat unrelated fixes that were found while
doing this.

Refs SHOOP-1636/SHOOP-1734


tulimaki:
Looks good

PR #243, opened 2015-11-05
Unvendor Classic Grey fonts (SHOOP-1759)

suutari-ai:
Remove the vendored copies of Font Awesome fonts and add a gulp rule
that copies the fonts from bower_components.

Also copy glyphicons-halflings-regular font from bowered Bootstrap,
because that is needed.

Refs SHOOP-1686/SHOOP-1759


akx:
Seems legit!

suutari-ai:
@akx And thanks for the Gulp advices!

PR #245, opened 2015-11-05
Make it possible to create orders via the admin

akx:


shoop/admin/modules/orders/json_order_creator.py
```diff
@@ -0,0 +1,187 @@
...
+#
+# Copyright (c) 2012-2015, Shoop Ltd. All rights reserved.
+#
+# This source code is licensed under the AGPLv3 license found in the
+# LICENSE file in the root directory of this source tree.
+from __future__ import unicode_literals
+from copy import deepcopy
+from django.core.exceptions import ValidationError, ObjectDoesNotExist
+from django.utils.translation import ugettext as _
+from shoop.core.models import (
```
tulimaki: Add blank line between django and shoop imports.

tulimaki:
Most of the new files is missing Shoop license information.

akx:
@tulimaki: I ran the license tool, isort and autopep8, but it seems that made your comments disappear. D:

akx:
Oh, wait, some cache magic got me a snapshot:

image

akx:
And to address the last two:

* Context: As far as I know, yes? Though in some cases, somehow, a request gets automagically converted into a context. Maybe @suutari-ai can clarify.
* Nowhere. We don't have a translation system in place for Javascript files.

shoop/admin/modules/orders/static_src/create/view/lines.js
```diff
@@ -0,0 +1,91 @@
...
+        return m("tr", [
+            m("td", selectBox(line.type, function () {
+                store.dispatch(setLineProperty(line.id, "type", this.value));
+            }, LINE_TYPES)),
+            m("td", line.sku),
+            m("td", {colspan: showPrice ? 1 : priceCells.length + 1}, text),
+            (showPrice ? priceCells : null),
+            m("td", m("button.btn.btn-xs.btn-danger", {
+                onclick: function () {
+                    store.dispatch(deleteLine(line.id));
```
Pikkupomo: `Uncaught TypeError: _actions.deleteLine is not a function` when clicking delete-button (trashcan). Points to this line

Pikkupomo:
If I add variation parent into order, only that is added without possibility to select the variation child I actually want to add

Pikkupomo:
* Order cannot be created if item has no supplier (product is not orderable). Error: `Could not create order: Order line has no supplier`. Error is in wrong color (blue) should be red.
* Tried to make discount-line, gave price of `-20` after saving and getting the error mentioned previously (supplier error) the price was 0 (not the original `-20`

akx:
> * If I add variation parent into order, only that is added without possibility to select the variation child I actually want to add

Variation parents aren't orderable in the first place. There's currently no filter in the product browser to only choose from products that are actually orderable, and that's out of scope for this PR.

> * Order cannot be created if item has no supplier (product is not orderable).

As it should be.

> Error: `Could not create order: Order line has no supplier`. Error is in wrong color (blue) should be red.

I suppose the UX for errors should be much better in the first place (for instance, they could be shown on the related line), but that could be improved later.

> * Tried to make discount-line, gave price of `-20` after saving and getting the error mentioned previously (supplier error) the price was 0 (not the original `-20`

Discount lines weren't in the original spec at all.

shoop/admin/modules/orders/static_src/create/view/lines.js
akx: Oops, yeah, was missing that action. Added.

PR #246, opened 2015-11-09
Better i18n extraction

akx:


_misc/gather_i18n.py
```diff
@@ -0,0 +1,82 @@
+#!/usr/bin/env python3
+from __future__ import unicode_literals
+
+import argparse
+import logging
+import click
```
Pikkupomo: I think this is missing from requirements

Pikkupomo:
When the missing requirement is fixed, LGTM and seems to extract alot of stuff

_misc/gather_i18n.py
akx: It's only a requirement for these development CLI tools. You can install it by hand, `pip install click` will do.

akx:
> alot of stuff

akx:
@Pikkupomo : removed the dependency on Click altogether.

PR #247, opened 2015-11-12
Themes: Remove default theme

tulimaki:
* Default theme moved to it's own repository as an addon
* Set Classic Gray as a default theme for tests
* Fix front/test_product_flow fill addresses to use correct form

Refs SHOOP-1761


shoop_tests/conftest.py
```diff
@@ -24,8 +24,8 @@ def pytest_configure(config):
 def pytest_runtest_call(item):
     # All tests are run with a default theme override `shoop.themes.default_theme.DefaultTheme`.
```
akx: This comment should be updated to mirror the actual state of the code.

suutari-ai:
Squash all three commits to single commit, since no commit should introduce any breakages.

suutari-ai:
While squashing, don't lose the commit messages, please. They're good IMO. (I.e. List the original commit messages in the squashed commit message with bullet points or something similar.)

PR #248, opened 2015-11-12
Move apply_request_middleware to testing

suutari-ai:
In `shoop.testing.factories` there was an import from `shoop_tests` for
`apply_request_middleware` function.  This caused `shoop.testing` to
depend on BeautifulSoup, since `shoop_tests.utils` imports `bs4`.  This
dependency is undesirable and not even stated in `setup.py`.

Fix this by moving the `apply_request_middleware` to
`shoop.testing.utils` and update its users accordingly.


tulimaki:
LGTM

PR #249, opened 2015-11-13
Enable js translations

tulimaki:
* Enable javascript translations for admin modules, notify and xtheme.

### DIFF
--- a/shoop/admin/urls.py
+++ b/shoop/admin/urls.py
@@ -11,8 +11,9 @@ from __future__ import unicode_literals
 import warnings
 
 import django.contrib.auth.views as auth_views
-from django.conf.urls import patterns
+from django.conf.urls import patterns, url
 from django.contrib.auth import logout as do_logout
+from django.views.i18n import javascript_catalog
 
 from shoop.admin.module_registry import get_module_urls
 from shoop.admin.utils.urls import AdminRegexURLPattern, admin_url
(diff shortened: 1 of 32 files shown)


shoop/admin/urls.py
```diff
@@ -52,10 +53,16 @@ def get_urls():
         ),
     ])
 
+    # Add Django javascript catalog url
+    urls.append(url(r'^jsi18n/$', javascript_catalog, {'packages': ('shoop',)}, name='js-catalog'))
+
     for u in urls:  # pragma: no cover
         if not isinstance(u, AdminRegexURLPattern):
             warnings.warn("Admin URL %r is not an AdminRegexURLPattern" % u)
```
akx: Won't the addition of the catalog above make this warning occur always? :(

shoop/admin/urls.py
```diff
@@ -52,10 +53,16 @@ def get_urls():
...
 
     return tuple(urls)
 
-urlpatterns = patterns('', *get_urls())
```
akx: Probably an unnecessary change.

shoop/xtheme/resources.py
```diff
@@ -132,7 +132,7 @@ def _render_resource(self, resource):
 
         # TODO: should this be extensible?
 
-        if resource.endswith(".js"):
+        if resource.endswith(".js") or resource.endswith("/jsi18n/"):
```
akx: This doesn't look like a good idea. Why not make the url something like `/i18n.js`?

shoop/admin/urls.py
```diff
@@ -56,6 +57,9 @@ def get_urls():
         if not isinstance(u, AdminRegexURLPattern):
             warnings.warn("Admin URL %r is not an AdminRegexURLPattern" % u)
 
+    # Add Django javascript catalog url
+    urls.append(url(r'^jsi18n.js', javascript_catalog, {'packages': ('shoop',)}, name='js-catalog'))
```
akx: Regex seems to be missing `$`?

akx:
LGTM!

PR #252, opened 2015-11-17
Unify AppConfigs and their name usages

suutari-ai:
Rename all `*AppConfig` classes in `__init__.py` files to just
`AppConfig` and make use of `__name__` when specifying strings in
`__init__.py` files that refer to names in the current package or in its
submodules.

### DIFF
--- a/doc/provides.rst
+++ b/doc/provides.rst
@@ -24,7 +24,7 @@ the keys and lists of loading specs as values for new provides to be discovered.
 
 .. code-block:: python
 
-   class PigeonAppConfig(AppConfig):
+   class AppConfig(shoop.apps.AppConfig):
 
        provides = {
            "shipping_method_module": [
--- a/shoop/addons/__init__.py
+++ b/shoop/addons/__init__.py
@@ -4,22 +4,22 @@
 #
 # This source code is licensed under the AGPLv3 license found in the
 # LICENSE file in the root directory of this source tree.
-from shoop.apps import AppConfig
+import shoop.apps
 from .manager import add_enabled_addons
 
 __all__ = ["add_enabled_addons"]
 
 
-class ShoopAddonsAppConfig(AppConfig):
-    name = "shoop.addons"
+class AppConfig(shoop.apps.AppConfig):
+    name = __name__
     verbose_name = "Shoop Addons"
     label = "shoop_addons"
 
     provides = {
         "admin_module": [
-            "shoop.addons.admin_module:AddonModule",
+            __name__ + ".admin_module:AddonModule",
         ]
     }
 
 
-default_app_config = "shoop.addons.ShoopAddonsAppConfig"
+default_app_config = __name__ + ".AppConfig"
--- a/shoop/admin/__init__.py
+++ b/shoop/admin/__init__.py
@@ -5,34 +5,34 @@
 #
 # This source code is licensed under the AGPLv3 license found in the
 # LICENSE file in the root directory of this source tree.
-from shoop.apps import AppConfig
+import shoop.apps
 from shoop.apps.settings import validate_templates_configuration
 
 
-class ShoopAdminAppConfig(AppConfig):
-    name = "shoop.admin"
+class AppConfig(shoop.apps.AppConfig):
+    name = __name__
     verbose_name = "Shoop Admin"
     label = "shoop_admin"
     required_installed_apps = ["bootstrap3"]
     provides = {
         "admin_module": [
-            "shoop.admin.modules.system:SystemModule",
-            "shoop.admin.modules.products:ProductModule",
-            "shoop.admin.modules.product_types:ProductTypeModule",
-            "shoop.admin.modules.media:MediaModule",
-            "shoop.admin.modules.orders:OrderModule",
-            "shoop.admin.modules.taxes:TaxModule",
-            "shoop.admin.modules.categories:CategoryModule",
-            "shoop.admin.modules.contacts:ContactModule",
-            "shoop.admin.modules.contact_groups:ContactGroupModule",
-            "shoop.admin.modules.users:UserModule",
-            "shoop.admin.modules.methods:MethodModule",
-            "shoop.admin.modules.attributes:AttributeModule",
-            "shoop.admin.modules.sales_units:SalesUnitModule",
-            "shoop.admin.modules.shops:ShopModule",
-            "shoop.admin.modules.demo:DemoModule",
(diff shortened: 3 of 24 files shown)


akx:
Hmm.

I don't really see the point in this. We lose descriptive `self.__class__.__name__`s, which can well be a Bad Thing when debugging (or when using things like `doccov`, come to think of it).

Also, the standard in the Django community is that AppConfigs have descriptive names. For examples of this, see Django Core ([1](https://github.com/django/django/blob/master/django/contrib/admin/apps.py), [2](https://github.com/django/django/blob/master/django/contrib/messages/apps.py#L5)), [Django-Jinja](https://github.com/niwinz/django-jinja/blob/master/django_jinja/apps.py#L7), [Django-Filer](https://github.com/stefanfoulis/django-filer/blob/develop/filer/apps.py#L7) to mention some examples of projects Shoop itself depends on. (To mention other popular Django projects: [Wagtail](https://github.com/torchbox/wagtail/blob/master/wagtail/wagtailcore/apps.py#L4),  [Taiga](https://github.com/taigaio/taiga-back/blob/master/taiga/base/apps.py#L20),  [Django-CMS](https://github.com/divio/django-cms/blob/develop/cms/apps.py#L7).)
In addition, you can notice the standard is to have the AppConfig object live in an `apps.py` file; now that's a change I could get behind.

suutari-ai:
The point is to reduce burden of addon developer. Naming is hard, and in this case the name doesn't even matter. So why force the addon developer to come up with irrelevant name? It's also easier to copy-paste the `__init__.py` for a new addon when there is no need to change the name.

I don't see your point why `parrots.CreateView` wouldn't be a good name.

suutari-ai:
Can you give an example where this matters in debugging?

I tried debugging this stuff with pdb, ipdb and winpdb. I had a local variable bound to `shoop.themes.classic_gray.AppConfig` and each occurence of the variable was showing the repr: `<AppConfig: shoop.themes.classic_gray>`. Also e.g. list(django.apps.apps.get_app_configs()) seems to print out nicely in the shell, since reprs are good.

suutari-ai:
The disagreement about this might be because I like to refer to names with their full dotted name (or at least the with the last module name), but you seem to prefer the short (undotted) name.

IMHO line like
```python
class MyCoolAddonAppConfig(AppConfig):
```
could cause confusion, because Django developers easily think that the `AppConfig` would be `django.apps.AppConfig`, but it really is `shoop.apps.AppConfig` in our case. That's why I think
```python
class AppConfig(shoop.apps.AppConfig):
```
is much better. And if you ever have to construct such appconfig manually, then you'd refer to it as `mycooladdon.AppConfig`.

Pikkupomo:
When this was suggested, I was interested in what is the common way of solve this issue (as there are multiple projects out there). Unluckily I was on the road and wasn't able to make that question, but @akx already answered it. Seems there are two ways of doing it `SampleConfig()` and `SampleAppConfig()`. 

Also,  the apps' `AppConfig`s live in `apps.py` which isn't the case in Shoop. This is actually something I've mentioned before, can't remember what was the answer to that :) 

I think that we should move all `AppConfig`s to apps.py instead of `__init__.py`.

About the naming
How about renaming `shoop.apps.AppConfig` to `shoop.apps.ShoopConfig`? Then name applications' AppConfigs to `<appname>Config(ShoopConfig)`. For example `class DiscountPricingConfig(ShoopConfig)` where `ShoopConfig` is `class ShoopConfig(django.apps.AppConfig)`. Dunno.

suutari-ai:
We can't rename shoop.apps.AppConfig without a major version bump.

I don't see why we would have to (or want to) mimic other projects with this. We should choose the better alternative, not the one that is more popular.

Also we're comparing apples and oranges here. `django.apps.AppConfig` is not `shoop.apps.AppConfig`. Shoop is not Django. What they do in Django projects doesn't always apply to Shoop projects.

Why do you care so much about name of a class that is never instantiated manually that you're willing to make addon developer's job harder? (They wouldn't even have to think about the app config naming if this was done in the way I propose.)

akx:
@Pikkupomo I don't mind our `shoop.apps.AppConfig` being called `AppConfig` -- it doesn't diverge too much from the Django `AppConfig` class anyway. And I'm also okay with @suutari-ai 's
```python
class MyCoolAddonConfig(shoop.apps.AppConfig):
class DiscountPricingConfig(shoop.apps.AppConfig):  # from @pikkupomo 's comment
```
if you want to change the import scheme.

But as for all of the app configs being called AppConfig -- it just feels _wrong_ to explicitly, willfully have several classes with the same name across the system. Also, I can imagine IDE users such as myself not having a very good time navigating across lists of classes like

* AppConfig (foo.bar.baz)
* AppConfig (baz.bar.quux)
* AppConfig (quux.baq.blop)

vs

* BazConfig (foo.bar.baz)
* BlopConfig (quux.baq.blop)
* QuuxConfig (baz.bar.quux)

As for using `__name__` everywhere, it just feels a little too magical; more so if we follow the `apps.py` module convention, leading to apps being called `shoop.admin.apps`... (Also, a mis-`sys.path`-configured system may be able to import the same module with two different names.)

akx:
> Shoop is not Django. What they do in Django projects doesn't always apply to Shoop projects.

But Shoop is a Django project, and I think we should play along the ecosystem standards. (Even if we've diverged from the general consensus in our choice of template engine -- but I think Jinja has very clear technical and productivity benefits.)

> Why do you care so much about name of a class that is never instantiated manually that you're willing to make addon developer's job harder? (They wouldn't even have to think about the app config naming if this was done in the way I propose.)

I really, _really_ don't think having to type

```python
class PigeonTransportConfig(shoop.apps.AppConfig):
    name = "pigeon_transport"
# ...
default_app_config = "pigeon_transport.apps.PigeonTransportConfig"
```

vs 

```python
class AppConfig(shoop.apps.AppConfig):
    name = __name__
# ...
default_app_config = "pigeon_transport.AppConfig"  # (assuming not following the `apps.py` convention either)
```

is that much of a mental hurdle for any developer worth their salt.

akx:
(Also, slightly OT, but:

> We can't rename shoop.apps.AppConfig without a major version bump.

Looking at https://github.com/shoopio/shoop/compare/v2.0.0...master there are several commits that already warrant a semver major bump. Excising the default theme from the repo is already a breaking change for current projects.)

suutari-ai:
OK, maybe foobar.FooBarAppConfig is the way to go. We should unify all AppConfigs to that format then. Will be another PR at some later time...

PR #254, opened 2015-11-18
Translation base and tools (SHOOP-1829)

Pikkupomo:
Add base translation file
Add `LOCALE_PATHS` to settings
Add i18n extractor management command
Classic Gray: Add language changer
Refs SHOOP-1829

### DIFF
--- a/shoop/core/management/commands/shoop_makemessages.py
+++ b/shoop/core/management/commands/shoop_makemessages.py
@@ -1,0 +1,44 @@
+# This file is part of Shoop.
+#
+# Copyright (c) 2012-2015, Shoop Ltd. All rights reserved.
+#
+# This source code is licensed under the AGPLv3 license found in the
+# LICENSE file in the root directory of this source tree.
+"""
+Extract translatable strings from shoop
+"""
+from __future__ import unicode_literals
+
+import logging
+import subprocess
+
+import os
+from django.conf import settings
+from django.core import management
+from django.core.management import BaseCommand
+
+
+logging.basicConfig(level=logging.INFO)
+
+
+class Command(BaseCommand):
+    def handle(self, *args, **options):
+        base_dir = os.path.join(settings.BASE_DIR, "..")
+        os.chdir(base_dir)
+        subprocess.check_call(["pybabel", "extract", "shoop", "-o" "shoop/locale/django.pot", "-F", "shoop/babel.cfg"])
+        subprocess.check_call([
+            "pybabel",
+            "update",
+            "-D", "django",
+            "-i", "shoop/locale/django.pot",
+            "-d", "shoop/locale",
+            "-l", "en",
+            "--ignore-obsolete"])
+
+        base_dir = os.path.join(settings.BASE_DIR, "../shoop")
+        os.chdir(base_dir)
+        logging.info("Running makemessages for .js files")
+        management.call_command("makemessages",
+                                domain="djangojs",
+                                locale=["en"],
+                                ignore=["node_modules", "bower_components", "doc"])
--- a/shoop/locale/en/LC_MESSAGES/djangojs.po
+++ b/shoop/locale/en/LC_MESSAGES/djangojs.po
@@ -1,0 +1,559 @@
+# English translations for PROJECT.
+# Copyright (C) 2015 ORGANIZATION
+# This file is distributed under the same license as the PROJECT project.
+# FIRST AUTHOR <EMAIL@ADDRESS>, 2015.
+#
+#, fuzzy
+msgid ""
+msgstr ""
+"Project-Id-Version: PROJECT VERSION\n"
+"Report-Msgid-Bugs-To: \n"
+"POT-Creation-Date: 2015-11-24 11:58+0200\n"
+"PO-Revision-Date: YEAR-MO-DA HO:MI+ZONE\n"
+"Last-Translator: FULL NAME <EMAIL@ADDRESS>\n"
+"Language-Team: en <LL@li.org>\n"
+"MIME-Version: 1.0\n"
+"Content-Type: text/plain; charset=UTF-8\n"
+"Content-Transfer-Encoding: 8bit\n"
+"Language: en\n"
+"Plural-Forms: nplurals=2; plural=(n != 1)\n"
+"Generated-By: Babel 2.1.1\n"
+
+#: .\admin\modules\media\static_src\media\browser\FileUpload.js:77
+#: .\admin\static\shoop_admin\js\media-browser.js:458
+msgid "Unexpected error while uploading files."
+msgstr ""
+
+#: .\admin\modules\media\static_src\media\browser\FileUpload.js:80
+#: .\admin\static\shoop_admin\js\media-browser.js:461
+msgid "Error:"
+msgstr ""
+
+#: .\admin\modules\media\static_src\media\browser\FileUpload.js:81
+#: .\admin\static\shoop_admin\js\media-browser.js:462
+msgid "Uploaded:"
+msgstr ""
+
+#: .\admin\modules\media\static_src\media\browser\actions\fileActions.js:14
+#: .\admin\static\shoop_admin\js\media-browser.js:627
+msgid "New file name?"
+msgstr ""
+
+#: .\admin\modules\media\static_src\media\browser\actions\fileActions.js:25
+#: .\admin\static\shoop_admin\js\media-browser.js:640
+#, c-format
+msgid "Are you sure you want to delete the file %s?"
+msgstr ""
+
+#: .\admin\modules\media\static_src\media\browser\actions\folderActions.js:13
+#: .\admin\modules\media\static_src\media\browser\actions\folderActions.js:32
+#: .\admin\static\shoop_admin\js\media-browser.js:678
+#: .\admin\static\shoop_admin\js\media-browser.js:702
+msgid "New folder name?"
+msgstr ""
+
+#: .\admin\modules\media\static_src\media\browser\actions\folderActions.js:44
+#: .\admin\static\shoop_admin\js\media-browser.js:718
+#, c-format
+msgid "Are you sure you want to delete the %s folder?"
+msgstr ""
+
+#: .\admin\modules\media\static_src\media\browser\components\folderTree.js:36
+#: .\admin\modules\media\static_src\media\browser\menus\folderContextMenu.js:17
+#: .\admin\static\shoop_admin\js\media-browser.js:259
+#: .\admin\static\shoop_admin\js\media-browser.js:1502
+msgid "New folder"
+msgstr ""
+
+#: .\admin\modules\media\static_src\media\browser\components\folderView.js:37
+#: .\admin\static\shoop_admin\js\media-browser.js:834
+msgid "Oldest first"
+msgstr ""
+
+#: .\admin\modules\media\static_src\media\browser\components\folderView.js:38
+#: .\admin\static\shoop_admin\js\media-browser.js:834
+msgid "Newest first"
+msgstr ""
+
+#: .\admin\modules\media\static_src\media\browser\components\folderView.js:39
+#: .\admin\static\shoop_admin\js\media-browser.js:834
+msgid "Smallest first"
+msgstr ""
+
+#: .\admin\modules\media\static_src\media\browser\components\folderView.js:40
+#: .\admin\static\shoop_admin\js\media-browser.js:834
+msgid "Largest first"
+msgstr ""
+
+#: .\admin\modules\media\static_src\media\browser\components\folderView.js:71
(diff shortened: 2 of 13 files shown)


suutari-ai:
Aren't these changes basicly for SHOOP-1829 (Translation support)?

suutari-ai:
You should separate the translation support (the scripts, management commands, language changer, message templates) from the Finnish translation. Separate PRs IMHO.

shoop/core/management/commands/shoop_extractmessages.py
```diff
@@ -0,0 +1,72 @@
...
+"""
+Extract translatable strings from shoop
+"""
+from __future__ import unicode_literals
+
+import logging
+
+import os
+from _misc import sanity_utils
+from _misc.gather_i18n import MAPPING, extract_from_file, write_po_wrap
```
suutari-ai: shoop should not depend on _misc.

The tooling used by this management command should be moved under shoop.utils or something like that.

shoop/locale/en/LC_MESSAGES/djangojs.po
```diff
@@ -0,0 +1,419 @@
...
+"POT-Creation-Date: 2015-11-18 12:59+0200\n"
+"PO-Revision-Date: YEAR-MO-DA HO:MI+ZONE\n"
+"Last-Translator: FULL NAME <EMAIL@ADDRESS>\n"
+"Language-Team: LANGUAGE <LL@li.org>\n"
+"MIME-Version: 1.0\n"
+"Content-Type: text/plain; charset=UTF-8\n"
+"Content-Transfer-Encoding: 8bit\n"
+"Generated-By: Babel 1.3\n"
+
+#: D:/projects/shoop/shoop/admin/modules/media/static_src/media/browser/components/responsiveUploadHint.js:0
```
akx: These shouldn't have `D:/projects/shoop/`...

shoop/core/management/commands/shoop_extractmessages.py
```diff
@@ -0,0 +1,72 @@
+# This file is part of Shoop.
+#
+# Copyright (c) 2012-2015, Shoop Ltd. All rights reserved.
+#
+# This source code is licensed under the AGPLv3 license found in the
+# LICENSE file in the root directory of this source tree.
+"""
```
akx: I'm not sure why the management command is necessary in the first place. [The script in `_misc` is documented, after all...](https://github.com/shoopio/shoop/blob/master/doc/dev_tools.rst#i18n-catalog-extraction)

Pikkupomo: we need to have 1 script that does the extracting, script under _misc doesn't extract everything, this combined extracting does. Management command feels more like "part of shoop" and gives more tools to tie the extraction into shoop only, and skip the other apps. The script under _misc extracts (if wanted) from everywhere and compiles them to one shoop-static.pot, which, in my opinion isn't what we need.

shoop/locale/en/LC_MESSAGES/djangojs.po
Pikkupomo: true, will fix

shoop/core/management/commands/shoop_extractmessages.py
Pikkupomo: okay will fix

shoop/core/management/commands/shoop_extractmessages.py
```diff
@@ -0,0 +1,72 @@
...
+logging.basicConfig(level=logging.INFO)
+
+
+class Command(BaseCommand):
+    def handle(self, *args, **options):
+
+        base_dir = os.path.join(settings.BASE_DIR, "../shoop")
+        os.chdir(base_dir)
+        logging.info("Running makemessages for .jinja and .py files")
+        management.call_command("makemessages",
```
akx: Why does this run `makemessages` at all? :O

shoop/core/management/commands/shoop_extractmessages.py
akx: Then embetter the external script to do what it should, imho.

I don't feel a management command is the right fit for this job; management commands are supposed to be run in the context of a finished project, and i18n extraction for the Shoop codebase shouldn't require a project (be it `shoop_workbench` or anything else).

akx:
Actually, come to think of it, maybe the i18n tool should live outside the main shoop repository altogether?

I mean, it'd be useful for addon developers (and perhaps other developers) too.

shoop/core/management/commands/shoop_makemessages.py
```diff
@@ -0,0 +1,44 @@
...
+            "-D", "django",
+            "-i", "shoop/locale/django.pot",
+            "-d", "shoop/locale",
+            "-l", "en",
+            "--ignore-obsolete"])
+
+        base_dir = os.path.join(settings.BASE_DIR, "../shoop")
+        os.chdir(base_dir)
+        logging.info("Running makemessages for .js files")
+        management.call_command("makemessages",
```
akx: Could we use `pybabel` for this too?

shoop/core/management/commands/shoop_makemessages.py
```diff
@@ -0,0 +1,44 @@
...
+import logging
+import subprocess
+
+import os
+from django.conf import settings
+from django.core import management
+from django.core.management import BaseCommand
+
+
+logging.basicConfig(level=logging.INFO)
```
akx: Don't configure logging on module level, please.

shoop/core/management/commands/shoop_makemessages.py
```diff
@@ -0,0 +1,44 @@
...
+
+
+class Command(BaseCommand):
```
akx: I still feel this shouldn't be a management command, for the reasons I outlined earlier (which got no reply):

> I don't feel a management command is the right fit for this job; management commands are supposed to be run in the context of a finished project, and i18n extraction for the Shoop codebase shouldn't require a project (be it `shoop_workbench` or anything else).

shoop/locale/en/LC_MESSAGES/djangojs.po
```diff
@@ -0,0 +1,559 @@
+# English translations for PROJECT.
```
akx: Do we actually need to store these source .po files in the repo at all?

shoop/themes/classic_gray/templates/classic_gray/shoop/front/includes/language_changer.jinja
```diff
@@ -0,0 +1,19 @@
+{% if LANGUAGES|length > 1 %}
+<form method="post" action="/set-language/" class="hidden">
```
akx: `/set-language/` should be CSRF exempt, imho, to avoid creating a session where one is not needed.

shoop/themes/classic_gray/static_src/js/custom.js
```diff
@@ -95,6 +95,10 @@ function updatePrice() {
     });
 }
 
+function changeLanguage() {
+    $("#language-field").attr("value", $(this).data("value")).closest("form").submit();
```
akx: Using `closest()` feels a little fragile. Add an ID to the form and use that?

shoop/themes/classic_gray/templates/classic_gray/shoop/front/includes/language_changer.jinja
```diff
@@ -0,0 +1,19 @@
...
+    {% csrf_token %}
+    <input type="hidden" name="language" id="language-field">
+</form>
+
+<li class="dropdown" id="language-changer">
+    <a aria-expanded="false" href="#" class="dropdown-toggle" data-toggle="dropdown">
+        <i class="menu-icon fa fa-globe hidden-xs"></i> {{ LANGUAGE_CODE|upper }}
```
akx: Using `LANGUAGE_CODE` for user consumption doesn't feel right.

shoop_setup_utils/commands.py
```diff
@@ -82,9 +86,29 @@ class BuildProductionResourcesCommand(BuildResourcesCommand):
...
+
+    def initialize_options(self):
+        pass
+
+    def finalize_options(self):
+        pass
+
+    def run(self):
+        os.chdir("shoop")
+        settings.configure()
```
akx: This just reinforces my opinion that the translation thing should definitely not be a management command.

shoop/core/management/commands/shoop_makemessages.py
suutari-ai: We decided to use makemessages for js, because it finds all the same translations but got line numbers right whereas pybabel didn't. Makemessages is also faster.

Anyway... those are quite minor issues and using only single tool to collect the messages would probably be a good idea. @Pikkupomo What do you think, should we switch pybabel only?

shoop/core/management/commands/shoop_makemessages.py
suutari-ai: IMHO this exactly the reason why it should be a management command provided by Shoop Core. Being installed with Shoop means it is available for addon developers very easily.

BUT:
* The shoop_makemessages command needs to be adapted for collecting messages from any given directory (not just shoop as it does now).
* Same for the "setup.py build_translations" command, so that addons could `import shoop_setuputils` in their setup.py and use it too.

shoop/locale/en/LC_MESSAGES/djangojs.po
suutari-ai: Yes. Transifex collects the messages from those.

akx: Okay. I thought they were manually uploaded to Tx.

shoop/core/management/commands/shoop_makemessages.py
akx: I recall `makemessages` relies on the gettext toolchain being available on the system, which usually isn't the case on Windows.

akx: And re line numbers: I don't think they actually matter, and moreover cause unnecessary changeset churn. The tool in `_misc` actually zeroes them for that reason.

suutari-ai: IMHO if we really need zeroing the line numbers, we should try to push that feature upstream, i.e. to Babel.

shoop/core/management/commands/shoop_makemessages.py
suutari-ai: Yep, at the start of the `handle` method would be better place for this.

shoop/themes/classic_gray/templates/classic_gray/shoop/front/includes/language_changer.jinja
suutari-ai: Wouldn't that enable changing language of the admin interface from another site?

shoop_setup_utils/commands.py
suutari-ai: Management command has the advantage that DJANGO_SETTINGS_MODULE is set. Compiling messages to mo does not need the settings module being set, but makemessages needs it, because the django_jinja extensions read settings (for some reason). Having the makemessages part in setup.py would be another option, but that would require setting DJANGO_SETTINGS_MODULE to some value, which would be `shoop_workbench.settings` for shoop, but there would be no way to guess it for addons.

shoop_workbench/settings/base_settings.py
```diff
@@ -178,6 +178,10 @@
     )
 }
 
+LOCALE_PATHS = (
+    os.path.join(BASE_DIR, "../shoop/locale"),
```
suutari-ai: Shouldn't this use os.path.join?

shoop/themes/classic_gray/templates/classic_gray/shoop/front/includes/language_changer.jinja
akx: Would that be that big of an issue?

shoop/core/management/commands/shoop_makemessages.py
akx: [But we've already implemented it.](https://github.com/shoopio/shoop/blob/master/_misc/gather_i18n.py#L44)

shoop/locale/en/LC_MESSAGES/djangojs.po
Pikkupomo: they are now @akx but it can be automatized, which we intend to do :)

akx: [Django-REST-Framework directly uses the tx client for uploads.](https://github.com/tomchristie/django-rest-framework/commit/7913947757b0e6bd1b8828db81933c32c498e20a) I think that might be better than storing generated files in the repo.

suutari-ai:
Replaced with PR #274.

PR #258, opened 2015-11-23
Don't use `from django.conf import settings` in app init modules.

akx:
When `fooapp.settings` is imported, that `settings` is injected into the
`fooapp` parent module, overriding the `django.conf` import, making it
impossible to actually access `django.conf.settings`.

Notably, this caused `shoop.front` to never actually honor
`SHOOP_FRONT_INSTALL_ERROR_HANDLERS`, which in kind made application
initialization begin importing the root urlconf, which in the case of e.g.
Django-CMS caused database accesses, causing exceptions during migration.

### DIFF
--- a/shoop/front/__init__.py
+++ b/shoop/front/__init__.py
@@ -5,7 +5,7 @@
 #
 # This source code is licensed under the AGPLv3 license found in the
 # LICENSE file in the root directory of this source tree.
-from django.conf import settings
+import django.conf
 from shoop.apps import AppConfig
 from shoop.apps.settings import validate_templates_configuration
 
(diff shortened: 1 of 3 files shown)


suutari-ai:
Thanks @akx!

akx:
For reference, client projects may work around this bug (as it exists in 2.0.0) by adding something like this to their `settings`:

```python
# TODO: Remove me when https://github.com/shoopio/shoop/pull/258 lands
if "shoop.front" in INSTALLED_APPS:
    import shoop.front.settings as sfs
    sfs.SHOOP_FRONT_INSTALL_ERROR_HANDLERS = False
```

PR #260, opened 2015-11-25
Admin: Manual orders

tulimaki:
* Order creation dummy templates and initial styles
* Admin browse widget: Add option to filter products based on shop
* Fix product edit view orderability checks
* Add footer styles for order-creator
* Refactor order creation
* Modify order creator actions and reducers for more complex views
* Refactor order creation frontend
* Refactor and add tests for order creator

Refs SHOOP-1799


shoop/admin/modules/orders/json_order_creator.py
```diff
@@ -121,27 +136,93 @@ def _process_lines(self, source, state):
...
 
+    def _create_address(self, address):
+        address_fields = [
+            "name", "tax_number", "phone", "email", "street", "street2", "postal_code", "city", "region", "country"
+        ]
+        fields = {}
+        for field in address_fields:
+            fields[field] = address.get(field, "")
+
+        address = Address.objects.filter(**fields).first()
```
akx: Isn't this just `get_or_create`?

shoop/admin/modules/orders/static_src/create/actions/index.js
```diff
@@ -51,7 +104,7 @@ function handleCreateResponse(dispatch, data) {
             location.href = url;
         } else {
             // Very, very unlikely that we'd ever get here
-            alert(interpolate(gettext("Order %s created."), [orderIdentifier]));
+            alert("Order" + orderIdentifier + " created.");
```
akx: Mismerge?

shoop/admin/modules/orders/static_src/create/reducers/lines.js
```diff
@@ -27,30 +35,60 @@ function setLineProperties(linesState, lineId, props) {
         if (line.id === lineId) {
             return _.assign({}, line, props);
         }
+
         return line;
     });
 }
 
+function getDiscountsAndTotal(quantity, baseUnitPrice, unitPrice, updateUnitPrice=false) {
+    var updates = {};
```
akx: Prefer consts over vars.

shoop/admin/modules/orders/static_src/create/reducers/lines.js
```diff
@@ -27,30 +35,60 @@ function setLineProperties(linesState, lineId, props) {
...
+    updates.step = product.purchaseMultiple;
+    updates.errorText = product.errorText;
     return setLineProperties(state, id, updates);
 }
 
 function setLineProperty(state, {payload}) {
     const {id, property, value} = payload;
     const line = _.detect(state, (sLine) => sLine.id === id);
-    const updates = {};
+    var updates = {};
```
akx: Prefer consts over vars.

shoop/admin/modules/orders/static_src/create/reducers/lines.js
```diff
@@ -68,24 +106,60 @@ function setLineProperty(state, {payload}) {
...
+                    updates = getDiscountsAndTotal(0, line.baseUnitPrice, 0);
                     updates.unitPrice = 0;
                     updates.quantity = 0;
                 }
+                updates.type = value;
                 break;
             case "quantity":
-                updates.quantity = Math.max(0, parseFloat(value));
+                var quantity = Math.max(0, parseFloat(value));
+                updates  = getDiscountsAndTotal(quantity, line.baseUnitPrice, line.unitPrice);
```
akx: Stray double space (here and further down)

shoop/admin/modules/orders/static_src/create/reducers/order.js
```diff
@@ -9,7 +9,31 @@
...
+function getMethodPrice(method) {
+    return  method && method.price ? method.price : 0;
+}
+
+function updateTotals(state, {payload}) {
+    const updates = {};
+    const {lines, methods} = payload();
+    var total = 0;
+    _.map(lines, (line) => {
+        total += parseFloat(line.total);
```
akx: `parseFloat` shouldn't be necessary at this point...

shoop/admin/modules/orders/static_src/create/view/confirm.js
```diff
@@ -0,0 +1,94 @@
...
+ */
+
+function renderHeaders() {
+    return  m("tr", [
+        m("th", gettext("SKU")),
+        m("th", gettext("Text")),
+        m("th", gettext("Quantity")),
+        m("th", gettext("Unit Price")),
+        m("th", gettext("Discount amount")),
+        m("th", gettext("Total (exclude taxes)")),
```
akx: > Total (excluding taxes)

shoop/admin/modules/orders/static_src/create/view/confirm.js
```diff
@@ -0,0 +1,94 @@
...
+            m("td", line.sku),
+            m("td", line.text),
+            m("td", line.quantity),
+            m("td", line.unitPrice),
+            m("td", line.discountAmount),
+            m("td", line.taxlessTotal),
+            m("td", line.taxPercentage),
+            m("td", line.taxfulTotal)
+        ]);
+    }).compact().value();
```
akx: `compact()` shouldn't be necessary (unless there can somehow be `null`s in the `lines` array?)

shoop/admin/modules/orders/static_src/create/view/confirm.js
```diff
@@ -0,0 +1,94 @@
...
+                m("strong", line),
+                m("br")
+            ];
+        } else {
+            return [
+                line,
+                m("br")
+            ];
+        }
+    }).compact().value();
```
akx: Use `flatten()` here (to turn `[[x,br],[y,br]]` into `[x,br,y,br]`

shoop/admin/modules/orders/static_src/create/view/customers.js
```diff
@@ -0,0 +1,117 @@
...
+                type: "text",
+                placeholder: field.label,
+                required: isRequired,
+                value: _.get(address, field.key, ""),
+                onchange: function () {
+                    store.dispatch(setAddressProperty(addressType, field.key, this.value));
+                }
+            })
+        ]);
+    }).compact().value();
```
akx: No need for compact here either. I don't see any route that would return nulls.

shoop/admin/modules/orders/views/create.py
```diff
@@ -34,6 +41,65 @@ def create_order_from_state(state, **kwargs):
...
 
+def create_source_from_state(state, **kwargs):
+    joc = JsonOrderCreator()
+    source = joc.create_source_from_state(state, **kwargs)
+    if not source:
+        raise ValidationError(list(joc.errors))
+    return source
+
+
+def get_address_json(address):
```
akx: Shouldn't this be `encode_address`?

shoop/admin/modules/orders/views/create.py
```diff
@@ -86,28 +144,84 @@ def handle_product_data(self, request):
...
-        price_info = product.get_price_info(context, quantity=1)
+        try:
+            shop_product = product.get_shop_instance(shop)
+        except ObjectDoesNotExist:
+            return {"errorText": _("Product %(product)s is not available in the %(shop)s shop.") % {
+                "product": product.name, "shop": shop.name}}
+
+        min_quantity = shop_product.minimum_purchase_quantity
+        # Make quantity to be at least minimum quantity
+        quantity = min_quantity if quantity < min_quantity else quantity
```
akx: Add parens around the rvalue, please?

shoop/admin/modules/orders/views/create.py
```diff
@@ -86,28 +144,84 @@ def handle_product_data(self, request):
...
-        ctx_request = RequestFactory().get("/")
-        ctx_request.shop = shop
-        ctx_request.customer = Contact.objects.filter(pk=customer_id).first()
-        ctx_request.user = AnonymousUser()
-        context = get_pricing_module().get_context_from_request(ctx_request)
-        price_info = product.get_price_info(context, quantity=1)
+        try:
+            shop_product = product.get_shop_instance(shop)
+        except ObjectDoesNotExist:
+            return {"errorText": _("Product %(product)s is not available in the %(shop)s shop.") % {
```
akx: Weird indentation here.

shoop/admin/modules/orders/views/create.py
```diff
@@ -86,28 +144,84 @@ def handle_product_data(self, request):
         product_id = request.GET["id"]
         shop_id = request.GET["shop_id"]
         customer_id = request.GET.get("customer_id")
+        quantity = decimal.Decimal(request.GET.get("quantity", 1))
         product = Product.objects.get(pk=product_id)
+        if not product:
```
akx: This code will never be reached; if `.get` errors out, it'll raise an ODNE, not return None.

shoop/admin/modules/orders/views/create.py
```diff
@@ -86,28 +144,84 @@ def handle_product_data(self, request):
...
+                "includesTax": price_info.base_price.includes_tax
             }
         }
 
+    def handle_customer_data(self, request):
+        customer = Contact.objects.filter(pk=request.GET["id"]).first()
+        return {
+            "id": customer.id,
+            "name": customer.name,
+            "isCompany": True if isinstance(customer, CompanyContact) else False,
```
akx: No need for `True if x else False`; either just `x` (or maybe preferably `bool(x)`)

shoop/admin/modules/orders/views/create.py
```diff
@@ -86,28 +144,84 @@ def handle_product_data(self, request):
...
+        min_quantity = shop_product.minimum_purchase_quantity
+        # Make quantity to be at least minimum quantity
+        quantity = min_quantity if quantity < min_quantity else quantity
+        customer = Contact.objects.filter(pk=customer_id).first() if customer_id else None
+        price_info = get_price_info(shop, customer, product, quantity)
+        error_text = None
+        supplier = shop_product.suppliers.first()  # TODO: Allow setting a supplier?
+        for message in shop_product.get_orderability_errors(
+                supplier=supplier, quantity=quantity, customer=customer, ignore_minimum=True):
+            error_text = str(message.args[0])
```
akx: Use something like [is already done in the product edit view](https://github.com/shoopio/shoop/blob/5a68a8a1ae8c42cfb9ec18912af1f5963bd97a3f/shoop/admin/modules/products/views/edit.py#L173-L183) to get _all_ orderability errors, not just the last one.

shoop/admin/modules/orders/views/create.py
```diff
@@ -86,28 +144,84 @@ def handle_product_data(self, request):
...
             "unitPrice": {
-                "value": price_info.price.value,  # TODO: This is always zero?!
-                "includesTax": price_info.price.includes_tax
+                "value": price_info.price.value,
+                "includesTax": price_info.base_price.includes_tax
             }
         }
 
+    def handle_customer_data(self, request):
+        customer = Contact.objects.filter(pk=request.GET["id"]).first()
```
akx: Use `.get()` here.

(Also, add error handling.)

shoop/admin/modules/orders/views/create.py
```diff
@@ -86,28 +144,84 @@ def handle_product_data(self, request):
...
+                "orderLines": [encode_line(line) for line in source.get_final_lines(with_taxes=True)],
+                "billingAddress": source.billing_address.as_string_list() if source.billing_address else None,
+                "shippingAddress": source.shipping_address.as_string_list() if source.shipping_address else None,
+            }
+        except ValidationError as exc:
+            message = _("Could not proceed with order:")
+            message += "\n" + "\n".join(force_text(err) for err in exc.messages)
+            return JsonResponse({"success": False, "errorMessage": message}, status=400)
+        except Exception:
+            return JsonResponse({
```
akx: No way of knowing what exception actually occurred in this code path :\

shoop/admin/modules/orders/views/create.py
```diff
@@ -117,13 +231,14 @@ def handle_create(self, request):
...
+        except ValidationError as exc:
             message = _("Could not create order:")
-            if isinstance(exc, ValidationError):  # pragma: no branch
-                message += "\n" + "\n".join(force_text(err) for err in exc.messages)
-            else:
-                message += " " + tb  # pragma: no cover
+            message += "\n" + "\n".join(force_text(err) for err in exc.messages)
             return JsonResponse({"success": False, "errorMessage": message}, status=400)
+        except Exception:
+            return JsonResponse({
```
akx: No way of knowing what exception actually occurred in this code path :\

shoop/admin/modules/products/views/list.py
```diff
@@ -28,7 +28,11 @@ class ProductListView(PicotableListView):
     ]
 
     def get_queryset(self):
-        return Product.objects.all_except_deleted()
+        shop = self.request.GET.get("shop")
+        qs = Product.objects.all_except_deleted()
+        if shop:
+            qs = qs.filter(shop_products__shop__pk=int(shop))
```
akx: Use `qs.filter(shop_products__shop_id=int(shop))` to avoid one more join.

shoop/admin/static_src/base/js/browse-widget.js
```diff
@@ -42,11 +42,13 @@ window.BrowseAPI = (function() {
      * Currently supported options are:
      * * `kind`: kind string (e.g. "product")
      * * `filter`: filter string (kind-dependent)
+     * * `shop`: shop id option for product-list
```
akx: This shouldn't be an additional option; please leverage `filter` instead (maybe put JSON there, like `{"shop": 1}`?)

shoop/admin/static_src/picotable/picotable.js
```diff
@@ -632,7 +642,13 @@ const Picotable = (function(m, storage) {
                 page: 0 | ctrl.vm.page(),
                 filters: ctrl.vm.filterValues()
             };
-            m.request({method: "GET", url: url, data: {"jq": JSON.stringify(data)}}).then(ctrl.vm.data, function() {
+            var params = parseGetParamsString(location.search);
```
akx: Use `const`.

shoop/admin/static_src/picotable/picotable.js
```diff
@@ -551,6 +551,16 @@ const Picotable = (function(m, storage) {
         ]);
     }
 
+    function parseGetParamsString(paramsString) {
+        var params = {};
```
akx: Use const.

shoop/admin/templates/shoop/admin/orders/create.jinja
```diff
@@ -1,18 +1,14 @@
...
-                    <div id="order-tool-container"></div>
-                </div>
-            </div>
-        </div>
-    </div>
+    <div id="order-tool-container"></div>
 {% endblock %}
 
 {% block extra_js %}
+    <script>var COUNTRIES = {{  countries|json }};</script>
```
akx: Why isn't this in `config` like the rest of the configuration?

shoop/admin/modules/orders/static_src/create/view/shops.js
```diff
@@ -0,0 +1,22 @@
...
+import {selectBox} from "./utils";
+
+export function shopSelectView(store) {
+    const {shop} = store.getState();
+    return m("div.form-group", [
+        m("label.control-label", gettext("Shop")),
+        selectBox(shop.selected.id, function () {
+            const newShop = _.find(shop.choices, {"id": parseInt(this.value)});
+            store.dispatch(setShop(newShop));
+            store.dispatch(updateLines(store));
```
akx: Passing the `store` to an action smells like an anti-pattern. What does `updateLines` do?

shoop/admin/modules/orders/static_src/create/reducers/lines.js
```diff
@@ -68,24 +106,60 @@ function setLineProperty(state, {payload}) {
...
+                var total = parseFloat(value);
+                updates  = getDiscountsAndTotal(line.quantity, line.baseUnitPrice, (total / line.quantity), true);
+                updates.total = total;
                 break;
         }
     }
     return setLineProperties(state, id, updates);
 }
 
+function updateLines(state, {payload}) {
```
akx: You could move this to the top-level reducer in `reducers/index.js`; that way you get direct access to the full state without having to pass the `store` as the payload (which has a terrible code smell).

Then you could change the payload to only have `dispatch` (`{dispatch: store.dispatch.bind(store)}`, and you'd end up with something like

```javascript
function updateLines(state, {payload}) {
    const {dispatch} = payload;
    state.lines.forEach((line) => {
        if (line.product) {
            dispatch(retrieveProductData({id: line.product.id, forLine: line.id, quantity: line.quantity}));
        }
    });
    return state;
}
```

akx: Actually, this is quite obviously an async action, so this logic should live in an async action creator.

So...

* Get rid of this reducer altogether
* Make the `updateLines` action look like this:

```javascript
const updateLines = () => {
    return (dispatch, getState) => {
        getState().lines.forEach((line) => {
            if (line.product) {
                dispatch(retrieveProductData({id: line.product.id, forLine: line.id, quantity: line.quantity}));
            }
        });
    }
}
```

This way action dispatching happens where it should -- in actions -- and not in the reducer :sunglasses:

shoop/admin/modules/orders/static_src/create/reducers/shop.js
```diff
@@ -10,6 +10,9 @@ import {handleActions} from "redux-actions";
 import _ from "lodash";
 
 export default handleActions({
-    setShopId: ((state, {payload}) => _.assign(state, {id: payload})),
     setShopChoices: ((state, {payload}) => _.assign(state, {choices: payload})),
-}, {choices: [], id: null});
+    setShop: ((state, {payload}) =>_.assign(state, {selected: payload}))
+}, {
+    choices: [],
+    selected: {}
```
akx: Should probably default to `null`

shoop/admin/modules/orders/views/create.py
```diff
@@ -34,6 +41,65 @@ def create_order_from_state(state, **kwargs):
...
+        "pricesIncludeTaxes": shop.prices_include_tax
+    }
+
+
+def encode_method(method):
+    is_shipping = isinstance(method, ShippingMethod)
+    return {
+        "id": method.pk,
+        "name": force_text(method),
+        "price": method.module_data.get("price"),
```
Pikkupomo: `module_data` is not always available it seems

shoop/admin/modules/products/views/list.py
suutari-ai: Shouldn't it be just `qs.filter(shop_products__shop=shop)`?

suutari-ai: Oh wait.. the shop variable was actually shop_id.. misleading naming IMHO.

shoop/admin/modules/orders/static_src/create/view/lines.js
```diff
@@ -6,87 +6,146 @@
...
+                            onSelect: (obj) => {
+                                store.dispatch(setLineProperty(line.id, "product", obj));
+                                store.dispatch(retrieveProductData(
+                                    {id: obj.id, forLine: line.id, quantity: line.quantity}
+                                ));
+                            }
+                        });
+                    }
+                }, (line.product ?
+                    [line.product.text, m("br"), m("small", "(" + line.sku + ")")] : gettext("No product selected"))
```
Pikkupomo: this could be a button with a text "select product" instead of "no product selected"

shoop/admin/modules/orders/json_order_creator.py
tulimaki: It could be. In some case I managed to create several addresses with same fields so get_or_create did give multiple objects error. I think, since several addresses with same fields is not restricted, this is safer. We don't want to throw error in this case.

shoop/admin/modules/orders/static_src/create/reducers/lines.js
tulimaki: Can't use const here since later on getDiscountAndTotals is assigned to this variable.

akx: Oh, so it is. Sorry!

shoop/admin/modules/orders/json_order_creator.py
akx: Check. :)

shoop/admin/static_src/picotable/picotable.js
akx: And actually, you can just use `m.route.parseQueryString` (from Mithril) instead of rolling your own, here. :)

shoop/admin/modules/orders/json_order_creator.py
suutari-ai: Do we really want to reference an existing address?

IMHO we should just always create a new one. (In that case this function is not needed, but `Address.objects.create(**adress)` would do.) Reason: Addresses might be changed, but Address in order should not change.

shoop/admin/modules/orders/json_order_creator.py
```diff
@@ -121,27 +136,93 @@ def _process_lines(self, source, state):
...
+        billing_address = customer_data.pop("billingAddress", {})
+        shipping_address = customer_data.pop("shippingAddress", {})
+        ship_to_billing_address = customer_data.pop("shipToBillingAddress", False)
+        is_company = customer_data.pop("isCompany", False)
+        save_address = customer_data.pop("saveAddress", False)
+        customer = self.safe_get_first(Contact, pk=customer_data.get("id")) if customer_data else None
+        if not customer:
+            customer = self._create_contact_from_address(billing_address, is_company)
+            if not customer:
+                return
```
suutari-ai: What's this return doing here?

tulimaki: Creating contact might be skipped if there is no tax_number for company. In that case this return ends order creation, which should be fine.

shoop/admin/modules/orders/json_order_creator.py
tulimaki: I think yes. I don't think it is ideal to have same address with same fields multiple times in db. Also address is immutable. Also this "get or create" is reached already in confirmation phase so .create(**fields) is not enough since we don't want to create same address twice in order creation. This "get or create" is possibly done multiple times per one order creation.

Pikkupomo: Order should cache addresses. The current implementation causes database to be flooded with addresses once used with one order. I think address should be "customers property" where only direct relation lives. All other places (if there is a fear for address data to change) should only cache these values for themselves. Also once order is made, it's data should be set in stone. If someone modifies something related to order, it should not change the order contents. ever. For me, this current design is faulty.

tl;dr; remove `Order.billing_address` and `Order.shipping_address` and replace them with address properties (`shipping_street`, `shipping_zip` etc)

suutari-ai: Having addressed denormalized to fields (i.e. order having shipping_street, shipping_zip etc.) is not very nice approach IMHO. What if we want to add a new field to address? We would have to add it to all places having addresses. Also setting/getting addresses would be much harder and error prone, since each field would have to be set/get separately.

I see that there is five options:
* (1) Make address objects used in orders immutable.
* (2) Make address a custom data type and implement a Django model field which will serialize the address to single db column on save and deserializes it back to Address object on load.
  - Would make querying orders by some address field much harder.
* (3) Have every order have its separate address instance.
  - Would "flood" the database as @Pikkupomo said.
* (4) Save addresses in denormalized form to orders (i.e. order having shipping_street, shipping_zip etc.)
  - This would flood the database exactly as much as (3) btw.
* (5) Just ignore it: Use same address ids if fields occasionally match and let addresses change on orders if they are changed elsewhere.
 - Not really an option IMHO.

I would prefer (1). For order creating it would mean that it should pick only immutable addresses or create a new immutable address if no such address exists. We already have immutability of Addresses, but IMHO it should be revised now: Possibly switching it to use ChangeProtected mixin like Shop and Tax do.

suutari-ai: It would also be nice to have a mechanism that forces addresses of any Order to be immutable. One possible way to implement that would be to have a separate model `ImmutableAddress` being derived from `Address` and make Order use ImmutableAddress FK fields. Any other idea for implementing that?

Pikkupomo: `1` will lead to `3` because customer cannot change their addresses after the address is once used. I think it's not viable option either

tulimaki: I added code to make addresses fetched here immutable if not already.

akx: I think:

* try to get an `is_immutable` address with the given fields
* if one doesn't exist, create an `is_immutable` address with the given fields and use that

would be the simplest, neatest solution here?

Pikkupomo:
LGTM but did @suutari-ai still have something open regarding Addresses?

suutari-ai:
Let's not make an issue about the immutable address stuff here.

IMHO ImmutableAddress as a separate class/model would be important addition for forcing Order addresses to always be immutable, but that's entirely separate issue from this one.

Good work @tulimaki! Thanks!

PR #261, opened 2015-11-25
Deny price unit changes of in-use Shop (SHOOP-1658)

suutari-ai:
Changing currency or taxfulness of Shop's prices while it has any orders
is not good, so deny it.

Also rename ImmutableMixin to ChangeProtected and improve it a bit and
make the non-field errors visible in ShopEditView.

### DIFF
--- a/shoop/core/models/_base.py
+++ b/shoop/core/models/_base.py
@@ -56,7 +56,7 @@ class ChangeProtected(object):
         super(ChangeProtected, self).clean(*args, **kwargs)
         if self.pk:
             changed_protected_fields = self._get_changed_protected_fields()
-            if changed_protected_fields and self._is_in_use():
+            if changed_protected_fields and self._are_changes_protected():
                 message = "{change_protect_message}: {fields}".format(
                     change_protect_message=self.change_protect_message,
                     fields=", ".join(sorted(changed_protected_fields)),
@@ -67,7 +67,15 @@ class ChangeProtected(object):
         self.clean()
         super(ChangeProtected, self).save(*args, **kwargs)
 
-    def _is_in_use(self):
+    def _are_changes_protected(self):
+        """
+        Check if changes of this object should be protected.
+
+        This can be overridden in the subclasses to make it possible to
+        avoid change protection e.g. if object is not in use yet.
+
+        The base class implementation just returns True.
+        """
         return True
 
     def _get_changed_protected_fields(self):
--- a/shoop/core/models/_base.py
+++ b/shoop/core/models/_base.py
@@ -49,13 +49,18 @@ class TranslatableShoopModel(ShoopModel, parler.models.TranslatableModel):
 
 class ChangeProtected(object):
     unprotected_fields = []
-    immutability_message = _("Cannot change immutable object that is in use")
+    change_protect_message = _("The following fields can not be changed")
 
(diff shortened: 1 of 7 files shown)


suutari-ai:
Oh wait.. just noticed that this is not yet strictly to the spec... gotta update this..

suutari-ai:
Actually the spec was stupid. Updated the issue description. This PR should be OK now.

shoop/core/models/_base.py
```diff
@@ -47,28 +47,37 @@ class Meta:
         abstract = True
 
 
-class ImmutableMixin(object):
+class ChangeProtected(object):
+    protected_fields = None
     unprotected_fields = []
-    immutability_message = _("Cannot change immutable object that is in use")
+    change_protect_message = _("Following fields cannot be changed")
```
akx: > The following fields can not be changed

akx:
This is also a semver major change.

shoop/core/models/_base.py
```diff
@@ -47,28 +47,37 @@ class Meta:
...
+                    fields=", ".join(sorted(changed_protected_fields)),
+                )
+                raise ValidationError(message)
 
     def save(self, *args, **kwargs):
         self.clean()
-        super(ImmutableMixin, self).save(*args, **kwargs)
+        super(ChangeProtected, self).save(*args, **kwargs)
 
     def _is_in_use(self):
```
akx: This isn't really a private function? I mean, it's supposed to be overridden in subclasses..

suutari-ai:
> This is also a semver major change.

How so?

shoop/core/models/_base.py
suutari-ai: I don't understand what you're after in this..

shoop/core/models/_base.py
suutari-ai: Updated.

shoop/core/models/_base.py
akx: It's an underscore-prefixed function, which hints that it's a private API, but subclasses of `ChangeProtected`s are supposed to override it. I think it should be just `is_in_use` (and it should have a proper docstring too).

akx:
> > This is also a semver major change.
> How so?

The publicly usable name `ImmutableMixin` is gone.

suutari-ai:
ImmutableMixin was never publicly usable. It's only available through an import from a private submodule (`shoop.core.models._base`).

akx:
@suutari-ai Oh, yeah, you're right. I didn't pay attention to the file name there... Sorry!

shoop/core/models/_base.py
suutari-ai: This PR didn't touch it.

Anyway, something being underscored does not mean it's private. It means that it's non-public. Quoting from PEP8:

> Use one leading underscore only for non-public methods and instance variables.

Being non-public does not mean it's not supposed to be overridden, it just means it's not supposed to be called from outside the class hierarchy. I.e. underscoring hints more about being protected than private (in C++/Java terms). But, yeah this is not really well defined anywhere and people's expectation might vary. Good docstring would help much.

suutari-ai: Renamed it and added a docstring. Better now?

PR #267, opened 2015-11-28
Protected fields

Pikkupomo:
Add better error message when trying to save shop that has protection triggered.
Add protection handling to `MultiLanguageModelForm`

Refs SHOOP-1896

### DIFF
--- a/shoop/admin/modules/shops/views/edit.py
+++ b/shoop/admin/modules/shops/views/edit.py
@@ -11,14 +11,17 @@ from django.conf import settings
 from django.utils.translation import ugettext_lazy as _
 
 from shoop.admin.forms.widgets import MediaChoiceWidget
-from shoop.admin.utils.views import CreateOrUpdateView
 from shoop.admin.toolbar import get_default_edit_toolbar
+from shoop.admin.utils.views import CreateOrUpdateView
 from shoop.core.models import Shop
+from shoop.core.utils.form_mixins import ProtectedFieldsMixin
 from shoop.utils.excs import Problem
 from shoop.utils.multilanguage_model_form import MultiLanguageModelForm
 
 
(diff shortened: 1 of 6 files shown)


shoop/utils/multilanguage_model_form.py
```diff
@@ -87,6 +87,12 @@ def __init__(self, **kwargs):
         kwargs["initial"] = object_data
         super(MultiLanguageModelForm, self).__init__(**kwargs)
 
+        if all([self.instance, self.instance.pk, hasattr(self.instance, "protected_fields")]):
```
akx: I'm not sure this is very good use of `all` :) Also, that oughta be `getattr` with an empty tuple as the default arg to be clear...

akx: You're using this same idiom (`if ...: for field in ...:`) in both these functions; might be worth to refactor it to

```python
def _get_protected_fields(self):
    if self.instance and self.instance.pk:
        return getattr(self.instance, "protected_fields", ())
    return ()
```

(with a suitable docstring, of course ;) )

shoop/admin/modules/shops/views/edit.py
```diff
@@ -6,19 +6,20 @@
 # This source code is licensed under the AGPLv3 license found in the
 # LICENSE file in the root directory of this source tree.
 from __future__ import unicode_literals
-
 from django.conf import settings
 from django.utils.translation import ugettext_lazy as _
-
 from shoop.admin.forms.widgets import MediaChoiceWidget
```
suutari-ai: Why are you removing the grouping of imports?

akx: Probably PyCharm being overzealous.

PR #268, opened 2015-11-30
Core: Enable creating immutable addresses

tulimaki:
Refs SHOOP-1897 / SHOOP-1799


akx:
Hmm. To be honest, it'd be more intuitive to just skip immutability validation in `Address.save()` when `self.pk = None`.

akx:
Seems like a cleaner patch! :)

PR #270, opened 2015-11-30
Taxes: Cleanup TODOs

tulimaki:
* Remove SHOOP_ORDER_LINE_TOTAL_DECIMALS and SHOOP_ORDER_TOTAL_DECIMALS
from shoop/core/settings.py. Instead of SHOOP_ORDER_LINE_TOTAL_DECIMALS use
2 in dashboard and orders cache_prices. Proper fix for this will be done
in SHOOP-1912.
* Documenting Orders.prices_include_tax will be done in SHOOP-933.
* Orders.customer_tax_group is not needed
* SHOOP-1911 will solve TODO in OrderSource.calculate_taxes_automatically
* From front/basket/objects.py remove temporary asserts from __init__.
Also get_final_lines was already moved to OrderSource
* Remove shoop/simple_pricing/models.py as not used
* Templates _confirm_table.jinja and front/macros.jinja already handled
taxes correctly
* Removed test_methods/test_tax as a bad test

Refs SHOOP-1623


shoop/admin/modules/orders/dashboard.py
```diff
@@ -47,7 +47,7 @@ def get_chart(self):
         bar_chart.add_data(
             _("Sales (%(currency)s)") % {"currency": self.currency},
             [
-                bankers_round(v["sum"], settings.SHOOP_ORDER_TOTAL_DECIMALS)
+                bankers_round(v["sum"], 2)
```
akx: This probably oughta have some sort of `# TODO: To be fixed in SHOOP-1912` thing, then?

shoop/core/models/orders.py
```diff
@@ -620,4 +619,4 @@ def get_status_display(self):
 
 
 def _round_price(value):
-    return bankers_round(value, settings.SHOOP_ORDER_TOTAL_DECIMALS)
+    return bankers_round(value, 2)
```
akx: This probably oughta have some sort of `# TODO: To be fixed in SHOOP-1912` thing, then?

shoop_tests/core/test_methods.py
```diff
@@ -152,23 +152,6 @@ def test_weight_limits():
     assert any(ve.code == "max_weight" for ve in sm.get_validation_errors(source))
 
 
-# TODO: (TAX) Taxing of shipping methods
-@pytest.mark.xfail # TODO: (TAX) Make this test not fail
```
akx: Would it be better to make this test actually not fail instead of removing it wholesale? :(

tulimaki: Well not sure what this is testing? Also ShippingMethod without tax_class can't be done at all. With @suutari-ai we though that since this test is poor and not sure what this is actually testing it should be removed.

suutari-ai:
Nice cleaning. Ah, I feel so fresh now! :-D

Thanks @tulimaki!

shoop_tests/core/test_methods.py
akx: Okay. :)

PR #271, opened 2015-11-30
Admin order creation: Ensure fields are not left empty

Pikkupomo:
Refs SHOOP-1899, SHOOP-1900, SHOOP-1901

### DIFF
--- a/shoop/admin/modules/orders/static_src/create/reducers/lines.js
+++ b/shoop/admin/modules/orders/static_src/create/reducers/lines.js
@@ -113,6 +113,7 @@ function setLineProperty(state, {payload}) {
                 break;
             case "type":
                 updates.type = value;
+                updates.errors = null;
                 if (value === "other" || value === "text") {
                     updates.product = null;
                     updates.sku = null;
--- a/shoop/admin/modules/orders/static_src/create/reducers/lines.js
+++ b/shoop/admin/modules/orders/static_src/create/reducers/lines.js
@@ -40,20 +40,26 @@ function setLineProperties(linesState, lineId, props) {
 
 function getDiscountsAndTotal(quantity, baseUnitPrice, unitPrice, updateUnitPrice=false) {
     const updates = {};
+
     if (updateUnitPrice) {
         updates.unitPrice = unitPrice;
     }
     var totalBeforeDiscount = baseUnitPrice * quantity;
     var total = +(unitPrice * quantity).toFixed(2);
     updates.total = total;
+
     if (baseUnitPrice < unitPrice || unitPrice < 0) {
         updates.discountPercent = 0;
         updates.discountAmount = 0;
         return updates;
     }
     var discountAmount = totalBeforeDiscount - total;
+    if (isNaN(discountAmount)) {
+        discountAmount = 0;
+    }
     updates.discountAmount = discountAmount;
     updates.discountPercent = ((discountAmount / totalBeforeDiscount) * 100).toFixed(2);
+
     return updates;
 }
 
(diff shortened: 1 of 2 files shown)


shoop/admin/modules/orders/static_src/create/reducers/lines.js
```diff
@@ -40,20 +40,26 @@ function setLineProperties(linesState, lineId, props) {
 
 function getDiscountsAndTotal(quantity, baseUnitPrice, unitPrice, updateUnitPrice=false) {
     const updates = {};
+
     if (updateUnitPrice) {
-        updates.unitPrice = unitPrice;
+        updates.unitPrice = "" + unitPrice;
```
akx: Er... is it really a good idea to store everything as strings in the actual data? :open_mouth:

shoop/admin/modules/orders/static_src/create/reducers/lines.js
```diff
@@ -83,6 +89,13 @@ function updateLineFromProduct(state, {payload}) {
     return setLineProperties(state, id, updates);
 }
 
+function ensureNumericValue(value, defaultValue=0) {
+    value = parseFloat(value);
+    if (isNaN(value))
```
akx: Braces, please

shoop/admin/modules/orders/static_src/create/reducers/lines.js
```diff
@@ -111,33 +124,44 @@ function setLineProperty(state, {payload}) {
...
+                    line.baseUnitPrice,
+                    ensureNumericValue(value, line.baseUnitPrice),
+                    true
+                );
                 break;
             case "discountPercent":
-                var discountPercent = Math.max(0, parseFloat(value));
+                var discountPercent = ensureNumericValue(value);
+                discountPercent = Math.max(0, discountPercent);
+                if (discountPercent > 100)
```
akx: Braces, please. Also, as you're using the idiom `Math.max` on the line above, just do

```javascript
const discountPercent = Math.min(100, Math.max(0, ensureNumericValue(discountPercent)));
```

akx:
The discount percent check should be another commit.

In addition, as mentioned in the code comment, I don't think it's correct to save numeric values as strings?

Also, the commit message doesn't explain why this change was necessary.

shoop/admin/modules/orders/static_src/create/reducers/lines.js
```diff
@@ -111,33 +124,44 @@ function setLineProperty(state, {payload}) {
                 updates.type = value;
                 break;
             case "quantity":
-                var quantity = Math.max(0, parseFloat(value));
+                const quantity = Math.max(0, ensureNumericValue(value, 1));
```
tulimaki: Actually if this is product line quantity should default as min quantity defined for product. That value should got with retrieveProductData.

Pikkupomo: I suppose this can be fixed in future. Imo not in the scope of this issue.

shoop/admin/modules/orders/static_src/create/reducers/lines.js
```diff
@@ -111,33 +126,41 @@ function setLineProperty(state, {payload}) {
...
+                    true
                 );
-                updates.discountAmount = discountAmount;
                 break;
             case "total":
-                var total = +parseFloat(value).toFixed(2);
-                updates = getDiscountsAndTotal(line.quantity, line.baseUnitPrice, (total / line.quantity), true);
-                updates.total = total;
+                const calculatedTotal = line.quantity * line.baseUnitPrice;
+                const total = +ensureNumericValue(value, calculatedTotal).toFixed(2);
```
akx: This should probably have a `TODO` regarding the hard-coded 2 decimal rounding?

shoop/admin/modules/orders/static_src/create/view/lines.js
```diff
@@ -15,7 +15,7 @@ function renderNumberCell(store, line, value, fieldId, canEditPrice, min=null) {
             type: "number",
             step: line.step,
             min: min != null ? min : "",
-            value: value,
+            value: "" + value,
```
akx: And a comment explaining why the string cast is necessary would be appreciated.

PR #273, opened 2015-11-30
Order creator improvements

tulimaki:
Admin: Make order creator confirmation page cancel to back
Admin: Fix order creator customer/address handling
Admin: Add order creator line text max length to 256
Admin: Make order creator check orderability errors creating OrderSource
Admin: Order creation make customer prices work for products

Refs SHOOP-1799

### DIFF
--- a/shoop/admin/modules/orders/json_order_creator.py
+++ b/shoop/admin/modules/orders/json_order_creator.py
@@ -149,7 +149,7 @@ class JsonOrderCreator(object):
         email = billing_address.get("email", "")
         fields = {"name": name, "phone": phone, "email": email}
         if is_company:
-            tax_number = billing_address.pop("tax_number", None)
+            tax_number = billing_address.get("tax_number", None)
             if not tax_number:
                 self.add_error(
                     ValidationError(_("Tax number is not set for company."), code="no_tax_number")
--- a/shoop/admin/modules/orders/json_order_creator.py
+++ b/shoop/admin/modules/orders/json_order_creator.py
@@ -98,8 +98,21 @@ class JsonOrderCreator(object):
                 "shop": source.shop
             }), code="no_shop_product"))
             return False
+        supplier = shop_product.suppliers.first()  # TODO: Allow setting a supplier?
+        is_orderable = True
+        for message in shop_product.get_orderability_errors(
+                supplier=supplier, quantity=sl_kwargs["quantity"], customer=source.customer):
+            self.add_error(ValidationError((_("Product %(product)s is not orderable: %(error)s") % {
+                "product": product,
+                "error": str(message.args[0])
+            }), code=str(message.args[1])))
+            is_orderable = False
+
+        if not is_orderable:
+            return False
+
         sl_kwargs["product"] = product
-        sl_kwargs["supplier"] = shop_product.suppliers.first()  # TODO: Allow setting a supplier?
+        sl_kwargs["supplier"] = supplier
         sl_kwargs["type"] = OrderLineType.PRODUCT
         sl_kwargs["sku"] = product.sku
         sl_kwargs["text"] = product.name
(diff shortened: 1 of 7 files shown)


shoop/admin/modules/orders/json_order_creator.py
```diff
@@ -98,8 +98,21 @@ def _process_product_line(self, source, sline, sl_kwargs):
...
             }), code="no_shop_product"))
             return False
+        supplier = shop_product.suppliers.first()  # TODO: Allow setting a supplier?
+        is_orderable = True
+        for message in shop_product.get_orderability_errors(
+                supplier=supplier, quantity=sl_kwargs["quantity"], customer=source.customer):
+            self.add_error(ValidationError((_("Product %(product)s is not orderable: %(error)s") % {
+                "product": product,
+                "error": str(message.args[0])
+            }), code="product_not_orderable"))
```
akx: Would probably be nice if this tried to use the `code` of the original `ValidationError` from `get_orderability_errors`?

shoop/admin/modules/orders/static_src/create/reducers/customer.js
```diff
@@ -31,7 +31,13 @@ function setAddressProperty(state, {payload}) {
 }
 
 export default handleActions({
-    setCustomer: ((state, {payload}) => payload),
+    setCustomer: ((state, {payload}) => _.assign(state, {
```
akx: This could just do `_.assign(state, payload)` to copy all properties from `payload` to the `state` without having to enumerate all of them by hand, couldn't it?

shoop/admin/modules/orders/static_src/create/view/customers.js
```diff
@@ -40,6 +40,8 @@ function renderAddress(store, shop, customer, address, addressType) {
 export function customerSelectView(store) {
     const {customer, shop} = store.getState();
     return m("div.form-group", [
+        m("p", gettext("By not selecting existing customer new customer is created based on billing address.")),
```
akx: Maybe show this only if no customer is selected? :) Also, in that case would just

> A new customer will be created.

do?

akx:
No test cases for the new backend behavior?

shoop/admin/modules/orders/json_order_creator.py
tulimaki: It does uses messages from from get_orderabilty_errors. Those messages is just like "This product is not visible" so I would like to add some product information to errors.

shoop/admin/modules/orders/static_src/create/reducers/customer.js
tulimaki: If I do that saveAddress and shipToBillingAddress is set to it's default value since those are missing from payload.

akx: If those keys aren't there in the payload, they won't be cleared.

image

shoop/admin/modules/orders/json_order_creator.py
akx: Yes, but the _code_ value :)

tulimaki:
Some new tests are added in separate PR.

shoop/admin/modules/orders/json_order_creator.py
```diff
@@ -98,8 +98,21 @@ def _process_product_line(self, source, sline, sl_kwargs):
...
             }), code="no_shop_product"))
             return False
+        supplier = shop_product.suppliers.first()  # TODO: Allow setting a supplier?
+        is_orderable = True
+        for message in shop_product.get_orderability_errors(
+                supplier=supplier, quantity=sl_kwargs["quantity"], customer=source.customer):
+            self.add_error(ValidationError((_("Product %(product)s is not orderable: %(error)s") % {
+                "product": product,
+                "error": str(message.args[0])
+            }), code=str(message.args[1])))
```
akx: `str(getattr(message, "code", "product_not_orderable"))` would be better.

PR #274, opened 2015-11-30
Translation extraction (SHOOP-1886)

suutari-ai:
Add tools for extracting and compiling translations.

 * makemessages monkey patched to work better with Jinja2.
 * Add shoop_makemessages helper.
 * Add "setup.py build_messages" command.

Split translations from a single monolith message catalog to one for
each app.

Make Javascript translation catalogs contain messages from all installed
apps to better support addons and such.

Also has couple message fixes.


suutari-ai:
This replaces PR #254.

akx:
> Split translations from a single monolith message catalog to one for each app.

Won't this mean that translators may need to translate the same string multiple times, as it exists in several catalogs (resources in Transifex speak)?

akx:
> Monkey patch the templatize function.

Hmm -- what was wrong with Jinja's own AST-based extractor? Regex monkey patching sounds fragile...

suutari-ai:
>>    Split translations from a single monolith message catalog to one for each app.

> Won't this mean that translators may need to translate the same string multiple times,
> as it exists in several catalogs (resources in Transifex speak)?

Should be no problem, if translator uses proper tools, since those will remember the earlier translations. Also, the same text does not always translate to same in different context.

suutari-ai:
> Imho it would be a good idea to use named placeholders everywhere.

Maybe so. At least we should not use messages like "foo %s bar %s baz" where the translator cannot change the order of the placeholders. That's a real problem. Naming/numbering of the placeholders is cosmetic.

We should do a separate clean-up for the translation messages later. This PR was not about that anyway. Those were just fixed as a test to see how that will affect the warnings yielded by "makemessages" and since the translations ended up to the *.po files, included the changes too.

Pikkupomo:
LGTM

PR #276, opened 2015-12-01
Add Travis CI

akx:
[Seems to work!](https://travis-ci.org/akx/shoop/builds/94130836)

Once the wheels have been cached, the install step is a snappy 10 to 18 seconds, so the build time is about 2 minutes altogether. Not bad!

The project has been set up on Travis, but build hooks are currently switched off there.

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -1,0 +1,13 @@
+sudo: false
+language: python
+cache:
+  directories:
+    - $HOME/.cache/pip
+python:
+  - "2.7"
+  - "3.4"
+install:
+  - pip install -U pip
+  - pip install coveralls -e .[everything]
+script: py.test -vvv --cov shoop shoop_tests
+after_success: coveralls
--- a/setup.py
+++ b/setup.py
@@ -89,7 +89,7 @@ REQUIRES = [
     'jsonfield==1.0.3',
     'Markdown==2.6.2',
     'pytz==2015.4',
-    'requests==2.7.0',
+    'requests>=2.7.0,<3',
     'six==1.9.0',
     'Jinja2==2.8',
     'pytoml==0.1.4'
--- a/tox.ini
+++ b/tox.ini
@@ -12,9 +12,9 @@ deps =
     beautifulsoup4==4.3.2
     mock==1.0.1
     pytest-cache==1.0
-    pytest==2.7.2
-    pytest-cov==1.8.1
-    pytest-django==2.8.0
+    pytest==2.8.4
+    pytest-cov==2.2.0
+    pytest-django==2.9.1
     # END testing deps
 commands = \
     py.test \


.travis.yml
```diff
@@ -0,0 +1,13 @@
...
+cache:
+  directories:
+    - $HOME/.cache/pip
+python:
+  - "2.7"
+  - "3.4"
+install:
+  - pip install -U pip
+  - pip install pytest==2.8.3 coveralls -e .[everything]
+script: py.test -vvv --nomigrations --cov shoop shoop_tests
```
suutari-ai: Ah.. of course it's faster... `--nomigrations` is cheating :P

akx: Should I remove it?

suutari-ai: IMHO yes, since migrations will not be tested otherwise.

setup.py
```diff
@@ -88,7 +88,7 @@
     'jsonfield==1.0.3',
     'Markdown==2.6.2',
     'pytz==2015.4',
-    'requests==2.7.0',
+    'requests==2.8.1',
```
suutari-ai: How does this relate to the PR?

tox.ini
```diff
@@ -12,9 +12,9 @@ deps =
     beautifulsoup4==4.3.2
     mock==1.0.1
     pytest-cache==1.0
-    pytest==2.7.2
-    pytest-cov==1.8.1
-    pytest-django==2.8.0
+    pytest==2.8.3
+    pytest-cov==2.2.0
+    pytest-django==2.9.1
```
suutari-ai: Are these updates needed for Travis or are you updating them just in case?

setup.py
akx: Travis already has Requests==2.8.1 installed, so installing Shoop was downgrading it without any good reason.

tox.ini
akx: There was a version conflict with the py library while installing the older version of pytest (since Travis comes with a version of pytest preinstalled), hence the upgrade. Should check whether the explicit install of the same version in `.travis.yml` is still needed though...

setup.py
suutari-ai: Maybe `requests>=2.7.0` would be more appropriate then?

tox.ini
suutari-ai: Can these also be specified as `>=`... I would prefer that, if possible... What do you think?

Also, comment the reason for the bumps in the commit message.

setup.py
akx: That will happily upgrade to versions which might be API incompatible.

tox.ini
akx: I think these versions should be locked down. Would avoid needless checks against the PyPI registry, when we already have the correct wheels cached.

akx:
Changes made and migrations re-enabled.

Still works: https://travis-ci.org/akx/shoop/builds/95535827

setup.py
```diff
@@ -89,7 +89,7 @@
     'jsonfield==1.0.3',
     'Markdown==2.6.2',
     'pytz==2015.4',
-    'requests==2.7.0',
+    'requests>=2.7.0,<3',
```
akx: This was also an excellent excuse to add a `<3` to the codebase.

suutari-ai:
Let's try this. There should be no harm done, since Jenkins is still running the testing too.

Thanks @akx!

akx:
@suutari-ai: Cool! I've hit the proverbial switch both on Travis and Coveralls, so future pull requests and pushes should now trigger builds.

PR #279, opened 2015-12-02
Admin: Fix admin JS

tulimaki:
Admin: Fix BasePopupChoiceWidget filter
Admin: drop strict-mode from dashboard
Admin js-packages: Enable es6 for picotable

Refs SHOOP-1839

### DIFF
--- a/shoop/admin/forms/widgets.py
+++ b/shoop/admin/forms/widgets.py
@@ -8,13 +8,13 @@
 from __future__ import unicode_literals
 
 from django.forms import HiddenInput, Widget
-from django.forms.utils import flatatt
 from django.utils.encoding import force_text
 from django.utils.html import escape
 from django.utils.safestring import mark_safe
 from django.utils.translation import ugettext_lazy as _
 from filer.models import File
 
+from shoop.admin.utils.forms import flatatt_filter
 from shoop.admin.utils.urls import NoModelUrl, get_model_url
 from shoop.core.models import Product
 
@@ -76,7 +76,7 @@ class BasePopupChoiceWidget(Widget):
             bits.insert(1, self.get_clear_markup())
 
         return mark_safe("<div %(attrs)s>%(content)s</div>" % {
-            "attrs": flatatt({
+            "attrs": flatatt_filter({
                 "class": "browse-widget %s-browse-widget" % self.browse_kind,
                 "data-browse-kind": self.browse_kind,
                 "data-clearable": self.clearable,
--- a/shoop/admin/gulp_bits/js-packages.js
+++ b/shoop/admin/gulp_bits/js-packages.js
@@ -32,6 +32,7 @@ module.exports = {
         ]
     },
     "dashboard": {
+        "es6": true,
         "files": bowerFiles("./static_src/dashboard")
     },
     "media-browser": {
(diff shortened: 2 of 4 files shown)


shoop/admin/gulp_bits/js-packages.js
```diff
@@ -67,6 +68,7 @@ module.exports = {
         ]
     },
     "order-creator": {
+        "es6": true,
```
akx: The es6 switch has no effect in webpack bundles.

akx:
Re the Dashboard change: `const` really shouldn't be used in non-transpiled code at all due to flaky cross-browser support at all. Would be better to enable `es6` for the file, and change the logic so Chartist (and likely all `bower_components`) don't get es6 transpiled. http://caniuse.com/#feat=const

suutari-ai:
Commit 8abf97a:
> Admin: Fix BasePopupChoiceWidget filter
>
> Since changes to filter in popup widgets we can't pass
> data-filter="None" to browse widget anymore since it is not
> JSON serializable.

The contents of this commit is OK, but the commit message is somehow misleading. I see no JSON serialization here, but does that even matter. There is no point in rendering data-filter attribute if there is no filter set anyway (no matter if it is JSON serialized or not). (Maybet that's what @akx meant too?)

suutari-ai:
Re @akx comment:
> Would be better to enable es6 for the file, and change the logic so Chartist (or anything in bower_components) don't get es6 transpiled.

I agree on this. @tulimaki/@akx: Would this be hard to do?

akx:
@suutari-ai: very easy. Just a gulp-if filter in the JS metatask.

shoop/admin/forms/widgets.py
```diff
@@ -75,14 +75,17 @@ def render(self, name, value, attrs=None):
...
+
         return mark_safe("<div %(attrs)s>%(content)s</div>" % {
-            "attrs": flatatt({
-                "class": "browse-widget %s-browse-widget" % self.browse_kind,
-                "data-browse-kind": self.browse_kind,
-                "data-clearable": self.clearable,
-                "data-empty-text": self.empty_text,
-                "data-filter": self.filter
-            }),
+            "attrs": flatatt(browser_attrs),
```
akx: It'd be better to just use `flatatt_filter` (in `shoop.admin.toolbar`), which automatically elides falsy values.

akx:
@suutari-ai, @tulimaki: re https://github.com/shoopio/shoop/commit/8abf97a61b7c8a42e97f84499ff408450bc2221e --

I also wanted to know what JSON serialization is suddenly having trouble.

But as mentioned in the inline code comment just now, we have an utility for eliding the attr, `flatatt_filter` :)

tulimaki:
@akx https://github.com/shoopio/shoop/commit/d53b733b6c058a45aba91fa525c7ef2453763063

suutari-ai:
@tulimaki: Include the commit ref about the JSON serialization having trouble to the commit message...

akx:
@tulimaki,right, so that would also be fixed by turning `"data-filter": self.filter` to `"data-filter": self.filter or ""`, to make it serialize into an empty attribute. But yeah, `flatatt_filter`.

suutari-ai:
@tulimaki Superb! Thanks!

PR #281, opened 2015-12-02
Admin list sorting

tulimaki:
Simple-cms: Fix admin module sorting with title
Admin: Fix order list sorting with taxful_total_price

Refs SHOOP-1836

### DIFF
--- a/shoop/simple_cms/admin_module/views.py
+++ b/shoop/simple_cms/admin_module/views.py
@@ -128,7 +128,10 @@ class PageEditView(CreateOrUpdateView):
 class PageListView(PicotableListView):
     model = Page
     columns = [
-        Column("title", _(u"Title"), linked=True, filter_config=TextFilter(operator="startswith")),
+        Column(
+            "title", _(u"Title"), sort_field="translations__title", display="title", linked=True,
+            filter_config=TextFilter(operator="startswith")
+        ),
         Column("available_from", _(u"Available from")),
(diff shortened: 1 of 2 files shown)


shoop/simple_cms/admin_module/views.py
```diff
@@ -128,7 +128,10 @@ def save_form(self, form):
 class PageListView(PicotableListView):
     model = Page
     columns = [
-        Column("title", _(u"Title"), linked=True, filter_config=TextFilter(operator="startswith")),
+        Column(
+            "title",_(u"Title"), sort_field="translations__title", display="title", linked=True,
```
suutari-ai: Repeating Jenkins here:
> missing whitespace after ','

shoop/simple_cms/admin_module/views.py
```diff
@@ -128,7 +128,10 @@ def save_form(self, form):
 class PageListView(PicotableListView):
     model = Page
     columns = [
-        Column("title", _(u"Title"), linked=True, filter_config=TextFilter(operator="startswith")),
+        Column(
+            "title", _(u"Title"), sort_field="translations__title", display="title", linked=True,
+            filter_config=TextFilter(operator="startswith")
+        ),
```
suutari-ai: Adding display="title" is not really needed, since default display is the same as id which is also "title" here, but that doesn't do any harm either.

PR #283, opened 2015-12-03
Admin Gulp: Really fix the es6 switch.

akx:
You can use `export DEBUG=*` before running `gulp js:dashboard` to verify Babel only traverses our components:

image


suutari-ai:
Nice that you included @tulimaki's commit as cherry-picked here too :)

suutari-ai:
We don't have any tests for the gulp stuff so no point in waiting Jenkins with this.

akx:
@suutari-ai : Had to, so I could see it worked :)

PR #284, opened 2015-12-03
MANIFEST.in: Include *.map to the Python package

suutari-ai:
Include the source maps of CSS and JS files in the Python
packages (including wheel) to make developers happy.


akx:
Happier, anyway. :D

PR #285, opened 2015-12-03
Do not forcibly update POT-Creation-Date with makemessages (SHOOP-1932)

suutari-ai:
Doing "shoop_makemessages" in the shoop directory updates all the
translation base files, but it unnecessarily updates also the
POT-Creation-Date header which causes all PO files to change even if
there is no change to messages.  This makes it hard to periodically
update the message files (with clean commits).

Update the Shoop's "makemessages" command to support skipping
POT-Creation-Date updating of PO files which don't have any changes to
messages and make "shoop_makemessages" to toggle that option on by
default.

### DIFF
--- a/shoop/core/management/commands/makemessages.py
+++ b/shoop/core/management/commands/makemessages.py
@@ -10,6 +10,8 @@ Patched version of Django's Makemessages that works with Jinja2.
 Works by monkey patching django.utils.translation.trans_real.templatize
 with our version.
 """
+from __future__ import unicode_literals
+
 import datetime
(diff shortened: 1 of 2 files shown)


suutari-ai:
Oh wait.. a bug..

suutari-ai:
doh.. another bug..

suutari-ai:
Ah, the second one was actually older bug. makemessages didn't work with Python2: added another commit to fix that one too.

tulimaki:
:lollipop:

PR #287, opened 2015-12-07
Product media with orders

tulimaki:
Core: Change ProductMedia url property to return correctly
Core: Add get_public_media-function for Product
Classic Gray: Add product media to templates

Refs SHOOP-1918

### DIFF
--- a/shoop/core/models/product_media.py
+++ b/shoop/core/models/product_media.py
@@ -82,13 +82,11 @@ class ProductMedia(TranslatableModel):
 
     @property
     def url(self):
-        if not self.public:
-            raise ValueError("`get_effective_url()` may not be used on non-public media")
-
-        if self.file_id:
-            return self.file.url
-        else:
+        if self.external_url:
             return self.external_url
+        if self.file:
+            return self.file.url
+        return ""
 
     @property
     def easy_thumbnails_thumbnailer(self):
--- a/shoop/core/models/products.py
+++ b/shoop/core/models/products.py
@@ -535,6 +535,9 @@ class Product(TaxableItem, AttributableMixin, TranslatableModel):
             ProductPackageLink.objects.filter(parent=self).values_list("child", flat=True)
         ))
 
+    def get_public_media(self):
+        return self.media.filter(enabled=True, public=True)
+
 
 ProductLogEntry = define_log_model(Product)
 
--- a/shoop/themes/classic_gray/templates/classic_gray/shoop/front/product/detail.jinja
+++ b/shoop/themes/classic_gray/templates/classic_gray/shoop/front/product/detail.jinja
@@ -81,13 +81,16 @@
     </div>
     <hr>
     <h3>{% trans %}Product Information{% endtrans %}</h3>
-
+    {% set media_files = product.get_public_media() %}
     <div class="product-tabs" role="tabpanel">
         <ul class="nav nav-tabs" role="tablist">
             <li role="presentation" class="active"><a href="#tab1" aria-controls="tab1" role="tab" data-toggle="tab">{% trans %}Description{% endtrans %}</a></li>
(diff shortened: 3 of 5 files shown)


shoop/core/models/products.py
```diff
@@ -535,6 +535,10 @@ def get_all_package_children(self):
             ProductPackageLink.objects.filter(parent=self).values_list("child", flat=True)
         ))
 
+    def get_public_media(self):
+        from .product_media import ProductMedia
+        return ProductMedia.objects.filter(product=self, enabled=True, public=True)
```
suutari-ai: Should use the `self.media` related manager:
```python
return self.media.filter(enabled=True, public=True)
```
Though a custom manager might be good addition too, i.e. this would be just:
```python
return self.media.enabled().public()
```
That would allow redefining public/enabled in the future (e.g. to make public() check if the current time is in date range or whatever).

shoop/themes/classic_gray/templates/classic_gray/shoop/personal_order_history/order_detail.jinja
```diff
@@ -1,4 +1,5 @@
 {% extends "shoop/front/base.jinja" %}
+{% from "shoop/front/macros.jinja" import render_ordered_files with context %}
```
suutari-ai: Does this really need the context?

shoop/core/models/products.py
suutari-ai: The custom manager stuff is out of the scope of this issue though, so do the first one...

shoop/themes/classic_gray/templates/classic_gray/shoop/personal_order_history/order_detail.jinja
tulimaki: Context is not neccessary here.

shoop/core/models/product_media.py
```diff
@@ -82,13 +82,11 @@ def effective_title(self):
...
-            raise ValueError("`get_effective_url()` may not be used on non-public media")
-
-        if self.file_id:
-            return self.file.url
-        else:
+        if self.external_url:
             return self.external_url
+        if self.file:
+            return self.file.url
+        return "/"
```
akx: Is this actually a good idea?

shoop/themes/classic_gray/templates/classic_gray/shoop/front/product/detail.jinja
```diff
@@ -106,6 +109,23 @@
                     {% include "shoop/front/product/_detail_attributes_section.jinja" %}
                 </div>
             {% endif %}
+            {% if media_files %}
```
akx: This segment (112..128) looks suspiciously similar to the body of the `render_ordered_files` macro. Could they maybe be unified?

shoop/core/models/product_media.py
suutari-ai: We talked about this yesterday. The non-raising fallback is good, since the product details page wants to show the media with url.

@tulimaki Though, maybe the fallback should be "" rather than "/", because then it would be possible to check if there is a url by simply iffing for non-emptyness. Agree?

shoop/themes/classic_gray/templates/classic_gray/shoop/front/product/detail.jinja
suutari-ai: They probably shouldn't, because rendering ordered media files vs. public media files on the product detail page may want different layout...

This block here could use a same macro though and if another theme wants to use something else, it can overwrite it. I don't know... IMHO this approach is OK too.

suutari-ai:
Good job @tulimaki!

The render_ordered_files macro vs product detail page files rendering can be unified later, if so wished.

PR #290, opened 2015-12-13
Add `custom_css` field to theme configuration

Pikkupomo:
This would allow some theme customization until proper tools are implemented. Currently customers are using the `Footer HTML` field for this purpose

Refs no ref

### DIFF
--- a/shoop/themes/classic_gray/__init__.py
+++ b/shoop/themes/classic_gray/__init__.py
@@ -14,6 +14,12 @@ from shoop.apps import AppConfig
 from shoop.xtheme.theme import Theme
 
 
+CUSTOM_CSS_HELP = _("""This field can be used to customize your theme to match your branding.
+The contents of this field will be rendered inside 'style'
+tag in the template header so use only plain CSS here.""")
+FOOTER_LINKS_HELP = _("One line per link in format 'http://example.com Example Link'")
+
+
 class ClassicGrayTheme(Theme):
     identifier = "shoop.themes.classic_gray"
     name = "Shoop Classic Gray Theme"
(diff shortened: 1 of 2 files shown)


shoop/themes/classic_gray/__init__.py
```diff
@@ -22,6 +22,8 @@ class ClassicGrayTheme(Theme):
 
     fields = [
         ("show_welcome_text", forms.BooleanField(required=False, initial=True, label=_("Show Frontpage Welcome Text"))),
+        ("custom_css", forms.CharField(required=False, label=_("Custom CSS"), widget=forms.Textarea,
+                                        help_text=_("Plain CSS without 'style' tag"))),
```
akx: The help text should be more informative, if you ask me. Maybe

> Enter custom CSS code (without the surrounding <style> tags)

?

shoop/themes/classic_gray/templates/classic_gray/shoop/front/base.jinja
```diff
@@ -14,6 +14,9 @@
     <link href='https://fonts.googleapis.com/css?family=Open+Sans:400,300,700' rel='stylesheet' type='text/css'>
     {# Include Styles #}
     <link rel="stylesheet" href="{{ STATIC_URL }}classic_gray/css/style.css">
+    <style>
```
akx: Might be better to avoid outputting the tag at all if it's empty?

akx:
Ps. It would be nice to have a Refs link to the Shoop forums thread that requested this feature (even if this really should've gone through JIRA).

shoop/themes/classic_gray/__init__.py
```diff
@@ -14,6 +14,12 @@
 from shoop.xtheme.theme import Theme
 
 
+CUSTOM_CSS_HELP = _("""This field can be used to customize your theme to match your branding.
```
akx: > You can enter custom CSS here to customize your shop to match your branding.
> The contents of this field will be rendered inside a 'style' tag.

should be enough?

akx:
Also, this feature is missing a test case.

PR #292, opened 2015-12-14
Code style fixes (and Travis configuration to enforce it)

akx:


shoop/admin/modules/orders/json_order_creator.py
```diff
@@ -13,8 +13,8 @@
 from django.utils.translation import ugettext as _
 
 from shoop.core.models import (
-    Address, Contact, CompanyContact, PersonContact, OrderLineType,
-    OrderStatus, PaymentMethod, Product, ShippingMethod, Shop
+    Address, CompanyContact, Contact, OrderLineType, OrderStatus, PaymentMethod, PersonContact, Product, ShippingMethod,
+    Shop
 )
```
suutari-ai: These would be much more readable with the 79 columns. Since they will be splitted to multiple lines anyway, why make it so wide then?

shoop/admin/utils/picotable.py
```diff
@@ -11,11 +11,12 @@
 
 import six
 from django.core.paginator import EmptyPage, Paginator
-from django.db.models import Count, Manager, Q, QuerySet
+from django.db.models import Q, Count, Manager, QuerySet
```
suutari-ai: Hmm.. Something doesn't seem to know alphabets here..

shoop/admin/modules/orders/json_order_creator.py
akx: Oh, hey, cool, there's actually a setting exactly for that.

suutari-ai:
Why doesn't Travis find the ensure_license_headers.main too complex even if Jenkins does? (I kinda agree with Jenkins here... the main would be better to be splitted to couple subfunctions.)

akx:
@suutari-ai Travis' flake8 only looks in `shoop/`, not the whole tree.

suutari-ai:
Ah, Jenkins' flake8 looks whole checkout and ignores are controlled by setup.cfg. IMHO this is the way it should be.

akx:
@suutari-ai Changed flake8 to run on the whole tree (sans `venv`s and `local_settings` to accommodate local runs with the same args)

akx:
retest this please

tox.ini
```diff
@@ -15,6 +15,7 @@ deps =
     pytest==2.8.4
     pytest-cov==2.2.0
     pytest-django==2.9.1
+    isort==4.2.2
```
suutari-ai: Kinda ironic that these are not sorted now :P

suutari-ai:
Good stuff! Thanks!

tox.ini
akx: Pff, well if they were, `pytest` would be below the other `pytest-` packages, an' I dun' like it!

akx:
Nps! :)

PR #293, opened 2015-12-15
Split Address to MutableAddress and ImmutableAddress

tulimaki:
* Core: Make core work with mutable and immutable addresses
* Make front work work with mutable and immutable addresses
* Admin: Fix contact edit to work with immutable address
* Tests: Fix tests to work with mutable and immutable addresses
* Fix tests that failed with mysql
* Admin: Make order creator work with mutable addresses

Refs SHOOP-1924

### DIFF
--- a/ChangeLog.rst
+++ b/ChangeLog.rst
@@ -9,6 +9,8 @@ Unreleased
   releasing next version, the "Unreleased" header will be replaced with
   appropriate version header and this help text will be removed.
 
+- Remove AddressManager
+- Split address into mutable and immutable address
 
 Version 2.0.0
 -------------
--- a/shoop/admin/modules/orders/json_order_creator.py
+++ b/shoop/admin/modules/orders/json_order_creator.py
@@ -13,7 +13,7 @@ from django.core.exceptions import ObjectDoesNotExist, ValidationError
 from django.utils.translation import ugettext as _
 
 from shoop.core.models import (
-    Address, CompanyContact, Contact, OrderLineType, OrderStatus,
+    CompanyContact, Contact, MutableAddress, OrderLineType, OrderStatus,
     PaymentMethod, PersonContact, Product, ShippingMethod, Shop
 )
 from shoop.core.order_creator import OrderCreator, OrderSource
@@ -170,26 +170,12 @@ class JsonOrderCreator(object):
                 return
 
             fields.update({"tax_number": tax_number})
-            customer = CompanyContact.objects.create(**fields)
+            customer = CompanyContact(**fields)
         else:
-            customer = PersonContact.objects.create(**fields)
+            customer = PersonContact(**fields)
         return customer
 
-    def _create_address(self, address):
-        address_fields = [
-            "name", "tax_number", "phone", "email", "street", "street2", "postal_code", "city", "region", "country"
-        ]
-        fields = {"is_immutable": True}
-        for field in address_fields:
-            fields[field] = address.get(field, "")
-
-        address = Address.objects.filter(**fields).first()
-        if not address:
-            address = Address.objects.create(**fields)
-
-        return address
-
-    def _initialize_source_from_state(self, state, creator):
+    def _initialize_source_from_state(self, state, creator, save):
         shop_data = state.pop("shop", None).get("selected", {})
         shop = self.safe_get_first(Shop, pk=shop_data.pop("id", None))
         if not shop:
@@ -199,21 +185,28 @@ class JsonOrderCreator(object):
         source = OrderSource(shop=shop)
 
         customer_data = state.pop("customer", None)
-        billing_address = customer_data.pop("billingAddress", {})
-        shipping_address = customer_data.pop("shippingAddress", {})
-        ship_to_billing_address = customer_data.pop("shipToBillingAddress", False)
+        billing_address_data = customer_data.pop("billingAddress", {})
+        shipping_address_data = (
+            billing_address_data
+            if customer_data.pop("shipToBillingAddress", False)
+            else customer_data.pop("shippingAddress", {}))
         is_company = customer_data.pop("isCompany", False)
         save_address = customer_data.pop("saveAddress", False)
         customer = self.safe_get_first(Contact, pk=customer_data.get("id")) if customer_data else None
         if not customer:
-            customer = self._create_contact_from_address(billing_address, is_company)
+            customer = self._create_contact_from_address(billing_address_data, is_company)
             if not customer:
                 return
+            if save:
+                customer.save()
 
-        billing_address = self._create_address(billing_address)
-        shipping_address = billing_address if ship_to_billing_address else self._create_address(shipping_address)
+        billing_address = MutableAddress.from_data(billing_address_data)
+        shipping_address = MutableAddress.from_data(shipping_address_data)
+        if save:
+            billing_address.save()
+            shipping_address.save()
 
-        if save_address:
+        if save and save_address:
             customer.default_billing_address = billing_address
             customer.default_shipping_address = shipping_address
             customer.save()
@@ -246,7 +239,7 @@ class JsonOrderCreator(object):
         if comment:
             order.add_log_entry(comment, kind=LogEntryKind.NOTE, user=order.creator)
 
-    def create_source_from_state(self, state, creator=None):
+    def create_source_from_state(self, state, creator=None, save=False):
         """
         Create an order source from a state dict unserialized from JSON.
 
@@ -254,6 +247,8 @@ class JsonOrderCreator(object):
         :type state: dict
         :param creator: Creator user
         :type creator: django.contrib.auth.models.User|None
+        :param save: Flag whether order customer and addresses is saved to database
+        :type save: boolean
         :return: The created order source, or None if something failed along the way
         :rtype: OrderSource|None
         """
@@ -264,7 +259,7 @@ class JsonOrderCreator(object):
         state = deepcopy(state)
 
         # First, initialize an OrderSource.
-        source = self._initialize_source_from_state(state, creator)
+        source = self._initialize_source_from_state(state, creator, save)
         if not source:
             return None
 
@@ -286,7 +281,7 @@ class JsonOrderCreator(object):
         :return: The created order, or None if something failed along the way
         :rtype: Order|None
         """
-        source = self.create_source_from_state(state, creator)
+        source = self.create_source_from_state(state, creator, save=True)
 
         # Then create an OrderCreator and try to get things done!
         creator = OrderCreator(request=None)
--- a/shoop/admin/modules/orders/views/create.py
+++ b/shoop/admin/modules/orders/views/create.py
@@ -16,8 +16,8 @@ from django.contrib.auth.models import AnonymousUser
 from django.core import serializers
 from django.core.exceptions import ObjectDoesNotExist, ValidationError
 from django.core.urlresolvers import reverse
+from django.db import transaction
 from django.db.models import Q
-from django.db.transaction import atomic
 from django.http.response import HttpResponse, JsonResponse
 from django.test.client import RequestFactory
 from django.utils.encoding import force_text
@@ -207,45 +207,45 @@ class OrderCreateView(TemplateView):
             "shippingAddress": encode_address(customer.default_shipping_address)
         }
 
-    @atomic
-    def handle_source_data(self, request):
-        try:
-            state = json.loads(request.body.decode("utf-8"))["state"]
-            source = create_source_from_state(state, creator=request.user)
-            # Calculate final lines for confirmation
-            source.calculate_taxes(force_recalculate=True)
-            return {
-                "customerId": source.customer.id,
-                "taxfulTotal": format_money(source.taxful_total_price.amount),
-                "taxlessTotal": format_money(source.taxless_total_price.amount),
-                "totalDiscountAmount": format_money(source.total_discount.amount),
-                "orderLines": [encode_line(line) for line in source.get_final_lines(with_taxes=True)],
-                "billingAddress": source.billing_address.as_string_list() if source.billing_address else None,
-                "shippingAddress": source.shipping_address.as_string_list() if source.shipping_address else None,
-            }
-        except Exception as exc:
-            message = _("Could not proceed with order:")
-            if isinstance(exc, ValidationError):  # pragma: no branch
-                message += "\n" + "\n".join(force_text(err) for err in exc.messages)
-            else:  # pragma: no branch
-                message += " " + str(exc)
-            return JsonResponse({"success": False, "errorMessage": message}, status=400)
+    @transaction.atomic
+    def _handle_source_data(self, request):
+        state = json.loads(request.body.decode("utf-8"))["state"]
+        source = create_source_from_state(state, creator=request.user)
+        # Calculate final lines for confirmation
+        source.calculate_taxes(force_recalculate=True)
+        return {
+            "taxfulTotal": format_money(source.taxful_total_price.amount),
+            "taxlessTotal": format_money(source.taxless_total_price.amount),
+            "totalDiscountAmount": format_money(source.total_discount.amount),
+            "orderLines": [encode_line(line) for line in source.get_final_lines(with_taxes=True)],
+            "billingAddress": source.billing_address.as_string_list() if source.billing_address else None,
+            "shippingAddress": source.shipping_address.as_string_list() if source.shipping_address else None,
+        }
 
-    @atomic
+    @transaction.atomic
+    def _handle_create(self, request):
+        state = json.loads(request.body.decode("utf-8"))["state"]
+        order = create_order_from_state(state, creator=request.user)
+        messages.success(request, _("Order %(identifier)s created.") % vars(order))
+        return JsonResponse({
+            "success": True,
+            "orderIdentifier": order.identifier,
+            "url": reverse("shoop_admin:order.list")
+        })
+
+    def handle_source_data(self, request):
+        return _handle_or_return_error(self._handle_source_data, request, _("Could not proceed with order:"))
+
     def handle_create(self, request):
-        try:
-            state = json.loads(request.body.decode("utf-8"))["state"]
-            order = create_order_from_state(state, creator=request.user)
-            messages.success(request, _("Order %(identifier)s created.") % vars(order))
-            return JsonResponse({
-                "success": True,
-                "orderIdentifier": order.identifier,
-                "url": reverse("shoop_admin:order.list")
-            })
-        except Exception as exc:
-            message = _("Could not create order:")
-            if isinstance(exc, ValidationError):  # pragma: no branch
-                message += "\n" + "\n".join(force_text(err) for err in exc.messages)
-            else:  # pragma: no branch
-                message += " " + str(exc)
-            return JsonResponse({"success": False, "errorMessage": message}, status=400)
+        return _handle_or_return_error(self._handle_create, request, _("Could not create order:"))
+
+
+def _handle_or_return_error(func, request, error_message):
+    try:
+        return func(request)
+    except Exception as exc:
+        if isinstance(exc, ValidationError):
+            error_message += "\n" + "\n".join(force_text(err) for err in exc.messages)
+        else:
+            error_message += " {}".format(exc)
+        return JsonResponse({"success": False, "errorMessage": error_message}, status=400)
--- a/shoop/core/migrations/0013_address_changes.py
+++ b/shoop/core/migrations/0013_address_changes.py
@@ -1,0 +1,123 @@
+# -*- coding: utf-8 -*-
+from __future__ import unicode_literals
+
+from django.db import models, migrations
+import django_countries.fields
+import shoop.core.utils.name_mixin
+import django.db.models.deletion
+import shoop.core.models._base
+import shoop.core.fields
+import shoop.utils.models
+
+
+def copy_order_addresses(apps, schema_editor):
+    """
+    For orders old shipping and billing address foreign keys is saved
+    to mutable_billing_address and mutable_shipping_address. Get or create
+    new immutable addresses based on those fields and set those to order
+    billing_address and shipping_address.
+    """
+    order_cls = apps.get_model("shoop", "Order")
+    immutable_address_cls = apps.get_model("shoop", "ImmutableAddress")
+
+    for order in order_cls.objects.all():
+        if order.mutable_billing_address:
+            order.billing_address = immutable_address_cls.objects.get_or_create(
+                **shoop.utils.models.get_data_dict(order.mutable_billing_address))[0]
+
+        if order.mutable_shipping_address:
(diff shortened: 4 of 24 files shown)


shoop/admin/modules/orders/json_order_creator.py
```diff
@@ -254,6 +247,8 @@ def create_source_from_state(self, state, creator=None):
         :type state: dict
         :param creator: Creator user
         :type creator: django.contrib.auth.models.User|None
+        :param save: Flag whether order customer or addresses is saved to db
```
akx: You can afford to spell "database" out :) Also, shouldn't it be customer _and_ addresses?

shoop/core/migrations/0013_address_changes.py
```diff
@@ -0,0 +1,166 @@
...
+import shoop.core.models._base
+import shoop.core.fields
+import shoop.utils.models
+
+
+def from_data(cls, data, immutable=False):
+    """
+    Create Address from a data dictionary.
+
+    :type data: dict[str,str]
```
akx: The `cls` parameter is undocumented.

shoop/core/migrations/0013_address_changes.py
```diff
@@ -0,0 +1,166 @@
...
+    :type immutable: boolean
+    :rtype: Unsaved Address
```
akx: `:rtype:` should be machine-readable. Use `:return:` to explain what you're actually returning.

shoop/core/migrations/0013_address_changes.py
```diff
@@ -0,0 +1,166 @@
...
+    """
+    if immutable:
```
akx: What's the point of the `immutable` arg if you're passing in `MutableAddress` and `ImmutableAddress` as `cls`?

shoop/core/migrations/0013_address_changes.py
```diff
@@ -0,0 +1,166 @@
...
+        immutable_address = from_data(address_cls, address_data, immutable=True)
+        immutable_address.save()
+        updated_field = {field: immutable_address}
+        objects.update(**updated_field)
+
+
+def copy_addresses(apps, schema_editor):
+    """
+    Copy existing addresses to new Mutable and Immutable addresses
+    For every address new MutableAddress is created with same id
```
akx: If you're copying IDs, shouldn't you also update the table's autoincrement counter? :confused:

shoop/core/models/_base.py
```diff
@@ -85,7 +85,9 @@ def _get_changed_protected_fields(self):
             protected_fields = [
                 x.name for x in self._meta.get_fields()
                 if not x.is_relation and x.name not in self.unprotected_fields]
-        in_db = type(self).objects.get(pk=self.pk)
+        in_db = type(self).objects.filter(pk=self.pk).first()
+        if not in_db:
+            return []  # To enable initializing custom id in creation
```
akx: Uh...?

shoop/core/models/addresses.py
```diff
@@ -173,6 +140,50 @@ def as_string_list(self, locale=None):
...
+    Address that can be changed.
+    """
+    pass
+
+
+class ImmutableAddress(ChangeProtected, Address):
+    """
+    Address that cannot be changed.
+    """
+    def __init__(self, *args, **kwargs):
```
akx: Unnecessary init?

shoop/core/models/addresses.py
```diff
@@ -173,6 +140,50 @@ def as_string_list(self, locale=None):
...
+        super(ImmutableAddress, self).__init__(*args, **kwargs)
+
+    def immutable(self, ignores=()):
```
akx: This implementation ignores the `ignores`?

shoop/core/models/addresses.py
```diff
@@ -173,6 +140,50 @@ def as_string_list(self, locale=None):
     def __iter__(self):
         return iter(self.as_string_list())
 
+    def immutable(self, ignores=()):
```
akx: Would `as_immutable()` maybe be a better name for this?

shoop/core/order_creator/creator.py
```diff
@@ -189,8 +182,8 @@ def create_order(self, order_source):
             creator=real_user_or_none(order_source.creator),
             orderer=(order_source.orderer or None),
             customer=(order_source.customer or None),
-            billing_address=order_source.billing_address,
-            shipping_address=order_source.shipping_address,
+            billing_address=order_source.billing_address.immutable() if order_source.billing_address else None,
```
akx: Parens around the rvalue, please :)

shoop/core/order_creator/creator.py
```diff
@@ -189,8 +182,8 @@ def create_order(self, order_source):
...
+            shipping_address=order_source.shipping_address.immutable() if order_source.shipping_address else None,
```
akx: Parens around the rvalue, please :)

shoop/core/models/addresses.py
akx: All in all, the semantics of the `immutable()` and `new_mutable()` (which have confusing names if you ask me) should be better documented. For instance, as I see it, this method may sometimes save an all new address? Maybe?

Also, I see no test cases focusing particularly on these new APIs; even if they might be covered by other tests, I'd still like to see them being explicitly tested.

shoop/core/models/addresses.py
```diff
@@ -173,6 +140,50 @@ def as_string_list(self, locale=None):
...
+        :rtype: MutableAddress
+        :return: Fresh unsaved MutableAddress with same data as self.
+        """
+        data = get_data_dict(self)
+        return MutableAddress.from_data(data)
+
+
+class MutableAddress(Address):
+    """
+    Address that can be changed.
```
akx: Could you document where these are used, please?

shoop/core/models/addresses.py
```diff
@@ -173,6 +140,50 @@ def as_string_list(self, locale=None):
...
+    """
+    pass
+
+
+class ImmutableAddress(ChangeProtected, Address):
+    """
+    Address that cannot be changed.
```
akx: Could you document where these are used, please? And how they are created?

shoop/core/models/addresses.py
```diff
@@ -115,7 +75,34 @@ class Address(NameMixin, models.Model):
...
+    def from_data(cls, data, ignores=()):
+        """
+        Create Address from a data dictionary.
+
+        :type data: dict[str,str]
+        :type ignores: Iterable[str]
+        :param ignores:
+          Fields names to ignore when searching for existing immutable
+          address.
+        :rtype: Address
```
akx: Maybe `:rtype: MutableAddress|ImmutableAddress`, because this will actually never return an `Address`.

shoop/core/models/addresses.py
```diff
@@ -115,7 +75,34 @@ class Address(NameMixin, models.Model):
...
+        """
+        immutable = (
+            issubclass(cls, ImmutableAddress) or
+            data.pop('immutable', False))
+        for field_name in ignores:
+            data.pop(field_name, None)
```
akx: If we fall through to line 105, the newly created address will lack some fields.

shoop/core/models/addresses.py
```diff
@@ -54,51 +55,10 @@ class Labels:
         ENABLED = _('enabled')
 
 
-class AddressManager(models.Manager):
```
akx: AddressManager's removal should occur in a commit of its own, to make it easier to find for the changelog.

shoop/core/migrations/0013_address_changes.py
```diff
@@ -0,0 +1,166 @@
...
+        mutable_address.id = address.id
+        mutable_address.save()
+
+        populate_immutable_addresses(
+            order_cls, "billing_address", immutable_address_cls, address, address_data_dict)
+        populate_immutable_addresses(
+            order_cls, "shipping_address", immutable_address_cls, address, address_data_dict)
+
+
+class Migration(migrations.Migration):
```
akx: Could this migration maybe rename the old single address table to either mutable or immutable? I mean something like


* Create `ImmutableOrder` table
* Rename `Order.billing_address` to `Order.mutable_billing_address`
* Rename `Order.shipping_address` to `Order.mutable_shipping_address`
* Add `Order.billing_address` FK (to immutable table)
* Add `Order.shipping_address` FK (to immutable table)
* Copy mutable addresses to immutable addresses and backfill FKs in Order
* Remove `Address.immutable`
* Rename `Address` table to `MutableAddress`
* Remove `Order.mutable_*`

shoop/core/migrations/0013_address_changes.py
suutari-ai: @akx is right. The autoincrement should be updated. Though I don't think there is any way to do that in a DBMS agnostic way...

This could be a big problem. Maybe the immutable addresses should be created without manually assigning any ids and just store a mapping "order_id -> immutable_address_id". This would mean that address ids in order has to be NULLed while creating the mapping and then finally when the foreign keys of order table are updated, their values should be set from the mapping.

akx: @suutari-ai : What about my proposed way of migrating this below?

shoop/core/migrations/0013_address_changes.py
suutari-ai: @akx: Thanks! This was a very good idea!

shoop/core/models/__init__.py
```diff
@@ -4,7 +4,7 @@
 #
 # This source code is licensed under the AGPLv3 license found in the
 # LICENSE file in the root directory of this source tree.
-from .addresses import Address, SavedAddress, SavedAddressRole, SavedAddressStatus
+from .addresses import MutableAddress, ImmutableAddress, SavedAddress, SavedAddressRole, SavedAddressStatus
```
suutari-ai: Shouldn't Immutable go before Mutable?

shoop/core/models/addresses.py
suutari-ai: Now that @tulimaki squashed the address splitting changes to a single commit, it's fine IMHO.

@tulimaki Just add the entry to ChangeLog.rst about this change within the "Split address into mutable and immutable address" commit, please.

shoop/core/models/addresses.py
```diff
@@ -115,7 +75,30 @@ class Address(NameMixin, models.Model):
...
+        verbose_name_plural = _('addresses')
+
+    @classmethod
+    def from_data(cls, data):
+        """
+        Create Address from a data dictionary.
+
+        :param data: data for address
+        :type data: dict[str,str]
+        :return: immutable address is saved address otherwise unsaved address object is returned
```
suutari-ai: Wrap the docstring, something around 72 columns is fine.

shoop/core/models/addresses.py
```diff
@@ -173,6 +136,46 @@ def as_string_list(self, locale=None):
...
+        :return: Fresh unsaved MutableAddress with same data as self.
+        """
+        data = get_data_dict(self)
+        return MutableAddress.from_data(data)
+
+
+class MutableAddress(Address):
+    """
+    Address that can be changed. Used for `Contact` addresses and for `SavedAddress`.
+    `MutableAddresses` could be created with self.from_data.
```
suutari-ai: Split the docstring to title and body:
```python
"""
Address that can be changed.

Used e.g. for addresses of `Contact` and for `SavedAddress`.
`MutableAddress` can be created with `MutableAddress.from_data`
or with the `new_mutable` method of `Address` objects.
"""
```

shoop/core/models/addresses.py
```diff
@@ -173,6 +136,46 @@ def as_string_list(self, locale=None):
...
+    """
+    pass
+
+
+class ImmutableAddress(ChangeProtected, Address):
+    """
+    Address that cannot be changed. Used for `Order` addresses.
+    `ImmutableAddresses` could be created with self.from_data or by
+    calling self.immutable for existing `MutableAddress`.
+    """
```
suutari-ai: * Split to title and body.
* Make sure that all references, that are backticked, are actually resolvable as a link by Sphinx, i.e. ``` `ImmutableAddresses` ``` does not work, because there is no class with that name, use ``` `ImmutableAddress` ``` instead.
* Better to just say ``` `ImmutableAddress.from_data` ``` than `self.from_data` to make it clear that it is a class method.
* Also ``` self.immutable for existing `MutableAddress` ``` -> ``` `immutable` method of existing `Address` ```

You can try generating the Sphinx documentation to see how these affect the generated docs. Though the warnings yielded from this will probably be hidden in between other warnings until I get the doc clean-up PR in.

shoop/testing/factories.py
```diff
@@ -206,9 +206,11 @@ def post(self, create, extracted, **kwargs):
             sp.categories = shop.categories.all()
 
 
-def get_address(**overrides):
+def get_address(immutable=False, **overrides):
     data = dict(DEFAULT_ADDRESS_DATA, **overrides)
-    return Address(**data)
+    if immutable:
+        return ImmutableAddress.objects.create(**data)
+    return MutableAddress.objects.create(**data)
```
suutari-ai: Shouldn't this use `Address.from_data(dict(data, immutable=immutable))`?

suutari-ai: no wait.. hmm.. that would return abstract `Address` object, if not immutable and `MutableAddress.from_data(data, immutable=True)` does not seem right... we should rethink the API a bit...

tulimaki: I think that cant be Address.from_data but yes it could use MutableAddress.from_data and ImmutableAddress.from_data.

suutari-ai: Now that I think about this. It's probably best to just drop the `Address.from_data` and add it as two separate class methods: `MutableAddress.from_data` and `ImmutableAddress.from_data`. That would make it less confusing.

```python
class ImmutableAddress(...):
    @classmethod
    def from_data(cls, data):
        """
        Create immutable address from a data dictionary.

        :param data: data for address
        :type data: dict[str,str]
        :return: Saved immutable address
        :rtype: Address
        """
        # Populate all known address fields even if not originally in `data`
        data_with_all_fields = get_data_dict(cls(**data))
        address = cls.objects.filter(**data_with_all_fields).first()
        return address if address else cls.objects.create(**data_with_all_fields)


class MutableAddress(...):
    @classmethod
    def from_data(cls, data):
        """
        Construct mutable address from a data dictionary.

        :param data: data for address
        :type data: dict[str,str]
        :return: Unsaved mutable address
        :rtype: MutableAddress
        """
        return cls(**data)
```
Then this `get_address` would use those...

Would this be clearer in your opinion?

shoop/testing/factories.py
```diff
@@ -449,8 +451,8 @@ def create_empty_order(prices_include_tax=False, shop=None):
         shop=(shop or get_shop(prices_include_tax=prices_include_tax)),
         payment_method=get_default_payment_method(),
         shipping_method=get_default_shipping_method(),
-        billing_address=get_address(name="Mony Doge"),
-        shipping_address=get_address(name="Shippy Doge"),
+        billing_address=get_address(immutable=True, name="Mony Doge"),
+        shipping_address=get_address(immutable=True, name="Shippy Doge"),
```
suutari-ai: actually, this could as easily be `get_address(name="Shippy Doge").immutable()`. So why bother with the immutable argument to get_address? Rather just make it always return MutableAddress.

shoop_tests/core/test_addresses.py
```diff
@@ -92,19 +85,54 @@ def test_address_ownership(admin_user):
...
+
+@pytest.mark.django_db
+def test_new_mutable_address():
+    address = get_address()
+    new_mutable = address.new_mutable()
+
+    # New address should be unsaved
+    assert new_mutable.pk == None
+    assert isinstance(new_mutable, MutableAddress)
+    assert get_data_dict(address).items() == get_data_dict(new_mutable).items()
```
suutari-ai: This test does not need db, I think.

shoop_tests/core/test_basic_order.py
```diff
@@ -17,8 +18,8 @@
 
 
 def create_order(request, creator, customer, product):
-    billing_address = get_address()
-    shipping_address = get_address(name="Shippy Doge")
+    billing_address = get_address(immutable=True)
+    shipping_address = get_address(immutable=True, name="Shippy Doge")
```
suutari-ai: Same here: `get_address().immutable()` and `get_address(name="Shippy Doge").immutable()` would do

shoop/admin/modules/orders/views/create.py
```diff
@@ -207,45 +207,45 @@ def handle_customer_data(self, request):
...
+
+
+def _handle_or_return_error(func, request, error_message):
+    try:
+        return func(request)
+    except Exception as exc:
+        if isinstance(exc, ValidationError):
+            error_message += "\n" + "\n".join(force_text(err) for err in exc.messages)
+        else:
+            error_message += " " + str(exc)
```
suutari-ai: `" " + str(exc)` should be `" {}".format(exc)` so it won't be converted to str and back to unicode in Py2 (possibly causing UnicodeEncodeError if error message has non-ASCII characters). (This was broken in the previous version too.

shoop/testing/factories.py
akx: I really think it should be `.as_immutable()`, btw.

shoop/core/models/addresses.py
```diff
@@ -173,6 +136,46 @@ def as_string_list(self, locale=None):
     def __iter__(self):
         return iter(self.as_string_list())
 
+    def immutable(self):
```
akx: I think this should be `get_immutable()`, or `as_immutable()`, or similar, so that methods remain verby (though I'm aware "as" is not a verb but a preposition here).

shoop/core/models/addresses.py
```diff
@@ -173,6 +136,46 @@ def as_string_list(self, locale=None):
...
+        """
+        Get or create saved ImmutableAddress from self.
+
+        :rtype: ImmutableAddress
+        :return: Saved ImmutableAddress with same data as self.
+        """
+        data = get_data_dict(self)
+        return ImmutableAddress.from_data(data)
+
+    def new_mutable(self):
```
akx: Similarly this feels like it should be `as_new_mutable`. Or `to_mutable`?

shoop/core/models/addresses.py
akx: Maybe explain the concept a little more:

> An address that can be changed (by an user, or an external process or similar).
>
> Mutable addresses are used for e.g. contacts' saved addresses.
> They are saved as new immutable addresses when used in e.g. orders.
>
> Mutable addresses can be created with `MutableAddress.from_data`
> or with the `new_mutable` method of `Address` objects.

shoop/core/models/addresses.py
akx: Again, a more verbose explanation of the semantics would be nice.

> An address that can not be changed.
>
> Immutable addresses are used for orders, etc., where subsequent
> edits to the original address (for example an user's default address)
> must not affect past business data.
>
> Immutable addresses can be created directly, with the `.from_data()`
> method, or by creating an immutable copy of an existing `MutableAddress`
> with the `Address.as_immutable()` method.

shoop_tests/core/test_categories.py
```diff
@@ -68,4 +68,4 @@ def test_category_wont_be_deleted():
     category.save()
     img.delete()
 
-    Category.objects.get(pk=1)
+    Category.objects.first()
```
akx: This change actually makes the test a no-op. Where `.get()` would have raised an ODNE when the object does not exist, this quietly returns None.

Change this to `Category.objects.get(pk=category.pk)` instead.

shoop_tests/core/test_shops.py
```diff
@@ -20,4 +20,4 @@ def test_shop_wont_be_deleted():
     shop.save()
     img.delete()
 
-    Shop.objects.get(pk=1)
+    Shop.objects.first()
```
akx: This change actually makes the test a no-op. Where `.get()` would have raised an ODNE when the object does not exist, this quietly returns None.

Change this to `Shop.objects.get(pk=shop.pk)` instead.

shoop/core/models/addresses.py
tulimaki: I will change this to to_immutable

shoop/core/models/addresses.py
tulimaki: I will change this to to_mutable

shoop/testing/factories.py
tulimaki: Yes, this is better. I implement this.

shoop/core/models/addresses.py
```diff
@@ -173,6 +117,82 @@ def as_string_list(self, locale=None):
...
+    """
+    @classmethod
+    def from_data(cls, data):
+        """
+        Create immutable address from a data dictionary.
+
+        :param data: data for address
+        :type data: dict[str,str]
+        :return: Saved immutable address
+        :rtype: Address
```
akx: The rtype is always ImmutableAddress

shoop/core/models/addresses.py
```diff
@@ -173,6 +117,82 @@ def as_string_list(self, locale=None):
...
+    must not affect past business data.
+
+    Immutable addresses can be created directly, with the `.from_data()`
+    method, or by creating an immutable copy of an existing `MutableAddress`
+    with the `Address.to_immutable()` method.
+    """
+    @classmethod
+    def from_data(cls, data):
+        """
+        Create immutable address from a data dictionary.
```
akx: This should have a mention of the fact that this might not create a new address.

suutari-ai: Yep, "create" is a not a good choice of a verb for this (my bad). Maybe `Get or create immutable address with given data.` would be better?

ChangeLog.rst
```diff
@@ -9,6 +9,8 @@ Unreleased
   releasing next version, the "Unreleased" header will be replaced with
   appropriate version header and this help text will be removed.
 
+- Deprecate AddressManager
```
suutari-ai: It's removed, not deprecated.

shoop/core/migrations/0013_address_changes.py
```diff
@@ -0,0 +1,118 @@
...
+    billing_address and shipping_address.
+    """
+    order_cls = apps.get_model("shoop", "Order")
+    immutable_address_cls = apps.get_model("shoop", "ImmutableAddress")
+
+    for order in order_cls.objects.all():
+        order.billing_address = immutable_address_cls.objects.get_or_create(
+            **shoop.utils.models.get_data_dict(order.mutable_billing_address))[0]
+        order.shipping_address = immutable_address_cls.objects.get_or_create(
+            **shoop.utils.models.get_data_dict(order.mutable_shipping_address))[0]
```
suutari-ai: Should check for None, since both fields are NULLable.

suutari-ai:
Nice work @tulimaki!

Thanks a lot!

PR #295, opened 2015-12-16
Wrap imports at column 79

suutari-ai:
I don't want to force our code base being PEP-8 incompatible.  Therefore
the forced length 120 for import lines is a bad thing.  Rather use 79
that complies with PEP-8.

I see no reason why lengthy import lines would be better than wrapped ones.


akx:
I don't see why import lines should be short when the rest of the code has a more relaxed line length limit.

akx:
Also, I honestly think the 91-character-wide original

```python
from shoop.admin.base import AdminModule, MenuEntry, SearchResult
from shoop.admin.utils.urls import derive_model_url, get_edit_and_list_urls, get_model_url
from shoop.core.models import Category
```

is more readable than the 79-character-wrapped alternative

```python
from shoop.admin.base import AdminModule, MenuEntry, SearchResult
from shoop.admin.utils.urls import (
    derive_model_url, get_edit_and_list_urls, get_model_url
)
from shoop.core.models import Category
```

suutari-ai:
They're equally readable to me, but the wider looks ugly when rest of the code (and especially docstrings) are somewhere about 50-79 columns wide.

I would prefer to not force this, i.e. let the code writer decide how to wrap the imports, but if we must force something, I think 79 is the way to go. Can isort be configured to not demand wrapping when ran with `-c`?

Ps. even more readable it would be this way:
```python
from shoop.admin.utils import urls
```

akx:
I didn't see anything wrong with the "120 max length, then wrap to 79" isort configuration that was in effect prior to this PR.

Besides, in practically any source file, the imports are just a single short section at the top of the file, and don't really compare to the width of the rest of the code imho.

Ps. 
> Ps. even more readable it would be this way:
> `from shoop.admin.utils import urls`
Then each invocation becomes uglier and more verbose. :crying_cat_face:

suutari-ai:
The problem with that was this.

I've written some code with perfectly fine imports like this:
```python
from foo import (
    bar, baz, bang, dang, ding, dippadappa, dippadappadoppa,
    dippadappadoppadippa, zappabappa, zippadappa,
)
```
but isort demands me to put all the imports to single line like this:
```python
from foo import bar, baz, bang, dang, ding, dippadappa, dippadappadoppa, dippadappadoppadippa, zappabappa, zippadappa
```
which is less readable and not PEP-8 compliant. I.e. if I don't rewrite the imports like that, isort will complain about them.

akx:
Aside: I changed Coveralls's settings to not fail the build unless coverage drops more than 5%.

akx:
I can, after looking at the data (see below), grudgingly accept wrapping imports at 79 chars.

However, please change back to the hanging closing parenthesis style -- after all you used it in your own comment, above :D 

image

Pikkupomo:
Super!

PR #297, opened 2015-12-30
Fix last shoop migration

jaywink:
The latest migration regarding renaming `Address` to `MutableAddress`, while applied correctly, caused `python manage.py migrate/makemigrations` to crash with:

    Traceback (most recent call last):
      File "/home/jaywink/.virtualenvs/kakara/lib/python3.4/site-packages/django/apps/config.py", line 159, in get_model
        return self.models[model_name.lower()]
    KeyError: 'mutableaddress'

Full paste: http://pastebin.com/iVXAZ0F7

Did a slight reordering and things seem to work. I haven't tried applying this migration to an old db though, since this migration doesn't want to reverse itself at all.

### DIFF
--- a/shoop/core/migrations/0013_address_changes.py
+++ b/shoop/core/migrations/0013_address_changes.py
@@ -101,6 +101,10 @@ class Migration(migrations.Migration):
             model_name='order',
             name='mutable_shipping_address',
         ),
+        migrations.RenameModel(
+            old_name='Address',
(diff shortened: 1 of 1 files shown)


jaywink:
Yeah of course the migration doesn't apply like this.

I'll close this and just fix my own fork manually, but it would probably make sense to figure out some proper fix for this before a release :)

PR #301, opened 2016-01-05
Make Xtheme plugins translatable (SHOOP-1941)

akx:


akx:
Assigned to @Pikkupomo because he was already reviewing this pre-rebase.

PR #303, opened 2016-01-07
Admin order creator: Don't encode unnecessary method extras

Pikkupomo:
Refs SHOOP-1992


shoop_tests/admin/test_order_creator.py
```diff
@@ -202,14 +201,6 @@ def test_order_creator_source_data(rf, admin_user):
     assert len(data.get("orderLines")) == 5
 
 
-@pytest.mark.django_db
-def test_encode_method_weights():
```
akx: Ah, hmm. Maybe it would have been better to rename this to `test_encode_method` instead of wholesale deleting it (my bad), because now we've lost test coverage over `encode_method`.

PR #306, opened 2016-01-11
Core: Fix migration 0013_address_changes (SHOOP-2003)

tulimaki:
* Do RenameModel first, which handles all foreign key changes to database.
* FakeAlterField applies only state forwards.

### DIFF
--- a/shoop/core/migrations/0013_address_changes.py
+++ b/shoop/core/migrations/0013_address_changes.py
@@ -33,6 +33,11 @@ def copy_order_addresses(apps, schema_editor):
             order.save()
 
 
+class FakeAlterField(migrations.AlterField):
+    def database_forwards(self, app_label, schema_editor, from_state, to_state):
+        pass
+
+
 class Migration(migrations.Migration):
 
     dependencies = [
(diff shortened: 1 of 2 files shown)


akx:
What happens if you try to rewind this migration? Considering FakeAlterField still does _something_ when moving backwards.

Also, it would be nice if the commit message explained what is being fixed, and why.

tulimaki:
Migrating backwards is not supported for this migration since copy order addresses goes only forward. And I think we want it to be that way.

akx:
Fair enough.

Still, the point about the commit message not explaining what is fixed and why a fix is needed stands.

tulimaki:
Updated commit message.

akx:
Hey, actually: can you add a test that attempts to run `makemigrations` (in some dry-run mode?) that ensures

* that makemigrations does not crash
* that makemigrations does not attempt to create new migrations (as they should have been done by the dev in the PR)

suutari-ai:
Is this tested with at least MySQL and SQLite now?

tulimaki:
Yes tested with both sqlite and mysql. Also added test that makemigration does not provide any new migrations and it goes through successfully.

Issue #298, opened 2015-12-30
Mithril.js 0.2.1 breaks admin UI


jaywink:
Editing views break with mithril.js 0.2.1 (with `"mithriljs": "~0.2.0",`). These are possibly fixed by 0.2.2-rc.1, according to changelog, but the way Shoop locks dependencies doesn't pull that in.

Locked my fork dep to 0.2.0 and admin edit views work again.

Question: why are Python dependencies locked to patch level but JavaScript not? Not locking dependencies causes each new environment to get a different version of the libraries, which can cause havoc with package maintainers breaking things at patch level. Could Shoop also lock javascript dependencies to patch level for increased stability?

http://nvie.com/posts/pin-your-packages/

akx:
The JavaScript ecosystem is (predominantly -- well, practically fully) Semver based, so the `"mithriljs": "~0.2.0"` dependency spec we have should only pull in versions that are compatible (as the 0.0.x level is patches only).

The Mithril 0.2.1 release was a human mishap in that sense, as it had, as I understand, multiple regressions which caused the trouble you're (or were) facing.

However, you're right in that we _should_ lock down JS packages whose version is < 1.0, as the Semver spec also specifies that within the 0.x.x series, any changes are allowed.

suutari-ai:
How does the admin UI break with Mithril 0.2.1?

suutari-ai:
Created an issue to Shoop backlog for this: SHOOP-2029

jaywink:
Sorry, could have specified or taken a screenshot, but I had already rolled back at this point. Basically for example creating a new product, the whole sidebar fails to exist. Select widgets etc don't work. In the console there were some errors about undefined stuff. Sorry I have no more exact errors, best to try it.

suutari-ai:
@jaywink Thanks. We also found out that product selector, image selector etc. don't work.

akx:
This is fixed by #307, so closing. Thanks again!

PR #307, opened 2016-01-13
Lock down JavaScript dependencies (SHOOP-2029)

suutari-ai:
Lock down JavaScript packages (NPM and Bower) to exact versions.  NPM
packages are locked with shrinkwraps and Bower packages by making the
version specifiers more strict in the bower.json files.

Especially Mithril is now locked to 0.2.0, since 0.2.1 does not work.
See issue #298 for details.


akx:
Also, btw, I just noticed `npm-shrinkwrap`'s docs say

> If you wish to lock down the specific bytes included in a package, for example to have 100% confidence in being able to reproduce a deployment or build, then you ought to check your dependencies into source control, or pursue some other mechanism that can verify contents rather than versions.

... Maybe it'd be okay just to pin the top-level versions and not add shrinkwraps, since even so they're not guaranteed? Would also lead to a patch smaller than 26,000 lines.

suutari-ai:
I still like more the shrinkwrap. Having the dependencies in source control would be cool too, but that would have to be another Git repo to keep Shoop's repo clean. That would also need some tooling that we have no time to write right now. Shrinkwrap works nicely out of the box and does almost what we want.

tulimaki:
Let's do this!

PR #308, opened 2016-01-13
Update Python package dependencies (SHOOP-1954)

suutari-ai:
Allow newer dependencies to be installed by not pinning to exact
versions in setup.py.  Note: A Shoop project can still use
requirements.txt to pin exact versions, we're just not forcing our
versions for them anymore.

Includes also couple related changes:
  * Unvendor django-parler-rest.
  * Do not install shoop_tests anymore (since that's not required by
    newer Tox.)
  * Some code changes needed for supporting newer versions of some
    Python packages.


suutari-ai:
PIP used for Shoop's Jenkins build at Spartacus was outdated. Updated it.. let's try again...

suutari-ai:
retest this please

suutari-ai:
The problem with Jenkins turned out to be too old setuptools (2.2). The used `~=` version specifiers need at least 8.0. Added setuptools requirement to tox.ini so that Jenkins will upgrade the setuptools if it's too old.

suutari-ai:
retest this please

PR #311, opened 2016-01-14
Fix coverage reporting of Tox and do doctests

suutari-ai:
Coverage reporting for Tox was broken on commit d7a4a6a97bf (which was
made, because there was some version of Tox that did not support
envsitepackagesdir variable).  Fix it.

Also enable doctests for Tox and Travis and fix some incorrectly
formatted doctests.

Refs SHOOP-1954

### DIFF
--- a/shoop/utils/iterables.py
+++ b/shoop/utils/iterables.py
@@ -29,11 +29,8 @@ def batch(iterable, count):
     """
     Yield batches of `count` items from the given iterable.
 
-    >>> for x in batch([1, 2, 3, 4, 5, 6, 7], 3):
-    >>>   print(x)
-    [1, 2, 3]
-    [4, 5, 6]
-    [7]
+    >>> tuple(x for x in batch([1, 2, 3, 4, 5, 6, 7], 3))
+    ([1, 2, 3], [4, 5, 6], [7])
 
     :param iterable: An iterable
     :type iterable: Iterable
--- a/shoop/utils/text.py
+++ b/shoop/utils/text.py
@@ -5,6 +5,8 @@
 #
 # This source code is licensed under the AGPLv3 license found in the
 # LICENSE file in the root directory of this source tree.
+from __future__ import unicode_literals
+
 import re
 import unicodedata
 
@@ -28,7 +30,7 @@ def flatten(str, whitespace="-"):
     Flatten the given text into lowercase ASCII, removing diacriticals etc.
     Replace runs of whitespace with the given whitespace replacement.
 
-    >>> print(flatten(u"hellö, wörld"))
+    >>> print(flatten("hellö, wörld"))
     hello,-world
 
     :param str: The string to massage
--- a/shoop/utils/text.py
+++ b/shoop/utils/text.py
@@ -28,8 +28,8 @@ def flatten(str, whitespace="-"):
     Flatten the given text into lowercase ASCII, removing diacriticals etc.
     Replace runs of whitespace with the given whitespace replacement.
 
-    >>> flatten(u"hellö, wörld")
-    "hello,-world"
+    >>> print(flatten(u"hellö, wörld"))
+    hello,-world
 
     :param str: The string to massage
     :type str: str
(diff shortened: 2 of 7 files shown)


shoop/utils/text.py
```diff
@@ -28,8 +30,8 @@ def flatten(str, whitespace="-"):
     Flatten the given text into lowercase ASCII, removing diacriticals etc.
     Replace runs of whitespace with the given whitespace replacement.
 
-    >>> flatten(u"hellö, wörld")
-    "hello,-world"
+    >>> print(flatten("hellö, wörld"))
```
akx: No need to use `print` in doctests.

Just that the retval should be single-quoted (as is `repr`'s preference).

shoop/utils/iterables.py
```diff
@@ -29,11 +29,8 @@ def batch(iterable, count):
     """
     Yield batches of `count` items from the given iterable.
 
-    >>> for x in batch([1, 2, 3, 4, 5, 6, 7], 3):
-    >>>   print(x)
-    [1, 2, 3]
-    [4, 5, 6]
-    [7]
+    >>> tuple(x for x in batch([1, 2, 3, 4, 5, 6, 7], 3))
```
akx: What was wrong with this doctest?

akx:
@suutari-ai: you might want to enable [these](https://github.com/python-babel/babel/blob/master/setup.cfg#L6) doctest flags :)

shoop/utils/text.py
suutari-ai: But the repr will be `u'hello,-world'` for Py2 and `'hello,-world'` for Py3. That's why I added the print.

shoop/utils/iterables.py
suutari-ai: I don't really know exactly. `flake8 --doctest` yielded an error about it.

Now that I added also `--show-source` to flake8 invocation I got more info:
```
(shoop-py3)tuomas@laatta:~/projects/shoop/project (tox-coverage-and-doctests) 0 $ flake8 --doctest --show-source 
./shoop/utils/iterables.py:32:51: F999 syntax error in doctest
    >>> for x in batch([1, 2, 3, 4, 5, 6, 7], 3):
                                                 
^
./shoop/utils/iterables.py:33:11: F999 syntax error in doctest
    >>>   print(x)
          ^
```

suutari-ai:
> @suutari-ai: you might want to enable these doctest flags :)

Most of the are probably good idea for us too, but since they are not yet needed and make the doctesting less strict I'd rather not add them yet.  I want to know what they mean before setting them. So far pytest defaults have been sensible and IMHO it's better to trust the defaults when I don't know better.

shoop/utils/iterables.py
suutari-ai: OK, now I got it... My fix was a bit overkill. It could have been:
```python
    >>> for x in batch([1, 2, 3, 4, 5, 6, 7], 3):
    ...     print(x)
```

i.e. the continuation lines should be starting with `...` (also indendation should be 4, but that's not a syntax error).

shoop/utils/text.py
akx: See my comment about Py.test's unicode compatibility doctest flag.

akx:
@suutari-ai well, setting the `ALLOW_UNICODE` flag would have obviated the need for f924f5b32833bda05f2b4cdb8b919333d6101ab8, you know...

suutari-ai:
@akx OK then... but imho it's better to have the docstrings as unicode objects on Py2 if they contain non-ASCII characters, therefore I prefer unicode_literals for that kind of cases rather than ALLOW_UNICODE flag for doctests.

akx:
Sure, `unicode_literals` should always be in use.

No matter what though, Python 2 (and hence the doctests on Python 2) will _output_ string literals embellished with that `u`.

That `ALLOW_UNICODE` flag provided by py.test disregards variations in those, allowing the same test to pass.

suutari-ai:
I take it, that by "output" you mean repr of unicode literals. Yep, exactly. That was the reason for using the print for flatten doctest (in c8e9556b3), but f924f5b is still a good idea.

Still gotta check how ALLOW_UNICODE works with binary data, i.e. bytes literals, on Py2/Py3. Let's get back to this later.

PR #315, opened 2016-01-19
Language changer for shop admin

shawnadelic:
Add language changer to shop admin so current language can be set/changed
before activating frontend theme.

Refs SHOOP-1943

Screenshots:

Desktop

Mobile

### DIFF
--- a/shoop/admin/static_src/base/js/language-changer.js
+++ b/shoop/admin/static_src/base/js/language-changer.js
@@ -1,0 +1,26 @@
+/**
+ * This file is part of Shoop.
+ *
+ * Copyright (c) 2012-2016, Shoop Ltd. All rights reserved.
+ *
+ * This source code is licensed under the AGPLv3 license found in the
+ * LICENSE file in the root directory of this source tree.
+ */
+function changeLanguage() {
+    const form = document.createElement("form");
+    form.method = "POST";
+    form.action = window.ShoopAdminConfig.browserUrls.setLanguage;
+    const input = document.createElement("input");
+    input.type = "hidden";
+    input.name = "language";
+    input.id = "language-field";
+    input.value = $(this).data("value");
+    form.appendChild(input);
+    document.body.appendChild(form);
+    form.submit();
+}
+
+$(function(){
+    "use strict";
+    $(".languages li a").click(changeLanguage);
+});
--- a/shoop/admin/templates/shoop/admin/base/_navigation.jinja
+++ b/shoop/admin/templates/shoop/admin/base/_navigation.jinja
@@ -35,14 +35,22 @@
 {% if front_url %}<a class="nav-btn shop-btn" href="{{ front_url }}" data-toggle="tooltip" data-placement="bottom" title="{% trans %}Visit shop{% endtrans %}"><i class="fa fa-shopping-cart"></i></a>{% endif %}
 
 <div class="dropdown">
-    <button class="dropdown-toggle" type="button" id="dropdownMenu" data-toggle="dropdown" aria-expanded="true" data-placement="bottom" title="{% trans %}User{% endtrans %}">
-        <i class="fa fa-user"></i>
+    <button class="dropdown-toggle" type="button" id="dropdownMenu" data-toggle="dropdown" aria-expanded="true" data-placement="bottom" title="{% trans %}Settings{% endtrans %}">
+        <i class="fa fa-cog"></i>
     </button>
     <div class="dropdown-menu" role="menu" aria-labelledby="dropdownMenu">
         {% if request.user.is_authenticated() %}
         <div class="username">
             <span class="icon"><i class="fa fa-user"></i></span>{% trans %}Hello{% endtrans %}, <span class="user">{{ request.user }}</span>!
         </div>
+        {% if LANGUAGES|length > 1 %}
+        <ul class="languages">
+            {% for code, name in LANGUAGES %}
+            <li><a href="#" class="language{% if LANGUAGE_CODE|lower == code|lower %} active{% endif %}" data-value="{{ code }}">{{ _("%(language)s",language=name)}}</a></li>
+            {% endfor %}
+        </ul>
+        {% endif %}
+        <div class="divider"></div>
         <div class="actions">
             <a href="{{ url("shoop_admin:logout") }}" class="btn btn-block btn-sm btn-info">
                 <i class="fa fa-sign-out"></i> {% trans %}Log out{% endtrans %}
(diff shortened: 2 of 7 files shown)


shawnadelic:
test this please

shawnadelic:
I just realized that someone might want to make sure this looks okay on smaller mobile device widths. I was testing out some other settings in my browser and it might throw off some of the icon spacing slightly on smaller widths (such as the iPhone 4).

juhakujala:
I would do this a bit differently. In my opinion this doesn't fit to the style of the navigation. Also there is a problem with mobile view: It should fit to 320px wide screens.

The reason why it doesn't fit is that there's no text or arrows in other navigation buttons. My suggestion would be that we use `fa-language` icon instead of the globe, text and arrow down. Then it should have a similar `data-toggle="tooltip"` than the rest of navigation elements. Current language could be highlighted in the dropdown menu with `@brand-primary` coloring.

But originally what I thought, was that the language between the shop and admin should be different. There should be an option to select the language for admin interface and the shop should be defined it's own languages. And by doing it this way, the language could be bound to the user and the language selector could be found from user settings in admin...

My option for changing the looks only would save some space and unify the looks but still doesn't fix the issue with smaller mobile devices...

akx:
@juhakujala: what if the languages lived in the user menu dropdown?

juhakujala:
@akx Thought about that too, but then it could look quite weird to just list all of the language options to user dropdown. And there can't be another dropdown/select inside a dropdown...

akx:
Would it though?

* :information_desk_person: My Profile
* :door: Log Out
* ===================
* :us: Switch to English
* :fish: Switch to Swedish
* :snowflake: Switch to Finnish
* :cn: Switch to Chinese (Simplified)

suutari-ai:
I love the fish and the snow flake for Swedish and Finnish :D

Seriously, IMHO @akx's idea is good.

akx:
@suutari-ai Github doesn't have the full `:flag-XX:` set of emoji, so I had to improvise. ;)

juhakujala:
I don't think that anyone would ever find the language selection from there... I'd say that it's not a good place to go with these. The account icon should then be changed to be something else. Something that describes all the possibilities under it...

akx:
An user will immediately notice language switching is there when they go to log out, though.

shoop/admin/static_src/base/js/language-changer.js
```diff
@@ -0,0 +1,26 @@
...
+ *
+ * Copyright (c) 2012-2016, Shoop Ltd. All rights reserved.
+ *
+ * This source code is licensed under the AGPLv3 license found in the
+ * LICENSE file in the root directory of this source tree.
+ */
+function changeLanguage() {
+    const form = document.createElement("form");
+    form.method = "POST";
+    form.action = "/set-language/";
```
RauliL: Seems like this hardcoded URL requires that Shoop front is being used?

juhakujala:
@akx but the point is that the user needs to set the language way before attempting to log out...

shoop/admin/static_src/base/js/language-changer.js
akx: Good catch. @shawnadelic: Add an endpoint for language changing in the admin too :)

akx:
@juhakujala Well, making the user menu's icon a cog/gear would fix that, wouldn't it?

akx:
Oh, also, adding a language dropdown box to the admin login screen might also be a good idea!
I think Magento does something like that.

juhakujala:
@akx Well maybe it fixes the problem right now but I think it's going to be changed in the future when we get a good solution for all these.
And a discrete dropdown in login screen could also help, but I'm not sure if it's part of this issue.

shoop/admin/static_src/base/js/language-changer.js
shawnadelic: Will do. :)

shawnadelic:
Thanks for the feedback, guys!

I went ahead and followed @juhakujala's suggestions. As noted, it throws off the responsiveness for small devices (screenshots below).

I also fixed /set-language/ front url as suggested by @RauliL

desktop
mobile

shawnadelic:
test this please

shawnadelic:
Forgot to include a screenshot of the mouseover w/ the current language highlighted. :)

hover

juhakujala:
@shawnadelic I would love to see it like this, but there's that major problem in 320px view and we can't go with it :confused: I think we are forced to put them to the "user" dropdown, but let's change the icon from `fa-user` to `fa-cog`. The tooltip text in it should also be changed according to these changes.

I think you should put the languages between the "log out" and "hello, user" elements, and use `<div class="divider"></div>` after the language list. to create a small separator line

shawnadelic:
Got it, will make those changes ASAP!

shoop/admin/templates/shoop/admin/base/_navigation.jinja
```diff
@@ -34,6 +34,20 @@
...
 {% if front_url %}<a class="nav-btn shop-btn" href="{{ front_url }}" data-toggle="tooltip" data-placement="bottom" title="{% trans %}Visit shop{% endtrans %}"><i class="fa fa-shopping-cart"></i></a>{% endif %}
 
+{% if LANGUAGES|length > 1 %}
+<div class="dropdown">
+    <button class="dropdown-toggle lang-btn" type="button" id="languageMenu" data-toggle="dropdown" aria-expanded="true" data-placement="bottom" title="{% trans %}Language{% endtrans %}">
+        <i class="menu-icon fa fa-language"></i>
+    </button>
+    <ul class="dropdown-menu languages" role="menu" aria-labelledby="languageMenu">
+        {% for code, name in LANGUAGES %}
+        <li><a href="#" class="language{% if LANGUAGE_CODE|upper == code|upper %} active{% endif %}" data-value="{{ code }}" data-set-language-url="{{ url("shoop_admin:set-language") }}">{{ name }}</a></li>
```
akx: Hmm -- instead of outputting the URL for every link (seems like a bit of a waste), it would probably be better to add it in [the admin configuration JavaScript object](https://github.com/shoopio/shoop/blob/master/shoop/admin/template_helpers/shoop_admin.py#L71-L85) (which is accessible as `window.ShoopAdminConfig` from JavaScript)?

shawnadelic: I was looking for something like that, thanks!

shawnadelic:
without_switch
with_switch

Should I include the "Switch to" text? I think it works fine without it (and looks a little less cluttered), but let me know what you think.

juhakujala:
I'm cool with the one without "Switch to". I think it's pretty self descriptive.

suutari-ai:
Good work @shawnadelic! Thanks!

PR #316, opened 2016-01-20
Change Log and Release Notes for 3.0.0 (SHOOP-1960)

suutari-ai:
Also move Change Log to doc and clean up the entries of last release to
be more in line with entries in the current one.

Also, remove gather_i18n.py script, because noticed that one while
writing the change log and we don't want to introduce a script that
should not be used.  Instead document the "shoop_makemessages".


doc/changelog.rst
```diff
@@ -0,0 +1,303 @@
...
+- Footer CMS Pages field are no longer required
+- Update label for footer links to avoid confusion
+
+Default Theme
+~~~~~~~~~~~~~
+
+- Remove Default theme from Shoop Base. Moved to
+  https://github.com/shoopio/shoop-simple-theme
+
+General/miscellaneous
```
akx: Many of these changes seem to be irrelevant for actual end users.

suutari-ai: And by end users you mean who?
* (D) Developers who are using Shoop for their project?
* (A) Shoop web shop administrators (who set-up Shoop for someone)?
* (M) Shoop web shop merchants (who sell stuff with Shoop)?
* (C)  Customers buying from Shoop web shop?

The change log is intended for D and A. Release notes is for D, A and M.

suutari-ai: One group is also Addon developers, but those can be mixed with (D) here.

suutari-ai: Ah, you were referring to the items listed in "General/miscellaneous" section... understood the context incorrectly and though you meant the items above your comment (IMHO you should comment on a line below (not above) the section you're referring to so that those items you're referring to would be visible in Github near your comment).

Sure, you're right. Will clean the misc section a bit.

akx: In general I don't think every commit needs to be listed in the change log anyway.

suutari-ai: Updated the misc section.

suutari-ai: > In general I don't think every commit needs to be listed in the change log anyway.

I agree. Change Log should list the relevant changes since last version. E.g. the Order creation tool which was added now had several commits and even several merges, but it is still just one entry in the change log. What are relevant changes then? That's a tough question, since it depends on who's reading it. I wouldn't filter the change log very much, because even if it's quite verbose, it's still much easier to read than the whole git log between two versions. For highlights we can use the release notes.

Pikkupomo:
For me, this looks good. I also like the level of detail the changelog has! nice work @suutari-ai

PR #317, opened 2016-01-20
Campaign groundwork (SHOOP-2045)

Pikkupomo:
Add support for upcoming campaign module.

* Add `load_modules` method for provides because campaigns will require multiple modules to be loaded
* Add `codes` property to `OrderSource`. This property can contain any type of code that affects `OrderSource`. The upcoming campaign module will use these codes for coupons
* Publish `OrderSource.get_product_lines()` because product lines must be easily available for upcoming campaigns
* Add `product_count` property to  `OrderSource`. This property is needed by upcoming campaigns. 
* Add `minimum_price` field to `ShopProduct`. This field will be used in upcoming campaign module to set a limit how low the product price can go.
* Require `shop` and `customer` in all pricing contexts. This change was made to simplify the usage of `PricingContext`. The current implementation required all subclasses to override the `get_context_from_request` method to add `shop` and `customer` there.
* Add core support for Campaign Module:
** Add two modules `CatalogCampaignModule` and `BasketCampaignModule`. 

Few notes about the naming
The original thought behind the naming was to have `CampaignDiscountModule` and `BasketDiscountModule`. Campaigns, however, won't always be giving discounts. Consider these two campaigns: `Buy n get x free` or `buy x get 20% off the whole basket`. Both are campaigns, but definitely not discounts per se. 

This PR will enable removing `discount_pricing` and adding the core for Campaigns feature

### DIFF
--- a/shoop/apps/provides.py
+++ b/shoop/apps/provides.py
@@ -169,19 +169,50 @@ def load_module(setting_name, provide_category):
     """
     Load a module from a module setting.
 
-    The value of the setting must be a module identifier for the given provide category.
+    The value of the setting must be a module
+    identifier for the given provide category.
 
     :param setting_name: The setting name for the identifier
     :type setting_name: str
-    :param provide_category: The provide category for the identifier lookup (e.g. `tax_module`)
+    :param provide_category:
+      The provide category for the identifier lookup (e.g. ``tax_module``)
     :type provide_category: str
     :return: An object.
     :rtype: object
     """
-    setting_value = getattr(settings, setting_name, None)
-    if not setting_value:
-        raise ImproperlyConfigured("The setting `%s` MUST be set." % setting_name)
+    setting_value = _get_settings_value(setting_name)
+    return _load_module(provide_category, setting_name, setting_value)
 
+
+def load_modules(setting_name, provide_category):
+    """
+    Load a list of modules from a module setting.
+
+    The value of the setting must be a list of module
+    identifiers for the given provide category.
+
+    The modules are returned in the same order they
+    are declared in the settings.
+
+    :param setting_name: The setting name for the identifier list
+    :type setting_name: str
+    :param provide_category:
+      The provide category for the identifier lookup (e.g. ``tax_module``)
+    :type provide_category: str
+    :return: A list of objects
+    :rtype: list[object]
+    """
+    setting_value = _get_settings_value(setting_name)
+    return [_load_module(provide_category, setting_name, x) for x in setting_value]
+
+
+def _get_settings_value(setting_name, required=True):
+    if not hasattr(settings, setting_name):
+        raise ImproperlyConfigured("The setting `%s` MUST be set." % setting_name)
+    return getattr(settings, setting_name, None)
+
+
+def _load_module(provide_category, setting_name, setting_value):
     object = get_identifier_to_object_map(provide_category).get(setting_value)
     if not object:
         raise ImproperlyConfigured(
--- a/shoop/core/models/_product_shops.py
+++ b/shoop/core/models/_product_shops.py
@@ -69,6 +69,9 @@ class ShopProduct(MoneyPropped, models.Model):
     default_price = PriceProperty('default_price_value', 'shop.currency', 'shop.prices_include_tax')
     default_price_value = MoneyValueField(verbose_name=_("default price"), null=True, blank=True)
 
+    minimum_price = PriceProperty('minimum_price_value', 'shop.currency', 'shop.prices_include_tax')
+    minimum_price_value = MoneyValueField(verbose_name=_("minimum price"), null=True, blank=True)
+
     class Meta:
         unique_together = (("shop", "product",),)
 
--- a/shoop/core/order_creator/_source.py
+++ b/shoop/core/order_creator/_source.py
@@ -15,7 +15,9 @@ from shoop.core.models import (
     OrderStatus, PaymentMethod, Product, ShippingMethod, Shop, Supplier,
     TaxClass
 )
-from shoop.core.pricing import Price, Priceful, TaxfulPrice, TaxlessPrice
+from shoop.core.pricing import (
+    get_basket_campaign_modules, Price, Priceful, TaxfulPrice, TaxlessPrice
+)
 from shoop.core.taxing import TaxableItem
 from shoop.utils.decorators import non_reentrant
 from shoop.utils.money import Money
@@ -311,6 +313,14 @@ class OrderSource(object):
                 raise TaxesNotCalculated('Taxes are not calculated')
             self.calculate_taxes()
 
+    def _get_basket_campaign_lines(self, lines):
+        """
+        Get basket campaign lines from available `CampaignModule`s
+        """
+        for module in get_basket_campaign_modules():
+            for line in module.get_basket_campaign_lines(self, lines):
+                yield line
+
     def uncache(self):
         """
         Uncache processed lines.
@@ -331,6 +341,7 @@ class OrderSource(object):
 
         lines.extend(self._compute_payment_method_lines())
         lines.extend(self._compute_shipping_method_lines())
+        lines.extend(self._get_basket_campaign_lines(lines))
 
         lines.extend(_collect_lines_from_signal(
             post_compute_source_lines.send(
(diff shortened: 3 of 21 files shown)


shoop/core/order_creator/_source.py
```diff
@@ -207,6 +210,15 @@ def status(self):
     def status(self, status):
         self.status_id = (status.id if status else None)
 
+    @property
+    def codes(self):
+        return self._codes
```
akx: Missing docstring.

shoop/core/order_creator/_source.py
```diff
@@ -207,6 +210,15 @@ def status(self):
...
+
+    @codes.setter
+    def codes(self, codes):
+        self._codes = codes
```
akx: This allows the type of `_codes` to change to something arbitrary.

shoop/core/order_creator/_source.py
```diff
@@ -264,6 +285,16 @@ def calculate_taxes_or_raise(self):
                 raise TaxesNotCalculated('Taxes are not calculated')
             self.calculate_taxes()
 
+    def _get_basket_campaign_lines(self, lines):
+        """
+        Yield `CAMPAIGN` type lines from available basket discount modules.
+
+        Also process `self.codes` in said modules.
```
akx: What does this line mean?

shoop/apps/provides.py
```diff
@@ -178,10 +178,44 @@ def load_module(setting_name, provide_category):
...
+    The modules are returned in the same order they are declared in the settings.
+
+    :param setting_name: The setting name for the identifier list
+    :type setting_name: str
+    :param provide_category: The provide category for the identifier lookup (e.g. `tax_module`)
+    :type provide_category: str
+    :return: A list of objects
+    :rtype: list[object]
+    """
+    setting_value = _get_settings_value(setting_name)
```
akx: I wonder if it's okay for this to error out if the list is empty...

shoop/core/pricing/__init__.py
```diff
@@ -61,7 +71,18 @@
     "PricingModule",
     "TaxfulPrice",
     "TaxlessPrice",
-    "get_pricing_module",
 ]
 
+
+def get_price_info(product, context, quantity):
+    pricing_module = get_pricing_module()
```
akx: Public API missing docstring.

shoop/simple_pricing/module.py
```diff
@@ -25,20 +25,13 @@ class SimplePricingModule(PricingModule):
     identifier = "simple_pricing"
     name = _("Simple Pricing")
 
-    pricing_context_class = SimplePricingContext
-
-    def get_context_from_request(self, request):
-        customer = getattr(request, "customer", None)
-
-        if not customer or customer.is_anonymous:
+    def get_customer_group_ids(self, customer):
```
akx: This should probably be a private (underscore-prefixed) API, or maybe preferably it should also live outside the class itself.

akx: In addition, it looks like it could use some sort of caching, considering it would cause N extra database queries for N product prices queried.

shoop/core/order_creator/_source.py
Pikkupomo: this is going to be fixed when campaigns arrive

shoop/simple_pricing/module.py
Pikkupomo: correct, I will change this

shoop/core/migrations/0014_product_minimum_price.py
```diff
@@ -0,0 +1,20 @@
+# -*- coding: utf-8 -*-
```
tulimaki: There is actually migration 0014 already added. Remake this migration.

shoop/core/models/_product_shops.py
```diff
@@ -65,6 +65,9 @@ class ShopProduct(MoneyPropped, models.Model):
     default_price = PriceProperty('default_price_value', 'shop.currency', 'shop.prices_include_tax')
     default_price_value = MoneyValueField(verbose_name=_("Default price"), null=True, blank=True)
 
+    minimum_price = PriceProperty('minimum_price_value', 'shop.currency', 'shop.prices_include_tax')
+    minimum_price_value = MoneyValueField(verbose_name=_("Minimum price"), null=True, blank=True)
```
tulimaki: All other verbose_names is in lowercase. That seems to be preferred way in other models too. Care to change that "Default price" also if not already done in master?

tulimaki:
Please rebase this

shoop/core/order_creator/_source.py
akx: Could you remove it from this PR to decrease churn then?

shoop/apps/provides.py
Pikkupomo: I think yes, it should be. If you set the settings value without actually setting any values, you should suffer from it.

shoop/core/models/_product_shops.py
Pikkupomo: good catch

shoop/apps/provides.py
```diff
@@ -178,10 +178,44 @@ def load_module(setting_name, provide_category):
...
+
+    The modules are returned in the same order they are declared in the settings.
+
+    :param setting_name: The setting name for the identifier list
+    :type setting_name: str
+    :param provide_category: The provide category for the identifier lookup (e.g. `tax_module`)
+    :type provide_category: str
+    :return: A list of objects
+    :rtype: list[object]
+    """
```
suutari-ai: Wrap the docstring.

shoop/apps/provides.py
```diff
@@ -178,10 +178,44 @@ def load_module(setting_name, provide_category):
...
+    for value in setting_value:
+        modules.append(_load_module(provide_category, setting_name, value))
+    return modules
+
+
+def _get_settings_value(setting_name, required=True):
     setting_value = getattr(settings, setting_name, None)
-    if not setting_value:
+    if required and not setting_value:
         raise ImproperlyConfigured("The setting `%s` MUST be set." % setting_name)
```
suutari-ai: Should rather just do `if not hasattr(settings, setting_name): raise ImproperlyConfigure(...)` and then the required param is not needed for anything, so just drop it.

shoop/apps/provides.py
suutari-ai: No, empty list is a proper value for a setting that is supposed to be a list. It should not be an error.

shoop/apps/provides.py
```diff
@@ -178,10 +178,44 @@ def load_module(setting_name, provide_category):
...
+    :type setting_name: str
+    :param provide_category: The provide category for the identifier lookup (e.g. `tax_module`)
+    :type provide_category: str
+    :return: A list of objects
+    :rtype: list[object]
+    """
+    setting_value = _get_settings_value(setting_name)
+
+    if not isinstance(setting_value, list):
+        raise ImproperlyConfigured("The setting `%s` MUST be a `list`." % setting_name)
```
suutari-ai: Drop this check it's not needed and e.g. tuple is a fine value too. If the setting value is not iterable it will error out later for sure, which is as good as this, but less demanding and less code.

shoop/apps/provides.py
```diff
@@ -178,10 +178,44 @@ def load_module(setting_name, provide_category):
...
+
+    modules = []
+    for value in setting_value:
+        modules.append(_load_module(provide_category, setting_name, value))
+    return modules
```
suutari-ai: Appending to a list is IMHO usually worse option than using a list comprehension. In this case the list comprehension would be as simple as:
```python
return [_load_module(provide_category, setting_name, x) for x in value_list]
```

shoop/core/pricing/_campaigns.py
```diff
@@ -0,0 +1,56 @@
...
+    :rtype: list[CatalogCampaignModule]
+    """
+
+    return _get_campaign_modules("SHOOP_CATALOG_CAMPAIGN_MODULES", "catalog_campaign_module")
+
+
+def _get_campaign_modules(setting_name, provide_category):
+    modules = []
+    if getattr(settings, setting_name, None):
+        # `load_modules` would error out at the setting being falsy,
```
akx: This comment is no longer necessary due to the changes in `load_modules`, so change the behavior here accordingly?

suutari-ai:
Replaced with PR #325.

PR #321, opened 2016-01-20
Simple printouts with weasyprint

tulimaki:
* Enhance order admin
* Add contact address for shop
* Add printouts for orders

Refs SHOOP-1767


shoop/utils/pdf_export.py
```diff
@@ -0,0 +1,26 @@
...
+
+try:
+    import weasyprint
+except ImportError:
+    weasyprint = None
+
+
+def render_html_to_pdf(html):
+    if not weasyprint:
+        raise NotImplementedError("weasyprint is not available")
```
akx: This isn't an implementation error. Would be better to raise a proper Problem that informs the user that the required library should be installed.

shoop/core/models/_shops.py
```diff
@@ -46,6 +46,8 @@ class Shop(ChangeProtected, TranslatableShoopModel):
     prices_include_tax = models.BooleanField(default=True, verbose_name=_("prices include tax"))
     logo = FilerImageField(verbose_name=_("logo"), blank=True, null=True, on_delete=models.SET_NULL)
     maintenance_mode = models.BooleanField(verbose_name=_("maintenance mode"), default=False)
+    contact_address = models.ForeignKey(
```
akx: Just as a note, this being a mutable address means printouts will be affected "retroactively" by a shop's address changes.

I.e.

* you print something for order X, get address A
* you change your address to B
* you print something for order X again and get address B

I think this is acceptable, just something I wanted to point out :)

shoop/printouts/admin_module/toolbar.py
```diff
@@ -0,0 +1,43 @@
...
+        self.order = order
+        self.items = self.get_menu_items()
+        super(SimplePrintoutsToolbarButton, self).__init__(self.items, **kwargs)
+
+    def get_menu_items(self):
+        shipment_items = []
+        for shipment_id in self.order.shipments.values_list("pk", flat=True):
+            shipment_items.append(
+                DropdownItem(
+                    text=_("PDF delivery slip %(id)s" % {"id": shipment_id}),
```
akx: Verbs for menu items maybe? `"Get Delivery Slip (Shipment %(id)s)"`

Also, the % interpolation must happen outside of the `_()` call.

shoop/printouts/admin_module/toolbar.py
```diff
@@ -0,0 +1,43 @@
...
+                    icon="fa fa-truck",
+                    url=reverse("shoop_admin:printouts.delivery_pdf", kwargs={"shipment_pk": shipment_id})
+                )
+            )
+
+        return shipment_items + [
+            DropdownItem(
+                text=_("PDF order confirmation"),
```
akx: Same here, "Get Order Confirmation"

shoop/printouts/admin_module/views.py
```diff
@@ -0,0 +1,57 @@
...
+
+from shoop.core.models import Order, OrderLine, OrderLineType, Shipment
+from shoop.utils import pdf_export
+
+
+def get_footer_information(shop):
+    address = shop.contact_address
+    if not address:
+        return shop.name
+    return "%s | %s %s, %s, %s | %s %s" % (
```
akx: Hmm, don't we have some sort of address formatting functionality somewhere?

akx:
So what's the rationale behind making printouts a separate app?

I mean, that way lies madn... well, at the very least that way lies a single app that must know about all sorts of object types and printouts for them.

tulimaki:
It's app since we want to make it optional. Also since weasyprint is dependency someone might want to disable it and use his own implementation for printouts.

akx:
It can be optional even if it isn't an app, can't it? I just don't like the idea of a single "omnibus" app that should know how to printout everything.

Besides, the bits that use weasyprint seem to be in `shoop.utils`, not in the optional app.

shoop/core/models/_shops.py
suutari-ai: As you said, it's not a problem. And the address being immutable doesn't even help, since you can still change the contact_address field in Shop to point a different immutable address. :)

shoop/printouts/admin_module/views.py
tulimaki: Could not find any :confused:

Pikkupomo:
like @akx said,  this now uses something inside `shoop/utils` which causes weasyprint requirement for shoop. If weasyprint isn't available an error is presented to user. How about you just return `HttpResponse(html)` if weasyprint isn't available? and perhaps move the weasyprint requirement inside this app?

akx:
Weasyprint isn't a hard requirement though; if it's missing, the webapp will still start okay, it just can't print PDFs.

tulimaki:
@Pikkupomo weasyprint code is not under app since I guess we would like to offer that same util to get PDFs from HTML for other apps too. Not sure if our pdf_exporter should fallback html if weasyprint is not available. If weasyprint is not available this wont work as planned and it's good to tell that to customer.

akx:
Either way, a separate single app (`shoop.printouts`) that should handle every kind of printout in the future sounds like a bad design.

tulimaki:
@akx could order_printouts be better or would you still prefer some other solution to do this.

akx:
I think you'll have to discuss the structure with the dev team.
I don't think I'm privileged to make suggestions here anymore.

tulimaki:
Renamed printouts to order_printouts and moved pdf_export under app.

shoop/admin/modules/orders/views/shipment.py
```diff
@@ -65,6 +65,7 @@ def get_form(self, form_class):
             if info["unshipped"] == 0:
                 attrs["disabled"] = "disabled"
             field = forms.DecimalField(
+                required=bool(info["unshipped"] != 0),
```
Pikkupomo: `info["unshipped"]` is used 4 times (atleast) in preceeding lines. I would like to see it being extracted once from `info`. for example `is_unshipped = info["unshipped"]`

akx: You're starting to learn, young grasshopper

Pikkupomo: :)

akx:
I briefly discussed this with @Pikkupomo and @suutari-ai yesterday though -- I think we came up with a structure that'd have Printouts being provideable classes. I don't know if they talked to you about that,  @tulimaki..

A provide that can add buttons to the order view in the admin is fine, but I think this should go one abstraction level higher, i.e. that you could ask for printouts that are applicable for a given context, for instance "order object, admin context, this request" would yield the order and shipment printouts...

If this is a temporary "simple" solution (like `simple_order_notification` or whatever it's called is), I suppose this is fine, but it could really use some TLC for further dev.

tulimaki:
@akx this was not discussed with me. Not sure we want to do anything more for this. If I understood correctly there is more printing related stuff coming and there is a couple other issues too that needs to solved and discussed.

Pikkupomo:
@akx due the time constraints we had to limit this to order printouts only now and make a more versatile implementation of printouts later on.

tulimaki:
Modified transifex config to include order_printouts translation file. @Pikkupomo

PR #322, opened 2016-01-21
Translate language names

shawnadelic:
Add localized language names to both frontend and admin language changers.

translated-default

translated-admin

A language's localized name defaults to its Django default translation when its language name is set equal to its country code (i.e., ``("en", "en")``), however this can be overridden by specifying a localized display name, as below with 'ENGLISH':

    LANGUAGES = [
        ('en', 'ENGLISH'),
        ('fi', 'fi'),
        ('ja', 'ja'),
        ('zh-Hans', 'zh-Hans'),
    ]

translated

### DIFF
--- a/shoop/core/templatetags/shoop_common.py
+++ b/shoop/core/templatetags/shoop_common.py
@@ -12,6 +12,8 @@ from json import dumps as json_dump
 
 from babel.dates import format_date, format_datetime, format_time
 from babel.numbers import format_decimal
+from django.conf import settings
+from django.utils import translation
 from django.utils.safestring import mark_safe
 from django.utils.timezone import localtime
 from django_jinja import library
@@ -21,6 +23,24 @@ from shoop.utils.i18n import (
     format_money, format_percent, get_current_babel_locale
 )
 from shoop.utils.serialization import ExtendedJSONEncoder
+
+
+@library.global_function
+def get_language_choices():
+    """
+    Get language choices as code and text in two languages.
+
+    :return:
+      Available language codes as tuples (code, name, local_name)
+      where name is in the currently active language, and local_name
+      is in the language of the item.
+    :rtype: Iterable[tuple[str,str,str]]
+    """
+    for (code, name) in settings.LANGUAGES:
+        lang_info = translation.get_language_info(code)
+        name_in_current_lang = translation.ugettext(name)
+        local_name = lang_info["name_local"]
+        yield (code, name_in_current_lang, local_name)
 
 
 @library.filter
--- a/shoop/admin/gulp_bits/js-packages.js
+++ b/shoop/admin/gulp_bits/js-packages.js
@@ -28,7 +28,8 @@ module.exports = {
             "js/remarkable-field.js",
             "js/form-utils.js",
             "js/jquery.form-submission-attributes.polyfill.js",
-            "js/datetimepicker.js"
+            "js/datetimepicker.js",
+            "js/language-changer.js"
         ]
     },
     "dashboard": {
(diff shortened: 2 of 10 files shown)


shoop/core/templatetags/shoop_common.py
```diff
@@ -23,6 +24,14 @@
 from shoop.utils.serialization import ExtendedJSONEncoder
 
 
+@library.global_function
+def get_localized_language_name(code, name):
+    if code == name:
+        return get_language_info(code.lower())["name_local"]
```
akx: So this will always return the language's name in that language... That's one possible behavior; the other would be to return the language's name in the current locale (`get_current_babel_locale().languages.get(code.lower())`).

I'm not sure which is better here though! The current behavior lets people fluent in that language switch to that language; the other behavior would let people fluent in the current UI language to easily understand what they're going to switch to...

akx:
Seeing as this contains code that also lives in #315, I think this shouldn't be merged before that one is.

shoop/core/templatetags/shoop_common.py
suutari-ai: I think it would be good idea to include them both in the language switcher, i.e. when current language is English, the choices would be:
 * Chinese (Simplified) (中文)
 * **English (English)**
 * Finnish (suomi)
 * Japanese (日本語)

@juhakujala As an UX expert, what do you think?

akx: I don't think ordering should need to be touched, since `LANGUAGES` is ordered already.

akx: Oh, and I think @suutari-ai 's suggestion is very sensible.

shawnadelic:
@suutari-ai 
I can definitely keep it standard to Django as well, was just basing it off of my interpretation of the following suggestion (from SHOOP-1940):

* Babel provides a function to retrieve a language name in a given locale
* So the actual names in `LANGUAGES` should not matter and you should use that Babel function instead
* Or, maybe better yet try to look up the language name via the Babel mapping
* So you can do `LANGUAGES = [('en', 'en'), ...]`
* In which case the translated language name would be automatically correct, but implicitly overridable

The only question that would come up would be whether to use the default localized language name from get_language_info or do the translation manually in case they want to override it. Not sure how likely the latter is, and it might be slightly hacky since the Django translation functions automatically translate to user's language, but we can definitely do it.

akx:
@suutari-ai: changing them honestly doesn't matter. Sure, Django provides some translations for language names, but since we're already depending on Babel which also does that (and likely with more coverage and accuracy), why not use it instead?

suutari-ai:
How about doing it like this?

In settings:

```python
# Use same names for languages as in django.conf.global_settings.LANGUAGES
# or provide a translation with gettext_noop.
LANGUAGES = [
    ('en', 'English'),
    ('fi', 'Finnish'),
    ('ja', 'Japanese'),
    ('zh-hans', 'Simplified Chinese'),
]
```

In templatetags:

```python
from django.utils import translation

def get_language_choices():
    """
    Get language choices as code and text in two languages.

    :return:
      Available language codes as tuples (code, name, local_name)
      where name is in the currently active language, and local_name
      is in the language of the item.
    :rtype: Iterable[tuple[str,str,str]]
    """    
    for (code, name) in settings.LANGUAGES:
        lang_info = translation.get_language_info(language_code)
        name_in_current_lang = translation.ugettext(name)
        local_name = lang_info["name_local"]
        yield (code, name_in_current_lang, local_name)
```

In the template:

```jinja
<ul class="dropdown-menu languages" role="menu">
    {% for (code, name, local_name) in get_language_choices() %}
        <li><a href="#" class="language" data-value="{{ code }}">{{ name }} ({{ local_name }})</a></li>
    {% endfor %}
</ul>
```

This would allow different kind of rendering for other templates, e.g. using only local_name or using local_name (name).

shawnadelic:
I think that's a good solution. Will make those changes.

suutari-ai:
Thanks @shawnadelic!

PR #323, opened 2016-01-21
Pull translations from Transifex (SHOOP-1957)

suutari-ai:
Also configure the Transifex client (in .tx/config) for easily
pulling/pushing translations with the "tx" command.  See
http://docs.transifex.com/client/ for details.


akx:
@Pikkupomo I guess the rest are mapped as-is.

suutari-ai:
Only cmn:zh_Hans, because others use the same language name on Transifex (the left side) as we use in our code base (the right side).

Generating entries for .tx/config automatically is out of the scope of this ticket. And we don't introduce new apps so often that it should even matter.

PR #325, opened 2016-01-27
Core changes for campaigns (SHOOP-2045)

suutari-ai:
This PR picks most of the code from PR #317, but some terminology is
changed and tests added, documentation improved etc.

See the commit messages and change log for details.


akx:
> Rename OrderLineType CAMPAIGN to DISCOUNT

Not all CAMPAIGN lines are DISCOUNTs.

doc/provides.rst
```diff
@@ -57,6 +57,8 @@ Provide Categories
 ``admin_module``
     Admin module classes. Practically all of the functionality in the admin is built
     via admin modules.
+``discount_module``
```
akx: This should be `price_modifier_module` to be symmetric with `order_source_modifier_module`, then.

shoop/core/locale/zh_Hans/LC_MESSAGES/django.po
```diff
@@ -747,6 +747,11 @@ msgstr "主分类"
 msgid "default price"
 msgstr "默认价格"
 
+#, fuzzy
+#| msgid "minimum purchase"
+msgid "minimum price"
+msgstr "最低购买"
```
akx: Fuzzy translations should NOT be included!

shoop/core/order_creator/_source.py
```diff
@@ -207,6 +210,47 @@ def status(self):
...
+    def codes(self):
+        return list(self._codes)
+
+    def add_code(self, code):
+        """
+        Add code to this OrderSource.
+
+        At this point it is expected that the customers
+        permission to use the code has already been
+        checked by the adding instance.
```
akx: `s/adding instance/caller/`

shoop/core/order_creator/_source.py
```diff
@@ -207,6 +210,47 @@ def status(self):
...
+
+        Code will be converted to text.
```
akx: The code

shoop/core/order_creator/_source.py
```diff
@@ -207,6 +210,47 @@ def status(self):
...
+
+        :param code: Code being added
```
akx: The code to add

shoop/core/order_creator/_source.py
```diff
@@ -207,6 +210,47 @@ def status(self):
...
+        :rtype: bool
+        """
+        code_text = force_text(code)
+        if code_text not in self._codes:
+            self._codes.append(code_text)
+            self.uncache()
+            return True
+        return False
+
+    def clean_codes(self):
```
akx: This _clears_ codes, it doesn't _clean_ them.

akx: Missing docstring.

shoop/core/order_creator/_source.py
```diff
@@ -207,6 +210,47 @@ def status(self):
...
+        if self._codes:
+            self._codes = []
+            self.uncache()
+            return True
+        return False
+
+    def remove_code(self, code):
```
akx: Missing docstring.

shoop/core/order_creator/_source.py
```diff
@@ -301,7 +355,16 @@ def _compute_shipping_method_lines(self):
             for line in self.shipping_method.get_source_lines(self):
                 yield line
 
-    def _get_product_lines(self):
+    def _add_lines_from_modifiers(self, lines):
+        """
+        Add lines from OrderSourceModifiers to given list of lines.
```
akx: If OSMs aren't allowed to modify the line list, make a copy of it here.

akx: (That still won't make it impossible for them to stick their grubby fingers into the lines themselves, of course.)

shoop/core/pricing/_context.py
```diff
@@ -41,24 +39,14 @@ class PricingContext(PricingContextable):
...
+        """
+        Initialize pricing context for shop and customer.
+
+        :type shop: shoop.core.models.Shop
+        :type customer: shoop.core.models.Contact
+        :type time: datetime.datetime|None
+        """
+        self.shop = shop
+        self.customer = customer
+        self.time = time if time is not None else now()
```
akx: Parens around the rvalue, please.

shoop/core/pricing/_utils.py
```diff
@@ -0,0 +1,102 @@
...
+    for module in get_discount_modules():
+        steps = module.get_pricing_steps_for_products(ctx, products, steps)
+    return steps
+
+
+def _get_module_and_context(context):
+    """
+    Get current pricing module and context converted to pricing context.
+
+    :rtype: tuple[PricingModule,PricingContext]
```
akx: Missing type declaration for context.

suutari-ai:
@akx: Thanks for the comment.

Discount lines are already needed e.g. for adding discounts to orders when creating orders via the order creator.

I don't see any use case why discounts added by campaigns would need a special line type. And if campaign needs to add some other kind of line than discount or product, it can add it with the "OTHER" enum value. Do you have an use case why campaigns would need their own line type?

shoop/core/settings.py
```diff
@@ -65,6 +65,17 @@
 #: :obj:`shoop.core.pricing` for details.
 SHOOP_PRICING_MODULE = "simple_pricing"
 
+#: List of identifiers of discount modules to use.
+#:
+#: Each discount module may change the price of a product.  See
```
akx: See above -- this isn't just a discounter, it's a price modifier.

akx:
@suutari-ai: I see it as a fairly common use-case that a campaign would add a "text" line that declares a campaign being used, or perhaps a campaign that adds a discount code to the order to be used for a subsequent order?

I really think it'd be nicer to just add a DISCOUNT type if that's needed in the admin order creator than to change the semantics of an existing line type. (Especially if you consider that there might already be user projects that use CAMPAIGN.) And to expand on that line of thought, I believe the enums we have now should be considered append-only.

shoop/front/basket/objects.py
```diff
@@ -171,6 +171,33 @@ def add_line(self, **kwargs):
...
+        self.dirty = bool(self.dirty or modified)
+        return modified
+
+    def remove_code(self, code):
+        modified = super(BaseBasket, self).remove_code(code)
+        self.dirty = bool(self.dirty or modified)
+        return modified
+
+    def create_line(self, **kwargs):
+        return BasketLine(source=self, **kwargs)
```
akx: Wouldn't it be cleaner to declare `line_class` on the class level and use that in the superclass?

shoop/core/order_creator/_source.py
```diff
@@ -207,6 +210,47 @@ def status(self):
     def status(self, status):
         self.status_id = (status.id if status else None)
 
+    @property
+    def codes(self):
```
akx: Missing docstring.

I do believe this used to be a set; why change it to a list?

shoop/core/locale/zh_Hans/LC_MESSAGES/django.po
akx: I made a separate issue out of this concern. https://github.com/shoopio/shoop/issues/326

suutari-ai:
> @suutari-ai: I see it as a fairly common use-case that a campaign would add a "text" line that declares a campaign being used,

Then it's a text line, i.e. maps to OTHER. Or should we have a line type TEXT maybe?

> or perhaps a campaign that adds a discount code to the order to be used for a subsequent order?

That should be stored separately to order.codes or something like that if needed. For information purposes it can be added as a text line too, but having a distinct line type with loose semantics how the codes are stored etc. is not a good idea.

> And to expand on that line of thought, I believe the enums we have now should be considered append-only.

Yep, but this is just a change of term and shouldn't cause much harm. There should be not so many users of campaign type lines, since it is not yet possible to add them from the front. If added from the order creator, those could be for discounts or for campaign text, but having the text in a discount line is not a big deal.

akx:
> For information purposes it can be added as a text line too, but having a distinct line type with loose semantics how the codes are stored etc. is not a good idea.

I meant a line like "Use code KUKKULUURUU in your next order for a discount!"

> Yep, but this is just a change of term and shouldn't cause much harm. 

It's a semver major change which will change how existing orders that might use the line type look. I'd really just like to see CAMPAIGN remain as-is and a DISCOUNT added as a new one.

suutari-ai:
> And to expand on that line of thought, I believe the enums we have now should be considered
 > append-only.

Also... this is exactly the reason why I don't want there to be any enum values in our enums that don't have a clear use case already. Since they cause unneeded cruft that is very hard to remove. Better to add the enum values when they are actually needed somewhere and then also define the related semantics. Sorry that I didn't catch this CAMPAIGN line type when it was introduced to Shoop, should have been more careful with the review back then.

suutari-ai:
> It's a semver major change which will change how existing orders that might use the line type look.

You're right. That'll be noted in the 4.0.0 release notes. And when updating to 4.0.0, the updater may change those lines to OTHER type if they are not discounts (if there is any).

> I'd really just like to see CAMPAIGN remain as-is and a DISCOUNT added as a new one.

I don't. You didn't provide (neither I can come up with) any use that can't be coped with a text line, so therefore the CAMPAIGN type has unclear purpose which does more harm than good.

akx:
> You're right. That'll be noted in the 4.0.0 release notes. And when updating to 4.0.0, the updater may change those lines to OTHER type if they are not discounts (if there is any).

Are you serious that this is the easier and nicer way in your opinion, vs. just adding the new type?

suutari-ai:
>>   You're right. That'll be noted in the 4.0.0 release notes. And when updating to 4.0.0, the updater may change those lines to OTHER type if they are not discounts (if there is any).

> Are you serious that this is the easier and nicer way in your opinion, vs. just adding the new type?

Yes. We're heading for 4.0.0 anyway and cleaning the semantics on the way is better than leaving some magical enum values that don't have a clear use case stay there.

shoop/core/locale/zh_Hans/LC_MESSAGES/django.po
suutari-ai: Gotta tune the params of shoop_makemessages to not include them by default. We already have many of them. Luckily those won't be included to the compiled MO files anyway, so these don't really matter.

akx:
> Yes. We're heading for 4.0.0 anyway and cleaning the semantics on the way is better than leaving some magical enum values that don't have a clear use case stay there.

Sigh, well, if you're sure. And if you've talked about this change with the rest of the team.

shoop/core/order_creator/_source.py
suutari-ai: To preserve ordering of the codes.

shoop/core/pricing/_discounts.py
```diff
@@ -0,0 +1,92 @@
...
+def get_discount_modules():
+    """
+    Get a list of configured discount module instances.
+
+    :rtype: list[DiscountModule]
+    """
+    return load_module_instances("SHOOP_DISCOUNT_MODULES", "discount_module")
+
+
+class DiscountModule(six.with_metaclass(abc.ABCMeta)):
```
akx: So yeah, as implied in other comments, this should be a PriceModifierModule, and the references to `discount`ing should be `modify`ing.

shoop/core/order_creator/_source.py
akx: Why is that necessary?

shoop/core/pricing/_discounts.py
suutari-ai: Nope, they are usually used for decreasing prices i.e. positive discounts, but you can use them for increasing prices too i.e. negative discounts. Both are discounts still.

akx: I don't think there's any sense (aside from a purely logical viewpoint) in the term "negative discount".

Pikkupomo:
Naming has been discussed in the team and the name `Discount` was chosen due the fact that they are more often discounts than anything else. 

For me this looks good and works as discussed. Good work and thank you for everyone participating :)

akx:
@Pikkupomo what about the missing docstrings, or my comments about the sudden non-setness of `codes`?

PR #328, opened 2016-01-27
Stock management fixes

tulimaki:
Front: Show better error messages in basket update
Simple supplier: Make SKU-sort in stock management work
Simple supplier: Make admin filter+stock adjustment work
Simple supplier: Add djangojs.po file to transifex config

Refs SHOOP-1876

### DIFF
--- a/shoop/front/basket/update_methods.py
+++ b/shoop/front/basket/update_methods.py
@@ -62,9 +62,10 @@ class BasketUpdateMethods(object):
         for linked_line in linked_lines:
             errors = list(self._get_orderability_errors(linked_line, new_quantity))
             if errors:
-                error_texts = ", ".join(six.text_type(error) for error in errors)
-                message = u"%s: %s" % (linked_line.get("text") or linked_line.get("name"), error_texts)
-                messages.warning(self.request, message)
+                for error in errors:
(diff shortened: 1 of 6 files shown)


shoop/front/basket/update_methods.py
```diff
@@ -14,6 +14,12 @@
 from shoop.utils.numbers import parse_decimal_string
 
 
+def get_error_text(error):
```
akx: What is this function supposed to do? It lacks a docstring.

shoop/front/basket/update_methods.py
```diff
@@ -14,6 +14,12 @@
...
+    if hasattr(error, "__iter__"):
```
akx: I think this might be a little too broad. What are you trying to do here?

tulimaki: Trying to check if the error is iterable. We don't want to show errors like ["error 1", "error 2"] I understood that this isn't always the case so added the check. I will add docstring and tests for this function. Also I will rethink if this is even a good place for this function.

akx: Okay. I mean, the thing is even `strings` have `__iter__`, so if you passed in a string, you'd get `s,o,m,e,t,h,i,n,g,b,a,d` out. :D

PR #330, opened 2016-01-28
Fix basket dirty flag not always being correctly set

akx:
Noticed this while working on an internal project.

Also, since my currently installed Py3 is 3.5, had to make things work with it too.


akx:
test this please

PR #332, opened 2016-01-29
Campaigns implementation (SHOOP-494)

Pikkupomo:
Add campaign module to Shoop.

This module will replace the basic discount modules like `discount_pricing` as the same rules can be created here. Campaigns are separated into two different flows: `CatalogCampaigns` and `BasketCampaigns` both of these have their own rulesets. 

Refs SHOOP-494

### DIFF
--- a/shoop/campaigns/admin_module/views.py
+++ b/shoop/campaigns/admin_module/views.py
@@ -1,0 +1,178 @@
+# This file is part of Shoop.
+#
+# Copyright (c) 2012-2016, Shoop Ltd. All rights reserved.
+#
+# This source code is licensed under the AGPLv3 license found in the
+# LICENSE file in the root directory of this source tree.
+from babel.dates import format_datetime
+from django.utils.timezone import localtime
+from django.utils.translation import ugettext_lazy as _
+
+from shoop.admin.base import MenuEntry
+from shoop.admin.toolbar import NewActionButton, Toolbar
+from shoop.admin.utils.picotable import ChoicesFilter, Column, TextFilter
+from shoop.admin.utils.views import CreateOrUpdateView, PicotableListView
+from shoop.campaigns.forms import (
+    BasketCampaignForm, CatalogCampaignForm, CouponForm
+)
+from shoop.campaigns.models.campaigns import (
+    BasketCampaign, CatalogCampaign, Coupon
+)
+from shoop.utils.i18n import format_percent, get_current_babel_locale
+
+
+class _Breadcrumbed(object):
+    def get_breadcrumb_parents(self):
+        return [MenuEntry(text=self.parent_name, url=self.parent_url)]
+
+
+class CampaignListView(PicotableListView):
+    columns = [
+        Column(
+            "name", _(u"Title"), sort_field="translations__name", display="name", linked=True,
+            filter_config=TextFilter(operator="startswith")
+        ),
+        Column("discount_percentage", _("Discount percentage"), display="format_percentage"),
+        Column("discount_amount", _("Discount amount")),
+        Column("start_datetime", _("Starts")),
+        Column("end_datetime", _("Ends")),
+        Column("active", _("Active"), filter_config=ChoicesFilter(choices=[(0, _("No")), (1, _("Yes"))])),
+    ]
+
+    def format_percentage(self, instance, *args, **kwargs):
+        if not instance.discount_percentage:
+            return ""
+        return format_percent(instance.discount_percentage)
+
+    def start_datetime(self, instance, *args, **kwargs):
+        if not instance.start_datetime:
+            return ""
+        return self._formatted_datetime(instance.start_datetime)
+
+    def end_datetime(self, instance, *args, **kwargs):
+        if not instance.end_datetime:
+            return ""
+        return self._formatted_datetime(instance.end_datetime)
+
+    def _formatted_datetime(self, dt):
+        return format_datetime(localtime(dt), locale=get_current_babel_locale())
+
+    def add_columns(self, column_id, columns, after=True):
+        # TODO: Make better
+        added = False
+        for idx, column in enumerate(self.columns):
+            if column.id == column_id:
+                found_idx = idx + 1 if after else idx
+                start = self.columns[:found_idx]
+                end = self.columns[found_idx:]
+                self.columns = start + columns + end
+                added = True
+                break
+
+        if not added:
+            self.columns += columns
+
+    def get_object_abstract(self, instance, item):
+        return [
+            {"text": "%s" % (instance or _("CatalogCampaign")), "class": "header"},
+        ]
+
+
+class CatalogCampaignListView(CampaignListView):
+    model = CatalogCampaign
+
+    def __init__(self, **kwargs):
+        new_columns = [
+            Column("conditions", _("Used Conditions")),
+            Column("filters", _("Used filters")),
+        ]
+        self.add_columns("name", new_columns)
+        super(CatalogCampaignListView, self).__init__(**kwargs)
+
+    def get_context_data(self, **kwargs):
+        context = super(CampaignListView, self).get_context_data(**kwargs)
+        context["toolbar"] = Toolbar([
+            NewActionButton("shoop_admin:catalog_campaigns.new", text=_("Create new Catalog Campaign")),
+        ])
+        return context
+
+
+class BasketCampaignListView(CampaignListView):
+    model = BasketCampaign
+
+    def __init__(self, **kwargs):
+        new_columns = [
+            Column("conditions", _("Used Conditions")),
+            Column("coupon", _("Discount Code")),
+        ]
+        self.add_columns("name", new_columns)
+        super(BasketCampaignListView, self).__init__(**kwargs)
+
+    def get_context_data(self, **kwargs):
+        context = super(CampaignListView, self).get_context_data(**kwargs)
+        context["toolbar"] = Toolbar([
+            NewActionButton("shoop_admin:basket_campaigns.new", text=_("Create new Basket Campaign")),
+        ])
+        return context
+
+
+class CampaignEditView(CreateOrUpdateView):
+    template_name = "shoop/campaigns/admin/edit_campaigns.jinja"
+
+    context_object_name = "campaign"
+    add_form_errors_as_messages = True
+
+    def get_form_kwargs(self):
+        kwargs = super(CampaignEditView, self).get_form_kwargs()
+        kwargs["request"] = self.request
+        return kwargs
+
+
+class CatalogCampaignEditView(_Breadcrumbed, CampaignEditView):
+    model = CatalogCampaign
+    form_class = CatalogCampaignForm
+    parent_name = _("Catalog Campaign")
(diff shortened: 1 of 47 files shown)


tulimaki:
You should add new .po files to transifex config.

shoop/core/order_creator/_source.py
```diff
@@ -288,6 +288,9 @@ def product_count(self):
         """
         return sum([line.quantity for line in self.get_product_lines()])
 
+    def create_line(self, **kwargs):
+        return SourceLine(source=self, **kwargs)
```
tulimaki: Can you add some information why this is added. Maybe in the commit message.

shoop/campaigns/forms.py
```diff
@@ -0,0 +1,233 @@
...
+
+    def clean_discount_amount_value(self):
+        amount = self.cleaned_data.get("discount_amount_value", 0)
+        if amount and amount < 0:
+            raise ValidationError(_("Discount amount cannot be negative."))
+        return amount
+
+    def clean_end_datetime(self):
+        end_datetime = self.cleaned_data.get("end_datetime")
+        # TODO: allow only end dates that are after start date
```
tulimaki: Is this something we should do now or is this intentionally leaved as TODO?

shoop/core/settings.py
```diff
@@ -69,12 +69,12 @@
 #:
 #: Each discount module may change the price of a product.  See
 #: `shoop.core.pricing.DiscountModule` for details.
-SHOOP_DISCOUNT_MODULES = []
+SHOOP_DISCOUNT_MODULES = ["catalog_campaigns"]
```
tulimaki: Do we really want to use this default at core?

shoop/core/settings.py
```diff
@@ -69,12 +69,12 @@
...
 
 #: List of identifiers of order source modifier modules.
 #:
 #: See `shoop.core.order_creator.OrderSourceModifierModule` for details.
-SHOOP_ORDER_SOURCE_MODIFIER_MODULES = []
+SHOOP_ORDER_SOURCE_MODIFIER_MODULES = ["basket_campaigns"]
```
tulimaki: Do we really want to use this default at core?

shoop/themes/classic_gray/templates/classic_gray/shoop/front/basket/default_basket.jinja
```diff
@@ -41,6 +41,39 @@
...
+                        else {
+                            alert(gettext("Coupon is not available."))
+                        }
+                    },
+                    error: function() {
+                        alert(gettext("Error when adding coupon."))
+                    }
+                });
+
+               // window.location.href = window.location.href + "?code=" + coupon;
```
tulimaki: remove this if not needed

shoop_workbench/settings/base_settings.py
```diff
@@ -58,7 +59,6 @@
     'filer',
     'registration',
     'rest_framework',
-
     'shoop.discount_pricing'
```
tulimaki: Should we just remove shoop.discount_princg from installed apps since it's not recommended to use it anymore.

shoop/themes/classic_gray/templates/classic_gray/shoop/front/basket/default_basket.jinja
```diff
@@ -41,6 +41,39 @@
...
+                $.ajax({
+                    url: "",
+                    method: "POST",
+                    data: data,
+                    success: function(response) {
+                        if (response.ok) {
+                            location.reload();
+                        }
+                        else {
+                            alert(gettext("Coupon is not available."))
```
tulimaki: These should be in some djangojs.po file.

shoop/campaigns/forms.py
Pikkupomo: I think it should be fixed now

shoop_workbench/settings/base_settings.py
Pikkupomo: there is another pr inc that removes whole discount pricing

shoop/themes/classic_gray/templates/classic_gray/shoop/front/basket/default_basket.jinja
Pikkupomo: hmm ok, so our makemessages won't generate these?

shoop/campaigns/admin_module/views.py
```diff
@@ -0,0 +1,163 @@
...
+
+class CampaignListView(PicotableListView):
+    columns = [
+        Column(
+            "name", _(u"Title"), sort_field="translations__name", display="name", linked=True,
+            filter_config=TextFilter(operator="startswith")
+        ),
+        Column("discount_percentage", _("Discount percentage"), display="format_percentage"),
+        Column("discount_amount", _("Discount amount")),
+        Column("start_datetime", _("Starts")),
```
tulimaki: No formatting needed for dates?

shoop/campaigns/models/basket_conditions.py
```diff
@@ -0,0 +1,88 @@
...
+    model = None
+    active = models.BooleanField(default=True)
+    name = _("Basket condition")
+
+    def matches(self, basket, lines):
+        return False
+
+    @property
+    def description(self):
+        return _("Define the required %s for this campaign." % self.name.lower())
```
tulimaki: Translators does not have any idea what these %s contains. Use %(name)s instead?

shoop/campaigns/models/catalog_filters.py
```diff
@@ -0,0 +1,85 @@
...
+
+    active = models.BooleanField(default=True, verbose_name=_("active"))
+
+    def filter_queryset(self, queryset):
+        raise NotImplementedError("Subclasses should implement `filter_queryset`")
+
+    @property
+    def description(self):
+        extra = "(s)" if self.model else ""
+        return _("Define the required %s%s for this campaign." % (self.name.lower(), extra))
```
tulimaki: Translators does not have any idea what these %s contains. Use %(name)s and %(extra)s instead?

shoop/campaigns/models/context_conditions.py
```diff
@@ -0,0 +1,51 @@
...
+
+    active = models.BooleanField(default=True)
+
+    def matches(self, context):
+        return False
+
+    @property
+    def description(self):
+        extra = "(s)" if self.model else ""
+        return _("Define the required %s%s for this campaign." % (self.name.lower(), extra))
```
tulimaki: Use %(name)s and %(extra)s instead?

shoop/themes/classic_gray/templates/classic_gray/shoop/front/basket/default_basket.jinja
tulimaki: Hmm you shouldn't probably use gettext inside jinja file. Why not just use default trans-tags?

shoop/campaigns/apps.py
```diff
@@ -0,0 +1,38 @@
...
+        ],
+        "admin_module": [
+            "shoop.campaigns.admin_module:CampaignAdminModule",
+        ],
+        "campaign_catalog_filter": [
+            "shoop.campaigns.models.catalog_filters:ProductTypeFilter",
+            "shoop.campaigns.models.catalog_filters:ProductFilter",
+            "shoop.campaigns.models.catalog_filters:CategoryFilter"
+        ],
+        "campaign_context_condition": [
```
tulimaki: Can't find these new campaign providers in doc/provides.rst can you add these.

shoop/campaigns/models/basket_conditions.py
Pikkupomo: agreed, will change

shoop/campaigns/apps.py
Pikkupomo: These are something this addon (module, app) provides, don't think they are part of the core documentation.

shoop/core/settings.py
Pikkupomo: I think we should. We are offering this app as a base discount / campaign implementation

shoop/core/order_creator/_source.py
Pikkupomo: added this

shoop/campaigns/models/catalog_filters.py
suutari-ai: No, it's still untranslatable. Better to just overwrite this description for each of the subclasses to make this easier for translators. Then this would be just e.g. `_("Define the required product types for this campaign")` for the `ProductTypeFilter` and so on for all of them.

shoop/campaigns/models/context_conditions.py
suutari-ai: I would rather say the same (as for CatalogFilter) for this: Provide the message in total for each of the Condition subclasses.

Also: The wording is a bit misleading. Maybe something like "Limit the campaign to members of selected contact groups." would be easier to understand, since the current wording suggests that a contact group is required for *creating* a campaign (rather than it being required for using the campaign)

shoop/campaigns/apps.py
suutari-ai: IMHO they should be in doc/provides.rst, since that is the documentation for the Shoop Base distribution not just Core. They can be in different section (with a subtitle Campaigns or something like that) in that file though.

suutari-ai:
All commits should have "Refs SHOOP-494".

Update changelog.rst in the commits, at least the core changes are relevant enough for the changelog, but IMHO we should add even quite minor stuff there now and then maybe remove some bits from it when releasing if something seems unrelevant at the time of release. (It's much easier to clean it up than collect it from the commits.)

suutari-ai:
Run generate_apidoc.py and include the changes to doc/api as a commit (or include them to the commit which adds shoop.campaigns).

shoop/themes/classic_gray/templates/classic_gray/shoop/front/base.jinja
```diff
@@ -14,6 +14,7 @@
     <link href='https://fonts.googleapis.com/css?family=Open+Sans:400,300,700' rel='stylesheet' type='text/css'>
     {# Include Styles #}
     <link rel="stylesheet" href="{{ STATIC_URL }}classic_gray/css/style.css">
+    <script type="text/javascript" src="{{ url("shoop:js-catalog") }}"></script>
```
tulimaki: This shouldn't be here since classic gray does not have any js translations.

shoop_tests/campaigns/__init__.py
```diff
@@ -0,0 +1,19 @@
+from django.utils.translation import activate
```
tulimaki: Tests files should also have license headers.

tulimaki:
Most of the commits still missing refs and also you should probably add this new feature to changelog.rst

shoop/campaigns/forms.py
Pikkupomo: SHOOP-2107 @tulimaki

suutari-ai:
Amazing work @Pikkupomo. Thanks a lot!

PR #337, opened 2016-02-02
Addons: Enable upgrade, migrations and collectstatic

tulimaki:
* Use upgrade flag while installing new addons. This allows
updating old addons.
* Run migrations command while reloading application
* Run collectstatic command while reloading application

Refs SHOOP-2106 / SHOOP-1770

### DIFF
--- a/shoop/addons/admin_module/views/reload.py
+++ b/shoop/addons/admin_module/views/reload.py
@@ -10,9 +10,11 @@ from __future__ import unicode_literals
 import time
 
 from django import forms
+from django.core.management import call_command
 from django.http.response import HttpResponse, JsonResponse
 from django.utils.translation import ugettext_lazy as _
 from django.views.generic import FormView
+from six import StringIO
 
 from shoop.addons.reloader import get_reload_method_classes
 from shoop.utils.excs import Problem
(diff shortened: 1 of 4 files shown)


shoop/addons/admin_module/views/reload.py
```diff
@@ -56,4 +64,8 @@ def form_valid(self, form):
     def get(self, request, *args, **kwargs):
         if request.GET.get("ping"):
             return JsonResponse({"pong": time.time()})
+        elif request.GET.get("migrate"):
+            return JsonResponse({"message": run_command("migrate")})
+        elif request.GET.get("collectstatic"):
+            return JsonResponse({"message": run_command("collectstatic")})
```
suutari-ai: IMHO all these operations (pip install && migrate && collectstatic) should happen in a single request. Splitting the process to multiple requests risks the addon installation being left in a state where migrations or collectstatic is not ran when some sort of network or web browser error occurs in the middle of installation.

What was the reason to split it to multiple requests? Maybe I'm missing something, because I can't really understand why it's done  this way, since doing it in single request should be simpler too, or is it?

tulimaki: Just what my plan was initially.

Well we have to reload the application twice. Before and after migrations and collectstatic. Currently the reload request does not return anything. That HttpResponse return is never reached because of reload. So doing migrations after reload seems impossible.

Also this way we offer shop admin more detailed information about what is done while reloading (like steps and log). All of the steps can be performed separately so in this way shop admin needs to just reload again if errors occur. Also I think it's cool to start migrating and doing collectstatic ONLY after the application is up again after initial reload. As I understood the reload itself might take while.

If we actually managed to do everythng in one request still debugging, would be harder in error cases.

PR #338, opened 2016-02-02
Convert CLA.md to CLA.rst

suutari-ai:
Since all the documentation and README etc. are in ReST format, it is
better to use the same format for this too.  It is also easier to read
in text form than the Markdown version was.

Refs SHOOP-2058


suutari-ai:
Just noticed that the section 3 is badly formatted. Don't merge yet.. will fix it..

suutari-ai:
OK, it should be good now.

PR #341, opened 2016-02-04
BasketStorage: fix mis-call of `delete` in `finalize`

akx:
`finalize` never called `delete` correctly (it obviously requires the `basket` argument).

This was noticed by yours truly and @RauliL while working on a client project.

### DIFF
--- a/doc/changelog.rst
+++ b/doc/changelog.rst
@@ -40,6 +40,7 @@ Admin
 Front
 ~~~~~
 
+- Fix bug: BasketStorage.finalize() never called delete() correctly
 - Check product quantity already in basket while adding
 - Process given coupon codes in basket
 - Process discounts from new campaign engine
--- a/shoop/front/basket/storage.py
+++ b/shoop/front/basket/storage.py
@@ -91,7 +91,7 @@ class BasketStorage(six.with_metaclass(abc.ABCMeta)):
         """
         pass
 
-    def finalize(self, basket):  # pragma: no cover
+    def finalize(self, basket):
         """
         Mark the basket as "finalized" (i.e. completed) in the storage.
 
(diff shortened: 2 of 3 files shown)


Pikkupomo:
@akx rebase this please

akx:
@Pikkupomo Rebased.

Pikkupomo:
Thanks, changes should include changelog additions also. We are currently logging every change in changelog and when version release comes, we will remove what seems unnecessary from there. right @suutari-ai?

akx:
The python-babel/babel project uses the GitHub wiki to track changelog changes, which will then be baked into the repo changelog at release time. That might work here too?

suutari-ai:
It's quite time consuming to backtrack relevant info for change log when parsing changes from the Git log. That's why we decided to switch to write the change log entries when doing the actual changes rather than writing them all in single bunch when releasing. Non-relevant stuff can be dropped at release time as @Pikkupomo said.

We still don't have to demand that each contributor writes the changelog entries, but it may be handled by the Shoop team, i.e. we could add the changelog entry when merging. After all, it's part of the release engineering process.

akx:
When I was doing releases, it really didn't take much longer than about half an hour to put the changelog together...

Anyway, I think using the Github wiki works pretty well for Babel; you can jot the changes down there after you've merged a PR without having to make more actual Git changes for the changelog.

If you like, I can add a changelog entry in this PR anyway; just say the word.

Pikkupomo:
@akx it's a current way we are doing things so you could add them now. We will see if there is a need later on, but now there is :) So add please :)

akx:
Changelog entry added, @Pikkupomo.

akx:
Rebased, you know, again.

PR #342, opened 2016-02-05
Core: option to create payments with REST API

tulimaki:
* Add create_payment detail route to OrderViewSet
* Add custom PaymentSerializer with amount_value,
currency, payment_identifier and description as fields.

Refs SHOOP-2117 / SHOOP-1973

### DIFF
--- a/shoop/core/api/orders.py
+++ b/shoop/core/api/orders.py
@@ -6,26 +6,38 @@
 # This source code is licensed under the AGPLv3 license found in the
 # LICENSE file in the root directory of this source tree.
 from django.utils.timezone import now
-from rest_framework.serializers import ModelSerializer
+from rest_framework import serializers, status
+from rest_framework.decorators import detail_route
+from rest_framework.response import Response
 from rest_framework.viewsets import ModelViewSet
 
-from shoop.core.models import MutableAddress, Order, OrderLine, OrderStatus
+from shoop.core.models import (
+    MutableAddress, Order, OrderLine, OrderStatus, Payment
+)
+from shoop.utils.money import Money
 
 
-class OrderLineSerializer(ModelSerializer):
+class OrderLineSerializer(serializers.ModelSerializer):
     class Meta:
         model = OrderLine
 
 
-class AddressSerializer(ModelSerializer):
+class AddressSerializer(serializers.ModelSerializer):
     class Meta:
         model = MutableAddress
 
 
-class OrderSerializer(ModelSerializer):
+class PaymentSerializer(serializers.ModelSerializer):
+    class Meta:
+        model = Payment
+        fields = ("payment_identifier", "amount_value", "description")
+
+
+class OrderSerializer(serializers.ModelSerializer):
     lines = OrderLineSerializer(many=True, read_only=True)
     billing_address = AddressSerializer(read_only=True)
     shipping_address = AddressSerializer(read_only=True)
+    payments = PaymentSerializer(many=True, read_only=True)
 
     def get_fields(self):
         fields = super(OrderSerializer, self).get_fields()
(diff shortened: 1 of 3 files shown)


akx:
This should mention it's about the REST API, not just "API".

Also, it's missing tests.

akx:
Are payments currently retrievable via the API?

Pikkupomo:
like @akx said, add tests please

tulimaki:
@akx payments are not retrievable via the REST API

shoop/core/api/orders.py
```diff
@@ -45,6 +48,49 @@ class Meta:
...
+        serializer = DefaultPaymentSerializer(data=request.data)
+        if serializer.is_valid():
+            data = serializer.validated_data
+            unpaid_value = (order.taxful_total_price_value - order.get_total_paid_amount().value)
+            order.create_payment(
+                Money(unpaid_value, order.currency),
+                data["payment_identifier"],
+                data.get("description", "")
+            )
+            return Response({'status': 'order is now fully paid'})
```
akx: I think a return code of 201 ("Created") would be more appropriate here?

shoop/core/api/orders.py
```diff
@@ -45,6 +48,49 @@ class Meta:
...
+        order = self.get_object()
+        serializer = PaymentSerializer(data=request.data)
+        if serializer.is_valid():
+            data = serializer.validated_data
+            order.create_payment(
+                Money(data["amount_value"], data["currency"]),
+                data["payment_identifier"],
+                data.get("description", "")
+            )
+            return Response({'status': 'payment created'})
```
akx: I think a return code of 201 ("Created") would be more appropriate here?

shoop/core/api/orders.py
```diff
@@ -45,6 +48,49 @@ class Meta:
...
+        else:
+            return Response(serializer.errors, status=status.HTTP_400_BAD_REQUEST)
+
+    @detail_route(methods=['post'])
+    def set_fully_paid(self, request, pk=None):
+        order = self.get_object()
+        if order.is_paid():
+            return Response({"status": "order is already fully paid"})
+        serializer = DefaultPaymentSerializer(data=request.data)
+        if serializer.is_valid():
```
akx: Couldn't this use the `create_payment` method's logic? Would seem DRYer...

tulimaki: What do you mean? create_payment and set_fully_paid uses different serializer since we don't want to use amount_value or currency here.

akx: Refactor the unserialization/creation out from `create_payment`, then

```python
request.data["currency"] =  order.currency
request.data["amount_value"] = unpaid_value
# then call the refactored-out logic with this augmented request
```

akx: Also, couldn't they use the same serializer, just with the currency field being optional (and read from the order in case it's omitted)?

tulimaki:
Ok, thanks, will check these ASAP. And I think that we can use the same serializer for both.

shoop/core/api/orders.py
tulimaki: Ok, thanks, will check these ASAP. And I think that we can use the same serializer for both.

akx:
@tulimaki Now that you have a simple serializer for payments, it should be trivial to add something like `payments = PaymentSerializer(many=True, read_only=True)` to the `OrderSerializer` to make the payments accessible via the REST API.

tulimaki:
Payments is now available via REST API. Since Payment didn't have currency I added currency property for it. I guess that it's OK.

tulimaki:
retest this please

shoop/core/models/_payments.py
```diff
@@ -34,3 +34,7 @@ class Payment(MoneyPropped, models.Model):
     class Meta:
         verbose_name = _('payment')
         verbose_name_plural = _('payments')
+
+    @property
+    def currency(self):
+        return self.amount.currency
```
suutari-ai: What's the point of this? Why can't the API user just use `payment.amount.currency` (or `payment.foreign_amount.currency`)?

shoop/core/api/orders.py
```diff
@@ -6,26 +6,37 @@
...
+class AddressSerializer(serializers.ModelSerializer):
     class Meta:
         model = MutableAddress
 
 
-class OrderSerializer(ModelSerializer):
+class PaymentSerializer(serializers.Serializer):
+    payment_identifier = serializers.CharField(max_length=50)
+    amount_value = serializers.DecimalField(decimal_places=9, max_digits=36, allow_null=True)
+    currency = serializers.CharField(max_length=10, allow_null=True)
```
suutari-ai: Currency of a payment should not be serialized, since it can't be different than the currency of the order. To allow creating payments in different currency, make the foreign_amount_value and foreign_amount_currency serializable.

Btw. Do we really need to respecify every field of the model here? Doesn't seem very DRY. I'm especially concerned about the field parameters which could get updated on the model, but are very easily forgotten here.

shoop/core/api/orders.py
```diff
@@ -45,6 +56,34 @@ class Meta:
...
+    if serializer.is_valid():
+        data = serializer.validated_data
+        order.create_payment(
+            Money(data["amount_value"], data["currency"]),
+            data["payment_identifier"],
+            data.get("description", "")
+        )
+        return Response({'status': 'payment created'}, status=status.HTTP_201_CREATED)
+    else:
+        return Response(serializer.errors, status=status.HTTP_400_BAD_REQUEST)
```
suutari-ai: This function shouldn't be public. It's just a helper for OrderViewSet. Also move it under OrderViewSet for making the code easier to read (less jumping up/down in the file).

suutari-ai:
This is cool now. Thanks, @tulimaki! :)

PR #345, opened 2016-02-08
Admin: Fix syntax error in order creator javascript

tulimaki:


doc/changelog.rst
```diff
@@ -30,6 +30,7 @@ Localization
 Admin
 ~~~~~
 
+- Fix syntax error in order creator
```
suutari-ai: This syntax error was not in 3.0.0, so mention about the fix should not be in the change log. (Change Log should be noting the changes since the last release, not back-forth things like "Implement X", "Fix X", "Fix X more".)

It would be cool if this kind of error would have been caught by some test. Any ideas how to implement that? (Definitely not a must for this PR, but just thinking...)

suutari-ai:
It would be cool if this kind of error would have been caught by some test. Any ideas how to implement that? (Definitely not a must for this PR, but just thinking...)

PR #348, opened 2016-02-10
Addons: Enable upgrade, migrations and collectstatic

tulimaki:
* Use upgrade flag while installing new addons. This allows
updating old addons.
* Run migrations command while reloading application
* Run collectstatic command while reloading application

Refs SHOOP-2106 / SHOOP-1770

### DIFF
--- a/shoop/apps/settings.py
+++ b/shoop/apps/settings.py
@@ -105,3 +105,13 @@ def validate_templates_configuration():
         "The `django_jinja` template backend was not encountered in your TEMPLATES configuration. "
         "See the Shoop or Django-Jinja documentation on more information how to configure things correctly."
     )
+
+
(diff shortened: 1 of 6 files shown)


tulimaki:
retest this please

shoop/apps/settings.py
```diff
@@ -105,3 +105,13 @@ def validate_templates_configuration():
         "The `django_jinja` template backend was not encountered in your TEMPLATES configuration. "
         "See the Shoop or Django-Jinja documentation on more information how to configure things correctly."
     )
+
+
+def reload_apps():
```
suutari-ai: Public API for this function should be in shoop.apps, not in shoop.apps.settings.

PR #349, opened 2016-02-10
Enhance stocks admin

tulimaki:
* Core: Add is_stocked method for Product model
* Simple Supplier: Show better information when product is not stocked

Refs SHOOP-2131 / SHOOP-1876

### DIFF
--- a/shoop/simple_supplier/templates/shoop/simple_supplier/admin/product_form_part.jinja
+++ b/shoop/simple_supplier/templates/shoop/simple_supplier/admin/product_form_part.jinja
@@ -1,26 +1,65 @@
 {% from "shoop/admin/macros/general.jinja" import content_block with context %}
 {% set ss_form = form["simple_supplier"] %}
-{% set product_qty = ss_form.products|length %}
-{% if product_qty %}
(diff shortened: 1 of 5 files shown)


shoop/simple_supplier/templates/shoop/simple_supplier/admin/product_form_part.jinja
```diff
@@ -1,26 +1,59 @@
...
+                <div class="col-md-6">
+                    {{ ss_form.get_stock_adjustment_form(supplier, product)|safe }}
                 </div>
-            {% endfor %}
         {% endfor %}
-    {% endcall %}
-{% endif %}
+    {% else %}
+        <p>
+            {% trans product_name=product.name, product_url=product_url, module_name=ss_form.module_name %}
```
akx: Translatable strings shouldn't have leading and trailing space. You can have Jinja get rid of those for you by adding a dash in the surrounding tag, i.e. `{% trans ... %}` would be `{% trans ... -%}`.

Pikkupomo:
changelog needed? other than that lgtm

PR #350, opened 2016-02-10
Informative basket error handling

tulimaki:
* Classic Gray: Show error messages while adding products to basket
* Front: Check product quantity already in basket while adding
* Core: Add shop product validation for OrderSource
* Admin: check product quantity on other lines when adding

Refs SHOOP-2132 / SHOOP-2130 / SHOOP-1876

### DIFF
--- a/shoop/admin/modules/orders/json_order_creator.py
+++ b/shoop/admin/modules/orders/json_order_creator.py
@@ -268,6 +268,12 @@ class JsonOrderCreator(object):
         if not self.is_valid:  # If we encountered any errors thus far, don't bother going forward
             return None
 
+        for error in source.get_validation_errors():
+            self.add_error(error)
+
+        if not self.is_valid:
+            return None
+
         return source
 
     def create_order_from_state(self, state, creator=None):
--- a/shoop/core/order_creator/_source.py
+++ b/shoop/core/order_creator/_source.py
@@ -7,8 +7,13 @@
 # LICENSE file in the root directory of this source tree.
 from __future__ import unicode_literals
 
+from collections import Counter
+
+from django.core.exceptions import ValidationError
 from django.utils.encoding import force_text
 from django.utils.timezone import now
+from django.utils.translation import ugettext_lazy as _
+from six import iteritems
 
 from shoop.core import taxing
 from shoop.core.models import (
@@ -316,6 +321,8 @@ class OrderSource(object):
             tax_module = taxing.get_tax_module()
             if with_taxes or tax_module.calculate_taxes_automatically:
                 self._calculate_taxes(lines, tax_module)
+        for error_message in self.get_validation_errors():
+            raise ValidationError(error_message.args[0], code="invalid_order_source")
         return lines
 
     def calculate_taxes(self, force_recalculate=False):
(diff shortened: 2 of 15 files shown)


tulimaki:
retest this please

tulimaki:
retest this please

tulimaki:
retest this please

shoop_tests/simple_supplier/test_order_source.py
```diff
@@ -0,0 +1,63 @@
...
+def test_order_source(rf, admin_user):
+    """
+    Test order source validation with stocked products.
+    """
+    shop = get_default_shop()
+    supplier = get_simple_supplier()
+    product = create_product("simple-test-product", shop, supplier)
+    product.stock_behavior = StockBehavior.STOCKED
+    product.save()
+    quantity = random.randint(100, 600)
```
suutari-ai: There's no point randomizing this.

shoop/core/order_creator/_source.py
```diff
@@ -392,6 +395,19 @@ def get_product_lines(self):
...
+    def get_products_and_quantities(self, supplier=None):
+        q_counter = Counter()
+        for line in self.get_lines():
+            if supplier and line.supplier != supplier:
+                continue
+            product = line.product
+            q_counter[product] += line.quantity
+
+        return dict(q_counter)
+
```
suutari-ai: `get_supplier` and `get_products_and_quantities`: Make these private and move them under the `get_validation_errors` since they are only used there for now and IMHO we shouldn't make them public before there is any real use for them. (I don't see they would be very useful for OrderSource API users.)

shoop/core/order_creator/_source.py
```diff
@@ -392,6 +395,19 @@ def get_product_lines(self):
         product_lines = [l for l in self.get_lines() if l.product]
         return product_lines
 
+    def get_suppliers(self):
+        return set([l.supplier for l in self.get_lines() if l.supplier])
```
suutari-ai: Using `get_lines` here doesn't account for product lines added by OrderSourceModifiers. They could adds some new product lines in the `get_final_lines` phase. (E.g. if there's a campaign that you get a rubber duck for free when buying two laptops, then `get_final_lines` could add the rubber duck to basket if there is two laptops in there.)

Therefore this and the `get_products_and_quantities` should use `get_final_lines` rather than `get_lines`, but does it create a cycle somewhere?

shoop/admin/modules/orders/json_order_creator.py
```diff
@@ -268,6 +268,12 @@ def create_source_from_state(self, state, creator=None, save=False):
         if not self.is_valid:  # If we encountered any errors thus far, don't bother going forward
             return None
 
+        for error in source.get_validation_errors():
```
suutari-ai: Should the validation errors be checked also in OrderCreator.create_order?

shoop/core/order_creator/_source.py
```diff
@@ -404,6 +420,13 @@ def get_validation_errors(self):
             for error in payment_method.get_validation_errors(source=self):
                 yield error
 
+        for supplier in self.get_suppliers():
+            for product, quantity in iteritems(self.get_products_and_quantities(supplier)):
+                shop_product = product.get_shop_instance(shop=self.shop)
+                for error in shop_product.get_orderability_errors(
+                        supplier=supplier, quantity=quantity, customer=self.customer):
+                    yield error
+
```
suutari-ai: IMHO the checks from `BaseBasket.get_validation_errors` should be moved here too, since now there is two places that check very similar things and there is no clear reason why other things are checked on the BaseBasket side and others on the OrderSource.

tulimaki: I think that checking that there is shipping and payment methods is just for BaseBasket. So I leave those there at least for now. ShopProduct checks after those is now unneccesary since it's already done in OrderSource validation, so I will remove shop product orderability checks from BaseBasket.

tulimaki:
So I figured that it is not a good thing to add new lines to OrderSource while validating it (so no get_final_lines). Added validation after new lines is processed in get_final_lines. That should be done since we have to be sure that those extra processors does not make OrderSource invalid. @suutari-ai

shoop/core/order_creator/_source.py
```diff
@@ -404,6 +411,35 @@ def get_validation_errors(self):
             for error in payment_method.get_validation_errors(source=self):
                 yield error
 
+        for supplier in self._get_suppliers():
+            for product, quantity in iteritems(self._get_products_and_quantities(supplier)):
+                shop_product = product.get_shop_instance(shop=self.shop)
+                if not shop_product:
```
suutari-ai: IIRC get_shop_instance raises if such shop product does not exists, so this is wrong.

PR #355, opened 2016-02-17
Simple CMS: Add parent for Page and list children on page (SHOOP-2144)

tulimaki:
Refs SHOOP-1966 / SHOOP-2144

### DIFF
--- a/shoop/simple_cms/models.py
+++ b/shoop/simple_cms/models.py
@@ -12,6 +12,8 @@ from django.db.models import Q
 from django.utils.encoding import force_text, python_2_unicode_compatible
 from django.utils.timezone import now
 from django.utils.translation import ugettext_lazy as _
+from mptt.managers import TreeManager
+from mptt.models import MPTTModel, TreeForeignKey
 from parler.managers import TranslatableQuerySet
 from parler.models import TranslatableModel, TranslatedFields
 
@@ -38,7 +40,7 @@ class PageQuerySet(TranslatableQuerySet):
 
 
 @python_2_unicode_compatible
-class Page(TranslatableModel):
+class Page(MPTTModel, TranslatableModel):
     available_from = models.DateTimeField(null=True, blank=True, verbose_name=_('available from'))
     available_to = models.DateTimeField(null=True, blank=True, verbose_name=_('available to'))
 
@@ -61,6 +63,9 @@ class Page(TranslatableModel):
     )
 
     visible_in_menu = models.BooleanField(verbose_name=_("visible in menu"), default=False)
+    parent = TreeForeignKey(
+        "self", blank=True, null=True, related_name="children", verbose_name=_("parent"))
+    list_children_on_page = models.BooleanField(verbose_name=_("list children on page"), default=False)
 
     translations = TranslatedFields(
         title=models.CharField(max_length=256, verbose_name=_('title')),
@@ -74,7 +79,7 @@ class Page(TranslatableModel):
         content=models.TextField(verbose_name=_('content')),
     )
 
-    objects = PageQuerySet.as_manager()
+    objects = TreeManager.from_queryset(PageQuerySet)()
 
     class Meta:
         ordering = ('-id',)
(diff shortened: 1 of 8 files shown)


tulimaki:
retest this please

shoop/simple_cms/templates/shoop/simple_cms/page.jinja
```diff
@@ -6,9 +6,21 @@
...
     </ol>
 {% endblock %}
 {% block content %}
     {{ page.get_html()|safe }}
+    {% if page.list_children_on_page %}
+        {% for child in page.children.all().order_by("-available_from")  %}
+            <div class="row">
+                <a href="/{{ child.url }}"><h2>{{ child.title }}</h2></a>
+                <small>{{ child.available_from|datetime }}</small>
+                <p>{{ child.get_html()|safe|striptags|truncate(256) }}</p>
```
akx: This shouldn't use `|safe|`, I think...

shoop/simple_cms/templates/shoop/simple_cms/page.jinja
```diff
@@ -6,9 +6,21 @@
...
         <li class="active">{{ page.title }}</li>
     </ol>
 {% endblock %}
 {% block content %}
     {{ page.get_html()|safe }}
+    {% if page.list_children_on_page %}
+        {% for child in page.children.all().order_by("-available_from")  %}
+            <div class="row">
+                <a href="/{{ child.url }}"><h2>{{ child.title }}</h2></a>
+                <small>{{ child.available_from|datetime }}</small>
```
akx: Is `available_from` really supposed to be an user-visible field? What's the UC?

tulimaki: Yes. Since the task is talking about articles and such I think it's important to show published date in the pages list.

akx: Okay. But not all articles will have `available_from` set; will this work if they're `None`?

tulimaki:
retest this please

Pikkupomo:
missing tests

tulimaki:
Added some tests and decided to render only visible children on parent page instead of all. @Pikkupomo

shoop/simple_cms/models.py
```diff
@@ -12,13 +12,15 @@
...
+from mptt.managers import TreeManager
+from mptt.models import MPTTModel, TreeForeignKey
+from parler.managers import TranslatableManager
 from parler.models import TranslatableModel, TranslatedFields
 
 from shoop.core.fields import InternalIdentifierField
 
 
-class PageQuerySet(TranslatableQuerySet):
+class PageManager(TranslatableManager, TreeManager):
```
suutari-ai: Is there such thing as TreeQuerySet? Would like to use query set rather than manager, because queryset can be piped (`Page.objects.filter(something).visible()`).

suutari-ai: OK, it seems that there is no such thing. But you can still let the PageQuerySet be as a queryset by using `from_queryset` (see https://docs.djangoproject.com/en/1.9/topics/db/managers/#from-queryset).

```
class PageQuerySet(TranslatableQuerySet):  # as before
    ...


class Page(MPTTModel, TranslatableModel):
    ...
    objects = TreeManager.from_queryset(PageQuerySet)()  # this is the trick!
```
```

File: shoop/simple_cms/models.py [hunk 4]
```diff
@@ -74,7 +79,7 @@ class Page(TranslatableModel):
         content=models.TextField(verbose_name=_('content')),
     )
 
-    objects = PageQuerySet.as_manager()
+    objects = PageManager()
```
suutari-ai: Check if this works with
```
objects = TreeManager.from_queryset(PageQuerySet)()
```
as noted above.

tulimaki: Seems to work. Thanks.

suutari-ai:
Good work, @tulimaki!

PR #356, opened 2016-02-17
Core: Move calculate taxes automatically under TaxModule

tulimaki:
Move calculate taxes automitacally from OrderSource to TaxModule.
Create setting SHOOP_CALCULATE_TAXES_AUTOMATICALLY.

Refs SHOOP-1911

### DIFF
--- a/doc/changelog.rst
+++ b/doc/changelog.rst
@@ -12,6 +12,7 @@ Unreleased
 Core
 ~~~~
 
+- Move calculate_taxes_automatically from ``OrderSource`` to ``TaxModule``
 - Add contact address for ``Shop``
 - Add update_stock calls for ``SimpleSupplierModule``
 - Rename ``simple_pricing`` to ``customer_group_pricing``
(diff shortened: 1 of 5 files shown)


suutari-ai:
The setting is not useful as it is implemented. The setting should not be read in the TaxModule, but rather TaxModule should return True or False based on its internal details. When some code would like to know if it should update taxes automatically, it should check the setting first and then the TaxModule.

Naming of the setting and TaxModule method is also non-optimal, since it's not decided by the TaxModule when taxes are calculated. TaxModule should just specify whether calculating taxes is cheap.

My suggestion:
 * Add `TaxModule.calculating_is_cheap` property. Set it to True for DefaultTaxModule.
 * Add setting `SHOOP_CALCULATE_TAXES_AUTOMATICALLY_IF_POSSIBLE`, default it to True. (added the `_IF_POSSIBLE` suffix to be more explicit that auto-calculate cannot be forced to be True, it can just be forced to be False)
 * Add `shoop.core.taxing.should_calculate_taxes_automatically` function defined as:
```python
def should_calculate_taxes_automatically():
    if not settings.SHOOP_CALCULATE_TAXES_AUTOMATICALLY_IF_POSSIBLE:
        return False
    return get_tax_module().calculating_is_cheap
```

suutari-ai:
Oh, and.. TaxModule.calculating_is_cheap should default to False in the TaxModule interface.

PR #357, opened 2016-02-18
Enable custom AUTH_USER_MODEL with username field other than `username`.

chessbr:
This enables the usage of a custom AUTH_USER_MODEL in Shoop using mock data.

### DIFF
--- a/shoop/testing/management/commands/shoop_populate_mock.py
+++ b/shoop/testing/management/commands/shoop_populate_mock.py
@@ -26,7 +26,8 @@ class Command(BaseCommand):
         translation.activate(settings.LANGUAGES[0][0])
         superuser_name = options.get("with_superuser")
         if superuser_name:
-            user = get_user_model().objects.filter(username=superuser_name).first()
+            filter_params = {get_user_model().USERNAME_FIELD: superuser_name}
+            user = get_user_model().objects.filter(**filter_params).first()
             if not user:
                 user = get_user_model().objects.create_superuser(
                     username=superuser_name,


shoop/testing/management/commands/shoop_populate_mock.py
```diff
@@ -26,7 +26,8 @@ def handle(self, *args, **options):
         translation.activate(settings.LANGUAGES[0][0])
         superuser_name = options.get("with_superuser")
         if superuser_name:
-            user = get_user_model().objects.filter(username=superuser_name).first()
+            filter_params = {get_user_model().USERNAME_FIELD: superuser_name}
+            user = get_user_model().objects.filter(**filter_params).first()
             if not user:
                 user = get_user_model().objects.create_superuser(
                     username=superuser_name,
```
suutari-ai: Does `create_superuser` still work with the hardcoded `username` parameter or should it be changed too?

chessbr: Yes, it works as soon as the `create_superuser` function has a `kwargs` parameter. In this case, the developer must keep in mind that any hardcoded keyword can be used when invoking this method.
If the custom user model `USERNAME_FIELD` is `email`, for example, the `create_superuser` here will fail, because you are setting the same parameter twice (1 harcoded and 1 through `USERNAME_FIELD`).

chessbr: news about this?

suutari-ai:
@chessbr Sorry that this got buried in the PR queue.

It looks good to me, so merging. Thanks!

PR #358, opened 2016-02-18
Core: Add default contact groups

tulimaki:
* Default contact groups for anonymous, person and company contacts
* Groups is assigned during save
* Make ContactGroup TranslatableShoopModel

Refs SHOOP-1981 / SHOOP-2141

### DIFF
--- a/doc/changelog.rst
+++ b/doc/changelog.rst
@@ -12,6 +12,7 @@ Unreleased
 Core
 ~~~~
 
+- Add contacts automatically to type specific default groups
 - ``OrderCreator`` no longer requires a request
 - Add ``order_creator_finished`` signal under ``order_creator``
 - Move calculate_taxes_automatically from ``OrderSource`` to ``TaxModule``
--- a/shoop/core/models/_contacts.py
+++ b/shoop/core/models/_contacts.py
@@ -12,18 +12,20 @@ from django.db import models
 from django.utils.encoding import python_2_unicode_compatible
 from django.utils.translation import ugettext_lazy as _
 from enumfields import Enum, EnumField
-from parler.models import TranslatableModel, TranslatedFields
+from parler.models import TranslatedFields
 from polymorphic.models import PolymorphicModel
 from timezone_field.fields import TimeZoneField
 
 from shoop.core.fields import InternalIdentifierField, LanguageField
 from shoop.core.utils.name_mixin import NameMixin
+from shoop.utils.text import force_text
 
+from ._base import TranslatableShoopModel
 from ._taxes import CustomerTaxGroup
 
 
 @python_2_unicode_compatible
-class ContactGroup(TranslatableModel):
+class ContactGroup(TranslatableShoopModel):
     identifier = InternalIdentifierField(unique=True)
     members = models.ManyToManyField("Contact", related_name="groups", verbose_name=_('members'), blank=True)
     show_pricing = models.BooleanField(verbose_name=_('show as pricing option'), default=True)
@@ -37,7 +39,7 @@ class ContactGroup(TranslatableModel):
         verbose_name_plural = _('contact groups')
 
     def __str__(self):
-        return self.safe_translation_getter("name", default="Group<%s>" % (self.identifier or self.id))
+        return force_text(self.safe_translation_getter("name", default="Group<%s>" % (self.identifier or self.id)))
 
 
 @python_2_unicode_compatible
@@ -45,6 +47,8 @@ class Contact(NameMixin, PolymorphicModel):
     is_anonymous = False
     is_all_seeing = False
     default_tax_group_getter = None
+    default_contact_group_identifier = None
+    default_contact_group_name = None
 
     created_on = models.DateTimeField(auto_now_add=True, editable=False, verbose_name=_('created on'))
     identifier = InternalIdentifierField(unique=True, null=True, blank=True)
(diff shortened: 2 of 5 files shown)


shoop/core/models/_contacts.py
```diff
@@ -92,9 +96,25 @@ def __init__(self, *args, **kwargs):
             kwargs.setdefault("tax_group", self.default_tax_group_getter())
         super(Contact, self).__init__(*args, **kwargs)
 
+    def save(self, *args, **kwargs):
+        super(Contact, self).save(*args, **kwargs)
+        if self.default_contact_group_identifier:
```
akx: This will happen on every save, not just the first one?
So one can't ever remove a contact from their default contact group?

shoop/core/models/_contacts.py
```diff
@@ -92,9 +96,25 @@ def __init__(self, *args, **kwargs):
...
+            self.groups.add(self.get_default_contact_group())
+
+    def get_default_contact_group(self):
```
akx: Missing docstring.

shoop/core/models/_contacts.py
```diff
@@ -92,9 +96,25 @@ def __init__(self, *args, **kwargs):
...
+        obj, created = ContactGroup.objects.get_or_create(
+            identifier=self.default_contact_group_identifier,
+            defaults={
+                "name": self.default_contact_group_name
```
akx: This will create the group with the name in some arbitrary language, depending on which i18n context is active during the request that creates the first contact of each type. Not sure that's what's wanted.

shoop/core/models/_contacts.py
```diff
@@ -150,10 +172,16 @@ def is_all_seeing(self):
...
     pk = id = None
     is_anonymous = True
+    default_contact_group_identifier = "default_anonymous_group"
+    default_contact_group_name = _("Anonymous Contacts")
 
     class Meta:
         managed = False  # This isn't something that should actually exist in the database
 
+    def __init__(self, *args, **kwargs):
+        self.get_default_contact_group()  # Make sure default anonymous contact group is created
```
akx: It seems somewhat scary that an `__init__` would have the side effect of creating an unrelated object in the database.

Would it be better to do this in `.groups`

shoop_tests/core/test_contacts.py
```diff
@@ -73,9 +74,8 @@ def test_anonymous_contact():
         a1.delete()
 
     assert isinstance(a1.groups, QuerySet)
-    assert a1.groups.count() == 0
-    assert len(a1.groups) == 0
-    assert not a1.groups
+    assert a1.groups.count() == 1
```
akx: This should probably assert that the single group returned is what you expect instead.

shoop_tests/core/test_customers.py
```diff
@@ -63,3 +60,22 @@ def test_customer_tax_group2():
...
+    assert default_group == new_contact.get_default_contact_group()
+    assert default_group.identifier == contact_cls.default_contact_group_identifier
+
+    some_other_contact = create_contact()
+    assert some_other_contact.groups.count() == 1
+
+
+@pytest.mark.django_db
+def test_default_groups():
+    default_group_basic_test(AnonymousContact, AnonymousContact)
```
akx: This would be better phrased as a parametrization of the single test, not a helper that runs three tests with parameters.

See [`pytest.parametrize`](https://pytest.org/latest/parametrize.html#pytest-mark-parametrize-parametrizing-test-functions).

shoop/core/models/_contacts.py
tulimaki: Good point. Default tax groups uses the same logic so I think this is ok.

tulimaki:
retest this please

shoop/core/models/_contacts.py
```diff
@@ -170,7 +207,8 @@ def delete(self, *args, **kwargs):
 
     @property
     def groups(self):
```
Pikkupomo: This really should be a top-level method, since now the `groups` is a required property but it is not required per se. 
Example: you add `ThisIsCustomContact(Contact)` and call the `groups` for that, does it have it?

Pikkupomo:
Ref should say  SHOOP-1981 /  SHOOP-2141 btw. Mainticket / sub

shoop/core/models/_contacts.py
tulimaki: Are you sure? I think in your examble groups works since it comes from ContactGroup relation.

suutari-ai: @tulimaki You're right. We totally missed this with @Pikkupomo yesterday. Please add a docstring to the AnonymousContact.groups property which clarifies this, e.g.

```python
@property
def groups(self):
    """
    Contact groups accessor for anonymous contact.

    The base class already has a `groups` property via `ContactGroup`
    related_name, but this overrides it for `AnonymousContact` so that
    it will return a queryset containing just the anonymous contact
    group rather than returning the original related manager, which
    cannot work since `AnonymousContact` is not in the database.

    This allows to use statements like this for all kinds of contacts,
    even `AnonymousContact`::

        some_contact.groups.all()

    :rtype: django.db.QuerySet
    """
```

doc/changelog.rst
```diff
@@ -12,6 +12,7 @@ Unreleased
 Core
 ~~~~
 
+- Add default contact groups
```
suutari-ai: Be more specific, e.g. "Add contacts automatically to automatically created type specific groups"

tulimaki:
Updated based on @suutari-ai comments. @Pikkupomo

shoop/core/models/_contacts.py
```diff
@@ -37,7 +39,7 @@ class Meta:
         verbose_name_plural = _('contact groups')
 
     def __str__(self):
-        return self.safe_translation_getter("name", default="Group<%s>" % (self.identifier or self.id))
+        return force_text(self.safe_translation_getter("name", default="Group<%s>" % (self.identifier or self.id)))
```
suutari-ai: Why would this need force_text? IMHO `__str__` should rather just be removed from here, since `TranslatableShoopModel` has a good default.

suutari-ai:
The comment about `__str__` is minor. I'll just merge this in and check if the `__str__` can be removed in another PR...

Very nice job, @tulimaki, thanks!

PR #359, opened 2016-02-19
pt_BR: Improve a couple of translation messages

jhess666:
I have improved some strings to better semantic of the message.


tulimaki:
Thanks!

PR #366, opened 2016-03-01
Campaigns: Add contact group and contact conditions (SHOOP-2098)

tulimaki:
Add contact group and contact conditions for basket and
context campaigns.

Refs SHOOP-1981 / SHOOP-2098

### DIFF
--- a/shoop/campaigns/models/basket_conditions.py
+++ b/shoop/campaigns/models/basket_conditions.py
@@ -10,7 +10,7 @@ from django.utils.translation import ugettext_lazy as _
 from polymorphic.models import PolymorphicModel
 
 from shoop.core.fields import MoneyValueField
-from shoop.core.models import Product
+from shoop.core.models import Contact, ContactGroup, Product
 from shoop.utils.properties import MoneyPropped, PriceProperty
 
 
@@ -38,7 +38,7 @@ class BasketTotalProductAmountCondition(BasketCondition):
 
     @property
     def description(self):
-        return _("Limit the campaign to match when basket has atleast the product count entered here.")
+        return _("Limit the campaign to match when basket has at least the product count entered here.")
 
     @property
     def value(self):
@@ -94,3 +94,51 @@ class ProductsInBasketCondition(BasketCondition):
     @values.setter
     def values(self, value):
         self.products = value
+
+
+class ContactGroupBasketCondition(BasketCondition):
+    model = ContactGroup
+    identifier = "basket_contact_group_condition"
+    name = _("Contact Group")
+
+    contact_groups = models.ManyToManyField(ContactGroup, verbose_name=_("contact groups"))
+
+    def matches(self, basket, lines=[]):
+        customers_groups = basket.customer.groups.all()
+        return self.contact_groups.filter(pk__in=customers_groups).exists()
+
+    @property
+    def description(self):
+        return _("Limit the campaign to members of the selected contact groups.")
+
+    @property
+    def values(self):
+        return self.contact_groups
+
+    @values.setter
+    def values(self, values):
+        self.contact_groups = values
+
+
+class ContactBasketCondition(BasketCondition):
+    model = Contact
+    identifier = "basket_contact_condition"
+    name = _("Contact")
+
+    contacts = models.ManyToManyField(Contact, verbose_name=_("contacts"))
+
+    def matches(self, basket, lines=[]):
+        customer = basket.customer
+        return bool(customer and self.contacts.filter(pk=customer.pk).exists())
+
+    @property
+    def description(self):
+        return _("Limit the campaign to selected contacts.")
+
+    @property
+    def values(self):
+        return self.contacts
+
+    @values.setter
+    def values(self, values):
+        self.contacts = values
--- a/shoop/campaigns/models/context_conditions.py
+++ b/shoop/campaigns/models/context_conditions.py
@@ -8,7 +8,7 @@ from django.db import models
 from django.utils.translation import ugettext_lazy as _
 from polymorphic.models import PolymorphicModel
 
-from shoop.core.models import ContactGroup
+from shoop.core.models import AnonymousContact, Contact, ContactGroup
 
 
 class ContextCondition(PolymorphicModel):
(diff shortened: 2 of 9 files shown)


Pikkupomo:
This is somewhat all over the place now. Simple prototype that should work:

`context_conditions.py`:

```
class ContactCondition(ContextCondition):
    model = Contact
    identifier = "contact_group_condition"
    name = _("Contact Group")
    description = _("Contact group")

    contacts = models.ManyToManyField(Contact, verbose_name=_("contact"))

    @property
    def description(self):
        return _("Limit the campaign to members of the selected contact groups.")

    def matches(self, context):
        if not context.customer:
            return False
        
        return self.contacts.filter(pk=context.customer.pk).exists()
    
    @property
    def values(self):
        return self.contacts

    @values.setter
    def values(self, values):
        self.contacts = values
```


`basket_conditions.py`:

```
class CustomerBasketCondition(BasketCondition):
    identifier = "basket_customers_condition"
    name = _("Customers in basket")

    model = Contact

    contacts = models.ManyToManyField(Contact, verbose_name=_("contacts"), blank=True)

    def matches(self, basket, lines):
        if not basket.customer:
            return False
        return self.contacts.filter(pk=basket.customer.pk).exists()

    @property
    def description(self):
        return _("Limit the campaign to have the selected customers in basket.")

    @property
    def values(self):
        return self.contacts

    @values.setter
    def values(self, value):
        self.contacts = value


class CustomerGroupBasketCondition(BasketCondition):
    identifier = "basket_customer_groups_condition"
    name = _("Customer Groups in basket")

    model = ContactGroup

    groups = models.ManyToManyField(ContactGroup, verbose_name=_("contact groups"), blank=True)

    def matches(self, basket, lines):
        if not basket.customer.groups:
            return False
        return self.groups.filter(pk_in=basket.customer.all()).exists()

    @property
    def description(self):
        return _("Limit the campaign to have the selected customer groups in basket.")

    @property
    def values(self):
        return self.groups

    @values.setter
    def values(self, value):
        self.groups = value
```

Why should it be more complex?

shoop/campaigns/models/basket_conditions.py
```diff
@@ -94,3 +94,53 @@ def values(self):
...
+    model = ContactGroup
+    identifier = "basket_contact_group_condition"
+    name = _("Contact Group")
+
+    contact_groups = models.ManyToManyField(ContactGroup, verbose_name=_("contact groups"))
+
+    def matches(self, basket, lines=[]):
+        if basket.customer:
+            contact_groups = basket.customer.groups.all().values_list("pk", flat=True)
+            return self.contact_groups.filter(pk__in=contact_groups).exists()
```
suutari-ai: There is no need to do `values_list` here.

Doing just:

```python
customers_groups = basket.customer.groups.all()
return self.contact_groups.filter(pk__in=customers_groups).exists()
```

would be faster, since it is only one query.

shoop/campaigns/models/catalog_filters.py
```diff
@@ -9,6 +9,7 @@
 from polymorphic.models import PolymorphicModel
 
 from shoop.core.models import Category, Product, ProductType
+from shoop.utils.text import force_text
```
suutari-ai: force_text should be imported from django.

shoop/campaigns/models/catalog_filters.py
```diff
@@ -19,6 +20,9 @@ class CatalogFilter(PolymorphicModel):
 
     active = models.BooleanField(default=True, verbose_name=_("active"))
 
+    def __str__(self):
+        return force_text(self.name)
```
suutari-ai: Name of the filter class in not very good string for the object, since str of the object should have the values of the object too (e.g. "Product type: Normal product" or "Product: SOME-SKU-123", not just "Product type" or "Product").  Why are you defining str here?

shoop_tests/campaigns/test_basket_product_conditions.py
```diff
@@ -18,7 +20,7 @@
 @pytest.mark.django_db
 def test_product_in_basket_condition(rf):
     request, shop, group = initialize_test(rf, False)
-    price = shop.create_price
+    shop.create_price
```
suutari-ai: Just remove the line if you don't need the price function in this test.

shoop_tests/campaigns/test_basket_product_conditions.py
```diff
@@ -38,7 +40,7 @@ def test_product_in_basket_condition(rf):
 @pytest.mark.django_db
 def test_basket_total_amount_condition(rf):
     request, shop, group = initialize_test(rf, False)
-    price = shop.create_price
+    shop.create_price
```
suutari-ai: Just remove the line if you don't need the price function in this test.

shoop_tests/campaigns/test_basket_product_conditions.py
```diff
@@ -56,7 +58,7 @@ def test_basket_total_amount_condition(rf):
 @pytest.mark.django_db
 def test_basket_total_value_condition(rf):
     request, shop, group = initialize_test(rf, False)
-    price = shop.create_price
+    shop.create_price
```
suutari-ai: Just remove the line if you don't need the price function in this test.

shoop_tests/campaigns/test_context_conditions.py
```diff
@@ -0,0 +1,89 @@
...
+    apply_request_middleware(request)
+    return request
+
+
+def create_random_product_and_campaign(shop, rules, original_price_value, discount_value):
+    campaign = CatalogCampaign.objects.create(
+        shop=shop, name="test", discount_amount_value=discount_value, active=True)
+
+    for rule in rules:
+        campaign.conditions.add(rule)
```
suutari-ai: Why are calling conditions rules? Rule is something that specifies what to do when some conditions match, i.e. campaign is a the rule here.

tulimaki: Ok, got confused since these are called "Campaing Rules" in admin.

shoop/campaigns/models/basket_conditions.py
tulimaki: I make this change also into ``ContactGroupCondition``.

shoop/campaigns/models/context_conditions.py
```diff
@@ -27,20 +27,19 @@ class ContactGroupCondition(ContextCondition):
...
     name = _("Contact Group")
-    description = _("Contact group")
 
     contact_groups = models.ManyToManyField(ContactGroup, verbose_name=_("contact groups"))
 
+    def matches(self, context):
+        if not context.customer:
+            return False
+        customers_groups = context.customer.groups.all()
+        return self.contact_groups.filter(pk__in=customers_groups).exists()
```
suutari-ai: This does not work for the AnonymousContact, since the are falsy objects and the code will go to `if not context.customer` branch rather than checking the anonymous contacts group as it should. The `if not context.customer` should probably be just removed or changed to:
```python
customer = (context.customer or AnonymousContact())
```
and then use `customer.groups.all()` below.

tulimaki: Oh, really good catch. Thanks!

suutari-ai:
Was about to say that this will need an entry to change log, but after checking better, it seems it doesn't since this is just improving a feature that is currently listed in the change log.

suutari-ai:
Good job, @tulimaki!

PR #365, opened 2016-02-29
Themes: Product highlights plugin improvements

shawnadelic:
Add option for highlights plugin to hide all out-of-stock products (true
by default).

Refs SHOOP-1969

PR #369, opened 2016-03-04
Themes: Add product cross sells plugin to classic_gray (SHOOP-1970)

shawnadelic:
Add a product cross sells xtheme plugin to classic_gray theme, replacing
hard-coded sections in theme templates.

Refs SHOOP-1970

### DIFF
--- a/shoop/themes/classic_gray/plugins.py
+++ b/shoop/themes/classic_gray/plugins.py
@@ -50,3 +50,37 @@ class ProductHighlightPlugin(TemplatedPlugin):
             "title": self.get_translated_value("title"),
             "products": products
         }
+
+
+class ProductCrossSellsPlugin(TemplatedPlugin):
+    identifier = "classic_gray.product_cross_sells"
+    name = _("Product Cross Sells")
+    template_name = "classic_gray/cross_sells_plugin.jinja"
+    required_context_variables = ["product"]
+    fields = [
+        ("title", TranslatableField(label=_("Title"), required=False, initial="")),
+        ("type", forms.ChoiceField(label=_("Type"), choices=[
+            ("related", "Related"),
+            ("recommended", "Recommended"),
+            ("computed", "Computed"),
+        ], initial="related")),
+        ("count", forms.IntegerField(label=_("Count"), min_value=1, initial=4)),
+        ("orderable_only", forms.BooleanField(label=_("Only show in-stock and orderable items"),
+                                              initial=True,
+                                              required=False))
+    ]
+
+    def get_context_data(self, context):
+        count = self.config.get("count", 4)
+        product = context.get("product", None)
+        orderable_only = self.config.get("orderable_only", True)
+        type = self.config.get("type", "related")
+
+        return {
+            "request": context["request"],
+            "title": self.get_translated_value("title"),
+            "product": product,
+            "type": type,
+            "count": count,
+            "orderable_only": orderable_only,
+        }
--- a/shoop/themes/classic_gray/templates/classic_gray/shoop/front/macros.jinja
+++ b/shoop/themes/classic_gray/templates/classic_gray/shoop/front/macros.jinja
@@ -34,21 +34,6 @@
     </div>
 {% endmacro %}
 
-{% macro render_cross_sell_products(product, relation_type="", title="") %}
-    {%- set cross_sell_products = shoop.product.get_product_cross_sells(product, relation_type) %}
-    {% if cross_sell_products %}
-        <hr>
-        {% if title %}<h3>{{ title }}</h3>{% endif %}
-        <div class="row">
-            {% for product in cross_sell_products %}
-                <div class="col-md-3">
-                    {{ product_box(product) }}
-                </div>
-            {% endfor %}
-        </div>
-    {% endif %}
-{% endmacro %}
-
 {% macro render_field(field, classes="") %}
     {% if field.field.widget.is_hidden %}
         {{ field.as_widget()|safe }}
--- a/shoop_tests/themes/classic_gray/test_plugins.py
+++ b/shoop_tests/themes/classic_gray/test_plugins.py
@@ -1,0 +1,31 @@
+# -*- coding: utf-8 -*-
+# This file is part of Shoop.
+#
+# Copyright (c) 2012-2016, Shoop Ltd. All rights reserved.
+#
+# This source code is licensed under the AGPLv3 license found in the
+# LICENSE file in the root directory of this source tree.
+import pytest
+
+from shoop.core.models import Product, ProductCrossSell, StockBehavior
+from shoop.testing.factories import create_product, get_default_shop
+from shoop.themes.classic_gray.plugins import ProductCrossSellsPlugin
+from shoop_tests.front.fixtures import get_jinja_context
+
+
+@pytest.mark.django_db
+def test_cross_sell_plugin_renders():
+    """
+    Test that the plugin renders a product
+    """
+    shop = get_default_shop()
+    product = create_product("test-sku", shop=shop, stock_behavior=StockBehavior.UNSTOCKED)
+    computed = create_product("test-computed-sku", shop=shop, stock_behavior=StockBehavior.UNSTOCKED)
+    type ="computed"
+
+    ProductCrossSell.objects.create(product1=product, product2=computed, type=type)
+    assert ProductCrossSell.objects.filter(product1=product, type=type).count() == 1
+
+    context = get_jinja_context(product=product)
+    rendered  = ProductCrossSellsPlugin({"type": type}).render(context)
+    assert computed.sku in rendered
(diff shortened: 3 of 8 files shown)


shoop/themes/classic_gray/templates/classic_gray/shoop/front/macros.jinja
```diff
@@ -34,8 +34,8 @@
     </div>
 {% endmacro %}
 
-{% macro render_cross_sell_products(product, relation_type="", title="") %}
-    {%- set cross_sell_products = shoop.product.get_product_cross_sells(product, relation_type) %}
+{% macro render_cross_sell_products(product, relation_type="", title="", count=4) %}
```
tulimaki: I don't think that render_croll_sell_products should be here. It should probably be under cross_sells_plugin.jinja.

shoop/themes/classic_gray/plugins.py
```diff
@@ -46,3 +46,32 @@ def get_context_data(self, context):
...
+    template_name = "classic_gray/cross_sells_plugin.jinja"
+    required_context_variables = ["product"]
+    fields = [
+        ("title", TranslatableField(label=_("Title"), required=False, initial="")),
+        ("type", forms.ChoiceField(label=_("Type"), choices=[
+            ("related", "Related"),
+            ("recommended", "Recommended"),
+            ("computed", "Computed"),
+        ], initial="newest")),
+        ("count", forms.IntegerField(label=_("Count"), min_value=1, initial=4)),
```
tulimaki: Should we have the orderable_only option for plugin too?

shoop_tests/themes/classic_gray/test_plugins.py
```diff
@@ -0,0 +1,76 @@
...
+from bs4 import BeautifulSoup
+
+from shoop.core.models import Product, ProductCrossSell
+from shoop.testing.factories import create_product, get_default_shop
+from shoop.themes.classic_gray import plugins
+from shoop_tests.front.fixtures import get_jinja_context
+
+
+@pytest.mark.django_db
+def test_cross_sell_plugin_type():
```
tulimaki: This test contains a lot of duplicate code which could be simplified with few functions.

shoop/themes/classic_gray/plugins.py
shawnadelic: We can, we would just need to add that to the get_product_cross_sells template helper, and so we'd probably want to replace list_visible with _get_list_products, similar to what we did with get_best_selling_products, etc.

Looking at what's currently in get_product_cross_sells, there shouldn't be an issues I don't think

shoop/front/template_helpers/general.py
```diff
@@ -16,41 +16,50 @@
...
         language=get_language(),
     )
 
+    if filter_dict:
+        products_qs = products_qs.filter(**filter_dict)
+    if ordering:
+        products_qs = products_qs.order_by(ordering)
+    if orderable_only:
+        products = []
+        for product in products_qs.all():
```
akx: This should maybe use `.iterator()`... otherwise you might be loading all list-visible products into Django memory, just to return a few of them.

shoop/front/template_helpers/general.py
```diff
@@ -16,41 +16,50 @@
 from shoop.utils.translation import cache_translations_for_tree
 
 
-def _get_list_products(request):
-    return Product.objects.list_visible(
+def _get_list_products(context, n_products, ordering=None, filter_dict=None, orderable_only=True):
```
akx: Docstring missing.

Why does this require `context` just to extract `request` out of it? :)

shoop/themes/classic_gray/plugins.py
```diff
@@ -26,18 +26,22 @@ class ProductHighlightPlugin(TemplatedPlugin):
...
+        ("orderable_only", forms.BooleanField(label=_("Only show in-stock and orderable items"),
+                                              initial=True,
+                                              required=False))
     ]
 
     def get_context_data(self, context):
         type = self.config.get("type", "newest")
         count = self.config.get("count", 4)
+        orderable_only = self.config.get("orderable_only", True)
```
akx: If the default value is True, this will change the behavior of shops currently using the plugin. Is that supposed to happen?

akx:
Hm, wait, what's #365 's relation to this PR?

shoop/front/template_helpers/general.py
shawnadelic: Can add a docstring, and you're right ``request`` should be sufficient

shoop/themes/classic_gray/plugins.py
shawnadelic: I believe @tulimaki felt that True would be the preferred default, but I'm sure you could make a case either way. The behavior can also be changed through Xtheme editor if we want to keep it True.

shawnadelic:
#365 changes the _get_list_products function so we can filter by orderability, so it's required in order to add orderability to the cross sells plugin.

shoop/themes/classic_gray/plugins.py
```diff
@@ -50,3 +50,37 @@ def get_context_data(self, context):
...
+    name = _("Product Cross Sells")
+    template_name = "classic_gray/cross_sells_plugin.jinja"
+    required_context_variables = ["product"]
+    fields = [
+        ("title", TranslatableField(label=_("Title"), required=False, initial="")),
+        ("type", forms.ChoiceField(label=_("Type"), choices=[
+            ("related", "Related"),
+            ("recommended", "Recommended"),
+            ("computed", "Computed"),
+        ], initial="newest")),
```
tulimaki: "newest" is not good default for this plugin

shoop/themes/classic_gray/plugins.py
```diff
@@ -50,3 +50,37 @@ def get_context_data(self, context):
...
+        ("orderable_only", forms.BooleanField(label=_("Only show in-stock and orderable items"),
+                                              initial=True,
+                                              required=False))
+    ]
+
+    def get_context_data(self, context):
+        count = self.config.get("count", 4)
+        product = context.get("product", None)
+        orderable_only = self.config.get("orderable_only", True)
+        type = self.config.get("type", "newest")
```
tulimaki: Cahnge the default type.

shoop_tests/themes/classic_gray/test_plugins.py
```diff
@@ -0,0 +1,67 @@
...
+    create_product, get_default_shop, get_default_supplier
+)
+from shoop.themes.classic_gray.plugins import ProductCrossSellsPlugin
+from shoop_tests.front.fixtures import get_jinja_context
+
+
+def _get_class_count(html, html_class):
+    parsed = BeautifulSoup(html)
+    return len(parsed.find_all(class_=html_class))
+
```
tulimaki: Missing blank line

shoop_tests/themes/classic_gray/test_plugins.py
```diff
@@ -0,0 +1,67 @@
...
+    # Create cross sell products and relations in different quantities
+    for type, count in type_counts:
+        _create_cross_sell_products(product, shop, type, count)
+        assert ProductCrossSell.objects.filter(product1=product, type=type).count() == count
+
+    # Make sure quantites returned by plugin match
+    for type, count in type_counts:
+        rendered_plugin = ProductCrossSellsPlugin({"type": type}).render(context)
+        assert _get_class_count(rendered_plugin, "product-box") == count
+
```
tulimaki: Missing blank line

shoop_tests/themes/classic_gray/test_plugins.py
```diff
@@ -0,0 +1,67 @@
...
+def test_cross_sell_plugin_count():
+    shop = get_default_shop()
+    product = create_product("test-sku", shop=shop, stock_behavior=StockBehavior.UNSTOCKED)
+    context = get_jinja_context(product=product)
+
+    _create_cross_sell_products(product, shop, "related", 5)
+    assert ProductCrossSell.objects.filter(product1=product, type="related").count() == 5
+
+    rendered_plugin = ProductCrossSellsPlugin({"type": "related", "count": 3}).render(context)
+    assert _get_class_count(rendered_plugin, "product-box") == 3
```
tulimaki: Instead of counting classes in rendered plugin you should probably just test ``shoop.product.get_product_cross_sells`` to return correct amount of products.

shoop_tests/themes/classic_gray/test_plugins.py
```diff
@@ -0,0 +1,67 @@
...
+
+    # Create cross sell products and relations in different quantities
+    for type, count in type_counts:
+        _create_cross_sell_products(product, shop, type, count)
+        assert ProductCrossSell.objects.filter(product1=product, type=type).count() == count
+
+    # Make sure quantites returned by plugin match
+    for type, count in type_counts:
+        rendered_plugin = ProductCrossSellsPlugin({"type": type}).render(context)
+        assert _get_class_count(rendered_plugin, "product-box") == count
```
tulimaki: Instead of counting classes in rendered plugin you should probably just test ``shoop.product.get_product_cross_sells`` to return correct amount of products.

tulimaki:
Thanks!

PR #370, opened 2016-03-05
Make incl. and excl. in price __str__ translatable

jaywink:
Noticed that the `Price.__str__` gives out an untranslated string.


suutari-ai:
Prices for should not be formatted with `__str__` but with `format_money` in Python code and with `{{ field | money }}` in Jinja templates. The `__str__` exists just for debugging convenience, e.g. for printing money amounts in log messages, debug prints or tracebacks and it's untranslated intentionally.

PR #371, opened 2016-03-07
XTheme: Show edit link in model choice fields (SHOOP-2170)

tulimaki:
Add custom XThemeModelChoiceField and XThemeModelChoiceWidget for
showing edit link to admin

Refs SHOOP-1892 / SHOOP-2170

### DIFF
--- a/shoop/xtheme/plugins/widgets.py
+++ b/shoop/xtheme/plugins/widgets.py
@@ -6,6 +6,7 @@
 # This source code is licensed under the AGPLv3 license found in the
 # LICENSE file in the root directory of this source tree.
 from django import forms
+from django.template.loader import render_to_string
 from django.utils.safestring import mark_safe
 from django.utils.translation import ugettext as _
 
@@ -90,3 +91,21 @@ class TranslatableFieldWidget(forms.Widget):
         else:
             out_dict[FALLBACK_LANGUAGE_CODE] = value
         return out_dict
+
+
+class XThemeModelChoiceWidget(forms.Select):
+    def render(self, name, value, attrs=None, choices=()):
+        return mark_safe(
+            render_to_string("shoop/xtheme/_model_widget.jinja", {
+                "name": name,
+                "selected_value": value,
+                "objects": self.choices,
+            })
+        )
+
+
+class XThemeModelChoiceField(forms.ModelChoiceField):
+    widget = XThemeModelChoiceWidget
+
+    def label_from_instance(self, obj):
+        return obj
--- a/shoop/xtheme/templates/shoop/xtheme/_model_widget.jinja
+++ b/shoop/xtheme/templates/shoop/xtheme/_model_widget.jinja
@@ -1,0 +1,11 @@
+<select id="id_{{ name }}" name="{{ name }}" class="xtheme-model-choice-widget">
+    {% for pk, object in objects %}
+        {% set url = (shoop_admin.model_url(object, "detail") if pk else "") %}
+        <option
+            value="{{ pk }}"
+            {% if pk == selected_value %}selected{% endif %}
+            data-admin-url="{{ url }}"
+         >{{ object }}</option>
+    {% endfor %}
+</select>
+<div id="extra_for_id_{{ name }}"></div>
--- a/doc/changelog.rst
+++ b/doc/changelog.rst
@@ -135,6 +135,7 @@ Front
 Xtheme
 ~~~~~~
 
+- Add custom XThemeModelChoiceField to show admin URLs
 - Enhance default text plugin editor to remarkable markdown editor
 - Add support for global/multi-view placeholders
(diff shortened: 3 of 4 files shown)


shoop/xtheme/plugins/widgets.py
```diff
@@ -90,3 +91,38 @@ def decompress(self, value):
...
             out_dict[FALLBACK_LANGUAGE_CODE] = value
         return out_dict
+
+
+class XThemeModelChoiceWidget(forms.Select):
+    def render(self, name, value, attrs=None, choices=()):
+        return mark_safe(
+            render_to_string("shoop/xtheme/_model_widget.jinja", {
+                "name": name,
+                "selected_object": [object for pk, object in self.choices if pk == int(value)][0] if value else None,
```
akx: If there are 15,000 products, this will iterate over all of them, then pick one. (I believe we have a helper called `first()` somewhere...)

Also, there's nothing guaranteeing pks are integers; comparing as strings would be safer.

It would really be better, all in all, that the ChoiceField passed the selected object itself down to the widget for rendering.

shoop/xtheme/plugins/widgets.py
```diff
@@ -90,3 +91,38 @@ def decompress(self, value):
...
+            **kwargs
+        )
+
+    def _get_custom_choices(self):
+        return XThemeModelChoiceIterator(self)
+
+    choices = property(_get_custom_choices, forms.ChoiceField._set_choices)
+
+
+class XThemeModelChoiceIterator(forms.models.ModelChoiceIterator):
```
akx: Why is this necessary?

shoop/xtheme/plugins/widgets.py
```diff
@@ -90,3 +91,38 @@ def decompress(self, value):
...
+            render_to_string("shoop/xtheme/_model_widget.jinja", {
+                "name": name,
+                "selected_object": [object for pk, object in self.choices if pk == int(value)][0] if value else None,
+                "objects": self.choices,
+            })
+        )
+
+
+class XThemeModelChoiceField(forms.ModelChoiceField):
+    def __init__(self, queryset, empty_label="---------", required=True, label=None, initial=None, *args, **kwargs):
```
akx: Would probably be better to just `*args, **kwargs` here (or really just `**kwargs` should be enough given the use).

Pikkupomo:
wrong ref @tulimaki?

shoop/xtheme/plugins/widgets.py
akx: ... Oh, wait, you're only using this to render `selected`?

Why not just pass `value` down to the template context and do `if value == object.pk` in the render loop there?

shoop/xtheme/templates/shoop/xtheme/_model_widget.jinja
```diff
@@ -0,0 +1,20 @@
+<select id="id_{{ name }}" name="{{ name }}" class="xtheme-model-choice-widget">
+    {% for pk, object in objects %}
+        {% if pk %}
+            {% set url = shoop_admin.model_url(object, "detail") %}
```
akx: Jinja supports ternary expressions; `{% set url = (shoop_admin.model_url(...) if pk else "") %}` is arguably clearer.

shoop/xtheme/plugins/widgets.py
tulimaki: It's not I think that overriding label_from_instance could be better solution for this.

shoop/xtheme/plugins/widgets.py
tulimaki: correct

shoop/xtheme/plugins/widgets.py
```diff
@@ -90,3 +91,27 @@ def decompress(self, value):
...
+                "objects": self.choices,
+            })
+        )
+
+
+class XThemeModelChoiceField(forms.ModelChoiceField):
+    widget = XThemeModelChoiceWidget
+
+    def __init__(self, **kwargs):
+        super(XThemeModelChoiceField, self).__init__(**kwargs)
```
akx: Unnecessarily overridden init?

tulimaki: yes

shoop/xtheme/plugins/widgets.py
```diff
@@ -90,3 +91,24 @@ def decompress(self, value):
...
+            })
+        )
+
+
+class XThemeModelChoiceField(forms.ModelChoiceField):
+    widget = XThemeModelChoiceWidget
+
+    def label_from_instance(self, obj):
+        """
+        Use objects in choices instead of strings
```
akx: This is a comment, not a docstring.

tulimaki:
retest this please

PR #351, opened 2016-02-10
Order signals and allow methods admin order creator

tulimaki:
Refs SHOOP-2114 / SHOOP-1973

PR #372, opened 2016-03-08
Move order_creator_finished signal etc. (SHOOP-2114)

suutari-ai:
* Move order_creator_finished signal from front to
  core.order_creator.signals
* Admin: Make all enabled shipping and payment methods available in
  order creation
* OrderCreator: Make request argument optional
  *  Remove request from OrderCreator init calls
  * Admin: Order creating: Populate Order.ip_address
* factories: Depend less on request
* Ensure pricing context gets shop and customer

First two of those commits are from @tulimaki (I only fixed the Refs format and rephrased change log item from "Enable all methods for order creator" to "Make all enabled shipping and payment methods available in order creator". The rest commits are related clean-ups/refactoring that we should do while moving the signal, since they fix a long standing TODO item and improves OrderCreator by making it request independent.

Replaces PR #351

### DIFF
--- a/shoop/core/order_creator/_creator.py
+++ b/shoop/core/order_creator/_creator.py
@@ -13,9 +13,9 @@ import six
 from django.utils.encoding import force_text
 
 from shoop.core.models import Order, OrderLine, OrderLineType
+from shoop.core.order_creator.signals import order_creator_finished
 from shoop.core.shortcuts import update_order_line_from_product
 from shoop.core.utils.users import real_user_or_none
-from shoop.front.signals import order_creator_finished
 from shoop.utils.numbers import bankers_round
 
 from ._source_modifier import get_order_source_modifier_modules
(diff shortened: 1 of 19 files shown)


shoop/core/order_creator/_creator.py
```diff
@@ -23,8 +23,25 @@
...
+          for initializing remote address: If request is given and
+          `remote_address` is not, then `remote_address` will
+          initialized to ``request.META.get("REMOTE_ADDRESS")``.
+        :type remote_address: str
+        :param remote_address:
+          Remote IP address as string (IPv4 or IPv6).  If given, will be
+          stored to `~shoop.core.models.Order.ip_address` field of
+          created orders.
+        """
+        self.remote_address = (
```
akx: Hmm. This should really be a property of the order source, not the creator.

suutari-ai: Now that you bring that up... yeah.. I agree. Didn't occur to me... I'll move it.

suutari-ai: It's now saved to OrderSource.ip_address and OrderCreator shouldn't be created with a request anymore.

tulimaki:
@suutari-ai Thanks!

PR #373, opened 2016-03-10
Couple tiny improvements for default ContactGroups (SHOOP-2141)

suutari-ai:
Fix couple things that popped into my mind while reviewing PR #358:

 * We shouldn't really define `__str__` for any of our models unless
   there is a very good reason, but rather make the version in
   TranslatableShoopModel work for most of the cases.
 * Rename `Contact.get_default_contact_group` to
   `Contact.get_default_group` and make it classmethod.
 * Add tests for ContactGroup str and repr.


tulimaki:
Thanks!

PR #375, opened 2016-03-14
Social Media: Add social media app and xtheme plugin (SHOOP-1972)

shawnadelic:
Add social media app and social media admin module for administration of social media links. Each link has a link type (Facebook, Twitter, etc) that maps it to a font-awesome icon class (but can be overridden).
    
Also add an xtheme plugin in ``classic_gray`` to display social links on the shop front.

image

Also add a generic snippets plugin used to create generic integrations which primarily require only including some JS/CSS resources and snippets of code.

Refs SHOOP-1972


shoop/themes/classic_gray/templates/classic_gray/plugins/social_media_links.jinja
```diff
@@ -0,0 +1,11 @@
+<section class="social-media-links">
+    <div class="row{% if alignment %} text-{{ alignment }}{% endif %}">
+        {% if topic %}<h1>{{ topic }}</h1>{% endif %}
```
juhakujala: This should be `<h2>` instead of `<h1>`. We should not be using more than one `<h1>` in one page and usually that is the page title.

shoop/themes/classic_gray/plugins/social_media_links.py
```diff
@@ -0,0 +1,46 @@
...
+class SocialMediaLinksPlugin(TemplatedPlugin):
+    identifier = "classic_gray.social_media_links"
+    name = _("Social Media Links")
+    template_name = "classic_gray/plugins/social_media_links.jinja"
+    fields = [
+        ("topic", TranslatableField(label=_("Topic"), required=False, initial="")),
+        ("text", TranslatableField(label=_("Text"), required=False, initial="")),
+        ("icon_size", forms.ChoiceField(label=_("Icon Size"), required=False, choices=[
+            ("", "Default"),
+            ("large", "Large"),
```
juhakujala: If I'm right, `fa-large` doesn't exist.

shoop/social_media/models.py
```diff
@@ -0,0 +1,75 @@
...
+        TWITTER = _("Twitter")
+        VIMEO = _("Vimeo")
+        VINE = _("Vine")
+        YELP = _("Yelp")
+        YOUTUBE = _("Youtube")
+
+
+@python_2_unicode_compatible
+class SocialMediaLink(ShoopModel):
+    type = EnumIntegerField(
```
akx: This can't be an EnumIntegerField if the enum values are strings.

shoop/social_media/models.py
```diff
@@ -0,0 +1,75 @@
...
+
+    class Meta:
+        verbose_name = _(u"Social Media Link")
+        verbose_name_plural = _(u"Social Media Links")
+        ordering = ["ordering"]
+
+    @property
+    def icon_class_name(self):
+        value = self.type.value
+        if value in self.icon_overrides:
```
akx: No need for the if/access; simply `return self.icon_overrides.get(value, value)` is what you want.

shoop/social_media/templates/shoop/admin/social_media/_edit_base_form.jinja
```diff
@@ -0,0 +1,9 @@
+{% from "shoop/admin/macros/general.jinja" import content_block %}
+{% from "shoop/admin/macros/multilanguage.jinja" import render_monolingual_fields %}
+
+{% set base_form = form["base"] %}
+{% set social_media_link = base_form.instance %}
```
akx: Unused.

shoop/themes/classic_gray/plugins/social_media_links.py
shawnadelic: Oops, looks like it's, fa-lg

shoop/xtheme/__init__.py
```diff
@@ -11,8 +11,10 @@
     get_current_theme, get_theme_by_identifier, set_current_theme, Theme
 )
 from .plugins._base import Plugin, templated_plugin_factory, TemplatedPlugin
+from .plugins.snippet import GenericSnippetsPlugin
```
tulimaki: This should be .plugins.snippets

shoop_tests/xtheme/test_plugins.py
```diff
@@ -0,0 +1,33 @@
...
+# This file is part of Shoop.
+#
+# Copyright (c) 2012-2016, Shoop Ltd. All rights reserved.
+#
+# This source code is licensed under the AGPLv3 license found in the
+# LICENSE file in the root directory of this source tree.
+import six
+
+from shoop.xtheme import Plugin, resources
+from shoop.xtheme.plugins.snippets import GenericSnippetsPlugin
```
tulimaki: I think you should import from shoop.xtheme.plugins instead of snippets.

tulimaki:
I think the correct ref for this is SHOOP-1972?

shoop/themes/classic_gray/plugins/__init__.py
```diff
@@ -0,0 +1,16 @@
+# This file is part of Shoop.
+#
+# Copyright (c) 2012-2016, Shoop Ltd. All rights reserved.
+#
+# This source code is licensed under the AGPLv3 license found in the
+# LICENSE file in the root directory of this source tree.
+
+from product_highlight import ProductHighlightPlugin
```
tulimaki: You should use ``from .product_highlight import ProductHighlightPlugin`` and same goes with the next to imports in this file.

shoop_tests/themes/classic_gray/test_plugins.py
```diff
@@ -0,0 +1,18 @@
+# -*- coding: utf-8 -*-
```
tulimaki: Missing the license header...

shoop/social_media/models.py
```diff
@@ -0,0 +1,80 @@
...
+class SocialMediaLink(ShoopModel):
+    type = EnumIntegerField(
+        SocialMediaLinkType, verbose_name=_("type"), default=SocialMediaLinkType.FACEBOOK,
+        help_text=_("Type of social media link.")
+    )
+    url = models.URLField(
+        verbose_name=_(u"URL"), max_length=200, help_text=_("Social media link URL.")
+    )
+    ordering = models.IntegerField(verbose_name=_(u"ordering"), default=0, blank=True, null=True)
+    icon_classes = {
```
tulimaki: Not very usefull. Is there a reason you can't use EnumField instead of EnumIntegerField and use these strings as enum values?

shoop_tests/social_media/test_social_media_links.py
```diff
@@ -0,0 +1,53 @@
...
+        SocialMediaLink.objects.create(type=None, url=url)
+
+
+@pytest.mark.django_db
+def test_url_required():
+    type = SocialMediaLinkType.TWITTER
+
+    with pytest.raises(IntegrityError) as error_info:
+        SocialMediaLink.objects.create(type=type, url=None)
+
```
tulimaki: Missing blank line

shoop/themes/classic_gray/plugins/social_media_links.py
```diff
@@ -0,0 +1,46 @@
...
+            ("4x", "4x"),
+            ("5x", "5x"),
+        ], initial="")),
+        ("alignment", forms.ChoiceField(label=_("Alignment"), required=False, choices=[
+            ("", "Default"),
+            ("left", "Left"),
+            ("center", "Center"),
+            ("right", "Right"),
+        ], initial="")),
+    ]
```
tulimaki: I think you should have ModelMultipleChoiceField here. Other solution would be to add enabled attribute to ``SocialMediaLink``s. Point is that we don't neccessary want to show every single link we have stored in database.

shoop/social_media/models.py
akx: Not a good idea. "facebook-square" isn't something you want to store in the database for "facebook".

tulimaki: Ok then! :lollipop:

tulimaki:
I think this is now in https://github.com/shoopio/shoop/pull/379 and https://github.com/shoopio/shoop/pull/380 so will close this for now at least.

PR #377, opened 2016-03-17
Dynamic taxness for price display (SHOOP-2166)

suutari-ai:
See the commit messages.

### DIFF
--- a/shoop/core/models/_contacts.py
+++ b/shoop/core/models/_contacts.py
@@ -17,24 +17,46 @@ from polymorphic.models import PolymorphicModel
 from timezone_field.fields import TimeZoneField
 
 from shoop.core.fields import InternalIdentifierField, LanguageField
+from shoop.core.pricing import PriceDisplayOptions
 from shoop.core.utils.name_mixin import NameMixin
 
 from ._base import TranslatableShoopModel
 from ._taxes import CustomerTaxGroup
 
 
+class ContactGroupQuerySet(models.QuerySet):
+    def with_price_display_options(self):
+        return self.filter(
+            models.Q(show_prices_including_taxes__isnull=False) |
+            models.Q(hide_prices__isnull=False))
+
+
 class ContactGroup(TranslatableShoopModel):
     identifier = InternalIdentifierField(unique=True)
     members = models.ManyToManyField("Contact", related_name="groups", verbose_name=_('members'), blank=True)
     show_pricing = models.BooleanField(verbose_name=_('show as pricing option'), default=True)
+    show_prices_including_taxes = models.NullBooleanField(
+        default=None, null=True, blank=True,
+        verbose_name=_("show prices including taxes"))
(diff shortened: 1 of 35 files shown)


shoop/core/models/_contacts.py
```diff
@@ -91,6 +113,12 @@ def __init__(self, *args, **kwargs):
             kwargs.setdefault("tax_group", self.default_tax_group_getter())
         super(Contact, self).__init__(*args, **kwargs)
 
+    def get_price_display_options(self):
+        group_with_options = self.groups.with_price_display_options().first()
```
tulimaki: I think first() is not correct here. IMO if contact has multiple contact groups set with pricing options, then default group should determine the options.

tulimaki:
I think the ``PriceDisplayOptions`` is not a good naming here. It's more like a price display mode and the options is to show_prices_including_taxes and hide_prices. Then based on these two options actual pricing mode is set.

tulimaki:
Can you add these new ``ContactGroup`` fields into admin also so these can be actually set from admin? I thnk the ``ContactGroupBaseForm`` is good place for these.

shoop/core/models/_contacts.py
suutari-ai: Sounds feasible. I will implement that.

suutari-ai:
> Can you add these new ContactGroup fields into admin also so these can be actually set from admin? I thnk the ContactGroupBaseForm is good place for these.

That's not in scope of SHOOP-2166, but rather SHOOP-2143, so let's do that in another PR. (Yeah, I know, this PR already has some parts of SHOOP-2143: the Classic Gray changes and contact group additions, but that's only because it was hard to check usability of these new filters without trying them with the templates...)

suutari-ai:
> I think the PriceDisplayOptions is not a good naming here. It's more like a price display mode and the options is to show_prices_including_taxes and hide_prices. Then based on these two options actual pricing mode is set.

I disagree. PriceDisplayMode sounds more like an enumeration, but the class is a collection of options rather than enumeration. Although currently there is just four different ways to show prices ((1) don't show them, (2) show them as taxful, (3) show them as taxless, (4) show them with original taxness), it might very well be that in the future PriceDisplayOptions class may have things like `taxless_price_template = '{price} (ALV 0%)'`, or `hide_base_prices = True  # to show discounted prices as if they were not discounted`, etc. I stand with the naming.

tulimaki:
You should add some tests for these if possible. At least ``shoop/core/utils/prices.py`` should have tests.

suutari-ai:
PR updated. Changes since Friday:
* Updated `Contact.get_price_display_options` as requested.
* Added tests.
* Moved PriceDisplayOptions to shoop.core.pricing.

suutari-ai:
retest this please

tulimaki:
@suutari-ai thank you so much!

PR #379, opened 2016-03-19
Classic Gray: Add social media xtheme plugin

shawnadelic:
Add an xtheme plugin in classic_gray to display social media links on the
shop front.

This is a refactor of PR #375.

Refs SHOOP-1972

### DIFF
--- a/doc/changelog.rst
+++ b/doc/changelog.rst
@@ -72,6 +72,7 @@ Classic Gray Theme
 - Show error messages while adding products to basket
 - Add Coupon use possibility to basket page
 - Add option to only show orderable products to highlights plugin
+- Add Xtheme plugin to display social media links on shop front
 
 Simple Supplier
 ~~~~~~~~~~~~~~~
--- a/shoop/themes/classic_gray/static_src/less/classic-gray/social-media-links.less
+++ b/shoop/themes/classic_gray/static_src/less/classic-gray/social-media-links.less
@@ -1,0 +1,12 @@
+.social-media-links ul {
+    padding-left: 0;
+
+    li {
+        display: inline;
+        list-style-type: none;
+    }
+
+    li + li {
+        margin-left: 10px;
+    }
+}
--- a/shoop/themes/classic_gray/static_src/less/style.less
+++ b/shoop/themes/classic_gray/static_src/less/style.less
@@ -77,3 +77,4 @@
 @import "classic-gray/loader.less";
 @import "classic-gray/user-account.less";
 @import "classic-gray/carousel.less";
+@import "classic-gray/social-media-links.less";
--- a/shoop/xtheme/__init__.py
+++ b/shoop/xtheme/__init__.py
@@ -33,6 +33,7 @@ class XThemeAppConfig(AppConfig):
         "xtheme_plugin": [
             "shoop.xtheme.plugins.text:TextPlugin",
             "shoop.xtheme.plugins.snippets:SnippetsPlugin",
+            "shoop.xtheme.plugins.social_media_links:SocialMediaLinksPlugin",
         ],
         "admin_module": [
             "shoop.xtheme.admin_module:XthemeAdminModule"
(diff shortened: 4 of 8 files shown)


shoop/themes/classic_gray/__init__.py
```diff
@@ -74,7 +74,8 @@ class ClassicGrayThemeAppConfig(AppConfig):
         "xtheme_plugin": [
             "shoop.themes.classic_gray.plugins:ProductHighlightPlugin",
             "shoop.themes.classic_gray.plugins:ProductCrossSellsPlugin",
-        ],
+            "shoop.themes.classic_gray.plugins:SocialMediaLinksPlugin",
+        ]
```
suutari-ai: Why the `],` -> `]` change? :)

shoop/themes/classic_gray/forms.py
```diff
@@ -0,0 +1,64 @@
...
+
+        Also adds an ordering field for each link type to change display order.
+        """
+        icon_classes = self.plugin.icon_classes
+        links = self.plugin.config.get("links", {})
+        for name, icon_class in sorted(icon_classes.items()):
+            url = links[name]["url"] if name in links else ""
+            ordering = links[name]["ordering"] if name in links else None
+            self.fields[name] = forms.URLField(
+                label=_(name), required=False,
```
suutari-ai: How is `_(name)` supposed to work? It doesn't. `_("message")` has two functions: (1) It translates the message from the source string to currently active language when called and (2) it marks the message for translation. Because of 2, the argument must always be a string literal, never a variable.

=> Move the underscore calls as part of the dict definition.

shoop/themes/classic_gray/plugins.py
```diff
@@ -84,3 +86,79 @@ def get_context_data(self, context):
...
+        "Instagram": "instagram",
+        "Linkedin": "linkedin-square",
+        "Pinterest": "pinterest",
+        "Tumbler": "tumblr",
+        "Twitter": "twitter",
+        "Vimeo": "vimeo",
+        "Vine": "vine",
+        "Yelp": "yelp",
+        "Youtube": "youtube",
+    }
```
suutari-ai: I don't see why this should be a dict. IMHO list of tuples would be better. Also, since there is a need for translated string, add it too.
```python
icon_data = [
    # (name, icon_class, label)
    ("Facebook", "facebook-square", _("Facebook")),
    ("Flickr", "flickr", _("Flickr")),
    ...
```

suutari-ai: Actually... is there really any need for translated string? I don't think any of these trade marked names will be different on any other language... maybe the translating should just be dropped?

shoop/themes/classic_gray/templates/classic_gray/social_media_links.jinja
```diff
@@ -0,0 +1,13 @@
+{% if links %}
+<section class="social-media-links">
+    <div class="row{% if alignment %} text-{{ alignment }}{% endif %}">
+        {% if topic %}<h2>{{ topic }}</h2>{% endif %}
+        {% if text %}<p>{{ text }}</p>{% endif %}
+        <ul>
+            {% for ordering, class, url in links %}
+                    <li><a href="{{ url }}"><i class="fa fa-{{ class }}{% if icon_size %} fa-{{ icon_size }}{% endif %}"></i></a></li>
```
suutari-ai: Indented too much.

shoop/themes/classic_gray/__init__.py
shawnadelic: Not intentional, I don't think :)

shoop/themes/classic_gray/plugins.py
shawnadelic: I ended up leaving the dictionary and getting rid of the translations, mostly due to the fact that I had to add an extra iteration through that every time I returned an icon class, which made it even harder to read. I don't think it should be an issue as-is since I got rid of the translations, but let me know if you disagree.

suutari-ai: It's good now.

PR #380, opened 2016-03-19
Xtheme: Add generic snippets plugin for simple integrations

shawnadelic:
Add a generic simple snippets plugin for doing simple integrations that
primarily require adding JavaScript or CSS resources to the page.

This is related to the PR #379, which is a refactor of PR #375.

Refs SHOOP-1972

### DIFF
--- a/doc/changelog.rst
+++ b/doc/changelog.rst
@@ -56,6 +56,7 @@ Front
 Xtheme
 ~~~~~~
 
+- Add generic snippets plugin for doing simple integrations
 
 Classic Gray Theme
 ~~~~~~~~~~~~~~~~~~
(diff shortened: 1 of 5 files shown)


akx:
This really sounds like a plugin that could live in Xtheme itself, not only in classic_gray.

shoop/xtheme/plugins/general.py
```diff
@@ -0,0 +1,33 @@
...
+from shoop.xtheme import Plugin, resources
+from shoop.xtheme.resources import add_resource, InlineMarkupResource
+
+
+class GenericSnippetsPlugin(Plugin):
+    """
+    Simple plugin class for including snippets and resources on the page, mostly for simple integrations.
+    """
+    fields = [
+        ("in_place", forms.CharField(label=_("In-Place Snippet"), widget=forms.Textarea, required=False)),
```
tulimaki: You probably need to add these to translation files?

PR #384, opened 2016-03-23
Always return non-None for OrderSource contacts (SHOOP-2246)

suutari-ai:
The customer, orderer and creator properties of OrderSource (or its
subclass BaseBasket) are used quite carelessly in some places, e.g. in
ContactGroupBasketCondition.matches.  Previously these properties could
have been None and then code like `basket.customer.groups.all()` would
fail to AttributeError.

Make OrderSource always return non-None objects for those three
attributes to fix ContactGroupBasketCondition.matches and possibly other
similar issues.

Add a test case for the issue in ContactGroupBasketCondition.matches.

Refs SHOOP-1981/SHOOP-2246

### DIFF
--- a/shoop_tests/campaigns/test_basket_contact_conditions.py
+++ b/shoop_tests/campaigns/test_basket_contact_conditions.py
@@ -12,7 +12,7 @@ from shoop.campaigns.models.campaigns import BasketCampaign
 from shoop.campaigns.models.basket_conditions import (
     ContactBasketCondition, ContactGroupBasketCondition
 )
-from shoop.core.models import AnonymousContact
+from shoop.core.models import AnonymousContact, Shop
 from shoop.front.basket import get_basket
 from shoop.testing.factories import (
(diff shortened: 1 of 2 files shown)


shoop_tests/campaigns/test_basket_contact_conditions.py
```diff
@@ -102,6 +102,16 @@ def test_group_basket_condition_with_anonymous_contact(rf):
 
 
 @pytest.mark.django_db
+def test_contact_group_basket_condition_with_none(rf):
+    request = rf.get("/")
+    request.shop = Shop()
+    basket = get_basket(request)
```
tulimaki: Can you assert the ``basket`.customer is None`` here so we can be sure that there is no customers?

suutari-ai: No, I can't, because `basket.customer` will actually return AnonymousContact after this fix.

PR #389, opened 2016-03-25
Compute product  bought with relations (SHOOP-2245)

tulimaki:
* Front: Fix for ``get_visible_products`` template helper
* Core: Add bought with relation to ``ProductCrossSellType``
* Classic Gray: Update cross-sells plugin
* Front: Add management command to generate bought with relations

Refs SHOOP-2245

### DIFF
--- a/doc/changelog.rst
+++ b/doc/changelog.rst
@@ -55,6 +55,7 @@ Admin
 Front
 ~~~~~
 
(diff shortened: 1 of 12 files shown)


shawnadelic:
Looked this over and didn't see anything. So long as that enum is working, everything seems good I think

PR #391, opened 2016-03-29
Classic Gray: Replace footer with xtheme plugins

shawnadelic:
Replace theme footer settings with xtheme plugins.

Refs SHOOP-2179


suutari-ai:
Good job, @shawnadelic!

PR #393, opened 2016-03-29
Fix choices of ProductCrossSellsPlugin type field (SHOOP-2305)

suutari-ai:
The type field of ProductCrossSellsPlugin used to use translated string as
field values as choices were given as (enum, string) pairs, where the
enum part was converted to translated string implicitly.  Fix this by
not defining the form field at all, but rather construct the form field
from the model field with `ProductCrossSell.type.field.formfield()`.
This will use the enum values as choice values.

Database may have old translated string values, use RELATED for those.

Also improve error handling of get_product_cross_sells (the second commit).


suutari-ai:
retest this please

PR #394, opened 2016-03-29
Xtheme: Fix bootstrap less dependency

shawnadelic:
Add local copy of less variables from admin to fix problems building
xtheme resources when admin resources have not been compiled yet.

This also allows for us to change less variables in xtheme styles.

Refs SHOOP-2287

### DIFF
--- a/shoop/xtheme/static_src/editor/style.less
+++ b/shoop/xtheme/static_src/editor/style.less
@@ -169,11 +169,10 @@ form {
     }
 }
 
-@import "../vendor/bower_components/bootstrap/less/mixins.less";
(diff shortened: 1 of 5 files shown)


shoop/xtheme/static_src/editor/style.less
```diff
@@ -169,11 +169,12 @@ form {
     }
 }
 
+@import "../vendor/bower_components/bootstrap/less/variables.less";
+@import "../../../admin/static_src/base/less/variables.less";
```
tulimaki: You shouldn't import any variables from admin. Also the vendor variables should be imported inside xhtheme variables.

PR #395, opened 2016-03-30
Add first and last name for ``PersonContact``

tulimaki:
Core: Add first and last name for ``PersonContact``
Admin: Enable edit for ``PersonContact`` first and last name
Front: Add first and last name for customer information edit

Refs SHOOP-832

### DIFF
--- a/shoop/core/models/_contacts.py
+++ b/shoop/core/models/_contacts.py
@@ -18,7 +18,6 @@ from timezone_field.fields import TimeZoneField
 
 from shoop.core.fields import InternalIdentifierField, LanguageField
 from shoop.core.pricing import PriceDisplayOptions
-from shoop.core.utils.name_mixin import NameMixin
 
 from ._base import TranslatableShoopModel
 from ._taxes import CustomerTaxGroup
@@ -78,7 +77,7 @@ class ContactGroup(TranslatableShoopModel):
 
 
 @python_2_unicode_compatible
-class Contact(NameMixin, PolymorphicModel):
+class Contact(PolymorphicModel):
     is_anonymous = False
     is_all_seeing = False
     default_tax_group_getter = None
@@ -130,6 +129,10 @@ class Contact(NameMixin, PolymorphicModel):
         if self.default_tax_group_getter:
             kwargs.setdefault("tax_group", self.default_tax_group_getter())
         super(Contact, self).__init__(*args, **kwargs)
+
+    @property
+    def full_name(self):
+        return (" ".join([self.prefix, self.name, self.suffix])).strip()
 
     def get_price_display_options(self):
         """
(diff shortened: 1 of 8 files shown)


shoop/core/models/_contacts.py
```diff
@@ -234,12 +239,18 @@ class Meta:
...
         if self.user_id and not self.pk:  # Copy things
             user = self.user
             if not self.name:
                 self.name = user.get_full_name()
             if not self.email:
                 self.email = user.email
+            if not self.first_name and not self.last_name:
+                self.first_name = user.first_name
+                self.last_name = user.last_name
+
```
suutari-ai: Shouldn't the assigning of self.name come after this? Now its done before (lines 242-243) and self.name won't be set if first and last name are copied from the user.

shoop/core/models/_contacts.py
```diff
@@ -234,12 +239,18 @@ class Meta:
         verbose_name_plural = _('persons')
 
     def save(self, *args, **kwargs):
+        if self.first_name or self.last_name:
+            self.name = (" ".join([self.first_name, self.last_name])).strip()
```
suutari-ai: Hmm.. now we can't have full names like 'John Fitzgerald "Jack" Kennedy' anymore. Should we have person.middle_names too? This would then be
```python
names = [self.first_name, self.middle_names, self.last_name]
if any(names):
    self.name = " ".join(x for x in names if x)
```
What do you think?

suutari-ai: OK, we probably shouldn't add any new features right now, so let's not add middle names. If we want that later, we can do it then. Until that it's still possible to set `person.first_name = 'John Fitzgerald "Jack"'` and `person.last_name = 'Kennedy'` to make full names with more than two names.

I would still prefer that form for setting the names even though stripping is cool it doesn't belong to model level.

suutari-ai:
Prefix and suffix should be removed from the base contact.

shoop/core/models/_contacts.py
suutari-ai: Actually, setting the `person.name` shouldn't be conditional, but it should always be set on save, even if there is no first and last name:

```python
names = [self.first_name, self.last_name]
self.name = " ".join(x for x in names if x)
```

suutari-ai:
> Prefix and suffix should be removed from the base contact (as done in @Pikkupomo's commit suutari-ai@47209c8).

I even did this in a branch, but then reconsidered as you can see from the commit message of my WIP commit for this: https://github.com/suutari-ai/shoop/commit/268d46194b97666fe6b79c1733c927e064e91f7a.

So let's not touch prefix or suffix for now.

suutari-ai:
* It should be impossible to make the name go out of sync with first_name and last_name. Therefore we should add a name property with getter and setter functions.
* Data migration should be done to set the first and last names correctly.
* Change Log entry is missing.

I have made changes to amend these, new PR coming soon.

PR #396, opened 2016-03-30
Add first and last name for PersonContact (SHOOP-832)

suutari-ai:
* Add first and last name for PersonContact
* Admin: Enable edit for PersonContact first and last name
* Add first and last name for customer information edit

This replaces PR #395.

The differences to PR #395 are:
 * Data migration for filling first_name and last_name
 * PersonContact.name property getter and setter
 * New tests
 * get_person_contact: Use user.first_name and user.last_name
 * Change Log entry

For full diff see `git diff 37fb7e4`.

### DIFF
--- a/shoop/core/models/_contacts.py
+++ b/shoop/core/models/_contacts.py
@@ -18,7 +18,6 @@ from timezone_field.fields import TimeZoneField
 
 from shoop.core.fields import InternalIdentifierField, LanguageField
 from shoop.core.pricing import PriceDisplayOptions
-from shoop.core.utils.name_mixin import NameMixin
 
 from ._base import TranslatableShoopModel
 from ._taxes import CustomerTaxGroup
@@ -78,7 +77,7 @@ class ContactGroup(TranslatableShoopModel):
 
 
 @python_2_unicode_compatible
-class Contact(NameMixin, PolymorphicModel):
+class Contact(PolymorphicModel):
     is_anonymous = False
     is_all_seeing = False
     default_tax_group_getter = None
@@ -130,6 +129,10 @@ class Contact(NameMixin, PolymorphicModel):
         if self.default_tax_group_getter:
             kwargs.setdefault("tax_group", self.default_tax_group_getter())
         super(Contact, self).__init__(*args, **kwargs)
+
+    @property
+    def full_name(self):
+        return (" ".join([self.prefix, self.name, self.suffix])).strip()
 
     def get_price_display_options(self):
         """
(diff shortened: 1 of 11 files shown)


suutari-ai:
@tulimaki: Most of the stuff here is from you. That all looks good to me now. You may merge this PR if it still looks good to you (with my changes).

chessbr:
Just a point about this PR. You are using `user.first_name` and `user.last_name` in several places assuming that the `AUTH_USER_MODEL` has this attributes. Well, I am using a custom `AUTH_USER_MODEL` and it has not theses attributes throwing `AttributeError` exceptions.
What do you think?

tulimaki:
@chessbr Good point, thanks! Pretty sure we are going to fix this issue, we probably shouldn't assume that those attributes is available. This task is currently in QA so for now I am going to hint those guys that there might an issue with this.

suutari-ai:
@chessbr You're right. This does not work for all user models.

Currently custom user models is not our priority so, I can't promise any schedule when we would fix that, but you may create a PR which adds necessary hasattr calls to protect these usages of user.first_name and user.last_name if you wish. And if you add some test cases which use a custom user model (without first_name or last_name probably), then it would be assured that the custom user model case is checked in the future changes too.

chessbr:
@suutari-ai OK, sounds good.
I will try to do that how ever I can't promise any schedule too. :) Things are going crazy here and TIME is rare!
Thanks.

suutari-ai:
Btw. we're also assuming that `user.email` and `user.is_active` work too. That might be bad assumption with custom user models.

chessbr:
I will assume the interface attributes as in https://docs.djangoproject.com/en/1.8/topics/auth/customizing/#django.contrib.auth.models.AbstractBaseUser

shoop/core/models/_contacts.py
```diff
@@ -296,6 +322,11 @@ def groups(self):
         return ContactGroup.objects.filter(identifier=self.default_contact_group_identifier)
 
 
+def _split_name(full_name):
+    names = full_name.rsplit(" ", 1)
```
akx: Are you sure `.rsplit()` is better than `.split()` here?

shoop_tests/core/test_contacts.py
```diff
@@ -106,6 +106,44 @@ def test_person_contact_creating_from_user(regular_user):
...
+    assert john.last_name == "Smith"
+
+
+@pytest.mark.django_db
+def test_person_name_create_with_name():
+    john = PersonContact.objects.create(name="John Smith")
+    assert PersonContact.objects.get(pk=john.pk).name == "John Smith"
+    assert john.name == "John Smith"
+    assert john.first_name == "John"
+    assert john.last_name == "Smith"
```
akx: None of these test cases address the case where there are more than two space-separated words in the name...

shoop/core/models/_contacts.py
suutari-ai: Both are wrong. Couple examples: "John F. Kennedy" -> first="John F.", last="Kennedy",  "Donald E. Knuth"

suutari-ai: We could do better with regular expressions and some additional data like list of common surnames or name articles and prepositions of course, but I'm no so sure if that's really needed.

PR #397, opened 2016-03-30
Simple CMS plugin fixes (SHOOP-2179)

tulimaki:
Simple CMS: Make plugin hide expired pages correctly
Simple CMS: Initalize pages-field in plugin correctly

Refs SHOOP-2179 / SHOOP-2315, SHOOP-2316

### DIFF
--- a/shoop/simple_cms/plugins.py
+++ b/shoop/simple_cms/plugins.py
@@ -32,7 +32,7 @@ class PageLinksConfigForm(GenericPluginForm):
         self.fields["pages"] = forms.ModelMultipleChoiceField(
             queryset=Page.objects.filter(visible_in_menu=True),
             required=False,
-            initial=self.plugin.config.get("categories", None),
+            initial=self.plugin.config.get("pages", None),
         )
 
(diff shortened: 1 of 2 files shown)


shoop/simple_cms/plugins.py
```diff
@@ -77,14 +77,14 @@ def get_context_data(self, context):
...
-            if selected_pages:
-                pages_qs = pages_qs.filter(id__in=selected_pages),
-            else:
-                pages_qs = pages_qs.none()
         if hide_expired:
             pages_qs = pages_qs.visible()
+
+        if not show_all_pages:
+            pages_qs = pages_qs.filter(id__in=selected_pages)
+
```
shawnadelic: Does this give you any problems when show_all_pages is turned off and you have no pages selected? I had a weird problem where id__in=[] was returning an queryset with an empty list or something strange, and the ``[]`` was rendering on the page.

If that hasn't been a problem for you, though, it should be fine.

shawnadelic:
This looks good, I think

PR #399, opened 2016-03-30
Xtheme: Enable global placeholder type (SHOOP-2219)

shawnadelic:
Enable a ``global`` parameter for Xtheme placeholders which would store
a site-wite placeholder configuration.

This is done by overriding the placeholder's view name as
``_XthemeGlobalView`` and should work as expected when editing,
rendering, etc.

Currently, this only applies to global placeholders that are within the
same template (but might be included in other templates), so if two
placeholders are marked as ``global`` but are not in the same template
they will not share the same settings.

Refs SHOOP-2219

### DIFF
--- a/shoop/xtheme/parsing.py
+++ b/shoop/xtheme/parsing.py
@@ -31,7 +31,8 @@ class NonConstant(ValueError):
 
 class NestingError(ValueError):
     """
-    Exception raised when a template's placeholder/column/row/plugin hierarchy is out of whack.
+    Exception raised when a template's placeholder/column/row/plugin
+    hierarchy is out of whack.
     """
 
 
@@ -45,7 +46,8 @@ def flatten_const_node_list(environment, node_list):
     :type node_list: list[jinja2.nodes.Node]
     :return: String of content
     :rtype: str
-    :raise Unflattenable: Raised when the node list can't be flattened into a constant
+    :raise Unflattenable: Raised when the node list can't be flattened into
+                          a constant
     """
     output = []
     eval_ctx = EvalContext(environment)
@@ -90,8 +92,8 @@ def parse_constantlike(environment, parser):
 
 class _PlaceholderManagingExtension(Extension):
     """
-    Superclass (could be mixin) with helpers for getting the currently active layout object
-    from a parser.
+    Superclass (could be mixin) with helpers for getting the currently
+    active layout object from a parser.
     """
     def _get_layout(self, parser, accept_none=False):
         """
@@ -99,12 +101,14 @@ class _PlaceholderManagingExtension(Extension):
 
         :param parser: Template parser
         :type parser: jinja2.parser.Parser
-        :param accept_none: Whether or not to accept the eventuality that there's no current layout.
-                            If False (the default), a `NestingError` is raised.
+        :param accept_none: Whether or not to accept the eventuality that
+                            there's no current layout. If False (the
+                            default), a `NestingError` is raised.
         :type accept_none: bool
         :return: The current layout
         :rtype: shoop.xtheme.view_config.Layout
-        :raises NestingError: Raised if there's no current layout and that's not okay.
+        :raises NestingError: Raised if there's no current layout and
+                              that's not okay.
         """
         cfg = getattr(parser, "_xtheme_placeholder_layout", None)
         if not accept_none and cfg is None:
@@ -163,14 +167,24 @@ def noop_node(lineno):
 
 class PlaceholderExtension(_PlaceholderManagingExtension):
     """
-    `PlaceholderExtension` manages `{% placeholder <NAME> %}` ... `{% endplaceholder %}`.
+    `PlaceholderExtension` manages `{% placeholder <NAME> [global] %}` ...
+      `{% endplaceholder %}`.
 
-    * The `name` can be any Jinja2 expression that can be folded into a constant, with the addition
-      of bare variable names such as `name` meaning the same as `"name"`. This makes it slightly
-      easier to write templates.
-    * The body of this block is actually discarded; only the inner `column`, `row` and `plugin`
-      directives have any meaning.  (A parser-time `Layout` object is created and populated during parsing
+    * The `name` can be any Jinja2 expression that can be folded into a
+      constant, with the addition of bare variable names such as `name`
+      meaning the same as `"name"`. This makes it slightly easier to write
+      templates.
+    * The body of this block is actually discarded; only the inner
+      `column`, `row` and `plugin` directives have any meaning.  (A
+      parser-time `Layout` object is created and populated during parsing
       of this block.)
+    * An optional ``global`` parameter can be used to specify that the
+      configuration for this placeholder should be "global" across
+      different views (although currently that only applies to placeholders
+      within the same template, i.e, if you defined the same global
+      placeholder in different templates they will not share configuration,
+      but an included or base template that is rendered by different views
+      will).
     """
     tags = set(['placeholder'])
 
--- a/shoop/xtheme/parsing.py
+++ b/shoop/xtheme/parsing.py
@@ -184,7 +184,14 @@ class PlaceholderExtension(_PlaceholderManagingExtension):
         :rtype: jinja2.nodes.Output
         """
         lineno = next(parser.stream).lineno
-        placeholder_name = six.text_type(parse_constantlike(self.environment, parser))
+        global_type = bool(parser.stream.look().value == "global")
+        if global_type:
+            # Do some special-case parsing for global placeholders
+            placeholder_name = six.text_type(parser.stream.current.value)
+            next(parser.stream)
+            next(parser.stream)
+        else:
+            placeholder_name = six.text_type(parse_constantlike(self.environment, parser))
         self._new_layout(parser, placeholder_name)
         parser.parse_statements(['name:endplaceholder'], drop_needle=True)
         # Body parsing will have, as a side effect, populated the current layout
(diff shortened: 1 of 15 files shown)


shoop/xtheme/layout.py
```diff
@@ -194,14 +194,17 @@ class Layout(object):
...
 
-    def __init__(self, placeholder_name, rows=None):
+    def __init__(self, placeholder_name, rows=None, global_type=False):
         """
         :param placeholder_name: The name of the placeholder. Could be None.
         :type placeholder_name: str|None
         :param rows: Optional iterable of LayoutRows to populate this Layout with.
         :type rows: Iterable[LayoutRow]|None
+        :param global_type: Optional boolean indicating whether this is global layout.
+        :type globla_type: bool|False
```
tulimaki: global

shoop/xtheme/parsing.py
```diff
@@ -184,25 +186,33 @@ def parse(self, parser):
...
         """
         lineno = next(parser.stream).lineno
+        global_type = False
+        placeholder_real_name = parser.stream.current.value
+        if parser.stream.look().value == "global":
+            global_type = True
+            next(parser.stream)
         placeholder_name = six.text_type(parse_constantlike(self.environment, parser))
-        self._new_layout(parser, placeholder_name)
+        placeholder_name = placeholder_real_name
```
tulimaki: What is happening here? You are always assigning parser.stream.current.value into placeholder_real_name.

Could this be like:
```
global_type = bool(parser.stream.look().value == "global")
if global_type:
    placeholder_name = parser.stream.current.value
    next(parser.stream)
else: 
    placeholder_name = six.text_type(parse_constantlike(self.environment, parser))
```

shoop/xtheme/rendering.py
```diff
@@ -60,9 +67,11 @@ def render_placeholder(context, placeholder_name, default_layout=None, template_
         context,
         placeholder_name,
         default_layout=default_layout,
-        template_name=template_name
+        template_name=template_name,
+        global_type=global_type,
     )
-    return renderer.render()
+    rendered = renderer.render()
```
tulimaki: Unnecessary change?

shoop/xtheme/views/editor.py
```diff
@@ -83,10 +88,15 @@ def _populate_vars(self):
...
             raise Problem(_("Unable to determine current theme."))
+        view_name = self.request.GET["view"]
+        global_type = self.request.GET.get("global_type", None)
+        if global_type:
+            view_name = XTHEME_GLOBAL_VIEW_NAME
         self.view_config = ViewConfig(
             theme=theme,
-            view_name=self.request.GET["view"],
-            draft=True
+            view_name=view_name,
```
tulimaki: This could be:
```
global_type = self.request.GET.get("global_type")
self.view_config = ViewConfig(
       theme=theme,
       view_name=(XTHEME_GLOBAL_VIEW_NAME if global_type else self.request.GET["view"]),
       draft=True,
       global_type=global_type
)
```

shoop/xtheme/view_config.py
```diff
@@ -31,9 +32,15 @@ def __init__(self, theme, view_name, draft):
...
+        :param global_type: Boolean indicating whether this is a global config
+        :type global_type: bool|False
         """
         self.theme = theme
-        self.view_name = force_text(view_name)
+        self.global_type = bool(global_type)
+        if not global_type:
+            self.view_name = force_text(view_name)
+        else:
+            self.view_name = XTHEME_GLOBAL_VIEW_NAME
```
tulimaki: Is this even necessary since you are already overring the view_name when calling this. Or just remove the view_name override in rendering.py and editor.py since you are still doing it here.

shoop/xtheme/view_config.py
```diff
@@ -52,7 +59,8 @@ def saved_view_config(self):
             self._saved_view_config = SavedViewConfig.objects.appropriate(
                 theme=self.theme,
                 view_name=self.view_name,
-                draft=self.draft
+                draft=self.draft,
+                global_type=bool(self.global_type),
```
tulimaki: This bool() shouldnt' be necessary.

shoop/xtheme/rendering.py
```diff
@@ -82,18 +91,27 @@ def __init__(self, context, placeholder_name, default_layout=None, template_name
...
         """
         self.context = context
-        self.view_config = get_view_config(context)
+        self.view_config = get_view_config(context, global_type=global_type)
         self.placeholder_name = placeholder_name
-        self.template_name = template_name
+        if global_type:
+            self.template_name = "_xtheme_global_template_name"
+        else:
+            self.template_name = template_name
```
tulimaki: maybe just ``self.template_name = ("_xtheme_global_template_name" if global_type else template_name)``

shoop/xtheme/parsing.py
```diff
@@ -171,6 +173,10 @@ class PlaceholderExtension(_PlaceholderManagingExtension):
     * The body of this block is actually discarded; only the inner `column`, `row` and `plugin`
       directives have any meaning.  (A parser-time `Layout` object is created and populated during parsing
       of this block.)
+    * An optional `global` parameter can be used to specify that the configuration for this placeholder
+      should be "global" across different views (although currently that only applies to placeholders within
+      the same template, i.e, if you defined the same global placeholder in different templates they will
+      not share configuration, but an included or base template that is rendered by different views will.
```
shawnadelic: I'm missing a ) here... :disappointed:

suutari-ai: Docstrings should be wrapped to narrower width (about column 72 or so, max 79) to make them easier to read and review. (Yeah, I know, the text above your added part is not wrapped so. You may change it too in a separate commit or leave it as is.)

suutari-ai: Another point: Docstrings are rendered by Sphinx in our API documentation and should be valid ReST. The backticks are used incorrectly here (in your added parts and in the parts above it). Singly backticked items are references to a class, function, module etc. and Sphinx tries to make them links, but for things that are not references to any Python object, you should use doubly backticked text. For example:
```rst
...optional ``global`` parameter...
...inner ``column``, ``row`` and ``plugin`` directives...
...A parser-time `~shoop.xtheme.layout.Layout` object... (Note: Sphinx will render that as single
                                                          word "Layout" which will be a link to
                                                          documentation of  shoop.xtheme.layout.Layout)
... see also `LayoutPartExtension` ... (objects in same module are usually resolved
                                        automatically by Sphinx and don't need the full module path)
```

You should check how the documentation renders out by building the HTML documentation as instructed in our README: https://github.com/shoopio/shoop/#documentation (Generated documentation for this should appear to: `doc/_build/html/api/shoop.xtheme.html` under your Shoop checkout)

shawnadelic: Cool, thanks!

shoop/xtheme/view_config.py
```diff
@@ -21,7 +21,7 @@ class ViewConfig(object):
     container for `SavedViewConfig` objects, and wraps the `SavedViewConfig` API.
     """
 
-    def __init__(self, theme, view_name, draft):
+    def __init__(self, theme, view_name, draft, global_type=False):
```
tulimaki: You still have global_type parameter here, but you don't use it in the rendering and editor? I think you probably should keep this and do the view_name override here. So if this init is called with global_type=True then the view_name is overriden into XTHEME_GLOBAL_VIEW_NAME.

shoop/xtheme/rendering.py
```diff
@@ -18,36 +18,42 @@
...
-    if config is None:
+    if (config is None) or (global_type is not config.global_type):
         view_object = context.get("view")
         if view_object:
             view_class = view_object.__class__
             view_name = view_class.__name__
         else:
             view_name = "UnknownView"
+        if global_type:
+            view_name = "_XthemeGlobalView"
```
tulimaki: Check my other comment on ViewConfig. Anyways this should be XTHEME_GLOBAL_VIEW_NAME instead of string. But I really do think that the view_name should be overriden in the ViewConfig if the global_type is True.

shoop/xtheme/models.py
```diff
@@ -34,9 +35,13 @@ def appropriate(self, theme, view_name, draft):
         :type view_name: str
         :param draft: Draft mode flag
         :type draft: bool
+        :param global_type: Boolean indicating global placeholder
+        :type global_type: bool|False
         :return: SavedViewConfig (possibly not saved)
         :rtype: SavedViewConfig
         """
+        if global_type:
+            view_name = XTHEME_GLOBAL_VIEW_NAME
```
tulimaki: Isn't the view_name already right since the view_name since this is used inside ViewConfig? Do we really need this?

shoop/xtheme/views/editor.py
```diff
@@ -83,10 +88,12 @@ def _populate_vars(self):
         theme = get_theme_by_identifier(self.request.GET["theme"])
         if not theme:
             raise Problem(_("Unable to determine current theme."))
+        view_name = self.request.GET["view"]
+        global_type = self.request.GET.get("global_type", None)
         self.view_config = ViewConfig(
             theme=theme,
-            view_name=self.request.GET["view"],
-            draft=True
+            view_name=(XTHEME_GLOBAL_VIEW_NAME if global_type else view_name),
```
tulimaki: No need to override the view_name here since ViewConfig init does that for you. Also missing the global_type parameter you need to pass for the ViewConfig.

PR #402, opened 2016-04-02
Front: Fix the ``compute_bought_with_relations`` management command

tulimaki:
Refs SHOOP-2294

### DIFF
--- a/shoop/front/management/commands/compute_bought_with_relations.py
+++ b/shoop/front/management/commands/compute_bought_with_relations.py
@@ -23,6 +23,10 @@ class Command(BaseCommand):
         ProductCrossSell.objects.filter(type=ProductCrossSellType.BOUGHT_WITH).delete()
 
         # Handle all ordered products
-        ordered_product_ids = OrderLine.objects.filter(type=OrderLineType.PRODUCT).values_list("pk", flat=True)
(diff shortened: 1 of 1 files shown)


shawnadelic:
Looks good

shoop/front/management/commands/compute_bought_with_relations.py
```diff
@@ -23,6 +23,10 @@ def handle(self, *args, **options):
         ProductCrossSell.objects.filter(type=ProductCrossSellType.BOUGHT_WITH).delete()
 
         # Handle all ordered products
-        ordered_product_ids = OrderLine.objects.filter(type=OrderLineType.PRODUCT).values_list("pk", flat=True)
+        ordered_product_ids = OrderLine.objects.filter(type=OrderLineType.PRODUCT).values_list("product_id", flat=True)
+        seen_product_ids = []
         for product_id in ordered_product_ids:
+            if product_id in seen_product_ids:
+                continue
+            seen_product_ids.append(product_id)
```
suutari-ai: Should use a set rather than list, because the `product_id in seen_product_ids` will get very slow with a list if there is many products, since it will do a linear lookup. (`seen_product_ids = set()` and `seen_product_ids.add(product_id)`)

PR #403, opened 2016-04-05
Admin: Hide initial method prices on order creator

shawnadelic:
Hide method prices for order creator until user reaches confirmation
view where they can be more easily calculated.

Refs SHOOP-2319

### DIFF
--- a/shoop/admin/modules/orders/static_src/create/reducers/order.js
+++ b/shoop/admin/modules/orders/static_src/create/reducers/order.js
@@ -9,10 +9,6 @@
 import {handleActions} from "redux-actions";
 import _ from "lodash";
 
-function getMethodPrice(method) {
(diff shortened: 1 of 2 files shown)


tulimaki:
You should probably remove prices from here also https://github.com/shoopio/shoop/blob/master/shoop/admin/modules/orders/views/create.py#L77

suutari-ai:
> You should probably remove prices from here also https://github.com/shoopio/shoop/blob/master/shoop/admin/modules/orders/views/create.py#L77

Yep, that is needed too. I will add it in another PR, since it's already done in the method-refactoring branch. Will add these changes to the new PR...

Thanks @shawnadelic!

suutari-ai:
Replaced by PR #404.

PR #404, opened 2016-04-05
Order creator: Hide initial method prices (SHOOP-2319)

suutari-ai:
Hide method prices for order creator until user reaches confirmation
view where they can be more easily calculated.


suutari-ai:
There is one commit from @shawnadelic which looks good to me. If the 609cb81 commit, which I added, looks good, please feel free to merge.

PR #405, opened 2016-04-06
Move compute_bought_with_relations management command

tulimaki:
Move compute_bought_with_relations command and related util from
front to core. In management command add handled product_ids into
set, instead of list.

Refs SHOOP-2326


doc/changelog.rst
```diff
@@ -58,6 +59,7 @@ Admin
 Front
 ~~~~~
 
+- Move compute_bought_with_relations management command under core
 - Add management command to generate bought with relations
```
suutari-ai: Just remove the "Add management command..." line, since change log should be written from the point of view of the last release.

suutari-ai:
This is good. Thanks @tulimaki!

PR #407, opened 2016-04-10
Translate missing Enums

chessbr:
Some Enums have not been translated yet. This should translate all Shoop enums.


chessbr:
Invalid user.

PR #409, opened 2016-04-11
Add attribute checks for user instance

chessbr:
One can set a different `AUTH_USER_MODEL` in settings which has not same
attributes of the default Django's User class, throwing exceptions.
Several `hasattr` and `getattr` calls were added to check and get attributes values.

### DIFF
--- a/shoop/front/apps/auth/forms.py
+++ b/shoop/front/apps/auth/forms.py
@@ -90,7 +90,7 @@ class RecoverPasswordForm(forms.Form):
             self.process_user(user)
 
     def process_user(self, user_to_recover):
-        if not user_to_recover.has_usable_password():
+        if not user_to_recover.has_usable_password() or not hasattr(user_to_recover, 'email'):
             return False
 
         context = {
--- a/shoop/admin/modules/users/views/detail.py
+++ b/shoop/admin/modules/users/views/detail.py
@@ -45,8 +45,8 @@ class BaseUserForm(forms.ModelForm):
             # Changing the password for an existing user requires more confirmation
             self.fields.pop("password")
             self.initial["permission_info"] = ", ".join(force_text(perm) for perm in [
-                _("staff") if self.instance.is_staff else "",
-                _("superuser") if self.instance.is_superuser else "",
+                _("staff") if getattr(self.instance, 'is_staff', None) else "",
+                _("superuser") if getattr(self.instance, 'is_superuser', None) else "",
             ] if perm) or _("No special permissions")
         else:
             self.fields.pop("permission_info")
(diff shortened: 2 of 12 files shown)


chessbr:
Hey folks, take a look. There are many calls to do the job. Maybe is not the best way to do it. Besides that, I didn't figure out how to create a test target to test Shoop with a different `AUTH_USER_MODEL`, because all the system may be testes with a different UserModel. Creating a new py.test command with other settings in `shoop_workbench` may do the trick.

As discussed earlier in Gitter:
> I was wondering if creating a `ShoopBaseUser` which inherits from a `ShoopUserMixin` interface class and Django's `AbstractBaseUser`. In this case `ShoopUserMixin` can contain a set of methods common to the whole Shoop system like methods `has_first_name()`, `has_last_name()` which may be implemented to return if the model has theses attributes. This way, when some want to place a custom UserModel, he must inherit from `ShoopUserMixin`, `AbstractBaseUser` and implement thoses required methods.. what do you think? Also, in `base_settings.py` the `AUTH_USER_MODEL` setting must contain the `ShoopBaseUser` by default.
Whatever Django project which wants to add Sshoop as an app, must set a UserModel which inherits from `ShoopUserMixin`.. otherwise, several exceptions will be raised.

shoop/front/apps/auth/forms.py
```diff
@@ -103,5 +103,5 @@ def process_user(self, user_to_recover):
         subject = loader.render_to_string(self.subject_template_name, context)
         subject = ''.join(subject.splitlines())  # Email subject *must not* contain newlines
         email = loader.render_to_string(self.email_template_name, context, request=self.request)
-        send_mail(subject, email, self.from_email, [user_to_recover.email])
+        send_mail(subject, email, self.from_email, [getattr(user_to_recover, 'email', '')])
```
tulimaki: Here we don't want to send the mail at all if the user model does not have email available.

tulimaki:
Looks pretty ok. @suutari-ai what do you think, do we need some extra tests around these changes, since these changes does not introduce anything new features to test? Also I don't think we need tests for custom user model for now, altought it would be nice?

shoop/front/apps/auth/forms.py
chessbr: You are right! I will fix that.

tulimaki:
ok to test

suutari-ai:
Nice work @chessbr! Thanks a lot!

Merged this in commit ebca260f115916033524c2ceaecf4aa81cfac029 (but modified commit message, that's why it does not show as merged here).

chessbr:
Great! Thanks!

PR #413, opened 2016-04-14
Contact group sales ranges (SHOOP-1975)

tulimaki:
* Add ``admin_contact_group_form_part`` provider
* Campaigns: Add contact group sales ranges
* Campaigns: Add form part for contact group sales ranges


suutari-ai:
Good work @tulimaki!

PR #415, opened 2016-04-14
Service provider admin (SHOOP-2061)

suutari-ai:
Add admin for service providers such as carriers and payment
processors.


suutari-ai:
The code here is mostly from @tulimaki. It looks good to me, but don't want to merge PR which I created myself, so @shawnadelic or @tulimaki, please merge this if OK.

PR #416, opened 2016-04-15
Admin: Add delete for service providers admin module (SHOOP-2297)

tulimaki:
Services linked to service provider is not removed, but the
service provider foreign key is set to null and service is disabled.

Refs SHOOP-2297

### DIFF
--- a/doc/changelog.rst
+++ b/doc/changelog.rst
@@ -49,6 +49,7 @@ Localization
 Admin
 ~~~~~
 
+- Add delete for service providers admin module
 - Remove shipping and payment method management
 - Add package mode for products
 - Enable ```merchant_notes`` editing for ``Contact``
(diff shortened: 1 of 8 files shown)


doc/changelog.rst
```diff
@@ -49,6 +49,7 @@ Localization
 Admin
 ~~~~~
 
+- Add delete for service providers admin module
```
shawnadelic: Is this needed since we only just added the service providers module?

suutari-ai: Yep, this changelog entry should be removed from this PR.

suutari-ai:
When the changelog entry is removed, this is good to go.

suutari-ai:
I removed the changelog entry (and added entry for service provider admin) and merged this with that change.

Thanks guys!

PR #418, opened 2016-04-15
Admin: Add carts module for monitoring abandoned carts

shawnadelic:
Add an admin carts module that allows merchants to monitor abanded
user carts, total sales, date since last update, and other
useful information.

Refs SHOOP-2237

### DIFF
--- a/doc/changelog.rst
+++ b/doc/changelog.rst
@@ -65,6 +65,7 @@ Admin
 Front
 ~~~~~
 
+- Add admin view for monitoring customer carts
 - Remove ``get_method_validation_errors`` signal
 - Fix bug at ``get_visible_products`` filter when orderable_only is False
 - Set template price display options from the customer
--- a/doc/api/shoop.front.admin_module.carts.rst
+++ b/doc/api/shoop.front.admin_module.carts.rst
@@ -1,0 +1,17 @@
+shoop.front.admin_module.carts package
+======================================
+
+Subpackages
+-----------
+
+.. toctree::
+
+    shoop.front.admin_module.carts.views
+
+Module contents
+---------------
+
+.. automodule:: shoop.front.admin_module.carts
+    :members:
+    :undoc-members:
+    :show-inheritance:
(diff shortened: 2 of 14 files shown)


tulimaki:
This admin module should live under front since stored basket is under front and this will break the admin if the front is not in installed apps.

shoop/admin/modules/carts/views/list.py
```diff
@@ -0,0 +1,67 @@
...
+        ),
+        Column(
+            "taxful_total_price", _(u"Total"), sort_field="taxful_total_price_value",
+            display="format_taxful_total_price", class_name="text-right",
+            filter_config=RangeFilter(field_type="number", filter_field="taxful_total_price_value")
+        ),
+    ]
+
+    def format_order_date(self, instance, *args, **kwargs):
+        return format_datetime(localtime(instance.created_on), locale=get_current_babel_locale())
```
tulimaki: This formatting should probably be in ``shoop.utils.dates`` as an util.

shawnadelic:
test this please

tulimaki:
Can you move all front admin-module related code under fron/admin_module/?

doc/changelog.rst
```diff
@@ -66,6 +66,7 @@ Front
 - Remove ``get_method_validation_errors`` signal
 - Fix bug at ``get_visible_products`` filter when orderable_only is False
 - Set template price display options from the customer
+- Add admin view for monitoring customer carts
```
suutari-ai: Move this entry to top, since it's newest.

suutari-ai:
Yeah, the admin module stuff should be under shoop/front/admin_modules, since now it's very confusing and hard to tell that these views are actually admin views.

Also, it's confusing that now we have used both terms basket and cart for the same thing. Can we do anything about it? I like "cart" more, since that's what we want to show in the UI, but that would need much refactoring.

suutari-ai:
OK the "basket" -> "cart" change ain't that difficult after all... almost got it done in 10 minutes: https://github.com/shoopio/shoop/compare/master...suutari-ai:basket-to-cart  ... just needs the migrations to be fixed... that can be difficult though. Maybe we shouldn't go that way yet...

PR #420, opened 2016-04-18
Admin: Enable deleting unused shipping and payment methods (SHOOP-2051)

tulimaki:
Do not allow delete if shipping or payment method is linked to order


shawnadelic:
Looks good :shipit:

PR #423, opened 2016-04-19
Campaigns: Remove only from groups with active sales range

tulimaki:
When assigning to contact groups based on sales ranges contacts shouldn't
be removed from contact groups which doesn't have sales range or the sales
range is not active i.e. min_value and max_value equals zero.

Refs SHOOP-1975 / SHOOP-2416


tulimaki:
retest this please

shawnadelic:
Looks good :shipit:

PR #424, opened 2016-04-19
Admin: In service admin set delete url to None in new mode

tulimaki:
New services couldn't be created since the delete url is not available
in new mode.

Refs SHOOP-2051 / SHOOP-2418


suutari-ai:
retest this please

PR #426, opened 2016-04-20
Core: Fix for updating order line from product in shortcuts (SHOOP-2443)

tulimaki:
Assing product name into order line text instead of name

### DIFF
--- a/doc/changelog.rst
+++ b/doc/changelog.rst
@@ -12,6 +12,7 @@ Unreleased
 Core
 ~~~~
 
+- Fix bug: Order line text is not set for package products
 - Add new Service API and implement shipping and payments with it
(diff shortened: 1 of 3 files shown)


suutari-ai:
The commit message and change log entry is not very descriptive. And isn't this SHOOP-2443?

How about using the ticket summary as change log entry too, i.e. "Fix bug: Order line text is not set for package products"

suutari-ai:
typo in commit message: Assing -> Assign. Also you're missing reasoning. Why we should set text instead of name? (Is there a name for order line or not?)

suutari-ai:
> Order line text attribute is much better option for product name or sku since order line does not have name attribute.

Much better indeed. LOL! :D

Thanks @tulimaki!

PR #431, opened 2016-04-25
Weight based pricing behavior (SHOOP-2048)

tulimaki:
Core: Add ``total_gross_weight`` property for ``Source``
Core: Add ``WeightBasedPricingBehaviorComponent``
Admin: Enable option to provide form parts for services
Admin: Add form parts for ``WeightBasedPricingBehaviorComponent``

### DIFF
--- a/shoop/admin/templates/shoop/admin/services/edit.jinja
+++ b/shoop/admin/templates/shoop/admin/services/edit.jinja
@@ -11,7 +11,7 @@
             {% if form_defs|length > 1 %}
             {% call content_block(_("Behavior"), ("fa-cubes")) %}
                 <div class="form-group">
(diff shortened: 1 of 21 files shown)


shoop/admin/templates/shoop/admin/services/edit.jinja
```diff
@@ -43,17 +44,17 @@
 {{ super() }}
 <script>
     $(function () {
-        $("#component-button").on("click", function (e) {
-            e.preventDefault();
+        $("#component-button .btn").on("click", function (event) {
```
juhakujala: Why this change? I think it's better to use the id straight at the clickable link element like it was...

Also using a class called `.btn` is not very good considering the previous version had only id selector which is unique and if someone decides to change the `<a>` classname to something different than btn, this stops working...

PR #432, opened 2016-04-25
Weight based pricing behavior (SHOOP-2048)

tulimaki:
Core: Add total_gross_weight property for Source
Core: Add WeightBasedPricingBehaviorComponent
Admin: Enable option to provide form parts for services
Admin: Add form parts for WeightBasedPricingBehaviorComponent
Core: Add out of range behavior for weight based pricing


shoop_tests/core/test_weight_based_pricing_behavior.py
```diff
@@ -0,0 +1,181 @@
...
+    costs = list(service.get_costs(source))
+    unavailability_reasons = list(service.get_unavailability_reasons(source))
+    assert (unavailability_reasons or costs) and not (unavailability_reasons and costs)
+    if costs:  # We have costs so let's test prices
+        assert len(costs) == 1
+        assert costs[0].price.value == target_price
+        assert costs[0].description == target_description
+
+
+
```
Pikkupomo: Remove extra line changes

shoop/core/models/_service_behavior.py
```diff
@@ -82,3 +84,70 @@ def get_unavailability_reasons(self, service, source):
         if self.max_weight:
             if weight > self.max_weight:
                 yield ValidationError(_("Maximum weight exceeded."), code="max_weight")
+
+
+class WeightBasedPriceRange(TranslatableModel):
```
Pikkupomo: If these are going to be something that are being offered from the "get go" I'd suggest this would be `PriceRangeService(PolymorphicModel)` or  something like 
```python 
class WeightBasedPriceRange(PriceRangeService):
	unit = "g"``` or something

suutari-ai:
* Remove the form part provide from documentation, since we're not sure yet if we want to support that in the future.
* Fix the JavaScript issues @juhakujala mentioned.
* We could probably use custom widgets for this, but let's examine that later.

suutari-ai:
The core side is great, but the admin side might need some work still, since the end result is IMHO far from being user friendly. Anyway, this is still good feature and we can work on the admin side details later. Good job!

PR #433, opened 2016-04-27
Classic gray: Fix initial config for cross sells plugin

shawnadelic:
Fix the inital configuriaton for the cross sells plugin.

Currently, configuration is stored as an enum, but initial settings
are received as a string. These are now mapped to the correct enum
type on initialization.

Refs SHOOP-2409


suutari-ai:
Good fix. Thanks, Shawn!

PR #438, opened 2016-04-28
Make tracking codes visible ( SHOOP-2281)

tulimaki:
Core: Add ``get_tracking_codes`` to ``shoop.core.models.Order``
Admin: Show tracking codes in order detail
Classic Gray Theme: Show tracking codes in order detail

### DIFF
--- a/shoop/core/models/_orders.py
+++ b/shoop/core/models/_orders.py
@@ -610,6 +610,9 @@ class Order(MoneyPropped, models.Model):
     def get_status_display(self):
         return force_text(self.status)
 
+    def get_tracking_codes(self):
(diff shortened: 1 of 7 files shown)


shoop/core/models/_orders.py
```diff
@@ -610,6 +610,9 @@ def get_unshipped_products(self):
     def get_status_display(self):
         return force_text(self.status)
 
+    def get_tracking_codes(self):
+        return [shipment.tracking_code for shipment in self.shipments.all() if shipment.tracking_code]
```
shawnadelic: This could maybe be ``for shipment in self.shipments.exclude(tracking_code=""))``?

Not sure if that's any better

shawnadelic:
Looks good other than possibly the above comment and fixing those merge conflicts

PR #441, opened 2016-04-29
Admin: Fix picotable sorting bug

shawnadelic:
Fix picotable bug when sorting querysets for translated fields.

Refs SHOOP-2398

### DIFF
--- a/shoop/admin/utils/picotable.py
+++ b/shoop/admin/utils/picotable.py
@@ -11,11 +11,12 @@ import json
 
 import six
 from django.core.paginator import EmptyPage, Paginator
-from django.db.models import Count, Manager, Q, QuerySet
+from django.db.models import Manager, Q, QuerySet
 from django.http.response import JsonResponse
 from django.template.defaultfilters import yesno
 from django.utils.encoding import force_text
 from django.utils.translation import ugettext_lazy as _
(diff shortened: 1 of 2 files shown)


tulimaki:
There is few problems in test. Populate if required only populates products with one language. This is issue since we faced problems most of the time only after multiple translations. Also you are not testing the actual issue here which was the duplicate items.

shoop/admin/utils/picotable.py
```diff
@@ -237,11 +238,9 @@ def to_json(self, context=None):
...
     def sort_queryset(self, queryset, desc=False):
         order_by = ("-" if desc else "") + self.sort_field
-        queryset = queryset.order_by(order_by)
-        if self.sort_field.startswith("translations__"):
-            # Ref http://archlinux.me/dusty/2010/12/07/django-dont-use-distinct-and-order_by-across-relations/
-            queryset = queryset.annotate(_dummy_=Count(self.sort_field))
-        return queryset
+        if "translations__" in self.sort_field:
+            queryset = queryset.translated(get_language())
+        return queryset.order_by(order_by)
```
suutari-ai: This creates another bug, since now some rows will be missing from the result. I.e. those that don't have a translation for currently active language.

PR #444, opened 2016-05-02
Shipment weights (SHOOP-2494)

tulimaki:
* Core: Prove that shipment weights is calculated correctly
* Core: Add signal for shipment created
* Core: Make ``create_shipment`` for orders atomic

### DIFF
--- a/doc/changelog.rst
+++ b/doc/changelog.rst
@@ -12,6 +12,7 @@ Unreleased
 Core
 ~~~~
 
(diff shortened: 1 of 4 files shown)


shawnadelic:
Reviewed, and this looks good IMO, although @suutari-ai should also have a look

suutari-ai:
Nice work, @tulimaki! Thanks!

PR #445, opened 2016-05-02
Simple Supplier: Fix product create view bug

shawnadelic:
Fix bug when creating new product with simple supplier set as
default by only rendering stock management form for saved products.

Refs SHOOP-2207

### DIFF
--- a/shoop/simple_supplier/admin_module/forms.py
+++ b/shoop/simple_supplier/admin_module/forms.py
@@ -39,7 +39,7 @@ class SimpleSupplierForm(forms.Form):
         return Supplier.objects.filter(shop_products__product=product, module_identifier="simple_supplier").distinct()
 
     def can_manage_stock(self):
-        return Supplier.objects.filter(module_identifier="simple_supplier").exists()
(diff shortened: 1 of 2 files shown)


shoop/simple_supplier/templates/shoop/simple_supplier/admin/product_form_part.jinja
```diff
@@ -47,7 +47,7 @@
 {% endmacro %}
 
 {% call content_block(_("Stock management"), "fa-cubes") %}
-    {% if ss_form.can_manage_stock() %}
+    {% if ss_form.product.pk and ss_form.can_manage_stock() %}
```
tulimaki: Instead of doing this you should probably skip the stock management tab all together when product is in new mode. So just yield this form in ``SimpleSupplierFormPart`` only when not in new mode.

tulimaki:
Can you still add changelog for this.

suutari-ai:
I'll add the changelog entry while merging this.

Good work, @shawnadelic! Thanks!

PR #446, opened 2016-05-03
Admin: Show service choice name in service provider admin

shawnadelic:
Show service choice name in service provider admin list view.


suutari-ai:
Thanks, Shawn!

suutari-ai:
@shawnadelic: Just found out that I missed a bug in this. If the shipping/payment method does not have a service provider, this will raise an exception. Can you fix this?

shawnadelic:
Yes, will do that

shawnadelic:
Here is what I came up with: https://github.com/shoopio/shoop/pull/452

PR #447, opened 2016-05-03
Improve shoop_setup_utils for addons (SHOOP-2507)

suutari-ai:
The shoop_setup_utils package provides a nice way to do resource
generating on "setup.py build" phase.  It is used here in Shoop, but it
is also useful for Shoop addons.

Improve it a bit so that is more useful for addons.  See the commit
messages for details.


tulimaki:
Thanks!

PR #449, opened 2016-05-03
Core: Change ``Shipment`` weight to grams

tulimaki:
Allow merchant use different measurement units for products. In case
merchant decides to use pounds we can't do any grams to kilograms
conversion in backend.

Unify Shoop weight based ``MeasurementField``s to use grams so we don't
need to do unit conversions.

Refs SHOOP-2494 / SHOOP-2511

### DIFF
--- a/shoop/core/models/_shipments.py
+++ b/shoop/core/models/_shipments.py
@@ -43,7 +43,7 @@ class Shipment(models.Model):
     tracking_code = models.CharField(max_length=64, blank=True, verbose_name=_("tracking code"))
     description = models.CharField(max_length=255, blank=True, verbose_name=_("description"))
     volume = MeasurementField(unit="m3", verbose_name=_("volume"))
-    weight = MeasurementField(unit="kg", verbose_name=_("weight"))
+    weight = MeasurementField(unit="g", verbose_name=_("weight"))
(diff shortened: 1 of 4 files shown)


shoop/core/models/_shipments.py
```diff
@@ -43,7 +43,7 @@ class Shipment(models.Model):
     tracking_code = models.CharField(max_length=64, blank=True, verbose_name=_("tracking code"))
     description = models.CharField(max_length=255, blank=True, verbose_name=_("description"))
     volume = MeasurementField(unit="m3", verbose_name=_("volume"))
-    weight = MeasurementField(unit="kg", verbose_name=_("weight"))
+    weight = MeasurementField(unit="g", verbose_name=_("weight"))
```
suutari-ai: I think the weight should be in kg as it used to be, but whoever is calculating that weight should do the conversion from grams to kilograms. I.e. the bug is not here, but in the calculation side..

suutari-ai: IMHO the bug is in the `cache_values` method line 73 below, which has `self.weight = total_weight`. That should rather be:
```python
self.weight = total_weight * 1000
```

akx: Yep, you're right; that's the correct fix.

PR #450, opened 2016-05-03
Admin: Hide invalid package product choices (SHOOP-2389)

shawnadelic:
When adding products to a package product, hide invalid choices,
including variation parents and package products.

Refs SHOOP-2389

Notes: There might be a better way to do this. The browse widget has a built-in "filter" parameter, however since we need to use Q objects to filter multiple valid modes, I had to implement a workaround by just indicating the "mode" in the filter setting.

### DIFF
--- a/shoop/admin/modules/products/forms/package_forms.py
+++ b/shoop/admin/modules/products/forms/package_forms.py
@@ -13,7 +13,7 @@ from django.contrib import messages
 from django.db.transaction import atomic
 from django.utils.translation import ugettext_lazy as _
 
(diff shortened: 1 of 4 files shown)


shoop/admin/modules/products/forms/package_forms.py
```diff
@@ -21,10 +23,14 @@
 from .parent_forms import ProductChildBaseFormSet
 
 
+class PackageProductChoiceWidget(ProductChoiceWidget):
```
tulimaki: This should probably be in ``shoop.admin.forms.widgets`` with all the other widgets.

PR #455, opened 2016-05-04
Move service checkout phase to Front (SHOOP-2502)

suutari-ai:
The `get_checkout_phase` methods in `ServiceProvider` and `Service`
clearly don't belong to Core, since even their signature says they
should return an object defined in Front.

Make the service API cleaner in Core by moving this checkout phase to
its own provide with an interface class.

Also move some code from `front/checkout/__init__.py` to private
submodules to allow importing CheckoutPhaseViewMixin from other
submodules of shoop.front.checkout.

We will need some provide for specifying similar views/forms for Admin
too, but that should be easier to implement when doing a concrete
implementation for one.


tulimaki:
Looks good!

PR #457, opened 2016-05-04
Admin/Core: Shipment creation extensions (SHOOP-2501)

tulimaki:
* Core: Add unique shipment key for ``Shipment``
* Admin: Allow shipment creation extensions

### DIFF
--- a/doc/changelog.rst
+++ b/doc/changelog.rst
@@ -12,6 +12,7 @@ Unreleased
 Core
 ~~~~
 
+- Add option use ``Order.create_shipment`` with unsaved ``Shipment``
 - Add identifier for ``Shipment``
 - Fix bug: Fix max length for service method names for ``Order``
 - Add payment methods to ``Order``
--- a/doc/changelog.rst
+++ b/doc/changelog.rst
@@ -61,6 +61,7 @@ Localization
 Admin
 ~~~~~
 
+- Allow shipment creation form extensions
 - Add payment creation view to ``Order`` admin
 - Enable order cancelation in Order edit view
 - Hide invalid choices for package products
--- a/doc/changelog.rst
+++ b/doc/changelog.rst
@@ -12,6 +12,7 @@ Unreleased
 Core
 ~~~~
 
+- Add identifier for ``Shipment``
 - Fix bug: Fix max length for service method names for ``Order``
 - Add payment methods to ``Order``
 - Add ``is_canceled`` and ``can_set_canceled`` to ``Order``
--- a/shoop/admin/modules/orders/views/shipment.py
+++ b/shoop/admin/modules/orders/views/shipment.py
@@ -7,6 +7,8 @@
 # LICENSE file in the root directory of this source tree.
 from __future__ import unicode_literals
 
+import abc
+
 import six
 from django import forms
 from django.contrib import messages
@@ -17,15 +19,70 @@ from django.views.generic import UpdateView
 from shoop.admin.toolbar import PostActionButton, Toolbar
 from shoop.admin.utils.forms import add_form_errors_as_messages
 from shoop.admin.utils.urls import get_model_url
+from shoop.apps.provides import get_provide_objects
 from shoop.core.excs import NoProductsToShipException
-from shoop.core.models import Order, Product, Supplier
+from shoop.core.models import Order, Product, Shipment, Supplier
+
+FORM_MODIFIER_PROVIDER_KEY = "admin_extend_create_shipment_form"
+
+
+class ShipmentFormModifier(six.with_metaclass(abc.ABCMeta)):
+    def get_extra_fields(self, order):
+        """
+        Extra fields for shipment creation view.
+
+        :param order: Order linked to form
+        :type order: shoop.core.models.Order
+        :return: List of extra fields that should be added to form.
+        Tuple should contain field name and Django form field.
+        :rtype: list[(str,django.forms.Field)]
+        """
+        pass
+
+    def clean_hook(self, form):
+        """
+        Extra clean for shipment creation form.
+
+        This hook will be called in `~Django.forms.Form.clean` method of
+        the form, after calling parent clean.  Implementor of this hook
+        may call `~Django.forms.Form.add_error` to add errors to form or
+        modify the ``form.cleaned_data`` dictionary.
+
+        :param form: Form that is currently cleaned
+        :type form: ShipmentForm
+        :rtype: None
+        """
+        pass
+
+    def form_valid_hook(self, form, shipment):
+        """
+        Extra form valid handler for shipment creation view.
+
+        This is called from ``OrderCreateShipmentView`` just
+        before the ``Order.create_shipment``
+
+        :param form: Form that is currently handled
+        :type form: ShipmentForm
+        :param shipment: Unsaved shipment
+        :type shipment: shoop.core.models.Shipment
+        :rtype: None
+        """
+        pass
+
+
+class ShipmentForm(forms.Form):
+    def clean(self):
+        cleaned_data = super(ShipmentForm, self).clean()
+        for extend_class in get_provide_objects(FORM_MODIFIER_PROVIDER_KEY):
+            extend_class().clean_hook(self)
+        return cleaned_data
 
 
 class OrderCreateShipmentView(UpdateView):
     model = Order
     template_name = "shoop/admin/orders/create_shipment.jinja"
     context_object_name = "order"
-    form_class = forms.Form  # Augmented manually
+    form_class = ShipmentForm  # Augmented manually
 
     def get_context_data(self, **kwargs):
         context = super(OrderCreateShipmentView, self).get_context_data(**kwargs)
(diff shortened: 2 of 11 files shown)


shoop/admin/modules/orders/views/shipment.py
```diff
@@ -46,13 +104,15 @@ def get_form_kwargs(self):
         return kwargs
 
     def get_form(self, form_class):
+        default_field_keys = []
```
akx: This smells like a set, not a list.

shoop/core/models/_shipments.py
```diff
@@ -45,6 +45,7 @@ class Shipment(models.Model):
     description = models.CharField(max_length=255, blank=True, verbose_name=_("description"))
     volume = MeasurementField(unit="m3", verbose_name=_("volume"))
     weight = MeasurementField(unit="kg", verbose_name=_("weight"))
+    shipment_key = models.CharField(max_length=64, unique=True, verbose_name=_("shipment key"))
```
akx: I think this should be called `identifier`, to keep in line with all of the other `identifier`s in Shoop. (Also, you could use `InternalIdentifierField`).

shoop/core/utils/shipments.py
```diff
@@ -0,0 +1,13 @@
...
+#
+# Copyright (c) 2012-2016, Shoop Ltd. All rights reserved.
+#
+# This source code is licensed under the AGPLv3 license found in the
+# LICENSE file in the root directory of this source tree.
+
+import random
+
+
+def generate_shipment_key(order):
```
akx: Shouldn't this logic just live in `Shipment.save`?

```python
if not self.identifier:
    self.identifier = "%s/%s/%s" % ...
```

doc/changelog.rst
```diff
@@ -12,6 +12,7 @@ Unreleased
 Core
 ~~~~
 
+- Add shipment key for ``shoop.core.models.Shipment``
```
suutari-ai: Would be better to refer it as just ``Shipment`` as below

shoop/admin/modules/orders/views/shipment.py
```diff
@@ -17,15 +19,71 @@
...
+        """
+        Extra clean for shipment creation form
+
+        :param cleaned_data: form cleaned data
+        :type cleaned_data: dict
+        :return: List of errors that should be handled.
+        Tuple should contain field name and error message.
+        :rtype: list(tuple)|None
+        """
+        pass
```
suutari-ai: Shouldn't this raise ValidationError on error and return None otherwise?

shoop/admin/modules/orders/views/shipment.py
```diff
@@ -17,15 +19,71 @@
...
+
+        :param form: Form that is currently handled
+        :type form: ShipmentForm
+        :param order: Order linked to form
+        :type order: shoop.core.models.Order
+        :param shipment_key: Generated shipment key for shipment
+        that is created when all form valid operations is done.
+        :rtype: None
+        """
+        pass
```
suutari-ai: Document when this is called in the shipment creation process. (After creating shipment? Before invoking the signal?)

shoop/admin/modules/orders/views/shipment.py
```diff
@@ -17,15 +19,71 @@
...
+    @abc.abstractmethod
+    def get_extra_fields(self, order):
+        """
+        Extra fields for shipment creation form
+
+        :param order: Order linked to form
+        :type order: shoop.core.models.Order
+        :return: List of extra fields that should be added to form.
+        Tuple should contain field name and Django form field.
+        :rtype: list(tuple)|None
```
suutari-ai: Should return just list always, no need to return None ever, since empty list is OK too. Also, specify the return type of the tuple components too, i.e.
```
:rtype: list[(str,django.forms.Field)]
```

akx: `:rtype: list[tuple[str, django.forms.Field]]` tbh :)

shoop/core/models/_shipments.py
suutari-ai: Order has key though. Hmm... but that's actually not the same usage either... yeah, identifier might be better for this.

shoop/core/utils/shipments.py
suutari-ai: No, it should be in `Shipment.__init__`, so that unsaved Shipments will have identifiers too. Init should generate an identifier if one is not specified in kwargs.

shoop/admin/modules/orders/views/shipment.py
suutari-ai: Actually, maybe we should just pass the form into this method and do the form.add_error in the method implementation if desired. Something like this:
```python
    def clean_hook(self, form, cleaned_data):
        """
        Extra clean for shipment creation form.

        This hook will be called in `~Django.forms.Form.clean` method of
        the form, after calling parent clean.  Implementor of this hook
        may call `~Django.forms.Form.add_error` to add errors to form or
        modify the `cleaned_data` dictionary.

        :type form: django.forms.Form
        :param form: form being processed
        :type cleaned_data: dict
        :param cleaned_data: form cleaned data, may be modified
        :rtype: None
        """
        pass
```

suutari-ai: This don't have to be abstract, since the default implementation `pass` might be just fine sometimes.

shoop/admin/modules/orders/views/shipment.py
suutari-ai: This shouldn't be abstract neither, since the default implementation `pass` is just fine sometimes.

shoop/admin/modules/orders/views/shipment.py
suutari-ai: Actually, tuples should be written as `(A,B)` rather than as `tuple[A,B]`. (Re: https://www.jetbrains.com/help/pycharm/2016.1/type-hinting-in-pycharm.html?origin=old_help )

suutari-ai: I've done this incorrectly many times myself though and there might even be some of those errors in Shoop too...

shoop/admin/modules/orders/views/shipment.py
tulimaki: ValidationError is not good for all situations. Like when adding field errors. That can't be done through ValidationError so I think the errors here is better.

shoop/admin/modules/orders/views/shipment.py
```diff
@@ -17,15 +19,68 @@
...
+        """
+        Extra clean for shipment creation form
+
+        :param cleaned_data: form cleaned data
+        :type cleaned_data: dict
+        :return: List of errors that should be handled.
+        Tuple should contain field name and error message.
+        :rtype: list[tuple(str,str)]
+        """
+        pass
```
suutari-ai: Still think that this should be like the one I wrote in this comment https://github.com/shoopio/shoop/pull/457#discussion_r62484593 i.e. it doesn't return anything but rather is given the form and the hook can then add errors or do whatever it wants with the form.

shoop/admin/modules/orders/views/shipment.py
```diff
@@ -17,15 +19,68 @@
...
+class ShipmentFormModifier(six.with_metaclass(abc.ABCMeta)):
+    def get_extra_fields(self, order):
+        """
+        Extra fields for shipment creation form
+
+        :param order: Order linked to form
+        :type order: shoop.core.models.Order
+        :return: List of extra fields that should be added to form.
+        Tuple should contain field name and Django form field.
+        :rtype: list[tuple(str,django.forms.Field)]
```
suutari-ai: `:rtype: list[(str,django.forms.Field)]`

shoop/admin/modules/orders/views/shipment.py
```diff
@@ -17,15 +19,68 @@
...
+        :type cleaned_data: dict
+        :return: List of errors that should be handled.
+        Tuple should contain field name and error message.
+        :rtype: list[tuple(str,str)]
+        """
+        pass
+
+    def form_valid_hook(self, form, shipment):
+        """
+        Extra form valid handler for shipment creation form
```
suutari-ai: Now that I look more carefully, I see that this is actually called from the view, not from the form. Therefore the last word in this title is incorrect. It should be "shipment creation view.". You're also missing the period: docstring title should end with period.

shoop/core/models/_shipments.py
```diff
@@ -45,12 +48,19 @@ class Shipment(models.Model):
...
 
     class Meta:
         verbose_name = _('shipment')
         verbose_name_plural = _('shipments')
 
+    def __init__(self, *args, **kwargs):
+        super(Shipment, self).__init__(*args, **kwargs)
+        if not self.identifier:
+            self.identifier = "%s/%s/%s" % (
+                self.order.pk, self.order.shipments.count(), random.randint(1, 1000000000))
```
suutari-ai: What if self.order is not set? Should take that into account.

suutari-ai: While at it, replace the random.randint with the thing from django that generates random strings securely...

suutari-ai: from django.utils.crypto import get_random_string

suutari-ai:
It's perfect now. Thanks, Panu!

PR #459, opened 2016-05-06
Add payment creation view to order admin (SHOOP-2399)

shawnadelic:
Add payment creation view to order admin.

payment_creator

Refs SHOOP-2399

### DIFF
--- a/shoop/core/models/_orders.py
+++ b/shoop/core/models/_orders.py
@@ -371,12 +371,22 @@ class Order(MoneyPropped, models.Model):
             self.payment_date = now()
             self.save()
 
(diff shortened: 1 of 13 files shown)


shoop/core/models/_orders.py
```diff
@@ -371,13 +371,23 @@ def _set_paid(self):
             self.payment_date = now()
             self.save()
 
+    def _set_partially_paid(self):
+        if self.payment_status != PaymentStatus.PARTIALLY_PAID:  # pragma: no branch
```
tulimaki: the ``pragma: no branch`` shouldn't be neccesary

PR #460, opened 2016-05-06
Front: Make required fields of address configurable (SHOOP-2503)

shawnadelic:
Add SHOOP_FRONT_REQUIRED_ADDRESS_FIELDS setting to define required
address settings in checkout address form in addition to those defined
at the model leve.

Refs SHOOP-2503

### DIFF
--- a/shoop/front/checkout/addresses.py
+++ b/shoop/front/checkout/addresses.py
@@ -31,6 +31,11 @@ class AddressForm(forms.ModelForm):
             # Set default country
             self.fields["country"].initial = settings.SHOOP_ADDRESS_HOME_COUNTRY
 
+        field_properties = settings.SHOOP_FRONT_ADDRESS_FIELD_PROPERTIES
+        for field, properties in field_properties.items():
+            for prop in properties:
+                setattr(self.fields[field], prop, properties[prop])
+
 
 class CompanyForm(TaxNumberCleanMixin, forms.ModelForm):
     company_name_field = "name"
--- a/shoop/front/settings.py
+++ b/shoop/front/settings.py
@@ -63,3 +63,16 @@ SHOOP_CHECKOUT_VIEW_SPEC = (
 #:
 #: Handled error cases are: 400, 403, 404, and 500
 SHOOP_FRONT_INSTALL_ERROR_HANDLERS = True
+
+#: A dictionary defining properties to override the default field properties of the
+#: checkout address form. Should map the field name (as a string) to a dictionary
+#: containing the overridding Django form field properties, as in the following
+#: example which makes the postal code a required field:
+#:
+#: SHOOP_FRONT_ADDRESS_FIELD_PROPERTIES = {
+#:    "postal_code": {"required": True}
+#: }
+#:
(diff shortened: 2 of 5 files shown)


shoop/front/settings.py
```diff
@@ -63,3 +63,7 @@
 #:
 #: Handled error cases are: 400, 403, 404, and 500
 SHOOP_FRONT_INSTALL_ERROR_HANDLERS = True
+
+#: A list of strings defining which address fields will be required at the form
+#: level in addition to those already required at the model level,
+SHOOP_FRONT_REQUIRED_ADDRESS_FIELDS = []
```
tulimaki: The name of this should indicate that these are in addition to required fields in model level.

shoop/front/checkout/addresses.py
```diff
@@ -31,6 +31,11 @@ def __init__(self, **kwargs):
             # Set default country
             self.fields["country"].initial = settings.SHOOP_ADDRESS_HOME_COUNTRY
 
+        required_address_fields = set(settings.SHOOP_FRONT_ADDITIONAL_REQUIRED_ADDRESS_FIELDS)
```
akx: Wouldn't it be better to just allow overriding any sort of property on the fields?

```python
SHOOP_FRONT_ADDRESS_FIELD_PROPERTIES = {
  "phone": {"required": False, "help_text": "But who was phone?"}
}
```

That would make it possible to make fields that are required as built-in non-required, etc.

shoop/front/settings.py
```diff
@@ -63,3 +63,11 @@
...
 #: Handled error cases are: 400, 403, 404, and 500
 SHOOP_FRONT_INSTALL_ERROR_HANDLERS = True
+
+#: A dictionary defining properties to override the default field properties in the
+#: checkout address form. For example, this can be used to set ``required`` to
+#: ``True`` to force a certain field to be required.
+#:
+#: The dictionary should map the field name (as a string) to a dictionary of the
+#: overridding properties.
+SHOOP_FRONT_ADDRESS_FIELD_PROPERTIES = {}
```
tulimaki: Can you some this some good default. Like the fields we have now required? I think making some fields like city unrequired is a little bit hard without migrations. But maybe that is not part of this task.

tulimaki: The comment for this setting should also address that issue.

shoop/front/settings.py
```diff
@@ -63,3 +63,16 @@
...
+#: checkout address form. For example, this can be used to set ``required`` to
+#: ``True`` to force a certain field to be required.
+#:
+#: The dictionary should map the field name (as a string) to a dictionary of the
+#: overridding properties.
+#:
+#: It should be noted, however, that overriding some settings (such as making a
+#: required field non-required) could create other validation issues.
+SHOOP_FRONT_ADDRESS_FIELD_PROPERTIES = {
+    "postal_code": {"required": True}
```
tulimaki: Instead of postal code I was thinking to add fields that is required before this change. Which would be name, street, city and country.

PR #461, opened 2016-05-06
Core: Change max length for service method names in `Order` (SHOOP-2489)

shawnadelic:
Make max length for shipping and payment method names in `Order` the
same as max length for model names.

Refs SHOOP-2489


tulimaki:
retest this please

PR #462, opened 2016-05-07
Admin: Order cancelation logic fix

shawnadelic:
I think there is an error in the logic for cancelling orders. The "else" here should only be executed when given a status of "canceled".

### DIFF
--- a/shoop/admin/modules/orders/views/detail.py
+++ b/shoop/admin/modules/orders/views/detail.py
@@ -101,8 +101,6 @@ class OrderSetStatusView(DetailView):
             raise Problem(_("Unable to set order as completed at this point"))
         if new_status.role == OrderStatusRole.CANCELED and not order.can_set_canceled():
             raise Problem(_("Paid, shipped, or canceled orders cannot be canceled"))
-        else:
-            order.set_canceled()
         order.status = new_status
         order.save(update_fields=("status",))
         message = _("Order status changed: %s to %s") % (old_status, new_status)


Pikkupomo:
Is there a test that proves this bug?

shawnadelic:
@Pikkupomo: Not yet, it could use one though so will probably do that first thing on Monday

shoop/admin/modules/orders/views/detail.py
```diff
@@ -99,10 +99,11 @@ def post(self, request, *args, **kwargs):
         old_status = order.status
         if new_status.role == OrderStatusRole.COMPLETE and not order.can_set_complete():
             raise Problem(_("Unable to set order as completed at this point"))
-        if new_status.role == OrderStatusRole.CANCELED and not order.can_set_canceled():
-            raise Problem(_("Paid, shipped, or canceled orders cannot be canceled"))
-        else:
-            order.set_canceled()
```
shawnadelic: This sets the status to cancelled anytime that new_status.role is not "canceled" (for example, if it's "complete").

shoop/admin/modules/orders/views/detail.py
```diff
@@ -99,10 +99,11 @@ def post(self, request, *args, **kwargs):
...
+        if new_status.role == OrderStatusRole.CANCELED:
+            if not order.can_set_canceled():
+                raise Problem(_("Paid, shipped, or canceled orders cannot be canceled"))
+            else:
+                order.set_canceled()
         order.status = new_status
```
shawnadelic: However, this line reverses that status change and sets it to the new status. Again, for example if new_status was "complete", the order would set to cancelled above then set to "complete" here, ultimately getting the same result for that case (despite being switched to "canceled" temprarily), and since we're currently only handling "complete" or "canceled" as statuses, there really aren't going to be any errors produced.

PR #466, opened 2016-05-13
Enable editing orders (SHOOP-2338, SHOOP-2534)

tulimaki:
* Refactor the `OrderCreator` for easier extensibility
* Add `OrderModifier` for basic `Order` modifying from `OrderSource`
* Core: Update shipping status correctly in ``Order.create_shipment``
* Core: Add ``Order.can_edit``
* Admin: Enable order editing

### DIFF
--- a/shoop/admin/modules/orders/views/edit.py
+++ b/shoop/admin/modules/orders/views/edit.py
@@ -19,13 +19,14 @@ from django.db import transaction
 from django.http.response import HttpResponse, JsonResponse
 from django.utils.encoding import force_text
 from django.utils.translation import ugettext as _
-from django.views.generic import TemplateView
 from django_countries import countries
 
 from shoop.admin.modules.orders.json_order_creator import JsonOrderCreator
+from shoop.admin.toolbar import Toolbar
+from shoop.admin.utils.views import CreateOrUpdateView
 from shoop.core.models import (
-    AnonymousContact, CompanyContact, Contact, Order, PaymentMethod, Product,
-    ShippingMethod, Shop, ShopStatus
+    AnonymousContact, CompanyContact, Contact, Order, OrderLineType,
+    PaymentMethod, Product, ShippingMethod, Shop, ShopStatus
 )
 from shoop.core.pricing import get_pricing_module
 from shoop.utils.i18n import (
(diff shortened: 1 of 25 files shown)


shoop/core/order_creator/_creator.py
```diff
@@ -22,21 +22,7 @@
...
-        :type request: django.http.HttpRequest|None
-        :param request:
-          Optional request object for backward compatibility.  Passing
-          non-None value is DEPRECATED.
-        """
-        if request is not None:
-            warnings.warn(
-                "Initializing OrderCreator with a request is deprecated",
-                DeprecationWarning, stacklevel=2)
+class OrderProcessor(object):
```
tulimaki: I think this new OrderProcessor class could be in it's own file like _processor

shoop/core/order_creator/_creator.py
```diff
@@ -248,3 +234,28 @@ def process_order_before_lines(self, source, order):
...
+            warnings.warn(
+                "Initializing OrderCreator with a request is deprecated",
+                DeprecationWarning, stacklevel=2)
+
+    def create_order(self, order_source):
+        data = self.get_source_base_data(order_source)
+        order = Order(**data)
+        order.save()
+        order = self.finalize_creation(order, order_source)
+        order_creator_finished.send(sender=type(self), order=order, source=order_source)
```
tulimaki: This signal is not actually called in same place as it use to be. Not sure if it's a problem or why it was called before process_order_after_lines before. What do you think @Pikkupomo?

shoop/core/order_creator/_creator.py
Pikkupomo: @suutari-ai said that same file is good for now because we don't want to refactor too much. I agree with that.

shoop/core/order_creator/_creator.py
Pikkupomo: maybe the original idea was to have it before processing lines because the "base order is finished" in that step. If @suutari-ai doesn't find anything to complain about this, I'm good with it.

shoop/admin/modules/orders/views/edit.py
```diff
@@ -99,32 +134,61 @@ def get_price_info(shop, customer, product, quantity):
...
+
+    def get_initial_order_data(self):
+        order = self.object
+        if not order.pk:
+            return {}
+        return {
+            "shop": encode_shop(order.shop),
+            "lines": [
+                get_line_data_for_edit(order.shop, line) for line in order.lines.filter(
+                    type__in=[OrderLineType.PRODUCT, OrderLineType.OTHER]
```
Pikkupomo: why no discount lines?

tulimaki: Discount lines is reprocessed/formed in confirm phase in order edit. Based on changes to original order the discount lines will change.

suutari-ai:
Two features in same PR. I want to merge the first two commits first.

PR #467, opened 2016-05-16
Add OrderModifier (SHOOP-2338)

suutari-ai:
Add OrderModifier class which can be used to modify existing orders using an OrderSource.

This is the first part of enabling editing orders.  Splitting the PR #466 to two pieces to make it easier to review.  The leftover commits will for the second part.

### DIFF
--- a/shoop/core/order_creator/_creator.py
+++ b/shoop/core/order_creator/_creator.py
@@ -22,21 +22,7 @@ from shoop.utils.numbers import bankers_round
 from ._source_modifier import get_order_source_modifier_modules
 
 
-class OrderCreator(object):
-
-    def __init__(self, request=None):
-        """
-        Initialize order creator.
-
-        :type request: django.http.HttpRequest|None
-        :param request:
-          Optional request object for backward compatibility.  Passing
-          non-None value is DEPRECATED.
-        """
-        if request is not None:
-            warnings.warn(
-                "Initializing OrderCreator with a request is deprecated",
-                DeprecationWarning, stacklevel=2)
+class OrderProcessor(object):
 
     def source_line_to_order_lines(self, order, source_line):
         """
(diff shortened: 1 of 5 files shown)


shoop/core/order_creator/_creator.py
```diff
@@ -220,9 +209,6 @@ def create_order(self, order_source):
 
         self._assign_code_usages(order_source, order)
 
-        order_creator_finished.send(
-            sender=type(self), order=order, source=order_source)
-
         order.save()
         self.process_order_after_lines(source=order_source, order=order)
```
suutari-ai: Yeah, as noted in https://github.com/shoopio/shoop/pull/466#discussion_r63204654 the signal is not triggered at same state as previously. It used to be triggered between _assign_code_usages and the second save, but after this change it is called after the second save and even after the process_order_after_lines hook. Anyway, as order_creator_finished signal is new in 4.0, this shouldn't matter for backward compatibility. I don't see any other reason neither. So this is fine. Actually this order seems even more logical as the original, i.e. triggering order_creator_finished being the last action in order creating.

shoop/core/order_creator/_modifier.py
```diff
@@ -0,0 +1,27 @@
...
+
+    @atomic
+    def update_order_from_source(self, order_source, order):
+        data = self.get_source_base_data(order_source)
+        Order.objects.filter(pk=order.pk).update(**data)
+
+        order = Order.objects.get(pk=order.pk)
+
+        for line in order.lines.all():
+            line.delete()
```
suutari-ai: This will make all rows linked to OrderLine vanish. Or if there is any model that links to OrderLine with on_delete=PROTECT FK then this will fail. I guess that is OK though, but should be documented so that developers using Shoop will know to not link to lines but to the order.

suutari-ai:
Nice work @Pikkupomo and @tulimaki. Thanks!

PR #469, opened 2016-05-17
Archiving products (SHOOP-1982)

shawnadelic:
Add product archiving to Core and Admin.

Refs SHOOP-1982

### DIFF
--- a/shoop/core/models/_products.py
+++ b/shoop/core/models/_products.py
@@ -122,10 +122,10 @@ class ProductQuerySet(TranslatableQuerySet):
         root = (self.language(language) if language else self)
 
         if customer and customer.is_all_seeing:
-            exclude_q = Q(deleted=True) | Q(mode=ProductMode.VARIATION_CHILD)
+            exclude_q = Q(deleted=True) | Q(archived=True) | Q(mode=ProductMode.VARIATION_CHILD)
             qs = root.all().exclude(exclude_q).filter(shop_products__shop=shop)
         else:
-            qs = root.all().exclude(deleted=True).filter(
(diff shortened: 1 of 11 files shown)


shoop/core/models/_products.py
```diff
@@ -163,6 +164,7 @@ class Product(TaxableItem, AttributableMixin, TranslatableModel):
     created_on = models.DateTimeField(auto_now_add=True, editable=False, verbose_name=_('created on'))
     modified_on = models.DateTimeField(auto_now=True, editable=False, verbose_name=_('modified on'))
     deleted = models.BooleanField(default=False, editable=False, db_index=True, verbose_name=_('deleted'))
+    archived = models.BooleanField(default=False, editable=False, db_index=True, verbose_name=_('archived'))
```
tulimaki: There shouldn't be need to create new attribute for archiving. I think you should use delete for archiving. Also use the soft delete functionality to archive/hide products.

suutari-ai:
Maybe you should have just renamed the "Delete" button to "Archive" in UI?

I don't think we need two flags for excluding products from listings. Just one should be enough. The name for that field could be "deleted" in the backend as it is now (to avoid a migration). Setting that flag is effectively just archiving, since nothing is really deleted and it could even be brought back by clearing the flag.

shawnadelic:
Sure, that would be easy enough, I had just assumed that "archive" would have a separate use case from "delete," but essentially they are doing the same thing.

suutari-ai:
The current functionality for "soft delete" is good as it is.

suutari-ai:
... and @kmn1 said that "Delete" is better term in the UI than "Archive", so no need to change that either.

PR #471, opened 2016-05-18
Front: Update UI for package products

shawnadelic:
Update UI for package products on product detail page and customer
basket display, listing products in package and their quantities.

Refs SHOOP-2274

### DIFF
--- a/shoop/front/views/product.py
+++ b/shoop/front/views/product.py
@@ -57,9 +57,9 @@ class ProductDetailView(DetailView):
         elif product.mode == ProductMode.PACKAGE_PARENT:
             children = (
                 product.get_all_package_children()
-                .language(language)
+                .translated()
                 .order_by("translations__name")
-                .all())
+            )
             context["package_children"] = cache_product_things(self.request, children)
 
         context["shop_product"] = self.shop_product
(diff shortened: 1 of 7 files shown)


shoop/front/views/product.py
```diff
@@ -57,8 +57,7 @@ def get_context_data(self, **kwargs):
         elif product.mode == ProductMode.PACKAGE_PARENT:
             children = (
                 product.get_all_package_children()
-                .language(language)
-                .order_by("translations__name")
+                .translated(get_language())
```
tulimaki: get_language is not neccessary here and the .all() shouldn't be either. What happened to the ordering here? Is the .translated even needed without the ordering?

shawnadelic: You're right, that left out the ordering.

Also, didn't realize but this doesn't have any ordering for the basket contents, so I added that too. It's a little bit hacky since we have to do the queryset in the template. Alternatively, we could pass an "ordering" parameter to get_all_package_children, but not sure if that's a good idea since we wouldn't know whether or not it was a translated field, etc.

shoop/themes/classic_gray/templates/classic_gray/shoop/front/basket/partials/_basket_content.jinja
```diff
@@ -28,6 +28,15 @@
                             {{ line.text }}
                         {% endif %}
                     </h4>
+                    {% set package_children = product.get_all_package_children().translated().order_by("translations__name") %}
```
tulimaki: Can you check if package parent here? Also can you make macro out of this too since you should probably use this same code in checkout confirmation too?

PR #472, opened 2016-05-19
Front: Test handle_add_var basket_command correctly

tulimaki:
``shoop.front.basket.commands.handle_add_var`` takes request, basket
and product_id as a parameters. Let's use corect product_id instead
of 1.


suutari-ai:
Why no refs?

PR #473, opened 2016-05-19
Front: Filter suppliable products correctly

tulimaki:
``Supplier.get_suppliable_products`` returns list of shop product pk's.

Missing tests since SHOOP-2576 will change this logic completely at the end of this week.


suutari-ai:
Why no refs?

PR #474, opened 2016-05-19
Core: Update stocks based on product id

tulimaki:
Instead of using ``ShipmentProduct`` id let's use product_id for
updating stock.


suutari-ai:
Why no refs?

PR #475, opened 2016-05-20
Enhanced multiselect dropdowns (SHOOP-2324)

tulimaki:
Admin: Add view and widget to support Select2 dropdowns
Admin: Enable multiselect dropdown with Select2
Admin: Remove support for selectpicker
Campaigns: Use Select2 for ManyToManyField's

### DIFF
--- a/shoop/campaigns/forms.py
+++ b/shoop/campaigns/forms.py
@@ -4,13 +4,18 @@
 #
 # This source code is licensed under the AGPLv3 license found in the
 # LICENSE file in the root directory of this source tree.
+
+import ast
+
 from django import forms
 from django.core.exceptions import ValidationError
-from django.db.models import OneToOneField, Q
+from django.db.models import ManyToManyField, OneToOneField, Q, QuerySet
 from django.forms import ChoiceField, ModelForm
+from django.utils.encoding import force_text
 from django.utils.translation import ugettext_lazy as _
 
 from shoop.admin.forms.fields import PercentageField
+from shoop.admin.forms.widgets import Select2Multiple
 from shoop.apps.provides import get_provide_objects
 from shoop.campaigns.models.campaigns import (
     BasketCampaign, CatalogCampaign, Coupon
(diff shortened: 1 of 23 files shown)


tulimaki:
retest this please

suutari-ai:
Why support for select pickers was removed? (Commit message doesn't say anything.)

shoop/campaigns/forms.py
```diff
@@ -77,13 +81,33 @@ def __init__(self, *args, **kwargs):
...
+                        formfield = field.formfield(**kwargs)
+                    self.fields[obj.identifier] = formfield
+
+    def _get_initial_value(self, obj):
+        initial_value = self.get_initial_value(obj.identifier)
+        if initial_value:
+            if not isinstance(initial_value, str):
+                initial_value = [o.pk for o in initial_value]
+            else:
+                initial_value = [initial_value]
```
suutari-ai: Got `TypeError: 'decimal.Decimal' object is not iterable` from the `initial_value = [o.pk for o in initial_value]` line of this.

suutari-ai:
* IMHO Products should be selectable by SKU too (and probably SKU should be shown in the selected item and in the dropdown too)
* The select2 fields should be styled (especially since all select fields are now using the select2):
   * Focus color should be same as for other input fields.
   * Padding should be fixed to be same above and below the items.
   * Select 2 fields should have same weight as other fields (when they fit to single line)

shoop/campaigns/forms.py
suutari-ai: How about this one instead?
```python
    def _get_initial_value(self, obj):
        initial_value = self.get_initial_value(obj.identifier)
        if initial_value is None:
            return None
        elif isinstance(initial_value, QuerySet):
            return [x.pk for x in initial_value]
        else:
            return [initial_value]
```

tulimaki:
* Added better styles for select2 selects
* Added option to search with SKU.
* Changed object text to force_text(obj) instead of objecte name. I think it's up to Product __str__ to display sku. And I think it would add unneccesary complexety to decide display based on the object type.
* Added fix to _get_initial_value in campaign forms

PR #478, opened 2016-05-21
Enable multiple service checkout phases

tulimaki:
Testing: Add test carrier with checkout phase
Front: Fix bug with multiple service checkout phases


suutari-ai:
Nice work! Thanks @shawnadelic and @tulimaki!

PR #479, opened 2016-05-21
Admin: Enable adding log entries to orders

shawnadelic:
Enable adding log entries to orders

Merchant can create `note` log entries for orders under `Log
Entries` in order detail view.

Refs SHOOP-2240


suutari-ai:
Thanks, @shawnadelic!

PR #481, opened 2016-05-23
Fix region_code handling in `TaxingContext`

Pikkupomo:
By default, form renders only `region` field, not `region_code` so taxing doesn't work.
No ref

### DIFF
--- a/shoop/core/taxing/_context.py
+++ b/shoop/core/taxing/_context.py
@@ -11,6 +11,6 @@ class TaxingContext(object):
         self.customer_tax_group = customer_tax_group
         self.customer_tax_number = customer_tax_number
         self.country_code = getattr(location, "country_code", None) or getattr(location, "country", None)
-        self.region_code = getattr(location, "region_code", None)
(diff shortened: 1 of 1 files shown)


shoop/core/taxing/_context.py
```diff
@@ -11,6 +11,6 @@ def __init__(self, customer_tax_group=None, customer_tax_number=None, location=N
         self.customer_tax_group = customer_tax_group
         self.customer_tax_number = customer_tax_number
         self.country_code = getattr(location, "country_code", None) or getattr(location, "country", None)
-        self.region_code = getattr(location, "region_code", None)
+        self.region_code = getattr(location, "region_code", None) or getattr(location, "region", None)
```
suutari-ai: I think this is not correct, since our Address interface has both region_code and region. If you want to access the name of the region in your tax module for some reason, then it's available as `taxing_context.location.region`. However, `taxing_context.region_code` should always be the code, not name of the region, e.g. CA instead of California.

Making sure region codes are set for addresses is an address validation issue. Address validation should happen (at least) in checkout (in Front) and order creation (in Admin).

PR #493, opened 2016-05-25
Preserve order data while editing (SHOOP-2587)

tulimaki:
Core: Add modified by and modified on for orders
Admin: Preserve data for unhandled fields in order edit
Test that coupon usages is added correctly
Campaigns: Preserve coupon codes in order edit

### DIFF
--- a/shoop/admin/modules/orders/json_order_creator.py
+++ b/shoop/admin/modules/orders/json_order_creator.py
@@ -286,6 +286,9 @@ class JsonOrderCreator(object):
         if not self.is_valid:
             return None
 
+        if order_to_update:
+            for code in order_to_update.codes:
+                source.add_code(code)
         return source
 
     def create_order_from_state(self, state, creator=None, ip_address=None):
--- a/shoop/admin/modules/orders/json_order_creator.py
+++ b/shoop/admin/modules/orders/json_order_creator.py
@@ -175,7 +175,7 @@ class JsonOrderCreator(object):
             customer = PersonContact(**fields)
         return customer
 
-    def _initialize_source_from_state(self, state, creator, ip_address, save):
+    def _initialize_source_from_state(self, state, creator, ip_address, save, order_to_update=None):
         shop_data = state.pop("shop", None).get("selected", {})
         shop = self.safe_get_first(Shop, pk=shop_data.pop("id", None))
         if not shop:
@@ -183,6 +183,8 @@ class JsonOrderCreator(object):
             return None
 
         source = OrderSource(shop=shop)
+        if order_to_update:
+            source.update_from_order(order_to_update)
 
         customer_data = state.pop("customer", None)
         billing_address_data = customer_data.pop("billingAddress", {})
@@ -192,13 +194,9 @@ class JsonOrderCreator(object):
             else customer_data.pop("shippingAddress", {}))
         is_company = customer_data.pop("isCompany", False)
         save_address = customer_data.pop("saveAddress", False)
-        customer = self.safe_get_first(Contact, pk=customer_data.get("id")) if customer_data else None
+        customer = self._get_customer(customer_data, billing_address_data, is_company, save)
         if not customer:
-            customer = self._create_contact_from_address(billing_address_data, is_company)
-            if not customer:
-                return
-            if save:
-                customer.save()
+            return
 
         billing_address = MutableAddress.from_data(billing_address_data)
         shipping_address = MutableAddress.from_data(shipping_address_data)
@@ -235,13 +233,22 @@ class JsonOrderCreator(object):
         )
         return source
 
+    def _get_customer(self, customer_data, billing_address_data, is_company, save):
+        customer = self.safe_get_first(Contact, pk=customer_data.get("id")) if customer_data else None
+        if not customer:
+            customer = self._create_contact_from_address(billing_address_data, is_company)
+            if not customer:
+                return
+            if save:
+                customer.save()
+        return customer
+
     def _postprocess_order(self, order, state):
         comment = (state.pop("comment", None) or "")
         if comment:
             order.add_log_entry(comment, kind=LogEntryKind.NOTE, user=order.creator)
 
-    def create_source_from_state(self, state, creator=None, ip_address=None,
-                                 save=False):
+    def create_source_from_state(self, state, creator=None, ip_address=None, save=False, order_to_update=None):
         """
         Create an order source from a state dict unserialized from JSON.
 
@@ -251,6 +258,8 @@ class JsonOrderCreator(object):
         :type creator: django.contrib.auth.models.User|None
         :param save: Flag whether order customer and addresses is saved to database
         :type save: boolean
+        :param order_to_update: Order object to edit
+        :type order_to_update: shoop.core.models.Order|None
         :return: The created order source, or None if something failed along the way
         :rtype: OrderSource|None
         """
@@ -262,7 +271,7 @@ class JsonOrderCreator(object):
 
         # First, initialize an OrderSource.
         source = self._initialize_source_from_state(
-            state, creator=creator, ip_address=ip_address, save=save)
+            state, creator=creator, ip_address=ip_address, save=save, order_to_update=order_to_update)
         if not source:
             return None
 
@@ -316,7 +325,7 @@ class JsonOrderCreator(object):
         :return: The created order, or None if something failed along the way
         :rtype: Order|None
         """
-        source = self.create_source_from_state(state, save=True)
+        source = self.create_source_from_state(state, order_to_update=order_to_update, save=True)
         if source:
             source.modified_by = modified_by
         modifier = OrderModifier()
--- a/shoop/admin/modules/orders/json_order_creator.py
+++ b/shoop/admin/modules/orders/json_order_creator.py
@@ -305,7 +305,7 @@ class JsonOrderCreator(object):
             self.add_error(exc)
             return
 
-    def update_order_from_state(self, state, order_to_update, creator=None, ip_address=None):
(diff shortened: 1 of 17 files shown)


suutari-ai:
retest this please

shoop/admin/modules/orders/json_order_creator.py
```diff
@@ -240,8 +240,7 @@ def _postprocess_order(self, order, state):
         if comment:
             order.add_log_entry(comment, kind=LogEntryKind.NOTE, user=order.creator)
 
-    def create_source_from_state(self, state, creator=None, ip_address=None,
-                                 save=False):
+    def create_source_from_state(self, state, creator=None, ip_address=None, order_to_update=None, save=False):
```
suutari-ai: missing documentation for the new params

shoop/admin/modules/orders/json_order_creator.py
```diff
@@ -277,6 +276,22 @@ def create_source_from_state(self, state, creator=None, ip_address=None,
...
+                ip_address=order_to_update.ip_address,
+                orderer=order_to_update.orderer,
+                order_date=order_to_update.order_date,
+                status=order_to_update.status,
+                marketing_permission=order_to_update.marketing_permission,
+                customer_comment=order_to_update.customer_comment,
+                payment_data=(order_to_update.payment_data or {}),
+                shipping_data=(order_to_update.shipping_data or {}),
+                extra_data=(order_to_update.extra_data or {}),
+            )
```
suutari-ai: This should use OrderSource.update_from_order, since otherwise any new fields added to ordersource won't be updated by this code.

shoop/admin/modules/orders/json_order_creator.py
```diff
@@ -277,6 +276,22 @@ def create_source_from_state(self, state, creator=None, ip_address=None,
...
+            for code in order_to_update.coupon_usages.values_list("coupon__code", flat=True):
+                source.add_code(code)
```
suutari-ai: You shouldn't set codes by coupon usages. Coupon usages is different as the codes in order_source. Coupons are something that shoop.campaigns uses, but order_source.codes is more general concept of shoop.core. It might also very well be that shoop.campaigns isn't enabled at all in INSTALLED_APPS and then there even isn't `order.coupon_usages` at all and your code would fail.

You should rather create a new field to order which stores the codes from order_source and then just use that to repopulate the order_source.codes, e.g. like this.

```python
for code in order_to_update.codes:
    source.add_code(code)
```

and in order_creator, do:

```python
order.codes = order_source.codes
```

`Order.codes` should probably be a property to protect it being accessed directly, like this:
```python
class Order(...):
    ...

    _codes = JSONField(...)

    @property
    def codes():
        return list(self._codes or [])

    @codes.setter
    def codes(self, value):
        codes = []
        for code in value:
            if not isinstance(code, six.text_type):
                raise TypeError('codes must be a list of strings')
            codes.append(code)
        self._codes = codes
```

shoop/core/order_creator/_modifier.py
```diff
@@ -17,9 +19,14 @@ class OrderModifier(OrderProcessor):
...
     def update_order_from_source(self, order_source, order):
         data = self.get_source_base_data(order_source)
+        data.update({
+            "modified_by": real_user_or_none(order_source.modified_by),
+            "modified_on": now()
+        })
         Order.objects.filter(pk=order.pk).update(**data)
 
         order = Order.objects.get(pk=order.pk)
+        order.coupon_usages.all().delete()  # Delete all coupon usages for this order
```
suutari-ai: This should happen in the shoop.campaings app (by signal probably), not here.

shoop/core/order_creator/_modifier.py
```diff
@@ -17,9 +19,14 @@ class OrderModifier(OrderProcessor):
     @atomic
     def update_order_from_source(self, order_source, order):
         data = self.get_source_base_data(order_source)
+        data.update({
+            "modified_by": real_user_or_none(order_source.modified_by),
+            "modified_on": now()
+        })
         Order.objects.filter(pk=order.pk).update(**data)
```
suutari-ai: Some data should never be updated in this process, e.g. creator or ip_address. You should drop them from the data dictionary before calling the `Order.update(**data)`

shoop_tests/admin/test_order_creator.py
```diff
@@ -259,15 +259,21 @@ def test_company_contact_creation(rf, admin_user):
...
     order = create_empty_order(shop)
+    order.payment_data = {"payment_data": True}
+    order.shipping_data = {"shipping_data": True}
+    order.extra_data = {"external_id": "123"}
     order.save()
     assert order.lines.count() == 0
     assert order.pk is not None
-    request = get_frontend_request_for_command(state, "finalize", admin_user)
+    assert order.modified_by is None  # Order is not edited
+    assert order.modified_on is None  # Order is not edited
```
suutari-ai: Shouldn't modified by be initialized to the same value as creator? I.e. these asserts should be
```python
assert order.modified_by == order.creator
assert order.modified_on == order.order_date
```

shoop_tests/admin/test_order_creator.py
```diff
@@ -279,11 +285,27 @@ def test_editing_existing_order(rf, admin_user):
...
     # Check that product line have right taxes
     for line in edited_order.lines.all():
         if line.type == OrderLineType.PRODUCT:
             assert [line_tax.tax.code for line_tax in line.taxes.all()] == ["test_code"]
             assert line.taxful_price.amount > line.taxless_price.amount
+
+    # Make sure order modification information is correct
+    assert order.creator != edited_order.modified_by
+    assert edited_order.modified_by == modifier
+    assert edited_order.modified_on is not None
```
suutari-ai: and this should be
```
assert edited_order.modified_on > edited_order.order_date
```

shoop_tests/admin/test_order_creator.py
```diff
@@ -279,11 +285,27 @@ def test_editing_existing_order(rf, admin_user):
 
     # Check that the product content is updated based on state
     assert edited_order.lines.count() == 5
-    assert edited_order.creator == admin_user
```
suutari-ai: Why this assert is removed? Shouldn't the creator still be admin_user? And that's important to check that it hasn't changed.

suutari-ai: Ah, sorry... now I see, that it was in fact just moved down. That's cool.

shoop_tests/campaigns/test_discount_codes.py
```diff
@@ -151,3 +155,8 @@ def test_campaign_with_coupons(rf):
     assert len(basket.get_final_lines()) == 3  # now basket has codes so they will be applied too
     assert OrderLineType.DISCOUNT in [l.type for l in basket.get_final_lines()]
 
+    basket.status = status
+    creator = OrderCreator(request)
+    order = creator.create_order(basket)
+    assert CouponUsage.objects.filter(order=order).count() == 1
+    assert CouponUsage.objects.filter(order=order, coupon__code=dc.code).count() == 1
```
suutari-ai: Good addition to this test.

shoop_tests/campaigns/test_order_edit_with_coupons.py
```diff
@@ -0,0 +1,145 @@
...
+        "shop": {
+            "selected": {
+                "id": shop.id,
+                "name": shop.safe_translation_getter("name"),
+                "currency": shop.currency,
+                "priceIncludeTaxes": shop.prices_include_tax
+            }
+        }
+    }
+    return state
```
suutari-ai: These were good tests. Nice work!

One nitpicking though: They're not really unit tests of the campaign module, but rather functional tests of the whole order editing process from coupons point of view, because they use many parts from non-core parts of Shoop too (shoop.admin, shoop.default_tax, shoop.front). Just move this file to shoop_tests/functional.

tulimaki:
retest this please

shoop/core/models/_orders.py
```diff
@@ -260,6 +261,7 @@ class Order(MoneyPropped, models.Model):
     require_verification = models.BooleanField(default=False, verbose_name=_('requires verification'))
     all_verified = models.BooleanField(default=False, verbose_name=_('all lines verified'))
     marketing_permission = models.BooleanField(default=True, verbose_name=_('marketing permission'))
+    _codes = JSONField(blank=True, null=True, verbose_name=_('codes'))
```
Pikkupomo: I think this should be `codes`

shoop/core/models/_orders.py
```diff
@@ -279,6 +281,19 @@ def __str__(self):  # pragma: no cover
         else:
             return "Order %s (%s)" % (self.identifier, name)
 
+    @property
+    def codes(self):
```
Pikkupomo: if the field is `codes` this could be function instead of property for example `get_codes`

shoop/core/models/_orders.py
```diff
@@ -279,6 +281,19 @@ def __str__(self):  # pragma: no cover
...
+        return list(self._codes or [])
+
+    @codes.setter
+    def codes(self, value):
```
Pikkupomo: there is nothing wrong with using `set_codes(values)` instead of setter.

shoop/core/models/_orders.py
tulimaki: I think the benefit we get here is that who ever read this sees that the attribute is private and shouldn't be touched directly.

suutari-ai: Yeah, and `Order.codes` can't be set to a dict for example, since a freely accessible JSONField would allow all kinds of misuse.

suutari-ai:
Good work @tulimaki!

PR #495, opened 2016-05-26
Add request parameter on get_payment_process_response method

chessbr:
Access to the user session may be useful for some kinds of payment processors.
Fix wrong variable name on payment_cancelled template.

### DIFF
--- a/shoop/core/models/_service_payment.py
+++ b/shoop/core/models/_service_payment.py
@@ -39,9 +39,9 @@ class PaymentMethod(Service):
     def can_delete(self):
         return not Order.objects.filter(payment_method=self).exists()
 
-    def get_payment_process_response(self, order, urls):
+    def get_payment_process_response(self, order, urls, request):
         self._make_sure_is_usable()
-        return self.provider.get_payment_process_response(self, order, urls)
+        return self.provider.get_payment_process_response(self, order, urls, request)
 
     def process_payment_return_request(self, order, request):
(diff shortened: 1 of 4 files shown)


tulimaki:
ok to test

tulimaki:
@chessbr for me this looks pretty harmless. Do you have some use case for this already? One minor issue for us is that we need to update some of our addons because of this change.

chessbr:
yeah.. I am developing a credit card payment gateway integration.. The gateway simply does not return to my `return_url` the `order_id` or anything else. This way I need to store the transaction id in the user session and take it when the payment gateway redirects to the `return_url`. I just can save this in the order extra data, but I don't think it will be good. Another case is when in the `get_payment_process_response` method I need to put a message of warning to the user, because some transactions could be denied here, before sending the user to the gateway payment authentication page. I don't know how are the gateway payments works there, but here we have one that some times does not need to redirect to their pages and then I got the error immediately, so I got a confirmed order without a valid payment.
The way that Shoop works now, I can see that the customer will be able to repay in his account in order history page - this will be needed.

tulimaki:
@chessbr still won't be merged like this. We are not convinced that we need that request as a parameter. Correct way for you to store the transaction id for later use would be to save it to ``CheckoutPhaseStorage`` which you can access through self.storage in checkout phase.

We will fix the payment_url in a seperate PR. Thanks for spotting that one.

PR #496, opened 2016-05-26
Admin: Create order detail section through provides

chessbr:
Fixed bug for custom user models


tulimaki:
ok to test

tulimaki:
Thanks! This is a great addition to our order admin.

PR #497, opened 2016-05-28
Front: Allow user to link company contact

shawnadelic:
Allow user to link a company contact from account page.

Refs SHOOP-2588


shoop/front/apps/customer_information/views.py
```diff
@@ -57,11 +36,49 @@ def form_valid(self, form):
         if shipping_address.pk != contact.default_shipping_address_id:  # Identity changed due to immutability
             contact.default_shipping_address = shipping_address
 
-        user.email = contact.email
-        user.first_name = contact.first_name
-        user.last_name = contact.last_name
-        user.save()
+        if not bool(get_company_contact(self.request.user)):  # Only update user details for non-company members
```
tulimaki: Why the user can't be updated here?

shoop/front/apps/customer_information/forms.py
```diff
@@ -0,0 +1,42 @@
...
+    def __init__(self, *args, **kwargs):
+        super(AddressForm, self).__init__(*args, **kwargs)
+        for field in ("email", "postal_code"):
+            self.fields[field].required = True
+
+
+class CompanyContactForm(forms.ModelForm):
+    class Meta:
+        model = CompanyContact
+        fields = ("name", "tax_number", "phone", "email")
```
tulimaki: I think marketing permission could be here too.

shoop/front/apps/customer_information/views.py
shawnadelic: Because then the username would switch every time the user edited either the person contact or the company contact, which I don't think was the intended behavior. We might be missing some kind of indication to the user which email would be linked to their account (in this case, the company email).

PR #501, opened 2016-05-30
Guide: Open search results in new tab

tulimaki:
Refs SHOOP-2608 / SHOOP-2644


shawnadelic:
Looks good, will merge once it passes

PR #502, opened 2016-05-31
Core: Fix decimal notation on model decimal fields (SHOOP-2616)

shawnadelic:
Fix decimal notation on model decimal fields so values do not display
in scientific notation (i.e., a value of `0` rendering as `0E-9`).

Refs SHOOP-2616

### DIFF
--- a/doc/changelog.rst
+++ b/doc/changelog.rst
@@ -12,6 +12,7 @@ Unreleased
 Core
 ~~~~
 
+- Fix bug: Do not display decimal values in scientific notation
 - Fix bug: Taxes of child order lines are filled incorrectly
 - Fix bug: Order line parent lines are not linked
 - Add modified on for ``Order``
--- a/shoop/core/fields/__init__.py
+++ b/shoop/core/fields/__init__.py
@@ -7,6 +7,8 @@
 # LICENSE file in the root directory of this source tree.
 from __future__ import unicode_literals
 
+import decimal
+
 import babel
 from django.core.exceptions import ImproperlyConfigured
 from django.core.validators import RegexValidator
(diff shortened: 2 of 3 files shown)


shoop/core/fields/__init__.py
```diff
@@ -60,23 +62,40 @@ def __init__(self, **kwargs):
...
+    """
+    DecimalField subclass that truncates zeroes to avoid displaying in
+    scientific notation.
+    """
+    def _remove_exponent(self, val):
+        context = decimal.Context(prec=self.max_digits)
+        if val == val.to_integral():
+            return val.quantize(decimal.Decimal(1), context=context)
+        else:
+            return val.normalize(context)
```
tulimaki: Decimals under 1 like 0.000000001 is still shown in scientific notation.

tulimaki:
Also I think the step shouldn't be in scientific notation either.

shoop/core/fields/__init__.py
```diff
@@ -60,23 +62,40 @@ def __init__(self, **kwargs):
...
+
+    def value_from_object(self, obj):
+        val = super(FormattedDecimalField, self).value_from_object(obj)
+        if isinstance(val, decimal.Decimal):
+            return self._remove_exponent(val)
```
suutari-ai: Maybe you could just use the new formatting operator (from py 2.7 / py 3), i.e. `'{:f}'.format(val)` or directly using the `format` function: `format(val, 'f')`.
```
>>> format(1/decimal.Decimal(10**100), 'f')
'0.0000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000001'
```

It will return a string,  so I'm not sure which function of the Field you should override, but I'm sure you'll figure that out. :)

doc/changelog.rst
```diff
@@ -12,6 +12,7 @@ Unreleased
 Core
 ~~~~
 
+- Fix bug: Do not display values in scientific notation
```
suutari-ai: Can you add word "decimal" to this? I.e. "... display decimal values in ..." To make it easier to find this changelog entry, since I guess "decimal" would be probable keyword people could use to search info about this bug)

shoop_tests/core/test_fields.py
```diff
@@ -7,10 +7,38 @@
...
+    product = get_default_product()
+    product.width = 0
+    product.save()
+    product = Product.objects.first()
+
+    form = TestModelForm(instance=product)
+    rendered_form = str(form)
+
+    assert "value=\"0E-9\"" not in rendered_form
+    assert "value=\"0\"" in rendered_form
```
suutari-ai: This test should be done without using the db, since it doesn't really need db for anything: just don't save the product and initialize its width to `Decimal('0E-9')`.

You should also parameterize this test for other values than just '0E-9' too, IMHO at least following should be tested `'0E-9', '0E-100', '1E-9', '-1E-9'` and also some values with many different digits in their value. Here's a list that I just tried in console:
```python
values_str = '0E-9 0E-30 1E-9 123E-10 -123E-10 1.12345666666666E20 1.23E-10000'
[(str(val), format(val, 'f')) for val in [decimal.Decimal(x) for x in values_str.split()]]
```

The last one is nasty, since its string representation contains 10004 characters. Also checked that it's possible to have even decimal values that generate 100 megabytes of characters when formatted: `len(format(decimal.Decimal('1.23E-100000000'), 'f'))`.  Probably the field should have some kind of maximum length. Maybe it could be truncated to something like 100 characters (`format(val, 'f')[:100]`). What do you think?

suutari-ai: Hmm.. having second thought about the truncation of long values, since that could lead to errors which are hard to find. Maybe we should just raise if the string representation would be too large. This could do it:

```python
def format_decimal(value, max_digits=100, exponent_limit=100)
    assert isinstance(value, decimal.Decimal)
    val = value.normalize()
    (sign, digits, exponent) = val.as_tuple()
    if exponent > exponent_limit:
        raise ValueError('Exponent too large for formatting: %r' % value)
    elif exponent < -exponent_limit:
        raise ValueError('Exponent too small for formatting: %r' % value)
    if len(digits) > max_digits:
        raise ValueError('Too many digits for formatting: %r' % value)
    return format(val, 'f')
```

suutari-ai:
Implementing this with the `FormattedDecimalField` class is a good idea, since it will fix the bug in many other places too.

suutari-ai:
Good work, Shawn. Thanks!

PR #503, opened 2016-06-01
OrderCreator: Fix parent line linking and taxing (SHOOP-2622)

suutari-ai:
Parent lines were never assigned for child lines, because the code was
simply incorrect.  It used `source_line.parent_line_id` even though package
products are not expanded in the source, i.e. the child lines don't exist
there. Rewrite the parent linkage code to fix it.

This also fixes how tax lines are set for the child lines.  Previously
child lines incorrectly got taxes of the parent line, since their source
line was set to the parent line.  Now child lines get taxes from the
corresponding child line in the source, if there is such source line. Child
lines that don't exist in the source won't get any taxes.  This works
correctly e.g. for the child product lines of a package product.

Also add test cases for the parent linkage and child line taxing bugs, plus
another test case for making sure that parent lines from order source are
also set in the order.

There's also a small docstring improvement / clean-up commit as a bonus.

### DIFF
--- a/doc/changelog.rst
+++ b/doc/changelog.rst
@@ -12,6 +12,8 @@ Unreleased
 Core
 ~~~~
 
+- Fix bug: Taxes of child order lines are filled incorrectly
+- Fix bug: Order line parent lines are not linked
 - Add modified on for ``Order``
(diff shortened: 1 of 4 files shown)


tulimaki:
Thanks!

PR #505, opened 2016-06-02
Add postal code range manager for TaxRules

Pikkupomo:
This will speed up tax loading alot.

Refs SHOOP-2607

### DIFF
--- a/shoop/default_tax/__init__.py
+++ b/shoop/default_tax/__init__.py
@@ -20,6 +20,7 @@ class AppConfig(shoop.apps.AppConfig):
         "tax_module": ["shoop.default_tax.module:DefaultTaxModule"],
         "admin_module": ["shoop.default_tax.admin_module:TaxRulesAdminModule"],
     }
-
+    def ready(self):
+        import shoop.default_tax.signals
 
 default_app_config = __name__ + ".AppConfig"
--- a/shoop/default_tax/migrations/0008_min_max_postalcode.py
+++ b/shoop/default_tax/migrations/0008_min_max_postalcode.py
@@ -1,0 +1,24 @@
+# -*- coding: utf-8 -*-
+from __future__ import unicode_literals
+
+from django.db import migrations, models
+
+
+class Migration(migrations.Migration):
+
+    dependencies = [
+        ('default_tax', '0007_verbose_names'),
+    ]
+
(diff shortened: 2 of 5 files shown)


suutari-ai:
OK, I think you got the general idea what I was after when proposing to add those min/max fields for postal code, country and region.

* Same should be done for country and region IMHO.

* The added min/max model fields should be private, since they are used only for caching and shouldn't be updated manually, i.e. prefix their names with an underscore. They don't even need public accessors, since their the queryset methods is the only API needed.

* You should use the `shoop.utils.patterns.Pattern` object for parsing the pattern.
  * Just add `get_alphabetical_limits` method to the class. It may return (min_string, max_string) or (None, None).
    * **Note** that the numerical/alphabetical matching might make unexpected values to match. See demonstration of this in PR #506. The value range should be alphabetic only to be able to do the matching in SQL. I don't see an easy way to generally calculate alphabetical limits for such numerical ranges having different length numbers as endpoints; maybe those should be just returned as (None, None), since usually postal code ranges have same length for both endpoints.

* Why don't you just override `TaxRule.save` instead of using the signal?

* The queryset should have a methods `may_match_postal_code(postal_code)`, `may_match_country_code(country_code)` and `may_match_region_code(region_code)` which should filter out those that cannot match given code, e.g. like this: 

  ```python
  null = Q(postal_codes_min=None)
  in_range = Q(postal_codes_min__lte=postalcode, postal_codes_max__gte=postalcode)
  return self.filter(null | in_range)
  ```

* The `get_taxes_of_effective_rules` should be updated to use those queryset filters:
  * API of the function should be updated: its `tax_rules` param should be a TaxRuleQuerySet. It's already called with a QuerySet, but the docstring promises it works also with any other iterable of TaxRule.
  * When the tax_rules param is a TaxRuleQuerySet, the function may use the new queryset methods to do some prefiltering at db level. This is the most important part of the optimization, since it is the part that avoids calling hundreds/thousands times `TaxRule.matches` method.

* You probably noticed that there's some problems which the test runners pointed out (coding style, missing license headers, unused import)

suutari-ai:
Closing this for now, since it's not yet in mergeable shape.

suutari-ai:
Continues in PR #508.

PR #507, opened 2016-06-02
Classic Gray: Fix "Retry payment" (SHOOP-2677)

suutari-ai:
The URL for "Retry payment" button was empty, because it used incorrect
property name `payment`.  Fix it to use the correct `payment_url` which
actually exists in `PaymentUrls` class.


suutari-ai:
Thanks for @chessbr for pointing this out in PR #495.

PR #508, opened 2016-06-02
Default Tax: Add min and max values to `TaxRule` (SHOOP-2616)

Pikkupomo:
Refs SHOOP-2616


tulimaki:
In commit messages can you be more clear why these changes is needed? Also can you add the changelog for these commits?

tulimaki:
Thanks!

PR #510, opened 2016-06-02
Hide cart and product order section when prices are hidden (SHOOP-2638)

tulimaki:
* Hide cart when prices are hidden
* Hide product order section when prices are hidden

### DIFF
--- a/shoop/themes/classic_gray/templates/classic_gray/navigation_basket_partial.jinja
+++ b/shoop/themes/classic_gray/templates/classic_gray/navigation_basket_partial.jinja
@@ -1,3 +1,4 @@
+{% if show_prices() %}
 <li class="dropdown cart">
     <a href="#" class="dropdown-toggle" data-toggle="dropdown">
         <i class="menu-icon fa fa-shopping-cart"></i>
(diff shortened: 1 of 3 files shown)


shoop/themes/classic_gray/templates/classic_gray/navigation_basket_partial.jinja
```diff
@@ -1,3 +1,4 @@
+{% if show_prices() %}
```
shawnadelic: Might be better to wrap this around and the <ul> in?

https://github.com/shoopio/shoop/blob/master/shoop/themes/classic_gray/templates/classic_gray/shoop/front/includes/_navigation.jinja#L104-L106

suutari-ai: Or we can wrap both, just to be sure.

suutari-ai:
Shawn is right that this `ul` element in `_navigation.jinja` should be wrapped too:
```jinja
<ul class="pull-right" id="navigation-basket-partial">
    {% include "classic_gray/navigation_basket_partial.jinja" with context %}
</ul>
```
I'll do that in a new PR (since can't add commits to this one).

PR #511, opened 2016-06-03
Admin: Add configurable default parameter to `ChoiceFilter` (SHOOP-2642)

shawnadelic:
Add configurable default parameter to picotable `ChoiceFilter`.

Refs SHOOP-2642

There is a small issue with the PR that I haven't been able to solve. Basically, the dropdown doesn't reflect the default selection value.

I have a WIP branch here that *should* work, but isn't currently:

https://github.com/shawnadelic/shoop/tree/picotable-default-initial-wip

### DIFF
--- a/doc/changelog.rst
+++ b/doc/changelog.rst
@@ -72,6 +72,7 @@ Localization
 Admin
 ~~~~~
 
+- Add default ``is_active`` filter to User admin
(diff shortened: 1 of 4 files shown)


tulimaki:
Ok, since there is a issue with default selection value I close this and let's reopen when the issue is solved.

PR #512, opened 2016-06-03
Admin: Enable contact deactivation (SHOOP-2236)

shawnadelic:
Add a login check so that inactive contacts cannot log into front.

Enable contact deactivation from Contact list view under Shoop Admin.

Refs SHOOP-2236

### DIFF
--- a/shoop/admin/modules/contacts/views/detail.py
+++ b/shoop/admin/modules/contacts/views/detail.py
@@ -8,19 +8,23 @@
 from __future__ import unicode_literals
 
 from django.conf import settings
+from django.contrib import messages
 from django.core.urlresolvers import reverse
 from django.db.models import Q
+from django.http.response import HttpResponseRedirect
 from django.utils.encoding import force_text
 from django.utils.translation import ugettext_lazy as _
 from django.views.generic import DetailView
 
 from shoop.admin.toolbar import PostActionButton, Toolbar, URLActionButton
 from shoop.core.models import CompanyContact, Contact, Order, PersonContact
+from shoop.utils.excs import Problem
 
 
 class ContactDetailToolbar(Toolbar):
-    def __init__(self, contact):
+    def __init__(self, contact, request):
         self.contact = contact
+        self.request = request
         self.user = getattr(self.contact, "user", None)
         super(ContactDetailToolbar, self).__init__()
         self.build()
@@ -58,6 +62,16 @@ class ContactDetailToolbar(Toolbar):
             extra_css_class="btn-gray btn-inverse",
         ))
 
+    def build_deactivate_button(self):
+        self.append(PostActionButton(
+            post_url=self.request.path,
+            name="set_is_active",
+            value="0" if self.contact.is_active else "1",
+            icon="fa fa-times-circle",
+            text=_(u"Deactivate Contact") if self.contact.is_active else _(u"Activate Contact"),
+            extra_css_class="btn-gray",
+        ))
+
     def build(self):
         self.append(URLActionButton(
             url=reverse("shoop_admin:contact.edit", kwargs={"pk": self.contact.pk}),
@@ -67,6 +81,7 @@ class ContactDetailToolbar(Toolbar):
         ))
         self.build_renew_password_button()
         self.build_new_user_button()
+        self.build_deactivate_button()
 
 
 class ContactDetailView(DetailView):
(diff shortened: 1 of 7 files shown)


shawnadelic:
Accidentally deleted this branch -_-

shoop/admin/modules/contacts/views/detail.py
```diff
@@ -58,6 +62,26 @@ def build_new_user_button(self):
             extra_css_class="btn-gray btn-inverse",
         ))
 
+    def build_deactivate_button(self):
+        if not self.contact.is_active:
```
tulimaki: This if else is not cool. Do it with help function or change the value/text based on the contact.is_active

shoop/admin/modules/contacts/views/detail.py
```diff
@@ -8,19 +8,23 @@
...
 from django.views.generic import DetailView
 
 from shoop.admin.toolbar import PostActionButton, Toolbar, URLActionButton
 from shoop.core.models import CompanyContact, Contact, Order, PersonContact
+from shoop.utils.excs import Problem
 
 
 class ContactDetailToolbar(Toolbar):
-    def __init__(self, contact):
+    def __init__(self, contact, path):
```
tulimaki: Instead of passing the path. Pass the request. The path is too specifig and the request might come handy in some point.

shoop/admin/modules/contacts/views/detail.py
```diff
@@ -90,9 +115,32 @@ def get_context_data(self, **kwargs):
...
         context["orders"] = Order.objects.filter(order_q).order_by("-id")
-        context["toolbar"] = ContactDetailToolbar(contact=self.object)
+        context["toolbar"] = ContactDetailToolbar(contact=self.object, path=self.request.path)
         context["title"] = "%s: %s" % (
             self.object._meta.verbose_name.title(),
             force_text(self.object)
         )
         return context
+
+    def _handle_set_is_active(self):
```
tulimaki: I would like to see test for this function.

tulimaki:
Thanks! :+1:

shoop/core/models/_contacts.py
```diff
@@ -359,7 +359,7 @@ def get_person_contact(user):
         'last_name': getattr(user, 'last_name', ''),
         'email': getattr(user, 'email', ''),
     }
-    return PersonContact.objects.get_or_create(user=user, defaults=defaults)[0]
+    return PersonContact.objects.filter(is_active=True).get_or_create(user=user, defaults=defaults)[0]
```
suutari-ai: Will this create a new contact if there already is inactive contact for the user?

shawnadelic: Yes, it looks like it will. Currently, I think this is the behavior if a contact doesn't exist.

For example, if you create a user from admin, they don't have a contact. However, on their first login a contact is created by this function in middleware that creates the contact. Not sure what is the correct behavior since (at least through our frontend) inactive contacts can't login. Maybe we should add a parameter that determines whether or not inactive contacts should be filtered out.

suutari-ai: It shouldn't create a new contact if there already is a contact for the user. The login was supposed to be disabled for those contacts and creating a new contact would allow them to login.

suutari-ai: I think it would better written as
```python
contact = PersonContact.objects.get_or_create(user=user, defaults=defaults)[0]
if contact.is_active:
    return contact
return None
```
And if we want to have that `allow_inactive`, then just add ` or allow_inactive` to the if. I don't know if it's needed, but wouldn't hurt though...

tulimaki: So since the in active contacts can't login this happens when the user is still logged in when it is deactivated. Also this will cause dublicate entry database error since the user is onetoone-field.

I think the correct fix here would be logout the user and show message that your account is deactivated.

I think this shouldn't never return None. Also user shouldn't be logged in without actual active contact in request.

Pikkupomo: I think @tulimaki is correct. If user is logged in and admin deactivates this user, next page load should make sure user is active, if not, user will be logged out and messages.error or such should be added to notify user why this happened.

Returning `None` here might cause a lot of problems too.

suutari-ai: > Also this will cause dublicate entry database error since the user is onetoone-field.

You're talking about the Shawn's implementation which does `.filter(...).get_or_create(...)`. Right? The snippet I posted, doesn't have that problem.

> I think the correct fix here would be to logout the user and show message that your account is deactivated.

That's not job of this function. It would be the middleware that should do that.

> I think this shouldn't never return None.

Hmm... I was about to disagree, because then it would be working differently than `get_company_contact`, but now that I looked for usages of `get_person_contact` I think you're right. It's probably wiser to always return a contact.

> Also user shouldn't be logged in without actual active contact in request.

Yeah, for Front. Login to Admin is OK without a customer in request. Preventing login to Front without active contact is implemented in `shoop.front.apps.auth.forms.EmailAuthenticationForm.confirm_login_allowed`. Though that currently does not seem to use this `get_person_contact` function, maybe it should?

suutari-ai: Ah, now I see that you already fixed this function in PR #542. Only open question is then the confirm_login_allowed, which IMHO should use the get_person_contact to avoid having two different ways for finding the person contact of an user.

PR #516, opened 2016-06-06
Hide cart and product order section when prices are hidden (SHOOP-2679)

suutari-ai:
* Hide cart when prices are hidden
* Hide product order section when prices are hidden
* Hide cart container ul element when cart is hidden.

This replaces PR #510.


suutari-ai:
Only difference to PR #510 is the last commit which just adds the if around the cart ul element as suggested by @shawnadelic. -> Merging this.

PR #517, opened 2016-06-06
Add Picotable  `ChoiceFilter` defaults and only show active users and contacts by default (SHOOP-2236, SHOOP-2642)

shawnadelic:
Refs SHOOP-2236, SHOOP-2642, SHOOP-2687

### DIFF
--- a/shoop/admin/static_src/picotable/picotable.js
+++ b/shoop/admin/static_src/picotable/picotable.js
@@ -228,6 +228,21 @@ const Picotable = (function(m, storage) {
 
     }
 
+    function getDefaultValues(ctrl) {
+        const data = ctrl.vm.data();
+        const filters = {};
+        for (var i = 0; i < data.columns.length; i++) {
+            if (data.columns[i].filter){
+                var value = data.columns[i].filter.defaultChoice;
+                if (value) {
+                    filters[data.columns[i].id] = value;
+                }
+            }
+        }
+
+        return filters;
+    }
+
     function buildColumnRangeFilter(ctrl, col, value) {
         var setFilterValueFromInput = function(which) {
             const filterObj = Util.extend({}, ctrl.getFilterValue(col.id) || {}); // Copy current filter object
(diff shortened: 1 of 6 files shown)


tulimaki:
I think there is a typo in third commit message.

shoop/admin/static_src/picotable/picotable.js
```diff
@@ -228,6 +228,20 @@ const Picotable = (function(m, storage) {
 
     }
 
+    function getDefaultValues(ctrl) {
+        var data = ctrl.vm.data();
+
+        var filters = {};
+        for (var i=0; i < data.columns.length; i++){
+            var value = data.columns[i].filter.defaultChoice;
+            if (value){
```
tulimaki: Missing whitespace

shoop/admin/static_src/picotable/picotable.js
```diff
@@ -228,6 +228,20 @@ const Picotable = (function(m, storage) {
 
     }
 
+    function getDefaultValues(ctrl) {
+        var data = ctrl.vm.data();
```
tulimaki: This should probably be const. Same goes with other variables in this function.

shoop/admin/static_src/picotable/picotable.js
shawnadelic: Do you mean between lines 234 and 235 or somewhere else?

shoop/admin/static_src/picotable/picotable.js
shawnadelic: What about line 236? Seems like that should stay var?

shoop/admin/static_src/picotable/picotable.js
tulimaki: I think it should be if (value) {

shoop/admin/static_src/picotable/picotable.js
tulimaki: maybe not in line 236

PR #519, opened 2016-06-06
Campaigns: Fix form population bug in Campaigns Admin

shawnadelic:
Fix form population bug in Campaigns Admin when populating
non-many-to-many fields.

Refs SHOOP-2617


tulimaki:
Thanks!

PR #522, opened 2016-06-07
Admin: Add product price rounding in order edit (SHOOP-2640)

tulimaki:
Round product prices to 2 decimals since the total is also rounded and more
than 2 decimals will cause problems with discounts.

### DIFF
--- a/shoop/admin/modules/orders/views/edit.py
+++ b/shoop/admin/modules/orders/views/edit.py
@@ -99,8 +99,8 @@ def get_line_data_for_edit(shop, line):
         "text": line.text,
         "quantity": line.quantity,
         "sku": line.sku,
-        "baseUnitPrice": line.base_unit_price.value,
-        "unitPrice": total_price / line.quantity if line.quantity else 0,
+        "baseUnitPrice": "{:.2f}".format(line.base_unit_price.value),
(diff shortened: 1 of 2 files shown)


shoop/admin/modules/orders/views/edit.py
```diff
@@ -99,8 +99,8 @@ def get_line_data_for_edit(shop, line):
         "text": line.text,
         "quantity": line.quantity,
         "sku": line.sku,
-        "baseUnitPrice": line.base_unit_price.value,
-        "unitPrice": total_price / line.quantity if line.quantity else 0,
+        "baseUnitPrice": "{:.2f}".format(line.base_unit_price.value),
+        "unitPrice": "{:.2f}".format((total_price / line.quantity) if line.quantity else 0),
```
suutari-ai: This is incorrect. Unit prices should not be rounded. Only the order total should be rounded. (And even then, it's not always 2 decimals, but rather currency dependent. The Currency class is not implemented yet though.)

You may render them rounded, if that's desired, but probably even that's not a good idea in the Admin's order editor, or at least it should show more decimals than just two to make the merchant aware that unit prices can use sub-cent precision.

PR #524, opened 2016-06-07
Admin: Sign in again in finnish language goes over the box (SHOOP-2691)

gurch101:
stack buttons vertically instead of horizontally.

Note layout bugs due to translations will likely become a more widespread issue. We should probably come up with a more general-purpose strategy (shrinking fonts/clipping/using ellipsis).


tulimaki:
Thanks! Also noted your concerns about these kind of issues with translations.

PR #527, opened 2016-06-08
Front: Allow user to change password from shop front (SHOOP-2704)

shawnadelic:
Note: I'm using a view function since that's what django.contrib.auth uses and it's simpler just to do it this way. :smile: 

Refs SHOOP-2704


suutari-ai:
Thanks, Shawn!

PR #528, opened 2016-06-09
Numbers in input fields increment/decrement in scientific notation (SHOOP-2746)

gurch101:
by default, django renders the step size in scientific notation when any decimal places
are present. This PR allows step size with up to 5 decimal places. If
FormattedDecimalFields need more decimal places, the step size becomes 1.


suutari-ai:
Good work, Gurchet! Thanks!

PR #531, opened 2016-06-09
Admin: Omit inactive contacts and users from multiselect (SHUUP-2745)

shawnadelic:
If a contact or user is marked as inactive, do not include in
multiselect search results.

Refs SHUUP-2745


suutari-ai:
Good work, Gurch! Nicely described the changes in the commit message, clean coding and a test case too. Great!

PR #536, opened 2016-06-10
Refactor campaign effects (SHUUP-2824)

Pikkupomo:
Campaigns gave the end results directly from campaign, this is not a good long term solution.

Refactored campaigns so that effects can be created separately from campaigns itself.

Refs SHUUP-2824

### DIFF
--- a/shoop/campaigns/forms.py
+++ b/shoop/campaigns/forms.py
@@ -14,7 +14,6 @@ from django.forms import ChoiceField, ModelForm
 from django.utils.encoding import force_text
 from django.utils.translation import ugettext_lazy as _
 
-from shoop.admin.forms.fields import PercentageField
 from shoop.admin.forms.widgets import Select2Multiple
 from shoop.apps.provides import get_provide_objects
 from shoop.campaigns.models.campaigns import (
@@ -26,8 +25,6 @@ CAMPAIGN_COMMON_FIELDS = [
     'name',
     'shop',
     "public_name",
-    'discount_percentage',
-    'discount_amount_value',
     'active',
     'start_datetime',
     'end_datetime'
@@ -36,13 +33,15 @@ CAMPAIGN_COMMON_FIELDS = [
 
 class CampaignFormMixin(object):
     rule_ids = set()
+    effect_ids = set()
     existing_rules = []
     rule_field = None
-    provide_keys = None
+    provide_keys = []
+    effect_keys = []
 
     def _get_values_for_identifier(self, key, identifier):
         """
-        Returns values from rule based on the key and the identifier.
+        Returns values from rule or effect based on the key and the identifier.
 
         Key can be either `condition` or `filter`.
 
@@ -50,21 +49,23 @@ class CampaignFormMixin(object):
           * `shoop.campaigns.basket_conditions.BasketCondition`
           * `shoop.campaigns.catalog_filters.CatalogFilter`
           * `shoop.campaigns.context_conditions.ContextCondition`
+          * `shoop.campaigns.basket_effects.BasketDiscountEffect`
+          * `shoop.campaigns.product_effects.ProductDiscountEffect`
 
-        Returns a queryset of objects or a single value found from conditions or filters.
+        Returns a queryset of objects or a single value found from conditions, filters or effects.
         Returns `None` if nothing is found
         """
         object_set = getattr(self.instance, "%ss" % key)
-        for condition_or_filter in object_set.all():
-            if condition_or_filter.identifier == identifier:
-                if condition_or_filter.model:
-                    return condition_or_filter.values.all()
+        for obj in object_set.all():
+            if obj.identifier == identifier:
+                if obj.model:
+                    return obj.values.all()
                 else:
-                    return condition_or_filter.value
+                    return obj.value
 
     def get_initial_value(self, identifier):
         if self.instance.pk:
-            for key in ["condition", "filter"]:
+            for key in ["condition", "filter", "effect"]:
                 if key in identifier:
                     return self._get_values_for_identifier(key, identifier)
 
@@ -73,36 +74,53 @@ class CampaignFormMixin(object):
         super(CampaignFormMixin, self).__init__(*args, **kwargs)
 
         for provide_key in self.provide_keys:
-            for cls in sorted(get_provide_objects(provide_key), key=lambda x: x.name):
-                obj = cls()
-                if obj.identifier in self.fields:
+            self._add_fields(provide_key)
+
+        for provide_key in self.effect_keys:
+            self._add_fields(provide_key, False)
+
+    def _add_fields(self, provide_key, is_provide=True):
+        for cls in sorted(get_provide_objects(provide_key), key=lambda x: x.name):
+            obj = cls()
+            if obj.identifier in self.fields:
+                continue
+
+            for field in obj._meta.get_fields(include_parents=False):
+                if isinstance(field, OneToOneField):
                     continue
 
-                for field in obj._meta.get_fields(include_parents=False):
-                    if isinstance(field, OneToOneField):
-                        continue
-
+                if is_provide:
                     self.rule_ids.add(obj.identifier)
-                    kwargs = {
-                        "required": False,  # rules are not required by default, validation is made later on
-                        "help_text": obj.description
-                    }
-                    if isinstance(field, ManyToManyField):
-                        kwargs.update({
-                            "initial": self._get_many_to_many_initial_values(obj),
-                            "widget": Select2Multiple(model=obj.model)
-                        })
-                        formfield = forms.CharField(**kwargs)
-                        value = kwargs.get("initial")
-                        if value:
-                            formfield.widget.choices = [
-                                (object.pk, force_text(object)) for object in obj.model.objects.filter(pk__in=value)]
-                    else:
-                        kwargs.update({
-                            "initial": self.get_initial_value(obj.identifier)
-                        })
-                        formfield = field.formfield(**kwargs)
-                    self.fields[obj.identifier] = formfield
+                else:
+                    self.effect_ids.add(obj.identifier)
+
+                kwargs = {
+                    "required": False,  # rules are not required by default, validation is made later on
+                    "help_text": obj.description
+                }
+
+                if hasattr(obj, "admin_form_class") and obj.admin_form_class:
+                    kwargs.update({
+                        "form_class": obj.admin_form_class
+                    })
+
+                if isinstance(field, ManyToManyField):
+                    kwargs.update({
+                        "initial": self._get_many_to_many_initial_values(obj),
+                        "widget": Select2Multiple(model=obj.model)
+                    })
+                    formfield = forms.CharField(**kwargs)
+                    value = kwargs.get("initial")
+                    if value:
+                        formfield.widget.choices = [
+                            (object.pk, force_text(object)) for object in obj.model.objects.filter(pk__in=value)]
+                else:
+                    kwargs.update({
+                        "initial": self.get_initial_value(obj.identifier)
+                    })
+                    formfield = field.formfield(**kwargs)
+
+                self.fields[obj.identifier] = formfield
 
     def _get_many_to_many_initial_values(self, obj):
         initial_value = self.get_initial_value(obj.identifier)
@@ -114,13 +132,19 @@ class CampaignFormMixin(object):
             return [initial_value]
 
     def clean(self):
-        """ Ensure at least one rule is set """
         data = super(CampaignFormMixin, self).clean()
-        if data.get("discount_percentage") and data.get("discount_amount_value"):
-            raise ValidationError(_("You should only define either discount percentage or amount."))
 
-        if not any([data.get("discount_percentage"), data.get("discount_amount_value")]):
-            raise ValidationError(_("You must define discount percentage or amount."))
+        # Ensure at least 1 effect is set
+        effect_count = 0
+        for key in self.effect_keys:
+            for cls in get_provide_objects(key):
+                if data[cls.identifier]:
+                    effect_count += 1
+
+                if effect_count > 1:
+                    raise ValidationError(_("You should only define one effect."))
+        if not effect_count:
+            raise ValidationError(_("At least one effect must be defined."))
 
         start_datetime = data.get("start_datetime")
         end_datetime = data.get("end_datetime")
@@ -128,18 +152,6 @@ class CampaignFormMixin(object):
             raise ValidationError(_("Campaign end date can't be before start date."), code="end_is_before_start")
 
         return data
-
-    def clean_discount_percentage(self):
-        percentage = self.cleaned_data.get("discount_percentage", 0)
-        if percentage and percentage < 0:
-            raise ValidationError(_("Discount percentage cannot be negative."))
-        return percentage
-
-    def clean_discount_amount_value(self):
-        amount = self.cleaned_data.get("discount_amount_value", 0)
-        if amount and amount < 0:
-            raise ValidationError(_("Discount amount cannot be negative."))
-        return amount
 
     def save(self, commit=True):
         instance = super(CampaignFormMixin, self).save(commit)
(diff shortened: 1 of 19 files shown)


shoop/campaigns/forms.py
```diff
@@ -73,36 +72,53 @@ def __init__(self, *args, **kwargs):
...
-                    if isinstance(field, OneToOneField):
-                        continue
+    def _add_fields(self, provide_key, is_provide=True):
+        for cls in sorted(get_provide_objects(provide_key), key=lambda x: x.name):
+            obj = cls()
+            if obj.identifier in self.fields:
+                continue
+
+            for field in obj._meta.get_fields(include_parents=False):
+                if isinstance(field, OneToOneField):
```
tulimaki: This probably could be in it's own function. This feels too complex.

shoop/campaigns/models/basket_effects.py
```diff
@@ -0,0 +1,90 @@
...
+from shoop.core.fields import MoneyValueField
+from shoop.utils.properties import PriceProperty
+
+
+class BasketEffect(PolymorphicModel):
+    identifier = None
+    model = None
+    admin_form_class = None
+    yields = False
+    name = _("Basket effect")
```
tulimaki: I think the name Basket effect here is unnesseray since I think the name should be always defined by subclass.

shoop/campaigns/models/campaigns.py
```diff
@@ -75,13 +68,12 @@ def is_available(self):
...
     def save(self, **kwargs):
-        if self.discount_percentage and self.discount_amount_value:
-            raise ValidationError(_("You should only define either discount percentage or amount."))
-
-        if not (self.discount_percentage or self.discount_amount_value):
-            raise ValidationError(_("You must define discount percentage or amount."))
-
         super(Campaign, self).save(**kwargs)
+        if not self.effects:
+            raise ValidationError(_("At least one effect must be defined."))
```
tulimaki: You shouldn't raise ValidationError from save. I think you could put these on clean.

Pikkupomo: They are on clean also, this is the same functionality there was before. This here is to prevent direct saving without effects.

shoop/campaigns/forms.py
```diff
@@ -114,13 +130,19 @@ def _get_many_to_many_initial_values(self, obj):
...
+        effect_count = 0
+        for key in self.effect_keys:
+            for cls in get_provide_objects(key):
+                if data[cls.identifier]:
+                    effect_count += 1
+
+                if effect_count > 1:
+                    raise ValidationError(_("You should only define one effect."))
+        if not effect_count:
+            raise ValidationError(_("At least one effect must be defined."))
```
suutari-ai: Why is effect a m2m field if there could be just one per campaign?

Though, I might see what you're after here: Currently just one effect per campaign is supported, but that might be changed in the future. And having it as m2m would avoid doing migrations then. Is this the reason for m2m?

shoop/campaigns/forms.py
```diff
@@ -180,11 +182,40 @@ def save(self, commit=True):
...
         return instance
 
+    def _save_effects(self, effect_key, instance):
+        for cls in get_provide_objects(effect_key):
+            if not self.cleaned_data[cls.identifier]:
+                continue
+            field_value = self.cleaned_data[cls.identifier]
+            effect = cls.objects.create()
+            if effect.model:
+                effect.values = effect.model.objects.filter(pk__in=ast.literal_eval(field_value))
```
suutari-ai: Hmm.. `ast.literal_eval` doesn't seem like a right tool here. Can't the field_value be just a list of ids? I.e. as comma separated string, so then `pks = map(int, field_value.split(','))` would do.

suutari-ai: OK, it seems that ast.literal_eval was there already. No need to change it then.

shoop/campaigns/migrations/0006_campaign_effects.py
```diff
@@ -0,0 +1,197 @@
...
+from __future__ import unicode_literals
+
+from django.db import migrations, models
+import shoop.core.fields
+
+cached_values = []
+
+# FORWARDS
+def cache_values(apps, schema_editor):
+    print("\n--- Caching values ---")
```
suutari-ai: I don't know if it's good idea to add print statements to migrations. Maybe it would be better to just leave them out to not mess with anything that doesn't expect writes to stdout during migrations. (It could cause failures e.g. if stdout is closed.)

shoop/campaigns/migrations/0006_campaign_effects.py
```diff
@@ -0,0 +1,197 @@
...
+
+    def cache_value(c, cache_type):
+        if not c.discount_amount_value and not c.discount_percentage:
+            return
+        cached_values.append({
+            "id": c.id,
+            "type": cache_type,
+            "amount": c.discount_amount_value,
+            "percentage": c.discount_percentage
+        })
```
suutari-ai: "Caching" these to memory might be bad idea, because (a) there could be so many campaigns that there is not enough memory and (b) if the migration is aborted for some reason, the values will be lost. Even transaction won't help since migration invokes DDL statements which mess up at least MySQL's transactions.

shoop/campaigns/migrations/0006_campaign_effects.py
```diff
@@ -0,0 +1,197 @@
...
+            model_name='basketcampaign',
+            name='effects',
+            field=models.ManyToManyField(to='campaigns.CampaignEffect', related_name='basket_campaigns', blank=True, verbose_name='effects', help_text='When campaign is triggered, these effects give end result.'),
+        ),
+        migrations.AddField(
+            model_name='catalogcampaign',
+            name='effects',
+            field=models.ManyToManyField(to='campaigns.CampaignEffect', related_name='catalog_campaigns', blank=True, verbose_name='effects', help_text='When campaign is triggered, these effects give end result.'),
+        ),
+        migrations.RunPython(add_effects, cache_effects)
```
suutari-ai: Why don't you just move the RemoveFields to bottom and do the migration in db just after everything new is created but before the fields are removed? That way you wouldn't need to story anything to memory and there should be no risk of losing data if migration is aborted.

shoop/campaigns/models/campaign_effects.py
```diff
@@ -0,0 +1,110 @@
...
+    admin_form_class = None
+    yields = False
+
+    def set_locals(self, **kwargs):
+        self.order_source = kwargs.get("order_source")
+        self.price_info = kwargs.get("price_info")
+        self.context = kwargs.get("context")
+        self.product = kwargs.get("product")
+        self._campaign = kwargs.get("campaign")
+        self.set_type()
```
suutari-ai: These shouldn't be here in CampaignEffect but rather in a separate class like `CampaignApplyingContext`. Then instance of that class should be passed to `apply`.

Also: Describe what are `price_info`, `context` and `product` in that list. I don't see which price, what context (pricing context? something else?) and which product they are referring to.

shoop/campaigns/models/campaign_effects.py
```diff
@@ -0,0 +1,110 @@
...
+
+    def set_type(self):
+        from shoop.campaigns.models import BasketCampaign
+        self.type = EffectType.BASKET if isinstance(self._campaign, BasketCampaign) else EffectType.CATALOG
```
suutari-ai: No need for set_type here, should be part of `CampaignApplyingContext`.

shoop/campaigns/models/campaign_effects.py
```diff
@@ -0,0 +1,110 @@
...
+    def apply(self, **kwargs):
+        raise NotImplementedError("Not implemented!")
+
+    @property
+    def shop_currency(self):
+        return self._campaign.shop.currency
+
+    @property
+    def shop_prices_include_tax(self):
+        return self._campaign.shop.prices_include_tax
```
suutari-ai: No need for shop_currency or shop_prices_include_tax here, maybe not even in `CampaignApplyingContext`. Prices should be created with `order_source.create_price`.

shoop/campaigns/models/campaign_effects.py
```diff
@@ -0,0 +1,110 @@
...
+    def description(self):
+        return _("Give discount amount.")
+
+    @property
+    def value(self):
+        return self.discount_amount_value
+
+    @value.setter
+    def value(self, value):
+        self.discount_amount_value = value
```
suutari-ai: Is the `value` property (getter and setter) really needed?

shoop/campaigns/models/campaign_effects.py
```diff
@@ -0,0 +1,110 @@
...
+        self.product = kwargs.get("product")
+        self._campaign = kwargs.get("campaign")
+        self.set_type()
+
+    def set_type(self):
+        from shoop.campaigns.models import BasketCampaign
+        self.type = EffectType.BASKET if isinstance(self._campaign, BasketCampaign) else EffectType.CATALOG
+
+    def apply(self, **kwargs):
+        raise NotImplementedError("Not implemented!")
```
suutari-ai: Should take the `CampaignApplyingContext` as parameter.

suutari-ai: Add docstring that describes the function, its input types and return type.

shoop/campaigns/models/campaign_effects.py
```diff
@@ -0,0 +1,110 @@
...
+        return self.discount_amount_value
+
+    @value.setter
+    def value(self, value):
+        self.discount_amount_value = value
+
+    def apply(self, **kwargs):
+        self.set_locals(**kwargs)
+        if self.type == EffectType.CATALOG:
+            create_price = self.context.shop.create_price
```
suutari-ai: Use `order_source.create_price` to create prices. That would be nicer for future multi-currency support.

shoop/campaigns/models/campaign_effects.py
```diff
@@ -0,0 +1,110 @@
...
+    @property
+    def shop_prices_include_tax(self):
+        return self._campaign.shop.prices_include_tax
+
+
+class DiscountAmount(CampaignEffect):
+    identifier = "discount_amount_effect"
+    name = _("Discount amount value")
+
+    discount_amount = PriceProperty("discount_amount_value", "shop_currency", "shop_prices_include_tax")
```
suutari-ai: Since CampaignEffect is m2m to campaign, this just doesn't work. Or needs some hacks like the `set_locals` (which shouldn't be needed).

Wouldn't it be better to have the campaign FK in CampaignEffect? Then this property could simply be expressed as

```python
discount_amount = PriceProperty(
    "discount_amount_value",
    "campaign.shop.currency", "campaign.shop.prices_include_tax")
```

This would still allow having multiple effects for a campaign (if we want that later), but it would tie an effect to single campaign, which IMHO is a good thing.

shoop/campaigns/models/campaigns.py
```diff
@@ -74,22 +66,16 @@ def is_available(self):
...
 
 class CatalogCampaign(Campaign):
     _queryset = None
 
     admin_url_suffix = "catalog_campaigns"
     conditions = models.ManyToManyField('ContextCondition', blank=True, related_name='campaign')
     filters = models.ManyToManyField('CatalogFilter', blank=True, related_name='campaign')
+    effects = models.ManyToManyField(
+        CampaignEffect, verbose_name=_("effects"), blank=True,
+        help_text=_("When campaign is triggered, these effects give end result."), related_name='catalog_campaigns')
```
suutari-ai: Effect should probably be linked to campaign through a `campaign` FK in `CampaignEffect` model. See my comments above.

That would mean changing the campaign to PolymorphicModel though, or using generic foreign key.

suutari-ai:
I think the idea here is very good: Campaign and its effect should be separate objects as you propose. 

As seen in the inline comments, there's still some implementation details that should be fixed though. Most of them should be quite easy to fix. Only changing campaign to be a polymorphic model could be a bit tricky, if migrations don't play along nicely. I may help you with that if needed...

shoop/campaigns/forms.py
Pikkupomo: yes, this is the very reason. We start by having one (to mimic current features) and move forward with multiple in future.

shoop/campaigns/models/campaign_effects.py
Pikkupomo: order source is available only in `BasketCampaigns`

shoop/campaigns/models/campaign_effects.py
Pikkupomo: `order_source` is not available through `CatalogCampaign`s

shoop/campaigns/forms.py
```diff
@@ -36,9 +33,11 @@
...
 class CampaignFormMixin(object):
     rule_ids = set()
+    effect_ids = set()
     existing_rules = []
     rule_field = None
-    provide_keys = None
+    provide_keys = []
+    effect_keys = []
 
     def _get_values_for_identifier(self, key, identifier):
```
tulimaki: Update docstring for this method.

shoop/campaigns/migrations/0006_effects.py
```diff
@@ -0,0 +1,164 @@
+# -*- coding: utf-8 -*-
+from __future__ import unicode_literals
+
+from django.db import migrations, models
+import shoop.core.fields
+
```
tulimaki: Use two blank lines here

shoop/campaigns/migrations/0006_effects.py
```diff
@@ -0,0 +1,164 @@
...
+        else:
+            new_ct = ContentType.objects.get_for_model(BasketDiscountPercentage)
+            e = BasketDiscountPercentage()
+            e.discount_percentage = c.discount_percentage
+        e.campaign = c
+        e.save()
+
+        BasketCampaignEffect.objects.filter(polymorphic_ctype__isnull=True, campaign_id=c.id).update(polymorphic_ctype=new_ct)
+
+    for c in CatalogCampaign.objects.using(db_alias).all():
```
tulimaki: Looks like you are using exactly same logic to handle both BasketCampaigns and CatalogCampaigns. I think you should make this forwards func little bit prettier.

shoop/campaigns/models/campaigns.py
```diff
@@ -74,14 +69,9 @@ def is_available(self):
...
-    def save(self, **kwargs):
-        if self.discount_percentage and self.discount_amount_value:
-            raise ValidationError(_("You should only define either discount percentage or amount."))
-
-        if not (self.discount_percentage or self.discount_amount_value):
-            raise ValidationError(_("You must define discount percentage or amount."))
-
-        super(Campaign, self).save(**kwargs)
+    @property
+    def type(self):
```
tulimaki: Where is type used?

Pikkupomo: nowhere now, it was added because @suutari-ai said campaigns should have such property.

shoop/campaigns/forms.py
Pikkupomo: Good catch, forgot that!

shoop/front/apps/simple_order_notification/templates.py
```diff
@@ -15,9 +15,7 @@
 For reference, here's a list of your order's contents.
 
 {% for line in order.lines.all() %}
-{%- if line.taxful_price %}
 * {{ line.quantity }} x {{ line.text }} - {{ line.taxful_price|money }}
```
tulimaki: This should probably be it's own commit.

tulimaki:
Why are translations in it's own commit? Also can you write commit message for these commits.

Pikkupomo:
Retest this please

shoop/campaigns/models/catalog_effects.py
```diff
@@ -0,0 +1,82 @@
...
+
+    def apply_for_product(self, applying_context):
+        """
+        Applies the effect based on given context or yields a new `OrderLine`
+
+        :type context: `shoop.campaigns.modules.CampaignApplyingContext`
+        :return: yields `OrderLine` or returns a `Price` based on context
+        :rtype: `OrderLine`, `Price`
+        """
+        raise NotImplementedError("Not implemented!")
```
suutari-ai: Now that the applying context has only 4 variables and one of them is redundant (campaign), maybe it would be better to just pass those variables as arguments and ditch the CampaignApplyingContext.

So this would be:
```python
def apply_for_product(self, context, product, price_info):
    """
    Apply this effect for given product in given context.

    :type context: shoop.core.pricing.PricingContext   # ?? check this
    :type product: shoop.core.models.Product
    :type price_info: shoop.core.pricing.PriceInfo
    :rtype: PriceInfo
    """
    pass
```

PR #540, opened 2016-06-13
Admin: Show stock on order creation (SHUUP-2730)

gurch101:
Show physical and logical stock counts to merchant when he is creating
an order.

Refs SHUUP-2730


tulimaki:
Thanks!

PR #541, opened 2016-06-13
Admin: Add product barcode filter (SHUUP-2681)

gurch101:
Add new field to table to filter products by barcode.

Refs SHUUP-2681


shoop/admin/modules/products/views/list.py
```diff
@@ -18,6 +18,7 @@
 class ProductListView(PicotableListView):
     model = Product
     columns = [
+        Column("barcode", _(u"Barcode"), display="barcode", filter_config=TextFilter(_("Filter by barcode..."))),
```
tulimaki: I think barcode shouldn't be first column. Maybe move this at least after name?

PR #542, opened 2016-06-13
Fix for deactivating contacts (SHUUP-2840)

tulimaki:
* Core: Do not create multiple contacts for user
* Front: Logout if user is linked to inactive contact

### DIFF
--- a/shoop/front/middleware.py
+++ b/shoop/front/middleware.py
@@ -6,12 +6,15 @@
 # This source code is licensed under the AGPLv3 license found in the
 # LICENSE file in the root directory of this source tree.
 from django.conf import settings
+from django.contrib import messages
+from django.contrib.auth import logout
 from django.contrib.auth.models import AnonymousUser
 from django.contrib.auth.signals import user_logged_in, user_logged_out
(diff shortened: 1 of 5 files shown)


shoop/front/middleware.py
```diff
@@ -77,6 +80,9 @@ def _set_shop(self, request):
 
     def _set_person(self, request):
         request.person = get_person_contact(request.user)
+        if not request.person.is_active:
+            messages.add_message(request, messages.INFO, _("Logged out since this account is inactive."))
+            logout(request)
```
suutari-ai: This should also set the request.person to AnonymousContact. I'll create a PR for that.

suutari-ai: Hmm.. it seems to happen already. Didn't first see why, but then found out that it's updated by the signal. I'll add a comment, since otherwise this code is hard to understand. Maybe it's still better to explicitly set it to anonymous too, just to be sure (signal might not be connected for some reason).

suutari-ai: PR #543.

PR #550, opened 2016-06-16
Tune order editor/creator prices (SHUUP-2853, SHUUP-2640)

tulimaki:
* Admin order creator: Return unit prices when fetching product data
* Admin order creator: Reduce discount amount precision to two decimals
* Admin order edit: Do not preserve lines with parent

### DIFF
--- a/shoop/admin/modules/orders/static_src/create/reducers/lines.js
+++ b/shoop/admin/modules/orders/static_src/create/reducers/lines.js
@@ -45,7 +45,7 @@ function getFormattedStockCounts(line) {
     return {
         physicalCount: physicalCount.toFixed(line.salesDecimals) + ' ' + line.salesUnit,
         logicalCount: logicalCount.toFixed(line.salesDecimals) + ' ' + line.salesUnit
-    }
+    };
 }
 
 function getDiscountsAndTotal(quantity, baseUnitPrice, unitPrice, updateUnitPrice=false) {
@@ -55,7 +55,7 @@ function getDiscountsAndTotal(quantity, baseUnitPrice, unitPrice, updateUnitPric
         updates.unitPrice = unitPrice;
     }
     var totalBeforeDiscount = baseUnitPrice * quantity;
-    var total = +(unitPrice * quantity).toFixed(2);
+    var total = +(unitPrice * quantity);
     updates.total = total;
     if (baseUnitPrice < unitPrice || unitPrice < 0) {
         updates.discountPercent = 0;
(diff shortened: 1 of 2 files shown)


shoop/admin/modules/orders/static_src/create/reducers/lines.js
```diff
@@ -54,15 +54,15 @@ function getDiscountsAndTotal(quantity, baseUnitPrice, unitPrice, updateUnitPric
...
+    var totalBeforeDiscount = (baseUnitPrice * quantity).toFixed(2);
     var total = +(unitPrice * quantity).toFixed(2);
     updates.total = total;
     if (baseUnitPrice < unitPrice || unitPrice < 0) {
         updates.discountPercent = 0;
         updates.discountAmount = 0;
         return updates;
     }
-    var discountAmount = totalBeforeDiscount - total;
+    var discountAmount = (totalBeforeDiscount - total).toFixed(2);
```
suutari-ai: The precision shouldn't be cut to 2 decimals, since line prices should use higher precision than currency's precision. It seems that the total is also cut to 2 decimals (in line 58), but that is incorrect too. Would it work better if you just remove that?

Order total should be rounded to currency's precision though. If that's not already done in JS, then it should be implemented.

Also, we eventually want currency's precision to be configurable too. So if the 2 decimals really needs to be hardcoded somewhere, please add `SHOOP-1912` tag to those lines.

suutari-ai:
I don't feel very good about doing these price calculations in JS code for two reasons:
 * (1) It uses floats, but money calculations should be done with proper decimal arithmetic.
 * (2) More importantly: It duplicates the same logic to two places: Python code and JS code. Business logic like this should be only in one place which should decide about details like rounding.

To resolve (1), we may add something like BigDecimal for JS (see e.g. http://stackoverflow.com/questions/1458633/how-to-deal-with-floating-point-number-precision-in-javascript), but it would still need the same logic be in two places, which is suboptimal. I don't think it's worth it. If we do something about this, then it's better to attack (2) right away.

To resolve (2), I see two options:
  * A: The order editor sends the numbers to backend (through AJAX) for doing calculations and then just uses the results as is. That would make the UI more laggy though, but probably it would still be usable.
 * B: Put the price calculation to its own JS module which will be used by backend too. That would be cool, but then we would need a JS engine available for the backend = new dependency. (Maybe PyExecJS with either Node.js or PyV8 could work.)

Anyway, we're not fixing this yet, so we'll just have to pick the best compromise. IMHO that would be replicating the price calculation logic of the backend in JS as closely as possible: Use as much precision as possible with floats, do rounding in same places and hope for the best.

suutari-ai:
The other two commits are fine. (Would have merged them if they weren't in the same PR.)

shoop/admin/modules/orders/static_src/create/reducers/lines.js
tulimaki: Ok, so. Yes it works if we remove the rounding hfrom JS all together (from lines). That just leads to a lot of decimal points that might confuse merchant. It seems that this is our compromise now based on your other comment.

Also yes the order total at the bottom of editor is using 2 decimals already which I am going leave like it is.

tulimaki:
retest this please

suutari-ai:
Good work @tulimaki and @Pikkupomo! Thanks!

PR #552, opened 2016-06-16
Add cash payment method (SHUUP-2723)

gurch101:
Add cash payments to the custom payment processor which sets created orders
to fully paid on creation. Also add rounding behavior component to round
order totals to the nearest 0.05.

Refs SHUUP-2723

### DIFF
--- a/shoop/admin/modules/orders/receivers.py
+++ b/shoop/admin/modules/orders/receivers.py
@@ -1,0 +1,30 @@
+# -*- coding: utf-8 -*-
+# This file is part of Shoop.
+#
+# Copyright (c) 2012-2016, Shoop Ltd. All rights reserved.
+#
+# This source code is licensed under the AGPLv3 license found in the
+# LICENSE file in the root directory of this source tree.
+from django.utils.timezone import now
+
+from shoop.core.models import CustomPaymentProcessor
+
+
+# TEMPORARY until we get admin orders also calling service methods
+
+
+def _create_cash_payment_for_order(order):
+    if not order.is_paid():
+        order.create_payment(
+            order.taxful_total_price,
+            payment_identifier="Cash-%s" % now().isoformat(),
+            description="Cash Payment"
+        )
+
+
+def handle_custom_payment_return_requests(sender, order, *args, **kwargs):
+    payment_processor = order.payment_method.payment_processor if order.payment_method else None
+    if isinstance(payment_processor, CustomPaymentProcessor):
+        service = order.payment_method.choice_identifier
+        if service == "cash":
+            _create_cash_payment_for_order(order)
--- a/shoop/core/models/_service_behavior.py
+++ b/shoop/core/models/_service_behavior.py
@@ -7,12 +7,16 @@
 
 from __future__ import unicode_literals
 
+import decimal
+
 from django.core.exceptions import ValidationError
 from django.db import models
 from django.utils.translation import ugettext_lazy as _
+from enumfields import Enum, EnumField
 from parler.models import TranslatableModel, TranslatedField, TranslatedFields
 
 from shoop.core.fields import MeasurementField, MoneyValueField
+from shoop.utils.numbers import nickel_round
 
 from ._service_base import (
     ServiceBehaviorComponent, ServiceCost,
@@ -168,3 +172,39 @@ class GroupAvailabilityBehaviorComponent(ServiceBehaviorComponent):
         groups_to_match = set(self.groups.all().values_list("pk", flat=True))
         if not bool(customer_groups & groups_to_match):
             yield ValidationError(_("Service is not available for any of the customers groups."))
+
+
+class RoundingMode(Enum):
+    ROUND_HALF_UP = decimal.ROUND_HALF_UP
+    ROUND_HALF_DOWN = decimal.ROUND_HALF_DOWN
+    ROUND_UP = decimal.ROUND_UP
+    ROUND_DOWN = decimal.ROUND_DOWN
+
+    class Labels:
+        ROUND_HALF_UP = _("round to nearest with ties going away from zero")
+        ROUND_HALF_DOWN = _("round to nearest with ties going towards zero")
+        ROUND_UP = _("round away from zero")
+        ROUND_DOWN = _("round towards zero")
+
+
+class RoundingBehaviorComponent(ServiceBehaviorComponent):
+    name = _("Rounding")
+    help_text = _("Round total order price to the nearest quant.")
+
+    quant = models.DecimalField(
+        max_digits=36, decimal_places=9, default=decimal.Decimal('0.05'),
+        verbose_name=_("rounding quant"))
+    mode = EnumField(
+        RoundingMode,
+        default=RoundingMode.ROUND_HALF_UP,
+        verbose_name=_("rounding mode"))
+
+    def get_costs(self, service, source):
+        total_price = source.total_price_of_products
+        rounded = nickel_round(total_price, self.quant, self.mode.value)
+        remainder = rounded - total_price
+        yield ServiceCost(remainder, _("rounding"))
+
+    def get_unavailability_reasons(self, service, source):
+        if not source.creator or not(source.creator.is_superuser or source.creator.is_staff):
+            yield ValidationError(_("Only administrators can process cash payments"))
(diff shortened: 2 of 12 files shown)


shoop/core/models/_service_behavior.py
```diff
@@ -164,3 +167,42 @@ def get_unavailability_reasons(self, service, source):
...
+
+class RoundingBehaviorComponent(ServiceBehaviorComponent):
+    name = _("Rounding")
+    help_text = _("Round total order price to the nearest five cents.")
+
+    rounding_behavior = EnumIntegerField(
+        RoundingBehavior,
+        default=RoundingBehavior.ROUND_DOWN,
+        verbose_name=_("rounding behavior"))
+
```
shawnadelic: This behavior component should probably not go in core, but should be provided by the AppConfig.

shoop/admin/modules/orders/receivers.py
```diff
@@ -0,0 +1,30 @@
...
+            payment_identifier="Cash-%s" % now().isoformat(),
+            description="Cash Payment"
+        )
+
+
+def handle_custom_payment_return_requests(sender, order, *args, **kwargs):
+    payment_processor = order.payment_method.payment_processor if order.payment_method else None
+    if isinstance(payment_processor, CustomPaymentProcessor):
+        service = order.payment_method.choice_identifier
+        if service == 'cash':
```
tulimaki: Add handling also for "manual" payment service. Use double quotes for cash.

shoop/core/models/_service_behavior.py
```diff
@@ -164,3 +167,42 @@ def get_unavailability_reasons(self, service, source):
...
+        rounded_amt = 0
+        if self.rounding_behavior == RoundingBehavior.ROUND_UP:
+            rounded_amt = self.round_up(total_price) - total_price
+        else:
+            rounded_amt = self.round_down(total_price) - total_price
+        rounded_amt = source.create_price(decimal.Decimal(rounded_amt))
+        yield ServiceCost(rounded_amt, self.rounding_behavior.label)
+
+    def get_unavailability_reasons(self, service, source):
+        if not source.creator or not source.creator.is_superuser:
```
tulimaki: Allow also is_staff here.

shoop/core/models/_service_behavior.py
```diff
@@ -168,3 +171,42 @@ def get_unavailability_reasons(self, service, source):
...
+    rounding_behavior = EnumIntegerField(
+        RoundingBehavior,
+        default=RoundingBehavior.ROUND_DOWN,
+        verbose_name=_("rounding behavior"))
+
+    def round_down(self, amount):
+        return (amount * 2).quantize(decimal.Decimal('.1'), rounding=decimal.ROUND_DOWN) / 2
+
+    def round_up(self, amount):
+        return (amount * 2).quantize(decimal.Decimal('.1'), rounding=decimal.ROUND_UP) / 2
```
suutari-ai: The rounding "quant", which seems to be 0.05 here, should be configurable: Add it as field to the model. Also the so called "nickel rounding" algorithm could be useful as a separate util. You should probably define the nickel rounding algorithm in shoop.utils.numbers. It could be something like this:
```python
def nickel_round(value, quant=Decimal('0.05'), rounding=decimal.ROUND_HALF_UP):
    """
    Round decimal value to nearest quant.

    For example:

    >>> assert nickel_round(Decimal('10.31')) == Decimal('10.30')
    >>> assert nickel_round(Decimal('10.33')) == Decimal('10.35')
    >>> assert nickel_round(Decimal('10.325')) == Decimal('10.35')
    >>> assert nickel_round(Decimal('10.3249')) == Decimal('10.30')
    >>> assert nickel_round(Decimal('10.31'), Decimal('0.125')) == Decimal('10.250')
    >>> assert nickel_round(Decimal('10.32'), Decimal('0.125')) == Decimal('10.375')
    """
    assert isinstance(value, Decimal)
    return (value / quant).quantize(1, rounding=rounding) * quant
```

suutari-ai: Btw. since our Money and Price classes are also subclasses of Decimal, the algorithm works fine with them too:
```python
In [12]: nickel_round(TaxfulPrice('10.241', 'EUR'))
Out[12]: TaxfulPrice('10.25', 'EUR')
```

suutari-ai: I created PR #555 for adding that function. You may merge that and rebase your PR on top.

shoop/core/models/_service_behavior.py
```diff
@@ -168,3 +172,39 @@ def get_unavailability_reasons(self, service, source):
...
+    class Labels:
+        ROUND_UP = _("round up")
+        ROUND_DOWN = _("round down")
+
+
+class RoundingBehaviorComponent(ServiceBehaviorComponent):
+    name = _("Rounding")
+    help_text = _("Round total order price to the nearest quant.")
+
+    rounding_quant = models.DecimalField(
```
tulimaki: This is never used.

shoop/core/models/_service_behavior.py
```diff
@@ -168,3 +172,39 @@ def get_unavailability_reasons(self, service, source):
...
+        max_digits=36, decimal_places=9,
+        verbose_name=_("rounding quant"))
```
suutari-ai: I think `Decimal('0.05')` would be a good default value for the `rounding_quant`.

shoop/core/models/_service_behavior.py
```diff
@@ -168,3 +172,39 @@ def get_unavailability_reasons(self, service, source):
...
+
+    def get_costs(self, service, source):
+        total_price = source.total_price_of_products.value
+        rounded_amt = 0
+        if self.rounding_behavior == RoundingBehavior.ROUND_UP:
+            rounded_amt = nickel_round(total_price, quant=self.rounding_quant, rounding=decimal.ROUND_UP)
+        else:
+            rounded_amt = nickel_round(total_price, quant=self.rounding_quant, rounding=decimal.ROUND_DOWN)
+        rounded_amt = source.create_price(decimal.Decimal(rounded_amt - total_price))
+        yield ServiceCost(rounded_amt, self.rounding_behavior.label)
```
suutari-ai: 1. You're assigning three different values to the `rounded_amt` variable. That makes it little hard to follow. Use different variables for different things.
2. You can operate with prices, no need to drop to bare decimals. `nickel_round(x, quant)` returns a `Price` if `x` is a `Price` and `quant` can still be bare `Decimal`.
3. Are these rounding modes "correct"? At least in Finnish stores prices are rounded with ROUND_HALF_UP method, e.g. 2,12 € will be rounded to 2,10 € and 2,13 € to 2,15 €. Neither ROUND_UP or ROUND_DOWN does that. We should provide  at least ROUND_HALF_UP too, but wouldn't hurt to provide some others too as long as they're supported by the decimal module.


I think this could be as simple as this:
```python
def get_costs(self, service source):
    total_price = source.total_price_of_products
    rounded = nickel_round(total_price, self.quant, self.mode)
    remainder = rounded - total_price
    yield ServiceCost(remainder, self.mode.label)
```
I'm assuming that the enum values for rounding mode use directly values of `decimal.ROUND_UP` etc. to avoid mapping them.

shoop/core/models/_service_behavior.py
```diff
@@ -168,3 +172,39 @@ def get_unavailability_reasons(self, service, source):
...
             yield ValidationError(_("Service is not available for any of the customers groups."))
+
+
+class RoundingBehavior(Enum):
+    ROUND_UP = 1
+    ROUND_DOWN = 2
+
+    class Labels:
+        ROUND_UP = _("round up")
+        ROUND_DOWN = _("round down")
```
suutari-ai: Provide also ROUND_HALF_UP and ROUND_HALF_DOWN and use the values of `decimal.ROUND_*` constants directly to avoid mapping those. They are strings , so the EnumIntegerField below has to be changed to EnumField too. Python's decimal module calls this rounding mode and IMHO that's better name than rounding behavior.

```python
class RoundingMode(Enum):
    ROUND_HALF_UP = decimal.ROUND_HALF_UP
    ROUND_HALF_DOWN = decimal.ROUND_HALF_DOWN
    ROUND_UP = decimal.ROUND_UP
    ...
    class Labels:
        ROUND_HALF_UP = _("round to nearest with ties going away from zero")
        ...
```

I don't think using those labels directly for the description of the cost is a good idea though...

suutari-ai:
Good job, @gurch101! Thanks!

PR #559, opened 2016-06-20
Add customer detail modal to admin order creator (SHUUP-2725)

gurch101:
Add new modal for admin order creation that shows detailed customer info,
order summary, and recent orders.

Refs SHUUP-2725

### DIFF
--- a/shoop/admin/modules/contacts/views/detail.py
+++ b/shoop/admin/modules/contacts/views/detail.py
@@ -10,14 +10,13 @@ from __future__ import unicode_literals
 from django.conf import settings
 from django.contrib import messages
 from django.core.urlresolvers import reverse
-from django.db.models import Q
 from django.http.response import HttpResponseRedirect
 from django.utils.encoding import force_text
 from django.utils.translation import ugettext_lazy as _
 from django.views.generic import DetailView
 
 from shoop.admin.toolbar import PostActionButton, Toolbar, URLActionButton
-from shoop.core.models import CompanyContact, Contact, Order, PersonContact
+from shoop.core.models import CompanyContact, Contact, PersonContact
 from shoop.utils.excs import Problem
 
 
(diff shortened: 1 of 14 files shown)


shoop/core/models/_contacts.py
```diff
@@ -281,6 +293,12 @@ def is_all_seeing(self):
         if self.user_id:
             return getattr(self.user, 'is_superuser', False)
 
+    def get_orders(self):
+        order_q = Q(orderer=self) | Q(customer=self)
```
tulimaki: Are you sure you want to mix these together. IMO you don't need this get_orders. You can already get already get the orders for contact throug related manager. And you should probably always use customer orders, orderer_orders and orders_created separately.

shoop/admin/modules/orders/views/edit.py
```diff
@@ -279,6 +281,47 @@ def handle_customer_data(self, request):
...
+        customer_id = request.GET["id"]
+        customer = Contact.objects.get(pk=customer_id)
+        companies = []
+        if isinstance(customer, PersonContact):
+            companies = sorted(customer.company_memberships.all(), key=(lambda x: force_text(x)))
+        recent_orders = customer.get_orders().order_by('-id')[:10]
+
+        order_summary = []
+        for dt in customer.get_orders().datetimes('order_date', 'year'):
+            summary = customer.get_orders().filter(order_date__year=dt.year).aggregate(
```
tulimaki: I think customer.customer_orders is enough here.

shoop/admin/modules/contacts/views/detail.py
```diff
@@ -103,18 +102,12 @@ def get_context_data(self, **kwargs):
...
             )
-        else:
-            order_q = Q(customer=self.object)
-        user = getattr(self.object, "user", None)
-        if user:
-            order_q |= Q(creator=user)
         context["contact_groups"] = sorted(
             self.object.groups.all(), key=(lambda x: force_text(x)))
-        context["orders"] = Order.objects.filter(order_q).order_by("-id")
+        context["orders"] = self.object.get_orders().order_by("-id")
```
tulimaki: I really think also this should be just just self.oject.customer_orders. In detail orders-tab orders as a customer and creator should be separated IMO. Since we don't even show customer/orderer/creator information in list.

PR #560, opened 2016-06-20
Add new campaign rules and add functionality to campaigns

Pikkupomo:
Add quantities to existing rules
Campaigns: Add product discount amount effect
Enable multiple effects for campaigns
Show quantity in basket for all lines

Refs SHUUP-2733, SHUUP-2734, SHUUP-2735

### DIFF
--- a/shoop/campaigns/modules.py
+++ b/shoop/campaigns/modules.py
@@ -36,13 +36,13 @@ class CatalogCampaignModule(DiscountModule):
         for campaign in CatalogCampaign.get_matching(context, shop_product):
             price = price_info.price
             # get first matching effect
-            effect = campaign.effects.first()
-            price -= effect.apply_for_product(context=context, product=product, price_info=price_info)
-            if best_discount is None:
-                best_discount = price
+            for effect in campaign.effects.all():
+                price -= effect.apply_for_product(context=context, product=product, price_info=price_info)
+                if best_discount is None:
+                    best_discount = price
 
-            if price < best_discount:
-                best_discount = price
+                if price < best_discount:
+                    best_discount = price
 
         if best_discount:
             if shop_product.minimum_price and best_discount < shop_product.minimum_price:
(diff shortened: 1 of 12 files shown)


shoop/campaigns/modules.py
```diff
@@ -111,21 +111,21 @@ def get_discount_line(campaign, amount, price_so_far):
...
-            # if campaign has coupon, match it to order_source.codes
-            if campaign.coupon:
-                # campaign was found because discount code matched. This line is always added
-                lines.append(get_discount_line(campaign, discount_amount, price_so_far))
-            elif best_discount is None or discount_amount > best_discount:
-                best_discount = discount_amount
-                best_discount_campaign = campaign
+            for effect in campaign.effects.all():
+                if not effect:
+                    continue
```
shawnadelic: I don't think this is necessary in the inner loop.

shoop/campaigns/modules.py
```diff
@@ -134,12 +134,12 @@ def get_discount_line(campaign, amount, price_so_far):
     def _handle_line_effects(self, matching_campaigns, order_source, original_lines):
         lines = []
         for campaign in matching_campaigns:
-            effect = campaign.effects.first()
-            if not effect:
-                continue
+            for effect in campaign.effects.all():
+                if not effect:
+                    continue
```
shawnadelic: Again, not sure if this is necessary

shoop/campaigns/modules.py
```diff
@@ -111,21 +111,21 @@ def get_discount_line(campaign, amount, price_so_far):
...
-                lines.append(get_discount_line(campaign, discount_amount, price_so_far))
-            elif best_discount is None or discount_amount > best_discount:
-                best_discount = discount_amount
-                best_discount_campaign = campaign
+            for effect in campaign.effects.all():
+                if not effect:
+                    continue
+
+                if hasattr(effect, "get_discount_lines"):
+                    continue
```
suutari-ai: Why are you mixing two kinds of effects in the same list? If the BasketLineEffects and BasketDiscountEffects were using two different related names, then there would be no need for this hasattr hack.

shoop/campaigns/modules.py
suutari-ai: Yep, not needed.

shoop/campaigns/modules.py
Pikkupomo: That's true, maybe I could rewrite this to match that schema.

PR #562, opened 2016-06-22
Ignore virtualenv files.

monokrome:


monokrome:
Sorry, not signing that.

PR #563, opened 2016-06-23
Admin: Add purchase orders (SHUUP-2699)

gurch101:
Add the ability to create purchase orders on manufacturers.
    * Add purchase order model
    * refactor client-side admin order creator
    * refactor admin `JsonOrderCreator` and `OrderEditorView`
    * Add purchase order list, create, and detail views
    * Add "create purchase order" button to manufacturer detail view

Refs SHUUP-2699

### DIFF
--- a/shoop/admin/modules/orders/json_order_creator.py
+++ b/shoop/admin/modules/orders/json_order_creator.py
@@ -13,10 +13,13 @@ from django.core.exceptions import ObjectDoesNotExist, ValidationError
 from django.utils.translation import ugettext as _
 
 from shoop.core.models import (
-    CompanyContact, Contact, MutableAddress, OrderLineType, OrderStatus,
-    PaymentMethod, PersonContact, Product, ShippingMethod, Shop
+    CompanyContact, Contact, Manufacturer, MutableAddress, OrderLineType,
+    OrderStatus, PaymentMethod, PersonContact, Product, ShippingMethod, Shop
 )
-from shoop.core.order_creator import OrderCreator, OrderModifier, OrderSource
+from shoop.core.order_creator import (
+    OrderCreator, OrderModifier, OrderSource, PurchaseOrderCreator,
+    PurchaseOrderSource
+)
 from shoop.utils.analog import LogEntryKind
 from shoop.utils.numbers import parse_decimal_string
 
@@ -99,17 +102,6 @@ class JsonOrderCreator(object):
             }), code="no_shop_product"))
             return False
         supplier = shop_product.suppliers.first()  # TODO: Allow setting a supplier?
-        is_orderable = True
-        for message in shop_product.get_orderability_errors(
-                supplier=supplier, quantity=sl_kwargs["quantity"], customer=source.customer):
-            self.add_error(ValidationError((_("Product %(product)s is not orderable: %(error)s") % {
-                "product": product,
-                "error": str(message.args[0])
-            }), code=str(message.args[1])))
-            is_orderable = False
-
-        if not is_orderable:
-            return False
 
         sl_kwargs["product"] = product
         sl_kwargs["supplier"] = supplier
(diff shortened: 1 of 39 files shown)


shoop/core/models/_orders.py
```diff
@@ -684,5 +684,42 @@ def can_edit(self):
 OrderLogEntry = define_log_model(Order)
 
 
+# note - @pikkupomo mentioned not to use a Polymorphic model due to migration troubles; hopefully multi-table
+# inheritance is better (seems to just be syntactic sugar around a one-to-one field with the parent)
+# One problem with defining the Purchase Order this way is upon PurchaseOrder creation - see _creator.py
+# another bummer is having to do Order.objects.exclude(purchaseorder__isnull=True) in order to get sales orders
+# although I suppose the latter can be fixed by adding a method to the query manager
+class PurchaseOrder(Order):
```
suutari-ai: Do we need them to have a concrete base class at all? What if we just move most of the stuff from Order to a new class which could be named e.g. `AbstractOrder`. Then just derive both `Order` and `PurchaseOrder` from the `AbstractOrder`. (Naming could also be `Order`/`SalesOrder`/`PurchaseOrder`, but that would force us to rename `Order` to `SalesOrder` everywhere, IMHO we shouldn't do that.)

Using polymorphic base class for a model is useful when you have a collection of objects which have the same interface, but possibly different implementation and you want to use them through the common interface. I don't think that's the case here.

shoop_tests/admin/test_order_creator.py
```diff
@@ -125,14 +143,44 @@ def test_order_creator_valid(rf, admin_user):
...
+    manufacturer = create_random_manufacturer()
+    contact = get_person_contact(admin_user)
+    supplier = get_simple_supplier()
+    state = get_frontend_order_state(type="purchase", manufacturer=manufacturer)
+    order = get_order_from_state(
+        state,
+        admin_user,
+        PurchaseOrderEditView
+    )
+    product_line = filter(lambda line: line["type"] == "product", state["lines"])[0]
```
suutari-ai: As you probably noticed from the tests, this does not work for Python 3, since filter doesn't return a list with Py3 but rather just an iterable. Use `first(filter(...))` (`from shoop.utils.iterable import first`) or `list(filter(...))[0]`.

shoop/simple_supplier/module.py
```diff
@@ -58,3 +58,7 @@ def update_stock(self, product_id):
         if latest_event:
             sv.stock_value_value = latest_event.purchase_price_value * sv.logical_count
         sv.save(update_fields=("logical_count", "physical_count", "stock_value_value"))
+
+    def get_last_purchase_price(self, product_id):
```
tulimaki: This is returning ``StockAdjustment``-object not a purchase price. I think this method is also missing from default module (which should return zero-price).

shoop/admin/modules/orders/views/edit.py
```diff
@@ -332,6 +317,92 @@ def handle_finalize(self, request):
...
+            )
+        return encode_manufacturer_data(manufacturer)
+
+    def handle_product_data(self, request):
+        product_data = super(PurchaseOrderEditView, self).handle_product_data(request)
+        product = Product.objects.filter(pk=product_data["id"]).first()
+        shop_id = request.GET["shop_id"]
+        shop = Shop.objects.get(pk=shop_id)
+        shop_product = product.get_shop_instance(shop)
+        supplier = shop_product.suppliers.first()  # TODO: Allow setting a supplier?
```
tulimaki: I think you might want to make supplier selectable for merchant or make this first simple supplier.

shoop/admin/modules/orders/views/edit.py
```diff
@@ -332,6 +317,92 @@ def handle_finalize(self, request):
...
+        latest_adjustment = supplier.get_last_purchase_price(product.id)
+        purchase_price = (latest_adjustment.purchase_price if latest_adjustment else None)
+        if purchase_price:
+            # note - I'm storing purchase price in base unit price to avoid having to change the OrderLine
+            # model - hacky, I know...
+            # i think adding a purchase_price field to OrderLine could be useful not just for purchase orders
+            # but also for sales orders to eventually generate interesting cost analysis reports
+            # (ie income statement/profit margin report).
+            # Do you think its worth adding the field? Doing so will however mean I will need
+            # to extend SourceLines so that the correct prices are used
```
tulimaki: I think you should use orderline base unit price to store the purchase price for this order. For sales order lines we can get the purchase price from supplier. And it's supplier job to calculate that based on stock adjustments and stock strategy (lifo, fifo, latest price etc...).

shoop/admin/modules/orders/json_order_creator.py
```diff
@@ -339,3 +295,102 @@ def update_order_from_state(self, state, order_to_update, modified_by=None):
...
+
+    def _get_manufacturer(self, manufacturer_data, save):
+        manufacturer = self.safe_get_first(Manufacturer, pk=manufacturer_data.get("id")) if manufacturer_data else None
+        if not manufacturer:
+            if not manufacturer_data.get('name', None):
+                self.add_error(
+                    ValidationError(_("Name is required for manufacturer"), code="no_name")
+                )
+                return
+            manufacturer = Manufacturer(name=manufacturer_data['name'], url=manufacturer_data.get('url', None))
```
tulimaki: Do not create new manufacturers while creatinf purchase orders. Since you are not adding new products for manufacturer you shouldn't add the manufacturer's either.

shoop/core/models/_orders.py
tulimaki: I think you should implement the AbstractOrder. Leave the SalesOrder rename for later.

tulimaki:
Split the commit little bit. For example core, supplier module, admin.

shoop/core/migrations/0029_add_purchase_orders.py
```diff
@@ -0,0 +1,28 @@
+# -*- coding: utf-8 -*-
+from __future__ import unicode_literals
+
+from django.db import migrations, models
+import django.db.models.deletion
+
+
+class Migration(migrations.Migration):
```
tulimaki: This migration is conflicting with current master.

shoop/simple_supplier/utils.py
```diff
@@ -35,17 +32,24 @@ def get_current_stock_value(supplier_id, product_id):
...
+        .filter(supplier_id=supplier_id, product_id=product_id, order__purchaseorder__isnull=True)
         .exclude(order__status__role=OrderStatusRole.CANCELED)
         .aggregate(total=Sum("quantity"))["total"] or 0)
     orders_sent = (
         ShipmentProduct.objects
         .filter(shipment__supplier=supplier_id, product_id=product_id)
         .aggregate(total=Sum("quantity"))["total"] or 0)
     return {
-        "logical_count": events - orders_bought,
+        "logical_count": events + pending_purchase_orders - orders_bought,
```
tulimaki: This logic seems to be missing tests.

tulimaki:
retest this please

shoop/core/models/_orders.py
suutari-ai: What happened to this? The multi-table inheritance is not the way to go IMHO.

PR #570, opened 2016-06-28
Admin: Only allow permissions for named modules

shawnadelic:
If admin modules do not have names set, do not allow as choices in
the `PermissionGroup` admin since there is no way to differentiate
modules when selecting modules to grant permissions.

Refs SHUUP-2901


Pikkupomo:
Tests will be added later on in another task made by @shawnadelic

PR #578, opened 2016-06-29
Admin: Add permission-related tests

shawnadelic:
Test that all url and module-level permissions are valid for enabled
modules.

Also test that only modules with defined names are shown as choices in
`PermissionGroup` admin.

Refs SHUUP-2909


suutari-ai:
Yo! Tests are always cool! Thanks, Shawn!

PR #582, opened 2016-06-30
Fix refund issues (SHUUP-2914, 2916)

shawnadelic:
- Fix handling of refunds for discount lines
- Replace buttons with dropdown in Orders admin
- Include quantities in refund order lines

Refs SHUUP-2914, 2916


suutari-ai:
Rebase to fresh master and add changelog entries (over the 0.4 line).

PR #584, opened 2016-07-01
Core: fix rounding component to use taxful total (SHUUP-2920)

gurch101:
* Add `get_post_tax_costs` method to `Service` which can be used to add order
lines after taxes have been calculated.
* Fix the rounding behavior component to calculate cost based on taxful amount.

### DIFF
--- a/doc/changelog.rst
+++ b/doc/changelog.rst
@@ -12,6 +12,7 @@ Unreleased
 Core
 ~~~~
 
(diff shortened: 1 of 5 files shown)


Pikkupomo:
This is not the proper way to do it since the plan was to remove behaviorcomponent in the process. Closing this now and I will introduce alternative solution tomorrow.

PR #586, opened 2016-07-01
Permission fixes (SHUUP-2898)

shawnadelic:
Add permission checks for:
- Toolbar buttons
- shuup.admin.utils.url.get_model_url
- shuup.admin.template_helpers.shuup_admin.model_url
- Dashboard blocks
- Fixes for various modules

Refs SHUUP--2898

### DIFF
--- a/shuup/admin/template_helpers/shuup_admin.py
+++ b/shuup/admin/template_helpers/shuup_admin.py
@@ -94,23 +94,24 @@ def get_breadcrumbs(context):
     return breadcrumbs
 
 
-def model_url(model, kind="detail", default=None):
+@contextfunction
+def model_url(context, model, kind="detail", default=None):
     """
     Get a model URL of the given kind for a model (instance or class).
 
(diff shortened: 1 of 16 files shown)


shuup/admin/template_helpers/shuup_admin.py
```diff
@@ -94,23 +94,24 @@ def get_breadcrumbs(context):
     return breadcrumbs
 
 
-def model_url(model, kind="detail", default=None):
+@contextfunction
+def model_url(context, model, kind="detail", default="#"):
```
gurch101: I think setting the default URL to # will make the UI appear broken to users. If we don't have the time to update stuff in templates (which is cool - I imagine that would take quite a bit more time), we should just leave things as-is and continue returning the "Can't view this page. You do not have the required permissions: set(['shuup.add_order'])". Maybe clean that permission string up by doing `','.join(permssions)`.

PR #596, opened 2016-07-06
Fix tax handling for refunds (SHUUP-2946)

gurch101:
Refund the applicable proportion of tax when creating a refund for a
line item by quantity or amount.

Refs SHUUP-2946

### DIFF
--- a/shuup/core/models/_orders.py
+++ b/shuup/core/models/_orders.py
@@ -551,6 +551,8 @@ class Order(MoneyPropped, models.Model):
             quantity = refund.get("quantity", 0)
             parent_line = refund.get("line")
             restock_products = refund.get("restock_products")
+            refund_line = None
+            percent_refunded = 0
 
             # TODO: Also raise this if the sum amount of refunds exceeds total,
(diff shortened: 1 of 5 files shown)


shuup/core/models/_orders.py
```diff
@@ -583,8 +586,22 @@ def create_refund(self, refund_data, created_by=None):
...
+            if refund_line and parent_line:
+                for line_tax in parent_line.taxes.all():
+                    refund_line.taxes.create(
+                        tax=line_tax.tax,
+                        name=_("Refund for %s" % line_tax.name),
+                        amount_value=-line_tax.amount_value * percent_refunded,
+                        base_amount_value=-line_tax.base_amount_value * percent_refunded,
+                        ordering=line_tax.ordering
+                    )
+
```
gurch101: Just want to highlight this bit - note that we can't properly refund taxes when the user refunds by amount without selecting a line first. In the future, we should either remove refunds by amount altogether or we should ensure amounts are associated with line items.

PR #598, opened 2016-07-06
Core: Prevent duplicate categories from all_visible-filter

tulimaki:
Remove duplicates in case category have multiple visibility groups set.

Remove order_by from all_visible-filter as unnecessary. The category
meta already defines the default ordering.


shawnadelic:
Looks good!

PR #600, opened 2016-07-07
Improve search

Pikkupomo:
Search didn't match single words, fix this by adding simple matching


Pikkupomo:
don't merge yet, I will add tests

PR #608, opened 2016-07-11
Fix admin content block styling (SHUUP-2992)

gurch101:
Wrap content blocks in untranslated section id so styling and behavior is
consistent across languages. Also fix variation variable delete functionality
across languages.

Refs SHUUP-2992

### DIFF
--- a/shuup/admin/templates/shuup/admin/macros/general.jinja
+++ b/shuup/admin/templates/shuup/admin/macros/general.jinja
@@ -20,8 +20,8 @@
     </div>
 {% endmacro %}
 
-{% macro content_block(heading, icon, extra_classes="") %}
-    <div class="content-block">
+{% macro content_block(heading, icon, extra_classes="", id="") %}
+    <div id="{{ id }}" class="content-block">
         <div class="title">
             {% if heading %}
                 <h2 class="block-title">{% if icon %} <i class="fa {{ icon }}"></i> {% endif %}{{ heading }}</h2>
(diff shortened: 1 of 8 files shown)


shuup/admin/templates/shuup/admin/products/_edit_media_form.jinja
```diff
@@ -90,17 +90,18 @@
 {%- set name = _("Product Images") if is_image_form else _("Product Media") %}
 
 {% call content_block(name, "fa-camera") %}
-    {{ media_form.management_form }}
-    {% for f in media_form %}
-        {{ render_media_form(f, media_form, loop.index, is_image_form) }}
-    {% endfor %}
-    <div class="hide" id="placeholder-panel">
-        {{ render_media_form(media_form.empty_form, media_form, "__prefix_name__", is_image_form) }}
+    {%- set target_panels = "product-images-section-wrapper" if is_image_form else "product-media-section-wrapper" %}
```
tulimaki: Instead of wrapping these content_blocks could you pass the untranslated id for content_block. I think it's bad idea that the contect block uses the title as id.

Also using content_block should work without wrapping everything to extra div.

shuup/admin/templates/shuup/admin/macros/general.jinja
```diff
@@ -20,8 +20,8 @@
     </div>
 {% endmacro %}
 
-{% macro content_block(heading, icon, extra_classes="") %}
-    <div class="content-block">
+{% macro content_block(id, heading, icon, extra_classes="") %}
+    <div id="{{id}}" class="content-block">
```
Pikkupomo: `{{ id }}` would be more in line with rest of the codebase :)

Pikkupomo:
rebase this please

tulimaki:
Also can you make this so that the id is optional. That way the existing addons wich has used this macro will still work.

PR #611, opened 2016-07-12
Campaigns: Add category products basket condition and line effect (SHUUP-3002)

tulimaki:
* Add condition to match with category products and quantity
* Add line effect that applies per product discount for products in
category

### DIFF
--- a/shuup/campaigns/models/basket_conditions.py
+++ b/shuup/campaigns/models/basket_conditions.py
@@ -11,7 +11,7 @@ from enumfields import Enum, EnumIntegerField
 from polymorphic.models import PolymorphicModel
 
 from shuup.core.fields import MoneyValueField
-from shuup.core.models import Contact, ContactGroup, Product
+from shuup.core.models import Category, Contact, ContactGroup, Product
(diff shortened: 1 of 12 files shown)


shuup/campaigns/models/basket_conditions.py
```diff
@@ -209,3 +209,28 @@ def values(self):
...
+    identifier = "basket_category_condition"
+    name = _("Category products in basket")
+
+    operator = EnumIntegerField(
+        ComparisonOperator, default=ComparisonOperator.GTE, verbose_name=_("operator"))
+    quantity = models.PositiveIntegerField(default=1, verbose_name=_("quantity"))
+    category = models.ForeignKey(Category, verbose_name=_("category"), blank=True)
+
+    def matches(self, basket, lines):
+        category_product_ids = self.category.shop_products.values_list("product_id", flat=True)
```
gurch101: number of products in a category will almost always be way larger than the number of items in a basket. IMO, maybe create a list of product ids in the basket, then filter and count the products in category based on that. Just a thought...

PR #614, opened 2016-07-13
Category products discounts (SHUUP-3002)

tulimaki:
* Add category products basket condition and line effect
* Optimize basket condition queries
* Optimize basket campaign matching queries

### DIFF
--- a/shuup/campaigns/admin_module/forms/_basket_effects.py
+++ b/shuup/campaigns/admin_module/forms/_basket_effects.py
@@ -11,7 +11,7 @@ from shuup.campaigns.models.basket_effects import (
     BasketDiscountAmount, BasketDiscountPercentage
 )
 from shuup.campaigns.models.basket_line_effects import (
-    DiscountFromProduct, FreeProductLine
+    DiscountFromCategoryProducts, DiscountFromProduct, FreeProductLine
 )
 
 from ._base import BaseEffectModelForm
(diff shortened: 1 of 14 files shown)


shuup/campaigns/admin_module/forms/_basket_effects.py
```diff
@@ -40,3 +40,21 @@ class Meta(BaseEffectModelForm.Meta):
...
     class Meta(BaseEffectModelForm.Meta):
         model = DiscountFromProduct
+
+
+class DiscountFromCategoryProductsForm(BaseEffectModelForm):
+    discount_percentage = PercentageField(
+        max_digits=6, decimal_places=5,
+        label=_("discount percentage"),
+        help_text=_("The discount percentage for this campaign."))
+
```
Pikkupomo: missing `discount_amount` field

shuup/campaigns/models/basket_line_effects.py
```diff
@@ -107,3 +109,42 @@ def get_discount_lines(self, order_source, original_lines):
...
+        help_text=_("Flat amount of discount."))
+    discount_percentage = models.DecimalField(
+        max_digits=6, decimal_places=5, blank=True, null=True,
+        verbose_name=_("discount percentage"),
+        help_text=_("The discount percentage for this campaign."))
+    category = models.ForeignKey(Category, verbose_name=_("category"))
+
+    @property
+    def description(self):
+        return _("Select discount amount and category. Discount is calculated for all category products in basket")
```
Pikkupomo: missing `.` from second sentence

shuup/campaigns/models/basket_conditions.py
```diff
@@ -209,3 +209,30 @@ def values(self):
...
+        category_product_ids = self.category.shop_products.filter(
+            product_id__in=product_id_to_qty.keys()
+        ).values_list(
+            "product_id", flat=True
+        )
+        product_count = sum(product_id_to_qty[product_id] for product_id in category_product_ids)
+        if self.operator == ComparisonOperator.EQUALS:
+            return bool(self.quantity == product_count)
+        else:
+            return bool(self.quantity <= product_count)
```
Pikkupomo: Strange that here this comparison is being flipped. In my opinion, it would be easier to read this by comparing `product_count >= self.quantity)` since the operator itself says `GTE` not `LTE` :) `ProductsInBasketCondition` does the `count >= self.quantity)` so this should follow the same pattern IMO.

shuup/campaigns/models/basket_conditions.py
```diff
@@ -209,3 +209,30 @@ def values(self):
...
+
+    @property
+    def description(self):
+        return _("Limit the campaign to have the products from selected category in basket.")
```
Pikkupomo: `to have` => `to match` imo.

shuup/campaigns/admin_module/forms/_basket_effects.py
tulimaki: No it's not missing it is a model form and the discount_percentage is just there to ovverride the fornfield to `PercentageField`

Pikkupomo: true :)

PR #619, opened 2016-07-15
Allow superadmins to login as customers (SHUUP-2968)

gurch101:
* Add "Login as User" button to user detail screen which allows superadmins
to login as customers.
* Set `order.creator` to impersonator
* Misc: Fix single page checkout view for customers not associated with a company
Refs SHUUP-2968

### DIFF
--- a/doc/changelog.rst
+++ b/doc/changelog.rst
@@ -36,6 +36,7 @@ Localization
 Admin
 ~~~~~
 
+- Allow superadmins to login as customer
 - Show orderability errors in package product management
 - Show stocks in package product management
 - Add link to order line product detail page in order editor
--- a/doc/changelog.rst
+++ b/doc/changelog.rst
@@ -77,6 +77,7 @@ Addons
 Front
 ~~~~~
 
+- Fix single page checkout for customers not associated with a company
 - Use contact default addresses for company creation
 - Use home country by default in customer information addresses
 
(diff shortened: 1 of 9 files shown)


doc/changelog.rst
```diff
@@ -30,6 +30,7 @@ Localization
 Admin
 ~~~~~
 
+- All superadmins to login as customer
```
shawnadelic: Assuming this is supposed to be Allow

shuup/admin/modules/users/views/detail.py
```diff
@@ -223,3 +243,38 @@ def post(self, request, *args, **kwargs):
...
+    def post(self, request, *args, **kwargs):
+        front_url = get_front_url()
+        user = self.get_object()
+        username_field = self.model.USERNAME_FIELD
+        person_contact = PersonContact.objects.filter(user=user).first()
+
+        if not front_url:
+            raise Problem(_("No shop configured."))
+        if not person_contact:
+            raise Problem(_("User is not bound to a person contact."))
```
shawnadelic: Is this 100% necessary? I know the shuup front middleware creates a contact if one doesn't exist for a user, although not sure if that applies in this case since we're logging in through admin.

tulimaki:
Could you add information about the impersonator user to `request.session` before login? Also use that information in `OrderSource` when setting creator.

We would like to separate orders that the actual customer did from the orders that some impersonator did for the customer.

doc/changelog.rst
```diff
@@ -31,8 +31,8 @@ Localization
 Admin
 ~~~~~
 
+- Allow superadmins to login as customer
 - Make contact views extendable
-- Make generic Section object for detail view sections
```
shawnadelic: Not sure if this was intentional

PR #630, opened 2016-07-19
Add orderable minimum (SHUUP-3049, SHUUP-3052)

gurch101:
Allow products to be sold after stock logical count is less than or equal to zero.

### DIFF
--- a/shuup/core/models/_product_shops.py
+++ b/shuup/core/models/_product_shops.py
@@ -42,6 +42,7 @@ class ShopProduct(MoneyPropped, models.Model):
     visibility_groups = models.ManyToManyField(
         "ContactGroup", related_name='visible_products', verbose_name=_('visible for groups'), blank=True
     )
+    backorder_maximum = QuantityField(default=0, blank=True, null=True, verbose_name=_('backorder maximum'))
     purchase_multiple = QuantityField(default=0, verbose_name=_('purchase multiple'))
     minimum_purchase_quantity = QuantityField(default=1, verbose_name=_('minimum purchase'))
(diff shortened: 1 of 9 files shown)


shuup/core/models/_product_shops.py
```diff
@@ -42,6 +42,7 @@ class ShopProduct(MoneyPropped, models.Model):
     visibility_groups = models.ManyToManyField(
         "ContactGroup", related_name='visible_products', verbose_name=_('visible for groups'), blank=True
     )
+    orderable_minimum = QuantityField(default=0, verbose_name=_('orderable minimum'))
```
tulimaki: Add help_text for this. Are you also sure that orderable_minimum is good name for this attrivbute? I am not sure.

shuup/core/suppliers/base.py
```diff
@@ -60,10 +60,11 @@ def get_orderability_errors(self, shop_product, quantity, customer):
         :rtype: iterable[ValidationError]
         """
         stock_status = self.get_stock_status(shop_product.product_id)
+        orderable_minimum = shop_product.orderable_minimum
         if stock_status.error:
             yield ValidationError(stock_status.error, code="stock_error")
         if shop_product.product.stock_behavior == StockBehavior.STOCKED:
-            if quantity > stock_status.logical_count:
+            if orderable_minimum != -1 and quantity > stock_status.logical_count + orderable_minimum:
```
tulimaki: Is there a way to prevent this magical -1 happening.

shuup/core/models/_product_shops.py
tulimaki: Could this be blank=True, null=Truem where the blank value would indicate the endles amount?

PR #633, opened 2016-07-19
Admin: Create a 'Save as a copy' button on product toolbar

chessbr:
Create a 'Save as a copy' button on product admin toolbar to make it easier to duplicate products.


chessbr:
Hey guys, I am proposing this feature to help users to create product variations much easier. The only task the user must do before clicking on this button is changing the SKU. What do you think?

chessbr:
Ouch, I've forgot to extract messages! Wait.

tulimaki:
Hi, @chessbr! This is something that was also in our list. Looks simple and good, thanks!

PR #635, opened 2016-07-19
Core: Round order total prices correctly (SHUUP-3109)

tulimaki:
When order and orderline prices is shown those should match. We are already
doing the bankers round for order totals based on non-rounded orderline prices.
This causes issue since now when we show these formatted prices in admin and
front the line prices and order total prices doesn't match.

So let's calculate the order totals based on the rounded orderline prices.

This is still not fixing the need for the currency class which will define
the price decimals.

Also missing the datamigration for the order-total prices, since this is
temporary fix before the currency class implementation which should also
fix the order totals based on the currency decimals.


Pikkupomo:
As a temporary fix i suppose this is ok

PR #643, opened 2016-07-22
Refund inventory, discount, and tax fixes (SHUUP-3114, SHUUP-3036)

gurch101:
* Add `StockAdjustmentType` `RESTOCK_LOGICAL` which allows restocking
unshipped products. If an order is partially shipped, restock from the
unshipped inventory first.

* Merge `OrderLineType.QUANTITY_REFUND` and `OrderLineType.AMOUNT_REFUND`
into `OrderLineType.REFUND`. Set refundable quantity and amount as default.

* Create refund order line with equal magnitude and opposite sign
when creating a full refund.

* Create tax lines with `tax_rate` instead of proportionally.

Refs SHUUP-3114, SHUUP-3036

### DIFF
--- a/shuup/admin/templates/shuup/admin/orders/create_refund.jinja
+++ b/shuup/admin/templates/shuup/admin/orders/create_refund.jinja
@@ -95,6 +95,8 @@
                     $refund.find("input[id$='amount']"), line && line["amount"] != 0);
                 setFormGroupVisibility(
                     $refund.find("input[id$='restock_products']"), line && line["type"] === "product");
+                $refund.find("input[id$='quantity']").val(line.quantity);
+                $refund.find("input[id$='amount']").val(line.amount);
             });
         }
 
--- a/shuup/admin/templates/shuup/admin/orders/create_refund.jinja
+++ b/shuup/admin/templates/shuup/admin/orders/create_refund.jinja
@@ -10,8 +10,8 @@
       {% endfor %}
       </select>
     </div>
-    {{ bs3.field(form.quantity, form_group_class="form-group hidden") }}
-    {{ bs3.field(form.amount, form_group_class="form-group hidden") }}
+    {{ bs3.field(form.quantity, form_group_class="form-group hidden", set_disabled=True) }}
+    {{ bs3.field(form.amount, form_group_class="form-group hidden", set_disabled=True) }}
     {{ bs3.field(form.restock_products, form_group_class="form-group hidden") }}
 {% endmacro %}
 
--- a/shuup/core/models/_orders.py
+++ b/shuup/core/models/_orders.py
@@ -538,10 +538,6 @@ class Order(MoneyPropped, models.Model):
         order's supplier the exact amount of the value of the `quantity`
         field.
 
-        If data for a refund line includes both an amount and a
-        quantity to refund, creates a separate refund line for each
-        refund type.
-
         :param refund_data: List of dicts containing refund data.
         :type refund_data: [dict]
         :param created_by: Refund creator's user instance, used for
@@ -562,10 +558,13 @@ class Order(MoneyPropped, models.Model):
             percent_refunded = 0
 
             assert parent_line
+            assert quantity
+            assert amount
+
             # TODO: Also raise this if the sum amount of refunds exceeds total,
             #       order amount, and do so before creating any order lines
             self.cache_prices()
-            if amount > self.taxful_total_price.amount:
+            if (amount > self.taxful_total_price.amount or amount > parent_line.max_refundable_amount):
                 raise RefundExceedsAmountException
 
             # If restocking products, calculate quantity of products to restock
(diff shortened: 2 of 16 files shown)


shuup/core/models/_orders.py
```diff
@@ -636,11 +636,12 @@ def create_full_refund(self, restock_products=False):
...
-            {"line": line, "quantity": line.quantity, "restock_products": restock_products}
-            for line in self.lines.all()
-            if (line.taxful_price or (line.type == OrderLineType.PRODUCT))
-        ]
+        line_data = [{
+            "line": line,
+            "quantity": line.quantity,
+            "amount": line.taxful_price.amount,
+            "restock_products": restock_products
+        } for line in self.lines.all() if (line.taxful_price or line.type == OrderLineType.PRODUCT)]
```
Pikkupomo: For me this check seems strange `line.taxful_price or line.type == OrderLineType.PRODUCT` what's the reasoning behind it and does it actually refund every line (methods as well) even if merchants sells taxless? What's the reason for `line.taxful_price` check, seems out of place.

shuup/core/models/_orders.py
```diff
@@ -797,7 +798,7 @@ def get_product_summary(self):
             products[product_id]['unshipped'] -= quantity
 
         refunded_prods = self.lines.refunds().filter(
-            type=OrderLineType.QUANTITY_REFUND,
+            type=OrderLineType.REFUND,
```
Pikkupomo: Again, it seems that only products are refunded? I understand that refunding in a sense of "returning items to stock" only considers products, but how about rest of the lines?

shuup/simple_supplier/module.py
```diff
@@ -56,8 +56,10 @@ def update_stock(self, product_id):
         sv, _ = StockCount.objects.get_or_create(supplier_id=supplier_id, product_id=product_id)
         sv.logical_count = values["logical_count"]
         sv.physical_count = values["physical_count"]
-        latest_event = StockAdjustment.objects.filter(
-            supplier=supplier_id, product=product_id).order_by("-created_on").first()
+        latest_event = (
+            StockAdjustment.objects
+            .filter(supplier=supplier_id, product=product_id, type=StockAdjustmentType.INVENTORY)
+            .order_by("-created_on").first())
```
Pikkupomo: is `.order_by("-created_on").first()` actually same as `.last()` in this queryset?

shuup/simple_supplier/utils.py
```diff
@@ -36,23 +37,26 @@ def get_current_stock_value(supplier_id, product_id):
     events = (
         StockAdjustment.objects
         .filter(supplier_id=supplier_id, product_id=product_id)
+        .exclude(type=StockAdjustmentType.RESTOCK_LOGICAL)
         .aggregate(total=Sum("delta"))["total"] or 0)
     orders_bought = (
```
Pikkupomo: strange name attribute as it has "nothing" to do with orders but mostly total quantity of bought products.

shuup/simple_supplier/utils.py
```diff
@@ -36,23 +37,26 @@ def get_current_stock_value(supplier_id, product_id):
...
-            Q(type=OrderLineType.AMOUNT_REFUND) |
-            Q(type=OrderLineType.QUANTITY_REFUND)
-        )
+            Q(type=OrderLineType.REFUND))
         .aggregate(total=Sum("quantity"))["total"] or 0)
+    orders_refunded_before_shipment = (
+        StockAdjustment.objects
+        .filter(supplier_id=supplier_id, product_id=product_id, type=StockAdjustmentType.RESTOCK_LOGICAL)
+        .aggregate(total=Sum("delta"))["total"] or 0)
     orders_sent = (
```
Pikkupomo: again "wrong" attribute name imo.

shuup/simple_supplier/utils.py
```diff
@@ -36,23 +37,26 @@ def get_current_stock_value(supplier_id, product_id):
...
         OrderLine.objects
         .filter(supplier_id=supplier_id, product_id=product_id)
         .exclude(
             Q(order__status__role=OrderStatusRole.CANCELED) |
-            Q(type=OrderLineType.AMOUNT_REFUND) |
-            Q(type=OrderLineType.QUANTITY_REFUND)
-        )
+            Q(type=OrderLineType.REFUND))
         .aggregate(total=Sum("quantity"))["total"] or 0)
+    orders_refunded_before_shipment = (
```
Pikkupomo: are adjustments per order? if not consider changing the attribute name

shuup/simple_supplier/utils.py
```diff
@@ -36,23 +37,26 @@ def get_current_stock_value(supplier_id, product_id):
...
     orders_sent = (
         ShipmentProduct.objects
         .filter(shipment__supplier=supplier_id, product_id=product_id)
         .aggregate(total=Sum("quantity"))["total"] or 0)
 
     return {
-        "logical_count": events - orders_bought,
+        "logical_count": events - orders_bought + orders_refunded_before_shipment,
         "physical_count": events - orders_sent
     }
```
Pikkupomo: This whole function seems very db consuming with a lot of queries generated (only counts but still). How is the caching handled and what if theme shows stocks in category product lists? Does this manage that? This just seems very complex for the task at hand or our db model has problems, not sure. There should be easier (and more efficient) way to handle this imo.

shuup/simple_supplier/utils.py
```diff
@@ -98,7 +102,7 @@ def get_stock_adjustment_div(request, supplier, product):
     :rtype: str
     """
     latest_adjustment = StockAdjustment.objects.filter(
-        product=product, supplier=supplier).order_by("-created_on").first()
+        product=product, supplier=supplier, type=StockAdjustmentType.INVENTORY).order_by("-created_on").first()
```
Pikkupomo: again, latest adjustment is usually last row added into database, hence the `.last()` should be enough here. But if there is a way to create `StockAdjustment` with custom `created_on`, ignore this :)

also you could have `ordering = "-created_on"` in object `Meta` imo and then use the `.first()` here.

Pikkupomo:
I would like to see a test cases with:
* Shop.prices_include_taxes = True and False:
  * order including few products with matching basket campaign, a priceful delivery and payment methods
  * Order including few products with matching catalog campaign, a priceful delivery and payment methods

shuup/admin/templates/shuup/admin/orders/create_refund.jinja
```diff
@@ -10,8 +10,8 @@
       {% endfor %}
       </select>
     </div>
-    {{ bs3.field(form.quantity, form_group_class="form-group hidden") }}
-    {{ bs3.field(form.amount, form_group_class="form-group hidden") }}
+    {{ bs3.field(form.quantity, form_group_class="form-group hidden", set_disabled=True) }}
```
gurch101: disabled as requested by @tulimaki until rounding issues have been resolved

shuup/simple_supplier/utils.py
gurch101: Front should be using `StockCount`/`get_stock_status` to get cached stock counts. 

Agreed that this function is db intensive but updating this (and likely the supplier API) is out-of-scope IMO.

gurch101:
really good call on those test suggestions @Pikkupomo - found a few bugs! added `functional/test_refunds.py`

PR #650, opened 2016-07-26
Core: Add method for modifying display options (SHUUP-2974)

shawnadelic:
Add `PriceDisplayOptions.set_option for modifying display options.

Refs SHUUP-2974


tulimaki:
Let's not do this. We shouldn't change this manually.

PR #661, opened 2016-07-28
Add soft deletion of categories (SHUUP-2994)

gurch101:
Allow categories to be soft deleted.
* Null out matching shop product and product categories.
* Set parent category to null for all children.
* send signal that other apps can act upon

Update campaigns that rely on category-based conditions, filters, and effects
when categories are soft deleted.

* Call `all_except_deleted` method on any `QueryManager`s that implement it for SelectView (currently Products and Categories).

Refs SHUUP-2994


shuup/core/models/_categories.py
```diff
@@ -129,6 +130,30 @@ def _get_slug_name(self):
...
 
+    def delete(self, using=None):
+        raise NotImplementedError("Not implemented: Use `soft_delete()` for categories.")
+
+    @atomic
+    def soft_delete(self, user=None):
+        if not self.status == CategoryStatus.DELETED:
+            for shop_product in self.primary_shop_products.all():
+                shop_product.primary_category = None
+                shop_product.categories.remove(self)
```
tulimaki: Unnecesary remove since the category is removed in the next phase from shop_product categories.

shuup/campaigns/signal_handlers.py
```diff
@@ -5,17 +5,41 @@
...
 
+def remove_category_from_campaigns(sender, category, **kwargs):
+    # raise an error if the category is part of an active campaign since this could result in a
+    # campaign that once applied to one category suddenly apply to the entire shop
+    if CategoryProductsBasketCondition.objects.filter(category=category, campaign__active=True).exists():
+        raise ValidationError(_("Category is used in the rules of an active basket campaign."))
+    if CategoryFilter.objects.filter(categories__id__exact=category.pk, campaign__active=True).exists():
+        raise ValidationError(_("Category is used in the rules of an active catalog campaign."))
+    for condition in CategoryProductsBasketCondition.objects.filter(category=category):
+        condition.delete()
```
tulimaki: Don't delete the conditions, filters or effects just set the campaign is_active False. Maybe the soft delete is just temporary or something so let's not start deleting here.

shuup/admin/modules/categories/views/edit.py
```diff
@@ -20,5 +23,11 @@ class CategoryEditView(SaveFormPartsMixin, FormPartsViewMixin, CreateOrUpdateVie
     base_form_part_classes = [CategoryBaseFormPart, CategoryProductFormPart]
     form_part_class_provide_key = "admin_category_form_part"
 
+    def get_toolbar(self):
+        save_form_id = self.get_save_form_id()
+        object = self.get_object()
+        delete_url = reverse_lazy("shuup_admin:category.delete", kwargs={"pk": object.pk})
```
tulimaki: How about new mode here?

PR #681, opened 2016-08-01
Simple Supplier: Add stock limit notification event (SHUUP-2862)

shawnadelic:
Add a stock limit notification event when stock for a product falls below
the set limit.

Refs SHUUP-2862

### DIFF
--- a/shuup/simple_supplier/forms.py
+++ b/shuup/simple_supplier/forms.py
@@ -27,3 +27,7 @@ class StockAdjustmentForm(forms.Form):
             raise ValidationError(_("Only non-zero values can be added to stock."), code="stock_delta_zero")
 
         return delta
+
+
+class AlertLimitForm(forms.Form):
+    alert_limit = forms.DecimalField(label=_("Alert limit"))
--- a/shuup/simple_supplier/module.py
+++ b/shuup/simple_supplier/module.py
@@ -5,6 +5,9 @@
 #
 # This source code is licensed under the AGPLv3 license found in the
 # LICENSE file in the root directory of this source tree.
+from django.conf import settings
+
+from shuup.core.models import Product, StockBehavior
 from shuup.core.stocks import ProductStockStatus
 from shuup.core.suppliers import BaseSupplierModule
 from shuup.core.suppliers.enums import StockAdjustmentType
@@ -62,4 +65,9 @@ class SimpleSupplierModule(BaseSupplierModule):
             .last())
         if latest_event:
             sv.stock_value_value = latest_event.purchase_price_value * sv.logical_count
+        product = Product.objects.get(pk=product_id)
+        if "shuup.notify" in settings.INSTALLED_APPS:
+            if (product.stock_behavior == StockBehavior.STOCKED) and (sv.physical_count < sv.alert_limit):
+                from .notify_events import AlertLimitReached
+                AlertLimitReached(supplier=self.supplier, product=product_id).run()
         sv.save(update_fields=("logical_count", "physical_count", "stock_value_value"))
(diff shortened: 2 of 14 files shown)


shuup/simple_supplier/module.py
```diff
@@ -63,3 +65,5 @@ def update_stock(self, product_id):
         if latest_event:
             sv.stock_value_value = latest_event.purchase_price_value * sv.logical_count
         sv.save(update_fields=("logical_count", "physical_count", "stock_value_value"))
+        product = Product.objects.get(pk=product_id)
+        stock_updated.send(sender=type(self), supplier=self.supplier, product=product)
```
tulimaki: Since this is SimpleSupplier specifig feature there is no need to send this signal. And it shouldn't be sent by simple supplier.

Instead you just can do the alert limit check here and send the notify event if needed.

shuup/simple_supplier/notify_events.py
```diff
@@ -0,0 +1,39 @@
...
+
+
+def _check_stock_level(supplier, product):
+    sc, _ = StockCount.objects.get_or_create(supplier=supplier, product=product)
+    if sc.physical_count < sc.alert_limit:
+        AlertLimitReached(supplier=supplier, product=product).run()
+
+
+@receiver(shipment_created)
+def check_order_stock_levels(order, **kwargs):
```
tulimaki: This might not be needed either since shipment save updates the stocks and also the alert limit should be handled with that stock update.

shuup/simple_supplier/forms.py
```diff
@@ -27,3 +27,7 @@ def clean_delta(self):
             raise ValidationError(_("Only non-zero values can be added to stock."), code="stock_delta_zero")
 
         return delta
+
+
+class AlertLimitForm(forms.Form):
+    alert_limit = forms.DecimalField(label=_("Alert limit"))
```
tulimaki: This field should include help_text to tell the merchant what this alret limit means. Alert notify event is sent when phycical count goes below alert limit. Or something like that.

shuup/simple_supplier/templates/shuup/simple_supplier/admin/add_stock_form.jinja
```diff
@@ -1,10 +1,27 @@
...
         <i class="fa fa-arrow-circle-up"></i> {% trans %}Add stock{% endtrans %}
     </button>
 </div>
+{% if alert_limit_form %}
+<hr>
+<div class="form" data-url="{{ url("shuup_admin:simple_supplier.alert_limits", supplier_id=supplier.id, product_id=product.id) }}">
+    {% csrf_token %}
+    {{ bs3.field(alert_limit_form.alert_limit, form_group_class=False ) }}
+    <span class="help-block">
+        {% trans %}
```
tulimaki: use {% trans -%} and  {%- endtrans %} avoid extra whitespaces and line breaks.

shuup/simple_supplier/templates/shuup/simple_supplier/admin/stock_information.jinja
```diff
@@ -4,6 +4,9 @@
         <p>{% trans %}Logical count{% endtrans %}: {{ stock.logical_count|floatformat(sales_decimals) }} {{ sales_unit }}</p>
         <p>{% trans %}Value per unit{% endtrans %}: {{ stock.stock_unit_price|money }}</p>
         <p>{% trans %}Total value{% endtrans %}: {{ stock.stock_value|money }}</p>
+    {% if alert_limit %}
+        <p>{% trans %}Alert limit{% endtrans %}: {{ stock.alert_limit|floatformat }}</p>
```
tulimaki: Include the variable in the translation

shawnadelic: Okay, should I do this for the rest of the stock information items too?

PR #687, opened 2016-08-02
Add simple browser tests (SHUUP-2698)

Pikkupomo:
Test checkout process in frontend
Test admin login and orders page load

Refs SHUUP-2698

### DIFF
--- a/shuup_tests/conftest.py
+++ b/shuup_tests/conftest.py
@@ -5,6 +5,8 @@
 #
 # This source code is licensed under the AGPLv3 license found in the
 # LICENSE file in the root directory of this source tree.
+import pytest
+
 from django.core.signals import setting_changed
 
 from shuup.apps.provides import clear_provides_cache
@@ -34,3 +36,7 @@ def pytest_runtest_teardown(item, nextitem):
     if hasattr(item.session, "_theme_overrider"):
         item.session._theme_overrider.__exit__(None, None, None)
         del item.session._theme_overrider
+
+@pytest.fixture(scope="session")
+def splinter_make_screenshot_on_failure():
+    return False
--- a/shuup_tests/front/test_checkout.py
+++ b/shuup_tests/front/test_checkout.py
@@ -10,8 +10,7 @@ from django.test import override_settings
 
 from shuup.core.models import get_person_contact
 from shuup.front.checkout.addresses import AddressesPhase, AddressForm
-from shuup.front.checkout.single_page import \
-    AddressForm as SinglePageAddressForm
+from shuup.front.checkout.single_page import AddressForm as SinglePageAddressForm
 from shuup.testing.factories import get_default_shop
 from shuup.testing.utils import apply_request_middleware
 
@@ -71,4 +70,3 @@ def test_address_phase_anonymous_user(rf):
     view_func = AddressesPhase.as_view()
     resp = view_func(request)
     assert 'company' in resp.context_data['form'].form_defs
-
--- a/.travis.yml
+++ b/.travis.yml
@@ -20,6 +20,12 @@ install:
   - pip install -U pip
   - pip install tox
   - pip install coveralls
+before_install:
+  - "export DISPLAY=:99.0"
+  - "sh -e /etc/init.d/xvfb start"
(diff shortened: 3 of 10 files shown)


shuup_tests/conftest.py
```diff
@@ -34,3 +34,6 @@ def pytest_runtest_teardown(item, nextitem):
     if hasattr(item.session, "_theme_overrider"):
         item.session._theme_overrider.__exit__(None, None, None)
         del item.session._theme_overrider
+
+# def pytest_addoption(parser):
```
tulimaki: No need to add commented code

shuup_tests/front/test_checkout.py
```diff
@@ -72,3 +77,102 @@ def test_address_phase_anonymous_user(rf):
...
+def create_orderable_product(name, sku, price):
+    supplier = get_default_supplier()
+    shop = get_default_shop()
+    product = create_product(sku=sku, shop=shop, supplier=supplier, default_price=price, name=name)
+    return product
+
+
+@pytest.mark.browser
+@pytest.mark.djangodb
+def test_browser_checkout(browser, live_server, settings):
```
tulimaki: I bet you could split this test a little bit and make it more readable.

shuup_tests/front/test_checkout.py
```diff
@@ -72,3 +77,102 @@ def test_address_phase_anonymous_user(rf):
...
+
+    # open product
+    browser.find_by_id("product-1").click()
+
+    # add product to basket
+    browser.find_by_id("add-to-cart-button").click()
+
+    # TODO: make this check better
+    # assume it was added to cart
+    assert browser.is_text_present("(1)")
```
tulimaki: Yeah, I think this assert needs some tuning if it's for proving that there is one product in the cart.

shuup_tests/front/test_checkout.py
```diff
@@ -72,3 +77,102 @@ def test_address_phase_anonymous_user(rf):
     resp = view_func(request)
     assert 'company' in resp.context_data['form'].form_defs
 
+
+# BROWSER TESTS BELOW
```
tulimaki: Unneccessary notification about incoming browser tests. you should probably separate the browser tests for checkout in it's own file if you also think that this file needs to be divided.

tulimaki:
Also this commit is missing changelog and documentation.

shuup_tests/front/test_checkout.py
```diff
@@ -72,3 +77,102 @@ def test_address_phase_anonymous_user(rf):
...
+    customer_city = "Test City"
+    customer_country = "US"
+
+    browser.fill("billing-name", customer_name)
+    browser.fill("billing-street", customer_street)
+    browser.fill("billing-city", customer_city)
+    browser.select("billing-country", customer_country)
+    # browser.execute_script('document.getElementById("same_as_billing").checked=true')  # same as billing
+
+    browser.find_by_css(".btn.btn-primary.btn-lg.pull-right").first.click()  # click "continue"
```
gurch101: selecting by generic bootstrap classes will make these tests brittle imo (plus harder to read without your comments :)). Is it possible to do something similar to this? "#addresses button[type='submit']"

shuup_tests/front/test_checkout.py
Pikkupomo: I will add comments to make it more readable

shuup_tests/front/test_checkout.py
Pikkupomo: I'm not 100% sure I want to divide browser tests into their own files. I was thinking about this, but I think we need to make it later IF the tests are long. This test here is exceptionally long as it proves the whole checkout process.

shuup_tests/front/test_checkout.py
Pikkupomo: Might be possible, not sure. Need to investigate. This is the reason I added the comments :)

suutari-ai:
I think browser tests should be under a separate directory, e.g. `shuup_tests/browser`.

suutari-ai:
Great stuff, @Pikkupomo! And now it seems to work on Travis too (and is skipped by Jenkins).

Let's try this.

PR #705, opened 2016-08-08
Admin: Update for order toolbar provide buttons (SHUUP-2544 / SHUUP-3204)

tulimaki:
* Add ``admin_order_toolbar_action_item`` provider for order toolbar
* Add deprecation warning for ``admin_order_toolbar_button`` usages

### DIFF
--- a/shuup/admin/modules/orders/views/detail.py
+++ b/shuup/admin/modules/orders/views/detail.py
@@ -7,18 +7,13 @@
 # LICENSE file in the root directory of this source tree.
 from __future__ import unicode_literals
 
-import warnings
-
 from django.contrib import messages
-from django.core.urlresolvers import reverse
 from django.http.response import HttpResponseRedirect
 from django.utils.encoding import force_text
 from django.utils.translation import ugettext as _
 from django.views.generic import DetailView
 
-from shuup.admin.toolbar import (
-    DropdownActionButton, PostActionButton, Toolbar, URLActionButton
-)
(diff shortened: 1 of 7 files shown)


shuup/admin/modules/orders/action_items.py
```diff
@@ -0,0 +1,49 @@
+# -*- coding: utf-8 -*-
```
gurch101: maybe consider creating a `toolbar.py` instead of `action_items.py` to follow suit with products.

shuup/admin/toolbar.py
```diff
@@ -277,6 +277,23 @@ def render(self, request):
             yield '<li class="dropdown-header">%s</li>' % self.text
 
 
+class ProvideActionItem(DropdownItem):
```
gurch101: I probably wouldn't use `Provide` in the class name since we don't do that with provided form parts/sections. Maybe just `ActionItem` or `DropdownActionItem`.

shuup/admin/modules/orders/views/detail.py
```diff
@@ -82,6 +84,9 @@ def get_toolbar(self):
         ))
 
         for button in get_provide_objects("admin_order_toolbar_button"):
+            warnings.warn(
+                "admin_order_toolbar_button provider is deprecated, use admin_order_toolbar_action_item instead",
+                DeprecationWarning, stacklevel=2)
```
gurch101: will setting stacklevel=2 actually have the intended effect here? isn't the caller in this case something internal to django?

PR #706, opened 2016-08-09
Admin: Use last product id + 1 as a default SKU

gurch101:
Set default SKU to product id + 1. Note this can result in a duplicate SKU error if the user has multiple "create product" tabs open or if multiple users are creating products simultaneously. Users are, however, provided proper errors in the event of a dup.

Refs SHUUP-2964


suutari-ai:
Nice feature! :+1:

PR #707, opened 2016-08-09
Admin: set shipment tracking code and description

chessbr:
User is able to set tracking codes and description when creating a shipment


tulimaki:
ok to test

PR #711, opened 2016-08-10
Fetch user IP using django ipware

chessbr:
When in production environment, it is strict necessary to fetch real user IP through other existing META keywords, set by some proxies.

### DIFF
--- a/shuup/utils/http.py
+++ b/shuup/utils/http.py
@@ -8,6 +8,8 @@
 import time
 
 import requests
+from django.conf import settings
+from ipware.ip import get_ip, get_real_ip
 
 
 def retry_request(n_retries=5, **kwargs):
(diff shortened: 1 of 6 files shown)


chessbr:
Can you rebuild isort target? It seems to be using outdated code.

shuup/utils/http.py
```diff
@@ -9,6 +9,9 @@
 
 import requests
 
+from django.conf import settings
+from ipware.ip import get_ip, get_real_ip
+
```
suutari-ai: `requests` is also third-party, so these new lines shouldn't be separated from it by blank line.

tulimaki:
ok to test

chessbr:
@akx proposed another solution: a simple middleware which rewrites the META header with the proper IP address. It is a nice pluggable solution which does not make hard changes to Shuup core and do not add another dependency.

Middleware code.
```
class IpwareMiddleware:
    def process_request(request):
        request.META['REMOTE_ADDR'] = get_real_ip(request)
```

PR #712, opened 2016-08-10
Contact creation refactor (SHUUP-3158)

gurch101:
* Add "New person" and "New company" toolbar buttons to list view
* Split out company contact and person contact edit into two separate forms
* Use select2multiple for contact group input
* Mark first name and last name as required for person contacts
* Show only required and related fields on contact creation
* Show all fields on contact edit
* Allow billing address to be used as shipping address when creating a contact. Also, put billing address before contact address for order and contact detail address sections.
* Add type filter to contact list view
* Make select2 multiple default height more in line with other form inputs. (note: it still isn't exact due to browser sub-pixel rendering but its better than before - before it was ~4 pixels taller, now it is <1px different).

### DIFF
--- a/shuup/admin/modules/contacts/views/edit.py
+++ b/shuup/admin/modules/contacts/views/edit.py
@@ -5,230 +5,45 @@
 #
 # This source code is licensed under the AGPLv3 license found in the
 # LICENSE file in the root directory of this source tree.
-
 from __future__ import unicode_literals
 
-from django import forms
-from django.contrib.auth import get_user_model
-from django.db.models.loading import get_model
 from django.db.transaction import atomic
-from django.forms import BaseModelForm
-from django.utils.encoding import force_text
-from django.utils.translation import ugettext_lazy as _
 
-from shuup.admin.form_part import (
-    FormPart, FormPartsViewMixin, SaveFormPartsMixin, TemplatedFormDef
+from shuup.admin.form_part import FormPartsViewMixin, SaveFormPartsMixin
+from shuup.admin.modules.contacts.form_parts import (
+    CompanyContactBaseFormPart, PersonContactBaseFormPart
 )
-from shuup.admin.forms.fields import Select2MultipleField
-from shuup.admin.forms.widgets import PersonContactChoiceWidget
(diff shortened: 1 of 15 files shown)


shuup/admin/modules/contacts/views/edit.py
```diff
@@ -32,144 +30,132 @@
...
+                self.fields["account_manager"].widget = PersonContactChoiceWidget(clearable=True)
+
+        def save(self, commit=True):
+            obj = super(ContactBaseForm, self).save(commit)
+            obj.groups = self.cleaned_data["groups"]
+            return obj
+    return ContactBaseForm
+
+
+def get_person_contact_base_form(mode, user=None):
```
tulimaki: Way too hacky stuff here. Just move `PersonContactBaseForm` and `CompanyContactBaseForm` under forms and if they something in common then but that stuff to `ContactBaseForm`. If you need to give mode for the form then pass that to __init__.

shuup/admin/modules/contacts/views/edit.py
```diff
@@ -32,144 +30,132 @@
...
-        return obj
+            self.fields["members"] = members_field
+    return CompanyContactBaseForm
 
 
 class ContactBaseFormPart(FormPart):
     priority = -1000  # Show this first, no matter what
 
-    def get_form_defs(self):
+    def get_contact_type(self):
```
tulimaki: Also form parts probably shouldn't live under views.

tulimaki:
You should take a second look for those contact forms. Also it is not necessary to use same views for the person and company contact. It actually might simplify things if you separate those, of course some of the functionality should live in the base classes.

shuup/admin/modules/contacts/views/edit.py
gurch101: I did this since I'm dynamically setting `fields` based on `mode`. I believe the alternative involves setting `Meta` inside of `__init__` which is just as messy imo. I suppose the easiest thing would be to just handle this `mode` logic inside the template. I will play around with this tomorrow...

PR #713, opened 2016-08-10
Admin: Add option for automatically adding product lines (SHUUP-3152)

shawnadelic:
Add an option in order creator automatically add the first matching
product line when using quick add.

This can be enabled by default by configuring the
`SHUUP_ADMIN_ORDER_CREATOR_QUICK_ADD_DEFAULT` setting.

Refs SHUUP-3152

### DIFF
--- a/shuup/admin/modules/orders/static_src/create/view/lines.js
+++ b/shuup/admin/modules/orders/static_src/create/view/lines.js
@@ -146,6 +146,11 @@ export function renderOrderLines(store, shop, lines) {
 }
 
 var select2 = {
+    focus: function() {
+        if (select2.element) {
+            select2.element.select2("open");
+        }
+    },
     clear: function() {
         if (select2.element) {
             select2.element.select2("val", null);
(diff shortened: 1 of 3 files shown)


tulimaki:
Can you get rid of the setting, it shouldn't be needed. I think here you only want the checkbox "Automatically add first matching product" (make it checked by default). And when that is checked the product line is added straight when the product is selected from the option. And like now after the line is added the select is cleared. If the checkbox is not checked then the adding happens from the plus icon.

Also add browser test for order create. At this point it's enough that it creates the order and uses this functionality.

shuup/admin/modules/orders/static_src/create/view/lines.js
```diff
@@ -175,16 +182,39 @@ var select2 = {
...
+        }
+        else {
+            store.dispatch(setLineProperty(line.id, "quantity", parseFloat(line.quantity) + 1));
+        }
+        select2.focus();
+    },
     currentProduct: m.prop(),
     changeProduct: function(id) {
         productQuickSelect.currentProduct(id);
+        if (id){
```
Pikkupomo: missing space

shuup/admin/modules/orders/static_src/create/view/lines.js
```diff
@@ -175,16 +182,39 @@ var select2 = {
...
+            productQuickSelect.quickAddProduct(productQuickSelect.store,id);
```
Pikkupomo: missing space after `,`

PR #714, opened 2016-08-11
Product creation/edit redesign (SHUUP-3168)

gurch101:
Products can now be created by simply adding a product name and pressing save. Note the product is not orderable since supplier is not defaulted.

* Fix `FormattedDecimalField` default value
* Allow content blocks to be collapsed on initialization
* Rearrange product creation and edit pages

### DIFF
--- a/shuup/core/fields/__init__.py
+++ b/shuup/core/fields/__init__.py
@@ -8,6 +8,7 @@
 from __future__ import unicode_literals
 
 import decimal
+import numbers
 
 import babel
 from django import forms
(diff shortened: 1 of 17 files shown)


shuup/core/fields/__init__.py
```diff
@@ -86,8 +87,8 @@ class FormattedDecimalField(models.DecimalField):
     """
     def value_from_object(self, obj):
         value = super(FormattedDecimalField, self).value_from_object(obj)
-        if isinstance(value, decimal.Decimal):
-            return self.format_decimal(value)
+        if isinstance(value, numbers.Number):
```
tulimaki: Can you please add tests around these changes?

tulimaki:
"Shipping & Payment" Under shop form should probably be collapsed. At least those are not set in most of the cases. Set first supplier to product suppliers by default, so the product is orderable after the name is set.

PR #715, opened 2016-08-11
Fix contact group availability behavior component

JoaRiski:
All customers belong to the "Anonymous" group by default (AKA. not signed in users),
thus it's pointless to check for the existance of the user in the database. This
also causes limiting to `not logged in users only` to be impossible.

Fix the issue explained above by removing a check for customer's existance in the database.

### DIFF
--- a/shuup/core/models/_service_behavior.py
+++ b/shuup/core/models/_service_behavior.py
@@ -163,7 +163,7 @@ class GroupAvailabilityBehaviorComponent(ServiceBehaviorComponent):
     groups = models.ManyToManyField("ContactGroup", verbose_name=_("groups"))
 
     def get_unavailability_reasons(self, service, source):
-        if not source.customer.pk:
+        if source.customer and not source.customer.pk:
             yield ValidationError(_("Customer does not belong to any group."))
             return
 


JoaRiski:
Not quite sure who to assign this to, so assigning to plenty of people

suutari-ai:
Hmm... this currently fails this test:

```python
    @pytest.mark.django_db
    def test_unsaved_contact(admin_user):
        payment_method = get_default_payment_method()
        _assign_component_for_service(payment_method, [PersonContact.get_default_group()])
        person = PersonContact(name="Kalle")
        source = _get_source_for_contact(admin_user, payment_method)
        source.customer = person
        assert not person.pk and not source.customer.pk
       _test_service_availability(source, payment_method, False)
```
I'm not sure if this test is even reasonable. It's added in commit d0c5d802 without reasoning why unsaved contacts should be even possible. IMHO `request.customer` should always be a saved contact or `AnonymousContact`. @tulimaki  Do you remember why we added support for `request.customer` being unsaved non-anonymous contact?

JoaRiski:
I've worked around the test by only skipping the `pk` check if the customer is not an instance of `AnonymousContact`

shuup/core/models/_service_behavior.py
```diff
@@ -163,7 +164,7 @@ class GroupAvailabilityBehaviorComponent(ServiceBehaviorComponent):
     groups = models.ManyToManyField("ContactGroup", verbose_name=_("groups"))
 
     def get_unavailability_reasons(self, service, source):
-        if not source.customer.pk:
+        if not issubclass(type(source.customer), AnonymousContact) and not source.customer.pk:
```
suutari-ai: If this is indeed what we want (still waiting for Panu's answer, since IMHO we should just remove the support for unsaved non-anonymous contacts in `request.customer`), then use `if source.customer and not source.customer.pk`. It will work, since `AnonymousContact` will evaluate as false in the `if`.

tulimaki: I really think that there is no reason to allow `request.customer` to be unsaved person or company contact.

But in case the order id done from admin the customer passed to behavior unavailability check might indeed be unsaved. Since the customer in admin order creation is saved only after confirm.

JoaRiski:
Changed the check to what @suutari-ai suggested.

tulimaki:
ok to test

PR #716, opened 2016-08-12
Add proper title to company creation (SHUUP-3222)

Pikkupomo:
Refs SHUUP-3222


tulimaki:
Do not submit djangojs file since there is not any new strings.

PR #719, opened 2016-08-15
Show affecting campaigns on product page (SHUUP-3008)

Pikkupomo:
Refs SHUUP-3008


shuup/core/templatetags/shuup_common.py
```diff
@@ -114,6 +114,9 @@ def datetime(value, kind="datetime", format="medium", tz=True):
 
     locale = get_current_babel_locale()
 
+    if value is None:
```
tulimaki: Based on docsting this function does not allow None as a parameter. So add it or/and can you give some extra information why this is needed?

shuup/campaigns/admin_module/sections.py
```diff
@@ -0,0 +1,34 @@
...
+
+    @staticmethod
+    def visible_for_object(product):
+        return bool(product.pk)
+
+    @staticmethod
+    def get_context_data(product):
+        ctx = {}
+        for shop in Shop.objects.all():
+            shop_product = product.get_shop_instance(shop)  # fails if not saved after shop creation?
```
tulimaki: IMO this shouldn't fail if product does not have instance for all shops.

shuup/campaigns/templates/shuup/campaigns/admin/_product_campaigns.jinja
```diff
@@ -0,0 +1,31 @@
+{% macro campaign_table(title, campaigns) %}
+<h3>{{ title }}</h3>
+<div class="table-responsive">
+    <table class="table table-striped table-bordered">
+        <thead>
+            <tr>
+                <th>{% trans %}Name{% endtrans %}</th>
```
tulimaki: Missing translations?

shuup/campaigns/models/campaigns.py
```diff
@@ -117,6 +119,13 @@ def rules_match(self, context, shop_product, matching_catalog_filters, matching_
         return True
 
     @classmethod
+    def get_for_product(cls, shop_product, limit=100):
+        matching_filters = get_matching_for_product(shop_product, provide_category="campaign_catalog_filter")
+        matching_conditions = get_matching_for_product(shop_product, provide_category="campaign_context_condition")
+        query_filter = Q(Q(filters__in=matching_filters) | Q(conditions__in=matching_conditions))
+        return cls.objects.filter(shop=shop_product.shop, active=True).filter(query_filter).distinct()[:limit]
```
tulimaki: Is this limit to 100 really necessary? If so this should be informed at the section.

tulimaki: It seems that you are filtering non active campaigns here. You should also check the dates then. This looks like a good spot for a custom queryset.

shuup/core/templatetags/shuup_common.py
Pikkupomo: Ok, will do. This is originally from @gurch101 branch. I will see if there is really use for this.

shuup/campaigns/admin_module/sections.py
Pikkupomo: agreed

shuup/campaigns/models/campaigns.py
Pikkupomo: I think we should not limit these at the moment. I will remove it. I think the proper way in future would be a pagination of sorts.

I created the queryset manager and it will be there in next pr.

tulimaki:
retest this please

shuup_tests/campaigns/test_availability.py
```diff
@@ -0,0 +1,99 @@
...
+
+    # end in future, no start date, match
+    campaign.end_datetime = future_datetime
+    campaign.save()
+    assert campaign.is_available()
+    assert CatalogCampaign.objects.available(shop).exists()
+
+
+
+# class CampaignQueryset(models.QuerySet):
```
tulimaki: No need to add this commented code.

shuup/admin/utils/views.py
```diff
@@ -165,3 +166,8 @@ def get_object_abstract(self, instance, item):
         return [
             {"text": "%s" % instance, "class": "header"},
         ]
+
+
+class _Breadcrumbed(object):
```
tulimaki: Could this be under admin breadcrumbs?

Pikkupomo:
adding some changes still

Pikkupomo:
ok everything is in

PR #720, opened 2016-08-15
Add test for category based discounts

JoaRiski:
Add a test for category based basket discounts where two products from the
same category should trigger a discount for both products


suutari-ai:
ok to test

suutari-ai:
Great, a new test. Thanks, @JoaRiski!

PR #721, opened 2016-08-15
Order admin: Show tax number in customer information (SHUUP-3151)

Pikkupomo:
Use `name_ext` if available in ´CompanyContact.full_name`

Refs SHUUP-3151


shuup/core/models/_contacts.py
```diff
@@ -216,6 +216,10 @@ class Meta:
         verbose_name = _('company')
         verbose_name_plural = _('companies')
 
+    @property
+    def full_name(self):
+        return self.name + (" / %s" % self.name_ext if self.name_ext else "")
```
tulimaki: How about `"%s / %s" % (self.name, self.name_ext or "")`?

shuup/admin/modules/orders/views/edit.py
```diff
@@ -340,6 +340,7 @@ def handle_customer_details(self, request):
                 "name": customer.full_name,
                 "phone_no": customer.phone,
                 "email": customer.email,
+                "tax_number": customer.tax_number if isinstance(customer, CompanyContact) else "",
```
tulimaki: I would prefer `getattr(customer, "tax_number", "")` instead of isinstance check.

shuup/core/models/_contacts.py
Pikkupomo: this would end up with "company / " if name_ext doesn't exists right?

shuup_tests/browser/admin/test_order_list.py
```diff
@@ -39,7 +39,7 @@ def _visit_orders_list_view(browser, live_server):
     url = reverse("shuup_admin:order.list")
     browser.visit("%s%s" % (live_server, url))
     assert browser.is_text_present("Orders")
-    time.sleep(0.5)  # Wait mithril for a sec
+    wait_until_appeared(browser, css_class=".picotable-item-info")
```
tulimaki: Change all sleep thingys.

Pikkupomo: I think the second wait is ok since the picotable is actually present already and the content don't currently have anything we can attach on with `wait_until_appeared`

PR #723, opened 2016-08-16
Send telemetry if no previous submission (SHUUP-3226)

Pikkupomo:
Refs SHUUP-3226


suutari-ai:
This should probably be configurable from the settings. Also the texts in Admin should be updated.

Pikkupomo:
Texts seem ok?

PR #724, opened 2016-08-16
Fix an issue where None comparison to int is failing

JoaRiski:
This comparision would raise an "unorderable types: NoneType()> int()" exception. Fix this by checking
for a value before comparing.

Traceback of said error:
image


tulimaki:
ok to test

tulimaki:
Replaced this with https://github.com/shuup/shuup/pull/726 so the fix can be merged. Thanks @JoaRiski!

PR #731, opened 2016-08-17
Front: Fix password recovery form bug with invalid email

shawnadelic:
Refs SHUUP-3209


shuup/front/apps/auth/forms.py
```diff
@@ -74,6 +74,9 @@ class RecoverPasswordForm(forms.Form):
     from_email = None
 
     def clean(self):
+        if self.errors:
```
tulimaki: Instead consider using this below:
```
username = data.get("username")
email = data.get("email")
```

PR #736, opened 2016-08-18
Front: Add tests for simple search

shawnadelic:
Add tests for simple search, including tests for finding words and
normalizing spaces in search queries.

Refs SHUUP-2990

### DIFF
--- a/doc/changelog.rst
+++ b/doc/changelog.rst
@@ -34,6 +34,8 @@ Addons
 Front
 ~~~~~
 
+- Add product SKU to searchable fields for simple search
+- Limit search results for simple search
 - Fix password recovery form bug with invalid email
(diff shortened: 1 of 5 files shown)


Pikkupomo:
Is the name only thing the search uses to find products?

shawnadelic:
Just double checked, and looked like it searches 'translations__name', 'translations__description',  and 'translations__keywords' currently. Will add a couple more tests for those, and also I think we should add SKU as a searchable field. I thought it was, but apparently not.

tulimaki:
Since you are touching tests here. I would add a small new feature to it. Could you limit the search reasults to 100 or 200 (maybe even 150). Now since we don't have any pagination involved here the popular search terms will make the shop crash.

PR #737, opened 2016-08-18
Misc performance fixes (SHUUP-3038)

gurch101:
Went through front with django debug toolbar and cleaned up some duplicate queries.

There are still quite a few places where things can be cleaned up but I stuck to the low hanging fruit :p

Refs SHUUP-3038

### DIFF
--- a/shuup/front/template_helpers/general.py
+++ b/shuup/front/template_helpers/general.py
@@ -44,13 +44,14 @@ def get_visible_products(context, n_products, ordering=None, filter_dict=None, o
     if ordering:
         products_qs = products_qs.order_by(ordering)
 
+    suppliers = Supplier.objects.all()
     if orderable_only:
         products = []
         for product in products_qs[:(n_products * 4)]:
             if len(products) == n_products:
                 break
             shop_product = product.get_shop_instance(shop)
-            for supplier in Supplier.objects.all():
+            for supplier in suppliers:
                 if shop_product.is_orderable(supplier, customer, shop_product.minimum_purchase_quantity):
                     products.append(product)
                     break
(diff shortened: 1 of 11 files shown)


shuup/xtheme/engine.py
```diff
@@ -66,6 +66,10 @@ class XthemeEnvironment(Environment):
 
     template_class = XthemeTemplate
 
+    def __init__(self, *args, **kwargs):
+        super(XthemeEnvironment, self).__init__(*args, **kwargs)
+        self.theme = get_current_theme()
```
shawnadelic: Are there any potential issues with storing the theme object here? Since it won't be changing, I don't think it will be an issue, but I'm not sure

shuup/front/template_helpers/general.py
```diff
@@ -68,10 +69,11 @@ def get_best_selling_products(context, n_products=12, cutoff_days=30, orderable_
     product_ids = [d[0] for d in data][:n_products]
 
     products = []
+    suppliers = Supplier.objects.all()
```
Pikkupomo: Now you are getting these even if `orderable_only` is `False`

tulimaki:
Also get rid of the template/macros fixes. I will adopt these to incoming template refactoring. Just wait 20 minutes.

shuup/front/template_helpers/general.py
```diff
@@ -151,3 +153,12 @@ def get_pagination_variables(context, objects, limit):
     variables["objects"] = page.object_list
 
     return variables
+
+
+@contextfunction
+def get_price_info(context, product, quantity=1):
```
tulimaki: Also the filters should be used instead of get_price_info. We should avoid the performance issue for those (like proper caching).

Pikkupomo:
I will close this and add my own branch which uses some of the fixes made here. The template fixes are however removed from that.

PR #738, opened 2016-08-19
Xtheme: Themes now own their own plugins (SHUUP-1806)

Pikkupomo:
Plugins provided with `xtheme_plugin` provide spec are considered as global plugins.

Refs SHUUP-1806

### DIFF
--- a/shuup/testing/__init__.py
+++ b/shuup/testing/__init__.py
@@ -39,8 +39,10 @@ class ShuupTestingAppConfig(AppConfig):
         ],
         "admin_contact_section": [
             "shuup.testing.admin_module.sections:MockContactSection",
-        ]
+        ],
+        "xtheme": [
+            __name__ + ".theme:ShuupTestingTheme",
+        ],
     }
-
 
 default_app_config = "shuup.testing.ShuupTestingAppConfig"
--- a/shuup/testing/theme/__init__.py
+++ b/shuup/testing/theme/__init__.py
@@ -1,0 +1,20 @@
+# -*- coding: utf-8 -*-
+# This file is part of Shuup.
+#
+# Copyright (c) 2012-2016, Shoop Ltd. All rights reserved.
+#
+# This source code is licensed under the AGPLv3 license found in the
+# LICENSE file in the root directory of this source tree.
+
+from django.utils.translation import ugettext_lazy as _
+
+from shuup.xtheme import Theme
+
+
+class ShuupTestingTheme(Theme):
+    identifier = "shuup_testing"
+    name = _("Shuup Testing Theme")
+    author = "Shuup"
+    template_dir = "testing/"
+
+    plugins = [__name__ + ".plugins.TestHighlightPlugin"]
--- a/shuup/xtheme/_theme.py
+++ b/shuup/xtheme/_theme.py
@@ -8,15 +8,17 @@
 import logging
 from contextlib import contextmanager
 
+from django.utils.translation import ugettext_lazy as _
+
 from shuup.apps.provides import (
     get_identifier_to_object_map, get_provide_objects
 )
+from shuup.utils.importing import load
 
 log = logging.getLogger(__name__)
 
 
 # TODO: Document how to create Xthemes
-
 class Theme(object):
     """
     Base class for all Xtheme themes.
@@ -51,6 +53,9 @@ class Theme(object):
     # This might not be used if you override `get_configuration_form`.
     # TODO: Document this
     fields = []
+
+    # List of plugin specs used in this template
+    plugins = []
 
     def __init__(self, settings_obj=None):
         """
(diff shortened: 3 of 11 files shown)


shuup/xtheme/_theme.py
```diff
@@ -161,6 +166,72 @@ def get_view(self, view_name):
         """
         return None
 
+    def get_all_plugin_choices(self, empty_label=None):
+        """
+        Get grouped list of 2-tuples (identifier and name) of all available Xtheme plugins.
+
+        Super handy for `<select>` boxen.
```
shawnadelic: typo

shuup/xtheme/_theme.py
```diff
@@ -161,6 +166,72 @@ def get_view(self, view_name):
...
+            choices.append(("", empty_label))
+        choices += [(_("Global plugins"), self._get_global_plugin_choices())]
+        choices += [(_("Theme plugins"), self._get_theme_plugin_choices())]
+        return choices
+
+    def _get_theme_plugin_choices(self, empty_label=None):
+        """
+        Get a sorted list of 2-tuples (identifier and name) of available theme specific Xtheme plugins.
+
+        Handy for `<select>` boxen.
```
shawnadelic: Same

shuup/xtheme/_theme.py
```diff
@@ -161,6 +166,72 @@ def get_view(self, view_name):
...
+                getattr(plugin, "name", None) or plugin.identifier
+            ))
+        choices.sort()
+        return choices
+
+    def _get_global_plugin_choices(self, empty_label=None):
+        """
+        Get a sorted list of 2-tuples (identifier and name) of available global Xtheme plugins.
+
+        Handy for `<select>` boxen.
```
shawnadelic: Same

shuup/testing/__init__.py
```diff
@@ -5,6 +5,7 @@
 #
 # This source code is licensed under the AGPLv3 license found in the
 # LICENSE file in the root directory of this source tree.
+
```
tulimaki: Remove the blank line here.

shuup/testing/theme/__init__.py
```diff
@@ -0,0 +1,24 @@
...
+
+class ShuupTestingTheme(Theme):
+    identifier = "shuup_testing"
+    name = _("Shuup Testing Theme")
+    author = "Shuup"
+    template_dir = "testing/"
+
+    plugins = [__name__ + ".plugins.TestHighlightPlugin"]
+
+    def get_view(self, view_name):
```
tulimaki: Do you really need declare views for ShuupTestingTheme? Are these used?

shuup/xtheme/_theme.py
```diff
@@ -161,6 +166,72 @@ def get_view(self, view_name):
...
+
+        :param empty_label: Label for the "empty" choice. If falsy, no empty choice is prepended
+        :type empty_label: str|None
+        :return: List of 2-tuples
+        :rtype: Iterable[tuple[str, str]]
+        """
+        choices = []
+        if empty_label:
```
tulimaki: Why this method needs it own empty_label?

shuup/xtheme/_theme.py
tulimaki: What this means?

shuup/xtheme/_theme.py
Pikkupomo: because it might be called alone. I will make this non-private also I think

shuup/testing/theme/__init__.py
Pikkupomo: It's more like proper themes this way. Theme itself should have this in baseclass anyway.

tulimaki: It has. And IMO theme shouldn't have front theme views enabled by default. Also the import should be `import shuup.front.themes.views as views` to make that actually work.

shuup/xtheme/_theme.py
tulimaki: Ok. Maybe there is a usecase to just get the theme plugin choices. I guess the `_get_global_plugin_choices` shouldn't be private either.

shuup_tests/xtheme/test_plugin_api.py
```diff
@@ -1,14 +1,20 @@
...
 from shuup_tests.xtheme.utils import get_jinja2_engine, plugin_override
 
 
 def test_plugin_choices():
     with plugin_override():
-        choice_identifiers = set(c[0] for c in Plugin.get_plugin_choices())
-        assert choice_identifiers == set(("inject", "text"))
+        theme = ShuupTestingTheme()
+        choice_identifiers = set()
+        for c, d in theme.get_all_plugin_choices():
```
shawnadelic: I think you could use better variable names here

PR #739, opened 2016-08-20
Xtheme: make `theme` properly go to cells (SHUUP-1806)

Pikkupomo:


shuup/xtheme/layout.py
```diff
@@ -68,7 +68,7 @@ def instantiate_plugin(self, theme=None):
             return plugin_class(config=self.config)
         return None
 
-    def render(self, context):
+    def render(self, context, theme=None):
```
tulimaki: Maybe the theme shouldn't be optional here?

PR #743, opened 2016-08-22
Simple Search: Fix bug in limiting query (SHUUP-2990 / SHUUP-3261)

tulimaki:
The original query caused errors since not all versions of MySQL
support LIMIT & IN/ALL/ANY/SOME subquery. By forcing the queryset to
list the product ids can be limited before caching the product ids.

### DIFF
--- a/shuup/front/apps/simple_search/views.py
+++ b/shuup/front/apps/simple_search/views.py
@@ -48,7 +48,7 @@ def get_compiled_query(query_string, needles):
     """
     Get compiled query
 
-    Complile query string into `Q` objects and return it
+    Compile query string into `Q` objects and return it
     """
     compiled_query = None
     for word in get_query_words(query_string):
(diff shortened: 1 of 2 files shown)


shuup/front/apps/simple_search/views.py
```diff
@@ -67,8 +67,8 @@ def get_search_product_ids(request, query, limit=150):
     if product_ids is None:
         entry_query = get_compiled_query(
             query, ['sku', 'translations__name', 'translations__description', 'translations__keywords'])
-        product_ids = Product.objects.filter(entry_query).distinct().values_list("pk", flat=True)[:limit]
-        cache.set(cache_key, product_ids, 60 * 5)
+        product_ids = list(Product.objects.filter(entry_query).distinct().values_list("pk", flat=True))
```
Pikkupomo: I think you should change this to `product_ids = list(Product.objects.filter(entry_query).distinct().values_list("pk", flat=True))[:limit]` so that the cached and returned ids are the same

PR #746, opened 2016-08-22
Simple Search: Set column width in less (SHUUP-3201)

tulimaki:
Add extra class `search-view` for product-box and set
column width in less to enable more customizable search
list.

### DIFF
--- a/doc/changelog.rst
+++ b/doc/changelog.rst
@@ -24,6 +24,8 @@ Addons
 Front
 ~~~~~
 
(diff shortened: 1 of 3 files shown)


doc/changelog.rst
```diff
@@ -24,6 +24,8 @@ Addons
 Front
 ~~~~~
 
+- Define simple search result list col width in less instead of template
```
Pikkupomo: this could say `column` instead of `col` imo

PR #747, opened 2016-08-22
Admin: require shipping address before creating shipment (SHUUP-3257)

hrayr-artunyan:
The order form already requires shipping information when creating an
order. This commit ensures that a shipping_address is set on the order
at the time of shipment creation, otherwise it raises a
NoShippingAddressException.

Refs SHUUP-3257

### DIFF
--- a/shuup_tests/admin/test_shipment_creator.py
+++ b/shuup_tests/admin/test_shipment_creator.py
@@ -16,13 +16,14 @@ from shuup.admin.modules.orders.views.shipment import (
     FORM_MODIFIER_PROVIDER_KEY, OrderCreateShipmentView, ShipmentFormModifier
 )
 from shuup.apps.provides import override_provides
+from shuup.core.excs import NoShippingAddressException
+
 from shuup.core.models import Order
 from shuup.testing.factories import (
     create_order_with_product, create_product, get_default_shop,
     get_default_supplier
 )
 from shuup.testing.utils import apply_request_middleware
-
 
 @pytest.mark.django_db
 def test_shipment_creating_view_get(rf, admin_user):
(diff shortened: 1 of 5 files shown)


shuup_tests/admin/test_shipment_creator.py
```diff
@@ -165,3 +166,24 @@ def form_valid_hook(self, form, shipment):
...
+    shop = get_default_shop()
+    supplier = get_default_supplier()
+    product = create_product(sku="test-sku", shop=shop, supplier=supplier, default_price=3.33)
+    order = create_order_with_product(product, supplier, quantity=1, taxless_base_unit_price=1, shop=shop)
+    
+    # remove shipping address 
+    order.shipping_address = None
+    order.save()
+
+    try:
```
tulimaki: You can replace this try..catch with 

```
with pytest.raises(NoShippingAddressException):
    order.create_shipment_of_all_products()
```

tulimaki:
ok to test

tulimaki:
add to whitelist

shuup_tests/admin/test_shipment_creator.py
gurch101: Any chance we can hide the "create shipment" toolbar button if there is no shipping address? Shipping addresses shouldn't be required for digital goods/marketplace items.

tulimaki: I think the update in `_orders.can_create_shipment` already does that. At least it should.

gurch101: yep you're right. Didn't see that. Sorry!

hrayr-artunyan:
@gurch101 are you able to produce an order without a shipping address at the UI level? I'm not sure if I'm missing a use case.

gurch101:
@hrayr-artunyan You currently can using the marketplace code (although the lack of a shipping addresses causes other issues in the marketplace as well - see SHUUP-3254). No biggie though your code looks good.

hrayr-artunyan:
Hm, does anyone know why https://travis-ci.org/shuup/shuup/builds/154285960 fails? Is this a known issue?

tulimaki:
Yeah that's the usual case with browser tests. There should be task for it and fixed soon @hrayr-artunyan

PR #754, opened 2016-08-24
CORE Overwrite empty language fields with default value (SH-90)

hrayr-artunyan:
When a user completes a form that includes translatable fields (e.g.
Product) and only fill out the default language portion (e.g. English)
the form will fill out the other language fields with that of the
english. This fix is meant to address limitations in django-parlers own
fallback mechanism.

This commit also includes a new management command
`shuup_fallback_language` that will perform the same functionality on
existing models. It's meant to be ran only once as a migration step.

### DIFF
--- a/doc/changelog.rst
+++ b/doc/changelog.rst
@@ -128,6 +128,8 @@ Released on 2016-08-12 03:00pm -0800.
 Core
 ~~~~
 
+- Fix `MultiLanguageModelForm` to overwrite empty language fields with default value
+- Add datamigration to add missing translations from default values 
 - Fix `FormattedDecimalField` default value for form fields
 - Combine `TreeManager` and `TranslatableManager` querysets for categories
 - Exclude deleted orders from valid queryset
--- a/shuup/core/migrations/0006_data_migrate_language_fallback.py
+++ b/shuup/core/migrations/0006_data_migrate_language_fallback.py
@@ -1,0 +1,87 @@
+# -*- coding: utf-8 -*-
+from __future__ import unicode_literals
+
+from django.apps import apps as django_apps
+from django.conf import settings
+from django.db import migrations
+from django.utils import translation
+from parler.models import TranslatableModel, TranslationDoesNotExist
+from django.db.utils import OperationalError
+
+
+def to_language_codes(languages):
+    languages = languages or (translation.get_language(),)
+    if languages and isinstance(languages[0], (list, tuple)):
+        # `languages` looks like a `settings.LANGUAGES`, so fix it
+        languages = [code for (code, name) in languages]
+    return languages
+
+
+def save_values_from_default_to_lang(default_translation, translation_object, fields):
+
+    for field in fields:
+        default_value = getattr(default_translation, field.name, '')
+        orig_value = getattr(translation_object, field.name)
+
+        if not orig_value and default_value != orig_value:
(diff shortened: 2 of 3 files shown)


Pikkupomo:
I think the management command should be a datamigration. 

The reason I think so is that:
* IMO management commands are commands that should be run many times (like collect translations etc) or ones that can be run only once but have re-usability on clean installations (shuup_init)
* This is a one time script that _should_ be run when this update is being run on the shuup. After this update it's useless to even have this script since all translated fields will be replaced with values on saving the object. And it has no use in clean installs after this update.

hrayr-artunyan:
"Retest this please"

hrayr-artunyan:
Retest this please

tulimaki:
retest this please

shuup/core/migrations/0005_data_migrate_language_fallback.py
```diff
@@ -0,0 +1,79 @@
...
+
+    translation_model_fields = [
+        f for f in translation_model._meta.fields
+        if f.name in base_fields and not f.unique
+    ]
+
+    for obj in objects:
+        # model being translatable does not guarantee translations exist
+        if obj.translations.count():
+            default_translation = obj.get_translation(default_lang)
```
tulimaki: This will fail if the object doesn't have translation in PARLER_DEFAULT_LANGUAGE_CODE. That happens some times since I think only currently active language is required most of the `MultiLanguageModelForm`.

hrayr-artunyan: The default_language is a required field in MultiLanguageModelForm isn't it?
`self.default_language = kwargs.pop("default_language", getattr(self, 'language', get_language()))`

Where is active language set?

tulimaki: I think not. Can you test it with for few admin modules like products, categories and shops. I think that most of the cases only the currently active language is required. That's probably a bug though. But maybe not for this task. But the migration should work in those "corrupted" databases too.

tulimaki: If I am right here on solution probably would be to check the default translation for one field with `safe_translation_getter`

shuup/core/migrations/0005_data_migrate_language_fallback.py
```diff
@@ -0,0 +1,79 @@
+# -*- coding: utf-8 -*-
+from __future__ import unicode_literals
+
+from django.apps import apps as django_apps
```
Pikkupomo: wouldn't `from django import apps` suffice?

hrayr-artunyan: @Pikkupomo, the function in which I reference apps, already passes apps as an argument, so there's a conflict. apps passed by the function is different from django.apps.apps

hrayr-artunyan:
retest this please

hrayr-artunyan:
retest this please

hrayr-artunyan:
retest this please

hrayr-artunyan:
retest this please

tulimaki:
For this we will be using different approach. We really need to follow parler here and copying the default taranslation is not the right way.

The "right" solution is already planned. Let's get back to this issue in few days.

Close for now at least.

PR #757, opened 2016-08-25
Admin: Fix missing shipping_address on order views (SH-120)

hrayr-artunyan:
Simply hides the shipping address in Order detail if it's missing,
though the order form should prevent this from happening in the first
place.

Refs SH-120


shuup/admin/templates/shuup/admin/orders/detail.jinja
```diff
@@ -47,22 +47,24 @@
             {% if order.shipping_address_id or order.billing_address_id %}
                 {% call content_block(_("Addresses"), "fa-map-marker") %}
                     <div class="row contact-addresses">
+                        {% if order.billing_address_id %}
```
tulimaki: I guess you don't need this if.

hrayr-artunyan: Ah, you're right. I rushed and didn't clean it up properly.

PR #764, opened 2016-08-27
Admin: Make currency field a dropdown in Shops admin

shawnadelic:
Refs SH-132


tulimaki:
Replaced with https://github.com/shuup/shuup/pull/767.

PR #765, opened 2016-08-28
Select columns in listview (SH-64)

Pikkupomo:
Refs SH-64

### DIFF
--- a/shuup/admin/modules/orders/views/list.py
+++ b/shuup/admin/modules/orders/views/list.py
@@ -42,7 +42,7 @@ class OrderStatusChoicesFilter(ChoicesFilter):
 
 class OrderListView(PicotableListView):
     model = Order
-    columns = [
+    default_columns = [
         Column("identifier", _(u"Order"), linked=True, filter_config=TextFilter(operator="startswith")),
         Column("order_date", _(u"Order Date"), display="format_order_date", filter_config=DateRangeFilter()),
         Column(
@@ -60,7 +60,7 @@ class OrderListView(PicotableListView):
         Column("payment_status", _(u"Payment Status"), filter_config=ChoicesFilter(choices=PaymentStatus.choices)),
         Column("shipping_status", _(u"Shipping Status"), filter_config=ChoicesFilter(choices=ShippingStatus.choices)),
         Column(
-            "taxful_total_price", _(u"Total"), sort_field="taxful_total_price_value",
+            "taxful_total_price_value", _(u"Total"), sort_field="taxful_total_price_value",
             display="format_taxful_total_price", class_name="text-right",
             filter_config=RangeFilter(field_type="number", filter_field="taxful_total_price_value")
         ),
@@ -81,6 +81,6 @@ class OrderListView(PicotableListView):
     def get_object_abstract(self, instance, item):
         return [
             {"text": "%s" % instance, "class": "header"},
-            {"title": _(u"Total"), "text": item["taxful_total_price"]},
-            {"title": _(u"Status"), "text": item["status"]}
+            {"title": _(u"Total"), "text": item.get("taxful_total_price_value")},
+            {"title": _(u"Status"), "text": item.get("status")}
         ]
--- a/shuup/admin/modules/settings/__init__.py
+++ b/shuup/admin/modules/settings/__init__.py
@@ -1,0 +1,118 @@
+# This file is part of Shuup.
+#
+# Copyright (c) 2012-2016, Shoop Ltd. All rights reserved.
+#
+# This source code is licensed under the AGPLv3 license found in the
+# LICENSE file in the root directory of this source tree.
+import six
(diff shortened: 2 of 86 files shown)


shuup/admin/modules/settings/views/__init__.py
```diff
@@ -0,0 +1,49 @@
...
+            key = self.settings.get_settings_key(col.id)
+            initial.update({
+                key: self.settings.get_config(col.id)
+            })
+        return initial
+
+    def form_valid(self, form):
+        for col, val in six.iteritems(form.cleaned_data):
+            self.settings.set_config(col, val, use_key=True)
+        messages.success(self.request, _("Settings saved"))
```
tulimaki: Missing translation

shuup/campaigns/admin_module/__init__.py
```diff
@@ -25,7 +25,7 @@ def get_urls(self):
         basket_campaign_urls = get_edit_and_list_urls(
             url_prefix="^campaigns/basket",
             view_template="shuup.campaigns.admin_module.views.BasketCampaign%sView",
-            name_template="basket_campaigns.%s",
+            name_template="basket_campaign.%s",
```
tulimaki: In case you are changing this you also need to change url references in `get_menu_entries()`

shuup/admin/modules/orders/views/list.py
```diff
@@ -60,7 +60,7 @@ class OrderListView(PicotableListView):
         Column("payment_status", _(u"Payment Status"), filter_config=ChoicesFilter(choices=PaymentStatus.choices)),
         Column("shipping_status", _(u"Shipping Status"), filter_config=ChoicesFilter(choices=ShippingStatus.choices)),
         Column(
-            "taxful_total_price", _(u"Total"), sort_field="taxful_total_price_value",
+            "taxful_total_price_value", _(u"Total"), sort_field="taxful_total_price_value",
```
tulimaki: Not sure if it's because of this change but unselecting total price causes KeyError at /sa/orders 'taxful_total_price_value'

shuup/admin/modules/settings/__init__.py
```diff
@@ -0,0 +1,125 @@
...
+        )
+        return self.handle_special_column(field, column)[0]
+
+    def handle_special_column(self, field, column):
+        # reverse handlers for convenience
+        is_special = False
+        for original_column in self.default_columns:
+            if column.id == original_column.id:
+                is_special = True
+                if getattr(original_column, "sortable"):
```
tulimaki: could this be?

```
for field in ["sortable", "display", "linked", "filter_config"]:
    if getattr(original_column, field):
        setattr(column, field, getattr(original_column, field))

```

shuup/admin/modules/settings/views/__init__.py
```diff
@@ -0,0 +1,49 @@
...
+
+
+class ListSettingsView(FormView):
+    form_class = ColumnSettingsForm
+    template_name = "shuup/admin/edit_settings.jinja"
+
+    def dispatch(self, request, *args, **kwargs):
+        model = request.GET.get("model")
+        module = request.GET.get("module")
+        mod = __import__(module, fromlist=[model])
```
tulimaki: This is not working in Python 2.7. TypeError at /sa/list-settings/ Item in ``from list'' not a string.

tulimaki:
`ViewSettings` and `ListSettingsView` needs some tests. Also some of the picotables with custom columns configurations should be tested with browser tests.

shuup/admin/utils/views.py
```diff
@@ -146,6 +147,12 @@ def check_and_raise_if_only_one_allowed(setting_name, obj):
 
 
 class PicotableListView(PicotableViewMixin, ListView):
+
+    def __init__(self):
+        super(PicotableListView, self).__init__()
+        self.settings = ViewSettings(self.model, self.default_columns)
+        self.columns = self.settings.columns
```
tulimaki: Not so various fix since this enables the whole deal? This change probably should be introduced before the actual columns->default_columns changes. Also the default_columns should probably be introduced in base class since it's required.

shuup/admin/modules/orders/views/list.py
Pikkupomo: Thanks for report, it's usually when `get_object_abstract` has hard coded keys, for example `item["taxful_total_price"]` will fix that

shuup/admin/modules/settings/__init__.py
Pikkupomo: I'm actually thinking that should we even limit these values and just overwrite anything that the original one defines?

shuup/admin/utils/views.py
Pikkupomo: true

tulimaki:
I would like to see you test this in the picotable browser test.

PR #770, opened 2016-08-29
Admin: Fix url bug in suppliers module

shawnadelic:
Refs SH-136


tulimaki:
Make everything here at Supplier `supplier` not `suppliers` see the naming convention from other admin modules. Also make sure that the `simple_supplier` app works correctly with new urls.

PR #771, opened 2016-08-29
Admin: Allow form group edit views to show errors as messages

shawnadelic:
Allow create-or-update views using form parts to show form errors as
messages.

Refs SH-24


shuup/admin/utils/views.py
```diff
@@ -101,7 +103,17 @@ def save_form(self, form):
 
     def form_invalid(self, form):
         if self.add_form_errors_as_messages:
-            add_form_errors_as_messages(self.request, form)
+            # If form is a form group, add form part errors individually
+            if isinstance(form, FormGroup):
+                for name, form_part in form.forms.items():
```
shawnadelic: Just realized name is actually unused here

PR #773, opened 2016-08-30
Front: Fix macro importing bug

shawnadelic:


shuup/front/templates/shuup/front/macros/basket.jinja
```diff
@@ -1,4 +1,5 @@
 {% extends "shuup/front/macros/theme/basket.jinja" %}
+{% from "shuup/front/macros/order.jinja" import render_package_children %}
```
tulimaki: Insteadn of importing this you should be able to use `{{ order_macros.render_package_children(product) }}` where every the render_package_children is used.

shawnadelic: Looks I still need to import order_macros, at least in basket.jinja

PR #775, opened 2016-08-31
Fixes for translation fallbacks and JS catalog loading (SH-17, SH-90)

tulimaki:
* MultiLanguageModelForm: Use Parler default as a default
* MultiLanguageModelForm: Avoid partially/empty translation objects
* Core: Fix bug in product and category slugs generation
* Admin: Add key prefix to JavaScript catalog page cache
* Admin: Add lang parameter for JS catalog load
* Xtheme: Skip adding JS-catalog for editing
* Tests: Add test for loading JS catalog

### DIFF
--- a/doc/changelog.rst
+++ b/doc/changelog.rst
@@ -13,6 +13,8 @@ Unreleased
 Core
 ~~~~
 
+- Fix bug in product and category slug generation
+
 Localization
 ~~~~~~~~~~~~
 
--- a/doc/changelog.rst
+++ b/doc/changelog.rst
@@ -21,6 +21,7 @@ Localization
 Admin
 ~~~~~
 
+- Add lang parameter for JS catalog load
 - Add key prefix to JavaScript catalog cache
 - Allow shop language to be set via admin
 - Allow form group edit views to show errors as messages
--- a/doc/changelog.rst
+++ b/doc/changelog.rst
@@ -21,6 +21,7 @@ Localization
 Admin
 ~~~~~
 
+- Add key prefix to JavaScript catalog cache
 - Allow shop language to be set via admin
 - Allow form group edit views to show errors as messages
 
--- a/doc/changelog.rst
+++ b/doc/changelog.rst
@@ -61,6 +61,8 @@ Guide
 General/miscellaneous
 ~~~~~~~~~~~~~~~~~~~~~
 
+- MultiLanguageModelForm: Use Parler default as a default
+
 SHUUP 0.4.4
 -----------
 
--- a/doc/changelog.rst
+++ b/doc/changelog.rst
@@ -61,6 +61,10 @@ Guide
 General/miscellaneous
 ~~~~~~~~~~~~~~~~~~~~~
 
+- MultiLanguageModelForm: Avoid partially/empty translation objects
+   - Delete untranslated objects from database
+   - Only set translation object to database if it is translated
+   - Ensure required fields if language is partially translated
 - MultiLanguageModelForm: Use Parler default as a default
 
 SHUUP 0.4.4
--- a/doc/changelog.rst
+++ b/doc/changelog.rst
@@ -32,6 +32,8 @@ Addons
 Front
 ~~~~~
 
+- Load JS catalog for superusers
+
 Xtheme
 ~~~~~~
 
--- a/doc/changelog.rst
+++ b/doc/changelog.rst
@@ -37,6 +37,8 @@ Front
 Xtheme
 ~~~~~~
 
+- Skip adding JS-catalog for editing
+
 Classic Gray Theme
 ~~~~~~~~~~~~~~~~~~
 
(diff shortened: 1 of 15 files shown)


shuup/utils/multilanguage_model_form.py
```diff
@@ -26,14 +27,18 @@ def to_language_codes(languages):
     return languages
 
 
+def get_default_language():
+    return getattr(settings, "PARLER_DEFAULT_LANGUAGE_CODE", get_language())
```
Pikkupomo: Should we even enforce this value? (`settings.PARLER_DEFAULT_LANGUAGE_CODE`) without the code default is again all over the place, right?

doc/changelog.rst
```diff
@@ -60,6 +68,12 @@ Guide
 General/miscellaneous
 ~~~~~~~~~~~~~~~~~~~~~
 
+- MultiLanguageModelForm: Avoid partially/empty translation objects
+  - Delete untranslated objects from database
+  - Only set translation object to database if it is translated
+  - Ensure required fields if language is partially translated
```
Pikkupomo: These won't render properly when viewing the file in html. I think you have to add more spaces or something

shuup/utils/multilanguage_model_form.py
```diff
@@ -54,6 +55,7 @@ def __init__(self, **kwargs):
         self.trans_field_map = defaultdict(dict)
         self.trans_name_map = defaultdict(dict)
         self.translated_field_names = []
+        self.required_if_translated_field_names = []
```
gurch101: can't this just be named `required_translated_field_names`? Also minor but if order doesn't matter, this should be a set.

shuup/utils/multilanguage_model_form.py
```diff
@@ -95,18 +99,38 @@ def __getitem__(self, key):
...
             (trans.language_code, trans)
             for trans
-            in translations_model.objects.filter(master=instance, language_code__in=self.languages)
+            in translations_model.objects.filter(master_id=instance.id, language_code__in=self.languages)
         )
         for lang, field_map in six.iteritems(self.trans_field_map):
+            translation_fields = dict((src_name, data.get(src_name)) for src_name in field_map)
+            translation = current_translations.get(lang)
+            # Add translation only if at least one translated field is given
+            if not [value for value in translation_fields.values() if value is not None and value != ""]:
```
gurch101: could this be
```
if not any(translation_fields.values()):
```

tulimaki:
Migration conflicts with https://github.com/shuup/shuup/pull/780 so be carefull when merging. I think I can change this when the language PR is merged.

shuup/admin/templates/shuup/admin/macros/multilanguage.jinja
```diff
@@ -40,6 +40,9 @@ When using multiple language_dependent_content_tabs in a single page, do remembe
                 {% endif %}
             </div>
         {% endfor %}
+        <p class="text-center">
+            {% trans %}Partially translated languages does not fallback to default language{% endtrans %}
```
gurch101: Are we sure we want this message to always be visible? Maybe this is something that should be documented in the guide.
If we do then maybe use this: "Partially translated fields do not fall back to the default language"

shuup/core/models/_categories.py
```diff
@@ -136,10 +136,10 @@ def is_visible(self, customer):
         return True
 
     @staticmethod
-    def _get_slug_name(self):
+    def _get_slug_name(self, translation):
```
gurch101: does this work? `self.safe_translation_getter('title', language_code=language_code)`
If you pass the `language_code` to this function then `generate_multilanguage_slugs` could be made much simpler

shuup/utils/multilanguage_model_form.py
```diff
@@ -125,15 +151,27 @@ def _save_translation(self, instance, translation):
...
+        # slug from translated name.
+        self._save_master(commit)
         return self.instance
 
     def _set_fields_for_language(self, language):
-        self.instance.set_current_language(language)
-        for field in self.translation_fields:
-            value = self.cleaned_data["%s__%s" % (field.name, language)]
-            field.save_form_data(self.instance, value)
+        original_language = self.instance.get_current_language()
```
gurch101: sorry for the late comment :) Looking at the django-parler docs, we should be able to use a context manager here:
```
from parler.utils.context import switch_language

with switch_language(self.instance, language):
    do stuff
```

shuup/core/models/_categories.py
tulimaki: The reason we are passing translation object here is that the product/category-object might not contain the translation. Which will lead in the situation that the `safe_translation_getter` does not return right `title` for the object.

At least the `shuup.utils.multlanguage_model_form:_save_translations` causes situation that the object that we are generating slug for does not have translation even if the translation object is already in the database. Not sure if that is correct either but probably we don't want to change this functionality. @Pikkupomo what do you think about this?

PR #776, opened 2016-08-31
Fix using "force_text()" on coupons

JoaRiski:
Coupons with unicode in their code would result in an UnicodeDecodeError when passed
to the "force_text()" function. Fix this issue.

### DIFF
--- a/shuup/campaigns/models/campaigns.py
+++ b/shuup/campaigns/models/campaigns.py
@@ -11,7 +11,7 @@ from django.conf import settings
 from django.core.exceptions import ValidationError
 from django.db import models
 from django.db.models import Q
-from django.utils.encoding import force_text
+from django.utils.encoding import force_text, python_2_unicode_compatible
 from django.utils.timezone import now
 from django.utils.translation import ugettext_lazy as _
 from enumfields import Enum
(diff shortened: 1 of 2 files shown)


shuup/campaigns/models/campaigns.py
```diff
@@ -397,3 +397,6 @@ def save(self, **kwargs):
 
     def __str__(self):
         return self.code
+
+    def __unicode__(self):
```
akx: Don't define `__unicode__`, use the `@python_2_unicode_compatible` decorator on the class.

shuup_tests/campaigns/test_coupons.py
```diff
@@ -0,0 +1,22 @@
...
+# This source code is licensed under the AGPLv3 license found in the
+# LICENSE file in the root directory of this source tree.
+import pytest
+
+from django.utils.encoding import force_text
+
+from shuup.campaigns.models.campaigns import Coupon
+
+@pytest.mark.django_db
+def test_context_contact_group_condition(rf):
```
akx: Is the name of the test correct?

JoaRiski: Whoops!

shuup/campaigns/models/campaigns.py
JoaRiski: Cheers!

PR #780, opened 2016-09-01
Allow shop languages to be set via admin (SH-189)

gurch101:
Refs SH-189

### DIFF
--- a/shuup/admin/modules/shops/views/edit.py
+++ b/shuup/admin/modules/shops/views/edit.py
@@ -12,6 +12,7 @@ from django.conf import settings
 from django.db.transaction import atomic
 from django.utils.translation import ugettext_lazy as _
 
+from shuup import configuration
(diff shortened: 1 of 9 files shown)


shuup/admin/modules/shops/views/edit.py
```diff
@@ -42,8 +44,21 @@ def __init__(self, **kwargs):
             required=True,
             label=_("Currency")
         )
+        self.fields["languages"] = forms.MultipleChoiceField(
+            choices=settings.LANGUAGES,
+            initial=initial_languages,
+            required=True,
+            label=_("Languages")
```
shawnadelic: Probably need to makemessages

tulimaki:
LGTM. Add the translation string and I think this can be merged. The admin language changer is probably not touched on purpose. I think it still can show all the options even if the shop has limited the languages.

PR #782, opened 2016-09-01
Package orderability changes (SH-145)

shawnadelic:
Fix handling of package orderability in Front and Core.

Also, display package orderability problems to customer in basket view to allow them to make changes (for example, if the stock changes after the product is placed into basket).

One test (test_basket_update_errors) has been removed since implementation has changed a bit regarding when orderability errors are shown.

### DIFF
--- a/shuup/front/basket/objects.py
+++ b/shuup/front/basket/objects.py
@@ -94,7 +94,9 @@ class BaseBasket(OrderSource):
         self.storage = get_storage()
         self._data = None
         self.dirty = False
-        self._lines_cache = None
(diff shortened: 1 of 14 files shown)


shuup/front/basket/objects.py
```diff
@@ -222,17 +223,52 @@ def remove_code(self, code):
         self.dirty = bool(self.dirty or modified)
         return modified
 
+    def _cache_lines(self):
```
shawnadelic: This whole method needs some work

PR #789, opened 2016-09-04
Created notify events and fixed payment order section method name

chessbr:
- Front: Create notify events for payment created, refund created, shipment created and shipment deleted
- Notify: changed Enum typology to return the enum identifier name instead of its value or translatable label


Pikkupomo:
ok to test

tulimaki:
Thank you @chessbr! The payment created signal was already also in our plans and the new notify events also looks like a nice addition to the current ones.

PR #791, opened 2016-09-05
Core: create address formatter setting

chessbr:
A new spec is created to customize an address formatter


tulimaki:
ok to test

tulimaki:
Thanks!

PR #793, opened 2016-09-06
Fix issue with PGSQL migration

Pikkupomo:
In some cases core migration 0005 caused old problem (with pgsql)
to arise where multiple changes in model are being done in same transaction.

No ref

### DIFF
--- a/doc/changelog.rst
+++ b/doc/changelog.rst
@@ -88,6 +88,8 @@ Regions
 General/miscellaneous
 ~~~~~~~~~~~~~~~~~~~~~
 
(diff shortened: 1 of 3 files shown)


Pikkupomo:
retest this please

tulimaki:
Missing changelog also could this PR also introduce Travis tests with postgresql?

Pikkupomo:
I added the changelog, but we can make the postgresql later. I created an issue for that task: https://shuupteam.atlassian.net/browse/SH-233

tulimaki:
Cancel this for now. There have been new migrations etch. @Pikkupomo reopen when you think this is mergeable.

PR #761, opened 2016-08-27
Front: Move Carousel Addon to Front (SH-11)

shawnadelic:
Move code from Shuup Carousel addon to Shuup base.

Also doing some minor refactoring for carousel code.

Refs SH-11

PR #795, opened 2016-09-07
Front: Add carousel app (SH-11)

tulimaki:
Move Shuup Carousel addon to Shuup base under front apps.

Note! Instances using shuup-carousel addon should be updated to use
this new app. There is no migration tools for old carousel and the old
carousels and slides needs to be copied manually to new app before removing
shuup-carousel addon from installed apps.

Modified from commit introduced in https://github.com/shuup/shuup/pull/761 by @shawnadelic

### DIFF
--- a/shuup/front/apps/carousel/forms.py
+++ b/shuup/front/apps/carousel/forms.py
@@ -1,0 +1,29 @@
+# -*- coding: utf-8 -*-
+# This file is part of Shuup.
+#
+# Copyright (c) 2012-2016, Shoop Ltd. All rights reserved.
+#
+# This source code is licensed under the AGPLv3 license found in the
+# LICENSE file in the root directory of this source tree.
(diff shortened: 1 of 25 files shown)


shuup/front/apps/carousel/forms.py
```diff
@@ -0,0 +1,35 @@
...
+        cleaned_data = super(CarouselConfigForm, self).clean()
+        carousel = cleaned_data.get("carousel")
+        cleaned_data["carousel"] = carousel.pk if hasattr(carousel, "pk") else None
+        return cleaned_data
+
+
+class BannerBoxConfigForm(CarouselConfigForm):
+    def populate(self):
+        # Use generic plugin form populate method first
+        GenericPluginForm.populate(self)
```
Pikkupomo: You could explain here why it's used first, mainly for generations to follow :)

PR #797, opened 2016-09-08
Order Printouts: Add option to send printouts as email attachments (SH-71)

tulimaki:
* Order printouts: Move prinouts to tab from toolbar
* Order Printouts: Add options to send printouts as email attachments


shuup/order_printouts/admin_module/section.py
```diff
@@ -0,0 +1,45 @@
...
+
+EMAIL_DEFAULT_BODY = _("""Important information regarding your order, see attachment.
+
+Best Regards,
+%(shop)s""")
+
+
+class PrintoutsSection(Section):
+    identifier = "printouts_section"
+    name = _("Default Printouts")
```
gurch101: should just be called Printouts (or confirmations), IMO.

shuup/order_printouts/admin_module/toolbar.py
```diff
@@ -1,43 +0,0 @@
...
-            shipment_items.append(
-                DropdownItem(
-                    text=_("Get Delivery Slip (Shipment %(id)s)") % {"id": shipment_id},
-                    icon="fa fa-truck",
-                    url=reverse("shuup_admin:printouts.delivery_pdf", kwargs={"shipment_pk": shipment_id})
-                )
-            )
-
-        return shipment_items + [
-            DropdownItem(
```
gurch101: Also add button to get HTML confirmation. PDF confirmation button should only be shown if weasyprint dependency is met.

PR #799, opened 2016-09-09
Core: Price rounding change (SH-129)

tulimaki:
* Round priceful taxless and taxful prices into two decimals
* Round taxes to two decimals when getting order line tax
amount.
* Limit source line base_unit_price and discount_amount to
9 decimals like it is in order line
* Limit priceful price to 9 decimals
* Remove uneccessary rounding while calculating order totals

One option would have been to round priceful price to two decimals,
but it will cause problems when calculating other price components
like discount rate since all the other price components would be
with full precision.

All spots with rounding to two decimals can be changed to round
based on currency decimals later.

See the extra test for refunds which will not be solved by
this PR in case that is a problem. I think we shouldn't refund
with higher precision that the currency allows.

### DIFF
--- a/shuup/admin/modules/sales_dashboard/dashboard.py
+++ b/shuup/admin/modules/sales_dashboard/dashboard.py
@@ -20,7 +20,6 @@ from shuup.core.pricing import TaxfulPrice
 from shuup.core.utils.query import group_by_period
 from shuup.utils.dates import get_year_and_month_format
 from shuup.utils.i18n import get_current_babel_locale
-from shuup.utils.numbers import bankers_round
 
 
 def get_orders_by_currency(currency):
@@ -47,10 +46,7 @@ class OrderValueChartDashboardBlock(DashboardChartBlock):
         ])
         bar_chart.add_data(
             _("Sales (%(currency)s)") % {"currency": self.currency},
-            [
-                bankers_round(v["sum"], 2)  # TODO: To be fixed in SHUUP-1912
-                for v in aggregate_data.values()
-            ]
+            [v["sum"] for v in aggregate_data.values()]
         )
         return bar_chart
 
--- a/shuup/core/fields/__init__.py
+++ b/shuup/core/fields/__init__.py
@@ -25,6 +25,8 @@ from shuup.utils.i18n import get_current_babel_locale
 
 IdentifierValidator = RegexValidator("[a-z][a-z_]+")
 
+MONEY_FIELD_DECIMAL_PLACES = 9
+
 
 class InternalIdentifierField(models.CharField):
 
@@ -109,7 +111,7 @@ class FormattedDecimalField(models.DecimalField):
 
 class MoneyValueField(FormattedDecimalField):
     def __init__(self, **kwargs):
-        kwargs.setdefault("decimal_places", 9)
+        kwargs.setdefault("decimal_places", MONEY_FIELD_DECIMAL_PLACES)
         kwargs.setdefault("max_digits", 36)
         super(MoneyValueField, self).__init__(**kwargs)
 
--- a/shuup/core/order_creator/_source.py
+++ b/shuup/core/order_creator/_source.py
@@ -18,6 +18,7 @@ from enumfields import Enum
 from six import iteritems
 
 from shuup.core import taxing
+from shuup.core.fields.utils import ensure_decimal_places
 from shuup.core.models import (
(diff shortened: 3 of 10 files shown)


gurch101:
LGTM. This is probably high impact enough to warrant a second set of eyes :)

shuup/admin/modules/sales_dashboard/dashboard.py
```diff
@@ -48,7 +47,7 @@ def get_chart(self):
         bar_chart.add_data(
             _("Sales (%(currency)s)") % {"currency": self.currency},
             [
-                bankers_round(v["sum"], 2)  # TODO: To be fixed in SHUUP-1912
+                v["sum"]
                 for v in aggregate_data.values()
```
Pikkupomo: now that the 1st line of this is way shorter, you could maybe use different notation `[v["sum"] for v in aggregate_data.values]` instead of multiple lines?

shuup/core/fields/__init__.py
```diff
@@ -25,6 +25,8 @@
 
 IdentifierValidator = RegexValidator("[a-z][a-z_]+")
 
+DECIMAL_FIELD_DECIMAL_PLACES = 9
```
Pikkupomo: Is this being added to _all_ decimal fields we are using? if not, use lower level values too `MONEY_FIELD_DECIMAL_PLACES = 9´ etc. If it's being used everywhere, nevermind.

Pikkupomo: Also this could be `SHUUP_DECIMAL_FIELD_PLACES` since this has nothing to do with Django I suppose?

shuup/core/order_creator/_source.py
```diff
@@ -569,9 +570,10 @@ def __init__(self, source, **kwargs):
             self.tax_class = tax_class
         self.supplier = kwargs.pop("supplier", None)
         self.quantity = kwargs.pop("quantity", 0)
-        self.base_unit_price = kwargs.pop("base_unit_price", source.zero_price)
-        self.discount_amount = (kwargs.pop("discount_amount", None) or
-                                source.zero_price)
+        self.base_unit_price = kwargs.pop("base_unit_price", source.zero_price).quantize(
+            decimal.Decimal(".1") ** DECIMAL_FIELD_DECIMAL_PLACES)
+        self.discount_amount = kwargs.pop("discount_amount", source.zero_price).quantize(
+            decimal.Decimal(".1") ** DECIMAL_FIELD_DECIMAL_PLACES)
```
Pikkupomo: Is this exponent usage here correct?

shuup/core/pricing/_priceful.py
```diff
@@ -53,9 +56,16 @@ def price(self):
...
+
+        base_unit_price      *        quantity -  discount_amount
+        940.234529877 EUR (excl. tax) 1.000000000 0E-9 EUR (excl. tax)
+        return 40.234529877000000000 EUR (excl. tax)
+
         :rtype: shuup.core.pricing.Price
         """
-        return self.base_unit_price * self.quantity - self.discount_amount
+        return (self.base_unit_price * self.quantity - self.discount_amount).quantize(
+            decimal.Decimal(".1") ** DECIMAL_FIELD_DECIMAL_PLACES)
```
Pikkupomo: second place this is used, should we use something like 
```python
 def ensure_decimal_places(value):
 	return value.quantize(decimal.Decimal(".1") ** DECIMAL_FIELD_DECIMAL_PLACES)
```

shuup/core/pricing/_priceful.py
```diff
@@ -172,21 +182,15 @@ def taxful_price(self):
...
     def taxless_price(self):
         """
         :rtype: TaxlessPrice
         """
         price = self.price
-        if price.includes_tax:
-            return TaxlessPrice(price.amount - self.tax_amount)
-        else:
-            return price
+        return bankers_round((TaxlessPrice(price.amount - self.tax_amount) if price.includes_tax else price), 2)
```
Pikkupomo: What's the use for `price` after this? if anyone ever uses that, (s)he will end up having wrong values. Should `price` be private?

tulimaki: Well price is used for example to calculate discount_rate and also these values. It probably can't be private since priceful is also used for priceinfo and there the price attribute is overriden.

PR #813, opened 2016-09-16
Admin: excluding form current instance from SKU unique validation

chessbr:


tulimaki:
ok to test

tulimaki:
Thank you very much!

PR #817, opened 2016-09-18
Bugfix for custom user model

chessbr:
When the AUTH_USER_MODEL class has not a username field it was impossible to reset passwords


tulimaki:
ok to test

PR #818, opened 2016-09-19
Auth: Enable user to configure emails sent by auth app to use HTML

chessbr:
A setting was created to make it possible to change EmailMessage
content_subtype from plain text to whatever is needed, especially html.


tulimaki:
ok to test

PR #826, opened 2016-09-26
Add shop setup wizard (SH-37)

gurch101:
* Add a shop setup wizard to admin which guides the merchant through the initial setup of the shop.
Includes setting up shop contact info/currency, one payment and shipping method, theme selection, and shop enabling. Wizard is shown automatically on admin login if any setup is required and the wizard has not been run before.
* Add shop setup page that shows steps remaining before shop can be "launched".
* Add quicklinks menu item for frequently used menu items


shuup/admin/utils/wizard.py
```diff
@@ -0,0 +1,25 @@
...
+    """
+    Check if shop wizard should be run.
+
+    :return: whether setup wizard needs to be run
+    :rtype: boolean
+    """
+    if getattr(settings, "SHUUP_ENABLE_MULTIPLE_SHOPS", False):
+        # setup wizard is only applicable in single shop mode
+        return True
+    shop = Shop.objects.first()
```
tulimaki: Something else should be checked here too. Maybe that there is at least one enabled payment and shipping method. Also theme should be active. Maybe it could even tell a reason why the wizard pop up.

tulimaki: Also the shop probabl should be enabled/maintenance mode off to awoid this wizard.

shuup/admin/modules/service_providers/forms.py
```diff
@@ -22,3 +24,35 @@ class CustomPaymentProcessorForm(ShuupAdminForm):
     class Meta:
         model = CustomPaymentProcessor
         exclude = ("identifier", )
+
+
+class ServiceWizardFormMixin(object):
+    def save(self):
+        provider = super(ServiceWizardFormMixin, self).save()
+        if Shop.objects.count() == 1 and TaxClass.objects.count() == 1 and not self.service_exists(provider):
```
tulimaki: Ok here you should probably include creating default class before creating service providers. Otherwise the services is never created.

tulimaki:
Instead of providing stuff to this wizard could it better to jsut define all the wizard steps in settings. That would allow us to customize the wizard process based on the project.

shuup/admin/modules/shops/forms.py
```diff
@@ -0,0 +1,102 @@
...
+            "postal_code", "city",
+            "region_code", "region",
+            "country"
+        )
+
+
+class ShopWizardForm(MultiLanguageModelForm):
+    class Meta:
+        model = Shop
+        fields = ["name", "currency"]
```
tulimaki: Could you add also the `prices_include_tax` here and add a help_text for it that explains what is the deal with that. And you actually define how to show prices in contact groups.

shuup/admin/views/wizard.py
```diff
@@ -0,0 +1,133 @@
...
+        for form in six.itervalues(self.forms):
+            form.save()
+
+
+class WizardView(FormPartsViewMixin, CreateOrUpdateView):
+    model = Shop
+    context_object_name = "shop"
+    template_name = "shuup/admin/wizard/wizard.jinja"
+    form_part_class_provide_key = "admin_setup_wizard_form_part"
+    wizard_panes = [
```
tulimaki: There is something weid about this. I think these should be defined in settings. And for example the shop module could introduce the `GENERAL_INFO_PANE_TYPE` and `FINAL_PANE_TYPE`.

tulimaki:
LGTM

PR #827, opened 2016-09-27
Pricing: Avoid rounding multiple times (SH-308)

tulimaki:
Fix issues with rounding stuff multiple times.

The rounded taxless or taxful total shouldn't be calculated based on rounded values like rounded tax_amount.

The `Priceful.tax_rate` shouldn't never be calculated based on rounded taxless or taxful total.

Priceful properties where the taxfulness is forced with `TaxfulFrom` or `TaxlessFrom` should never be calculated based on rounded taxless or taxful total.

Add `Priceful.raw_taxful_price` and `Priceful.raw_taxless_price` which could be used to calculate the properties mentioned above.

Skip ensuring the `Priceful.price` since the priceful properties should stay untouched until those are shown somewhere. Only `OrderSourceLine` should ensure things to make sure the key components that is saved later to database with certain precision will match.

Ensure decimal limit in `OrderSource.tax_amount` the order source tax amount should have same precision as the order line tax amount does.

At tax summary round line taxless based on values before calculating SUM from those. This is same functionality that is used for order lines. Also the taxless line total in tax summary should match with order taxless total.

At tax summary round line taxful based on the raw based on value. Here we can't use already rounded based on value since we shouldn't add unrouded tax amount to already rounded value. Also the lines total in tax summary should match with order taxful total.

### DIFF
--- a/shuup_tests/core/test_rounding.py
+++ b/shuup_tests/core/test_rounding.py
@@ -13,7 +13,10 @@ from shuup.core.models import OrderLine
 from shuup.core.models import OrderLineType
 from shuup.core.models import Shop
 from shuup.core.models import ShopStatus
-from shuup.testing.factories import create_empty_order
+from shuup.testing.factories import (
+    add_product_to_order, create_empty_order, create_product,
+    get_default_shop, get_default_supplier
+)
 from shuup.utils.numbers import bankers_round
 from shuup_tests.utils.basketish_order_source import BasketishOrderSource
 
@@ -39,7 +42,6 @@ def test_rounding(prices):
 
     order = create_empty_order(prices_include_tax=False)
     order.save()
-    currency = order.shop.currency
     for x, price in enumerate(prices):
         ol = OrderLine(
             order=order,
@@ -59,6 +61,10 @@ def test_rounding(prices):
 
         # make sure the line taxless price is rounded
         assert order_line.taxless_price == order.shop.create_price(bankers_round(price, 2))
+
+        # Check that total prices calculated from priceful parts still matches
+        assert _get_taxless_price(order_line) == order_line.taxless_price
+        assert _get_taxful_price(order_line) == order_line.taxful_price
 
         # make sure the line price is rounded
         assert order_line.price == order.shop.create_price(price)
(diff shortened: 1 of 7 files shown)


shuup_tests/functional/test_discount_taxing.py
```diff
@@ -205,7 +209,7 @@ def test_3prods_services_2discounts(taxes):
             'Tax-C   0.50   45.333333333   22.666666667   68.000000000',
             'Tax-A   0.25   16.320000000    4.080000000   20.400000000',
             'Tax-B   0.20   88.541666667   17.708333333  106.250000000',
-            'Total   0.30  150.190000000   44.455000000  194.650000000']
+            'Total   0.30  150.180000000   44.455000000  194.650000000']
```
gurch101: if this is supposed to represent the sum of the previous three lines, shouldn't this be 150.19 as before (45.33 + 16.32 + 88.54)? wouldn't this show up as order lines not summing to the order total?

shuup_tests/core/test_rounding.py
```diff
@@ -100,8 +106,52 @@ def test_order_source_rounding(prices):
...
+            base_unit_price=order.shop.create_price(price)
+        )
+        ol.save()
+    order.cache_prices()
+    for x, order_line in enumerate(order.lines.all().order_by("ordering")):
+        # Check that total prices calculated from priceful parts still matches
+        assert _get_taxless_price(order_line) == order_line.taxless_price
+        assert _get_taxful_price(order_line) == order_line.taxful_price
+
+
```
gurch101: i think it would be a good idea to also add tests to ensure all the invariants in `Priceful` are still met (block starting at line 44 in `_priceful.py`)

shuup_tests/functional/test_discount_taxing.py
tulimaki: Well not sure how this should be. This basically happens in this case the tax_amount in source used to have over 9 decimals. And when ensuring tax amount to 9 decimals we go 1 cent. But without ensuring the 9 decimals source line taxes does not match order line taxes and will cause some other places to fail.

shuup_tests/core/test_rounding.py
```diff
@@ -100,8 +106,52 @@ def test_order_source_rounding(prices):
...
     # make sure order total is rounded
     assert source.taxless_total_price == source.shop.create_price(bankers_round(expected, 2))
+
+
+@pytest.mark.parametrize("prices", PRICE_SPEC)
+@pytest.mark.django_db
+def test_rounding_with_taxes(prices):
+    shop = get_default_shop()
+    supplier = get_default_supplier()
+    expected = 0
```
JoaRiski: Is this variable actually used anywhere besides for filling up?

tulimaki:
Well the sum(orderlines) is always order total. That is how it used to be also. Now what it seems is that the tax-summary doesn't match. The goal of this is that also all the priceful components can be used to calculate stuff without mutliple roundings.

Also it would be nice if source -> order maintaines the total price. There the main issue now seems to be that prices in lines is max 9 decimals precision and in source the precisin seems to be able to be higher (it shouldn't).

shuup_tests/functional/test_discount_taxing.py
tulimaki: This issue should now be fixed.

shuup_tests/core/test_rounding.py
tulimaki: Add price check here and some other to `functionality/test_discount_taxing.py`

tulimaki:
Based on the feedback I ended up tuning the tax summary. Prices there should be rounded like those are for order line. Also the tax summary lines total should really match the order total. The match can be only achieved in case the line taxless and taxfull totals is also rounded with same logic that in order lines and orders.

Pikkupomo:
I also tested this locally and it seems to be properly working now, problem was indeed that tax summary didn't round or ensure prices.

PR #831, opened 2016-10-04
Campaigns: Fix some matching issues (SH-330)

tulimaki:
* Avoid matching inactive filters and conditions
* Fix bug in product type catalog filter matching

### DIFF
--- a/shuup/campaigns/models/campaigns.py
+++ b/shuup/campaigns/models/campaigns.py
@@ -118,6 +118,11 @@ class CatalogCampaign(Campaign):
     def __str__(self):
         return force_text(_("Catalog Campaign: %(name)s" % dict(name=self.name)))
 
+    def save(self, *args, **kwargs):
+        super(CatalogCampaign, self).save(*args, **kwargs)
+        self.filters.update(active=self.active)
+        self.conditions.update(active=self.active)
+
     def rules_match(self, context, shop_product, matching_catalog_filters, matching_context_conditions):
         if not self.is_available():
             return False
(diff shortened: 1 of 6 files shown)


shuup/campaigns/models/campaigns.py
```diff
@@ -118,6 +118,12 @@ class CatalogCampaign(Campaign):
     def __str__(self):
         return force_text(_("Catalog Campaign: %(name)s" % dict(name=self.name)))
 
+    def save(self, *args, **kwargs):
+        super(CatalogCampaign, self).save(*args, **kwargs)
+        if not self.active:
+            self.filters.update(active=self.active)
+            self.conditions.update(active=self.active)
```
Pikkupomo: Should this work other way around too?

shuup/campaigns/models/campaigns.py
```diff
@@ -204,6 +210,11 @@ class BasketCampaign(Campaign):
     def __str__(self):
         return force_text(_("Basket Campaign: %(name)s" % dict(name=self.name)))
 
+    def save(self, *args, **kwargs):
+        super(BasketCampaign, self).save(*args, **kwargs)
+        if not self.active:
+            self.conditions.update(active=self.active)
```
Pikkupomo: Should this work other way around too?

shuup/campaigns/models/campaigns.py
tulimaki: Hmm yes maybe better to just always update the `active` to be same with the campaign.

tulimaki:
Added the fix for the saves.

PR #833, opened 2016-10-06
Set zero sum orders as paid on creation

JoaRiski:
It isn't possible to create payments on orders with zero as their total price
However, said orders should still be marked as fully paid

Circumvent the issue by marking zero sum orders as paid on creation

### DIFF
--- a/shuup/core/models/_orders.py
+++ b/shuup/core/models/_orders.py
@@ -229,7 +229,7 @@ class Order(MoneyPropped, models.Model):
     payment_data = JSONField(blank=True, null=True, verbose_name=_('payment data'))
 
     shipping_method = UnsavedForeignKey(
-        "ShippingMethod", related_name='shipping_orders',  blank=True, null=True,
+        "ShippingMethod", related_name='shipping_orders', blank=True, null=True,
         default=None, on_delete=models.PROTECT,
         verbose_name=_('shipping method'))
     shipping_method_name = models.CharField(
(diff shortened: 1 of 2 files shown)


JoaRiski:
I'll fix the tests and add a new one soon

tulimaki:
I don't agree with this. What if the order is zero price for a reason, maybe it get's the price after seller jave had a look at it. Maybe it's some service which price is defined later in some process.

If this is I guess the set complete could set the paid status to fully paid or something. Not even sure why there needs to be payment for order that really does not have any payments in it.

@Pikkupomo what do you think?

Pikkupomo:
Order can be zero priced for various reasons. A simple example comes to my mind where a seller creates order for customer X just with one text line `Add product Y and Z to this order when they are in stock and ship the orders to customer`. It's hard to see why this order should be paid at all.

Zero total => cannot be paid per se. This is something that *shouldn't* happen automatically.

JoaRiski:
Another option is indeed by doing that when the order is set complete, which is also what I originally considered, but there doesn't seem to be any guaranteed way of catching when an order is set as complete. Only place this could be done reliably in is the `save` function, but I'm not sure if it's the right place.

What are your thoughts on this? @tulimaki @Pikkupomo

tulimaki:
Like I said not sure if it's right to set something paid that is not.

Could you add `set_complete` method for order. Also benefit would be that the order get's control on that state change. That way you can add this logic there. There shouldn't be many places setting order complete now and make those use this new method.

JoaRiski:
@tulimaki Issue with this is that there's no "sure" way of always catching the event when an order is set as complete. Only feasible way of doing this is catching it in the `save` function as far as I see it. Would you consider that acceptable?

PR #842, opened 2016-10-11
Front: Enable modifying prouduct lists queryset and filter with product variations (SH-284 / SH-346)

tulimaki:
* Enable modyfying products queryset
* Enable option to filter with variation values

### DIFF
--- a/shuup/front/views/category.py
+++ b/shuup/front/views/category.py
@@ -11,7 +11,8 @@ from django.views.generic import DetailView
 
 from shuup.core.models import Category, Product
 from shuup.front.utils.sorts_and_filters import (
-    get_query_filters, post_filter_products, ProductListForm, sort_products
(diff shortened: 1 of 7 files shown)


shuup/front/views/category.py
```diff
@@ -43,7 +44,9 @@ def get_context_data(self, **kwargs):
             shop_products__shop=self.request.shop,
             variation_parent__isnull=True,
             shop_products__categories=category
-        ).filter(get_query_filters(self.request, category, data=data)).distinct()
+        ).filter(get_query_filters(self.request, category, data=data))
+        products = get_product_queryset(products, self.request, category, data)
+        products = products.distinct()
```
Pikkupomo: I think `product = get_product_queryset(products, self.request, category, data).distinct()` might work as well. or does it? :)

PR #848, opened 2016-10-11
Front: Fill search results with words in query (SH-357, SH-254)

tulimaki:
* Pass data for `ProductListFormModifier.sort_products`
* Add search results from words in query string
* Add default sorting based on distance between product name and full query distance

### DIFF
--- a/shuup/front/apps/simple_search/forms.py
+++ b/shuup/front/apps/simple_search/forms.py
@@ -8,6 +8,7 @@
 from __future__ import unicode_literals
 
 import re
+from difflib import SequenceMatcher
 
 from django import forms
 from django.db.models import Q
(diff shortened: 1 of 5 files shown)


shuup/front/apps/simple_search/forms.py
```diff
@@ -91,3 +107,19 @@ def get_filters(self, request, data):
...
+        sort = data.get("sort")
+        if sort:  # Sort only if not sort available
+            return products
+
+        query_str = data.get("q")
+        if not query_str:  # Do not sort if no query string
+            return products
+
+        def _get_product_distance_to_query_str(product):
+            return Levenshtein.distance(product.name, query_str)
```
gurch101: not sure using Levenshtein is the right approach since this is usually used for spell check and can degrade to O(n^2). Plus since were matching with `icontains`, we won't have substitutions anyways. Why not use something like `ORDER BY CASE sku CONTAINS query THEN 1 CASE translations__name CONTAINS query THEN 2....etc`. I'm *pretty sure* this is SQL92 compatible. http://stackoverflow.com/questions/10329849/django-order-by-specific-order

PR #853, opened 2016-10-17
Fix issue with selectpickers

Pikkupomo:
Selectpickers were not working when product page content was loaded with ajax.

Move extrajs definition outside if...else and initialize selectpickers in category list js

no ref

### DIFF
--- a/shuup/front/templates/shuup/front/product/category.jinja
+++ b/shuup/front/templates/shuup/front/product/category.jinja
@@ -23,8 +23,8 @@
         {{ category_macros.render_information() }}
         {{ category_macros.render_products_section() }}
     {% endblock %}
-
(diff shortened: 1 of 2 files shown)


shuup/front/templates/shuup/front/product/category.jinja
```diff
@@ -23,8 +23,7 @@
         {{ category_macros.render_information() }}
         {{ category_macros.render_products_section() }}
     {% endblock %}
-
-    {% block extrajs %}
-        {{ category_macros.render_extrajs() }}
-    {% endblock %}
 {% endif %}
+{% block extrajs %}
```
tulimaki: Add extra line between blocks.

PR #857, opened 2016-10-21
Add checkout view with register and login phases (SH-376)

tulimaki:
* Core: Populate some unfilled customer fields from order
* Front: Add is_visible_for_user method for checkout view phase
* Front: Add checkout view with option to login and register
* Front: Enable login and register phases for vertical process

### DIFF
--- a/shuup/front/checkout/checkout_method.py
+++ b/shuup/front/checkout/checkout_method.py
@@ -1,0 +1,98 @@
+# -*- coding: utf-8 -*-
+# This file is part of Shuup.
+#
+# Copyright (c) 2012-2016, Shoop Commerce Ltd. All rights reserved.
+#
+# This source code is licensed under the AGPLv3 license found in the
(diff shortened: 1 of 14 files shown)


shuup/front/checkout/checkout_method.py
```diff
@@ -0,0 +1,98 @@
...
+        return
+
+
+class RegisterPhase(CheckoutPhaseViewMixin, RegistrationNoActivationView):
+    identifier = "register"
+    title = _("Register")
+    template_name = "shuup/front/checkout/register.jinja"
+
+    def is_visible_for_user(self):
+        checkout_method_choice_is_registered = bool(self.storage.get(CHECKOUT_CHOICE_STORAGE_KEY, None))
```
gurch101: can't this be defaulted if the user is already logged in? User shouldn't have to log in twice.

PR #866, opened 2016-10-23
Make two strings available to translations

itbabu:
The strings are: `stored baskets` in the dashboard cart list  and `Cart saved!` while saving a basket.

Note: I had to add a new migration


tulimaki:
ok to test

PR #869, opened 2016-10-24
Use correct model in ProductCrossSellFormSet

finmaster19:
ProductCrossSellFormSet is using wrong model in ProductCrossSellEditView, which can cause errors in some shops. Change ProductCrossSellFormSet model to use ProductCrossSell instead of ProductMedia model.

Now when trying to delete a product from product's cross-sell, form errors are:
[{'id': ['Select a valid choice. That choice is not one of the available choices.']}, {'id': ['Select a valid choice. That choice is not one of the available choices.']},... and nothing happens. After I changed to use ProductCrossSell model everything works correct.

### DIFF
--- a/doc/changelog.rst
+++ b/doc/changelog.rst
@@ -21,6 +21,7 @@ Localization
 Admin
 ~~~~~
 
+- Fix bug in product cross-sell editview
 - Allow product attribute form extension through provides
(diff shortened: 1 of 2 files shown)


tulimaki:
ok to test

tulimaki:
`products/views/edit_cross_sell.py Imports are incorrectly sorted.`

PR #878, opened 2016-10-26
Category extensions (PP-5)

gurch101:
Add mechanism to automatically populate products into categories.

work done entirely by @Pikkupomo. I merely cleaned up tests.

Refs PP-5


shuup/category_extensions/management/commands/update_populated_categories.py
```diff
@@ -0,0 +1,23 @@
...
+
+from shuup.category_extensions.models.category_populator import \
+    CategoryPopulator
+from shuup.core.models import ShopProduct
+
+
+class Command(BaseCommand):
+
+    def handle(self, *args, **options):
+        for shop_product in ShopProduct.objects.all():
```
tulimaki: Imo we can't do this. Is there some filters we can use instead?

One approach would be to use `CategoryPopulator` in main loop.

tulimaki:
Add changelog, translations and .tx entry.

PR #880, opened 2016-10-26
Front: Show variation parents in highlight plugin (PP-49)

gurch101:
Show variation parents in product highlight plugins to avoid the variants from flooding the results.

thanks to @tulimaki and @Pikkupomo for their input!

Refs PP-49


tulimaki:
LGTM, will merge in case the tests passes.

PR #881, opened 2016-10-26
Front: Remove flex from product and search product lists (PP-45)

Pikkupomo:
Safari was not wrapping the product lists in categories and search results properly

Refs PP-45

### DIFF
--- a/shuup/front/static_src/less/front/product-list.less
+++ b/shuup/front/static_src/less/front/product-list.less
@@ -188,28 +188,48 @@
 }
 
 .product-list-view {
-    display: flex;
-    flex-wrap: wrap;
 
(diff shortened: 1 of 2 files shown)


tulimaki:
Now the search view only shows three (3) products per row which used to be five (5) before? Since there is no sidebar not sure if the three is good idea.

shuup/front/static_src/less/front/product-list.less
```diff
@@ -188,28 +188,50 @@
 }
 
 .product-list-view {
-    display: flex;
-    flex-wrap: wrap;
 
+    //.single-product {
```
tulimaki: Remove this!

PR #886, opened 2016-10-31
Fix problems with `Priceful.tax_rate` precision  (SH-341)

Pikkupomo:
Fix a bug in postal code when using wildcard
Ensure decimal places in `Priceful` objects' `tax_rate`

Refs SH-341


tulimaki:
Did you check that all calculations is now using this new `raw_tax_rate` property?

shuup/default_tax/models.py
```diff
@@ -17,11 +17,12 @@
...
 class TaxRuleQuerySet(models.QuerySet):
     def may_match_postal_code(self, postalcode):
-        null = Q(_postal_codes_min__isnull=True)
+        null = Q(Q(_postal_codes_min__isnull=True) | Q(_postal_codes_min=""))
+        star = Q(_postal_codes_min="*")
         in_range = Q()
         if postalcode:
             in_range = Q(_postal_codes_min__lte=postalcode, _postal_codes_max__gte=postalcode)
-        return self.filter(null | in_range)
+        return self.filter(star | null | in_range)
```
suutari-ai: I think this should be fixed by making `Pattern.get_alphabetical_limits()` return `(None, None)` for `*`. Then there would be no need to check `*` pattern here.

shuup/default_tax/models.py
```diff
@@ -86,6 +87,8 @@ def matches(self, taxing_context):
         return True
 
     def save(self, *args, **kwargs):
+        self._postal_codes_min = None
+        self._postal_codes_max = None
         if self.postal_codes_pattern:
             min_value, max_value = Pattern(self.postal_codes_pattern).get_alphabetical_limits()
             self._postal_codes_min = min_value
```
suutari-ai: You may just remove the `if self.postal_codes_pattern:` line, since `_postal_codes_min` and `_postal_codes_max` should always be set by and `get_alphabetical_limits` should always return the correct value for them.

shuup_tests/default_tax/test_default_tax.py
```diff
@@ -182,6 +182,23 @@ def test_rule_min_max():
     postal_code = None
     assert TaxRule.objects.may_match_postal_code(postal_code).count() == 1
 
+@pytest.mark.django_db
+def test_wildcard_postalcode():
+    tax = create_tax("test-1", rate=Decimal("0.12"))
+    tax.save()
+    rule = TaxRule.objects.create(postal_codes_pattern="*", tax=tax)
+    assert rule._postal_codes_min == "*"
+    assert rule._postal_codes_max == "*"
```
suutari-ai: IMHO these asserts are wrong. You shouldn't assert about the inner state of the object. What matters is that it matches correctly (the assert below), it doesn't matter how (the asserts here)...

shuup/default_tax/models.py
suutari-ai: I think this should be fixed by making Pattern.get_alphabetical_limits() return (None, None) for *. Then there would be no need to check * pattern here.

shuup/default_tax/models.py
suutari-ai: You may just remove the `if self.postal_codes_pattern:` line, since `_postal_codes_min` and `_postal_codes_max` should always be set and `get_alphabetical_limits` should always return the correct values for them.

shuup_tests/default_tax/test_default_tax.py
suutari-ai: IMHO these asserts are wrong. You shouldn't assert about the inner state of the object. What matters is that it matches correctly (the assert below), it doesn't matter how (the asserts here)...

shuup/core/pricing/_priceful.py
```diff
@@ -158,6 +159,10 @@ def unit_discount_amount(self):
 
     @property
     def tax_rate(self):
+        return ensure_decimal_places(self.raw_tax_rate)
```
suutari-ai: I don't think Priceful should do any rounding...

tulimaki: I think this is ok if we don't use the `tax_rate` to calculate anything and only for showing the final value. The `raw_tax_rate` should be used to do calculations.

suutari-ai: Showing the values should done with formatting functions anyway. You don't want to show "0.123456789", but "12.346%" on "12,346 %" (depending on locale). So the rounding should happen in `format_percent` or the `|percent` Jinja-filter, not in Priceful which should just do the pure numbers without rounding.

Same applies for prices and other money amounts. We don't want to show the pure decimal value 12.4559, but rather "$12.46" or "12,46 $" etc. For money amounts there is `format_money` Python function and `|money` Jinja filter which should do the rounding.

I don't see any good reason to do rounding in Priceful level.

tulimaki:
Also this is back in the drawing board. Will not be included in next release. For me it's ok if we let this hang here until this is fixed.

shuup/core/pricing/_priceful.py
suutari-ai: How it fails? Sounds like a bug in `format_percent` then. How about fixing that instead?

suutari-ai:
The first commit (47a77bd) is good btw. :+1:

tulimaki:
Separated tax fixes to https://github.com/shuup/shuup/pull/941 so we can merge those.

shuup/core/pricing/_priceful.py
suutari-ai: Shouldn't 0.1976500000123497 be formatted as "19.765%"? If it isn't then format_percent doesn't work correctly. (It should round the value with ROUND_HALF_UP or ROUND_HALF_EVEN, certainly not ROUND_UP which would cause the incorrect 19.766%)

PR #889, opened 2016-11-02
Fix order rounding

Pikkupomo:
This should also fix the full refunds of products with 3 decimal points in price

SH-338

### DIFF
--- a/shuup/core/models/_orders.py
+++ b/shuup/core/models/_orders.py
@@ -42,6 +42,7 @@ from shuup.core.signals import (
 from shuup.utils.analog import define_log_model, LogEntryKind
 from shuup.utils.excs import Problem
 from shuup.utils.money import Money
+from shuup.utils.numbers import bankers_round
 from shuup.utils.properties import (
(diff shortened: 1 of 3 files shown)


shuup/core/models/_orders.py
```diff
@@ -302,10 +303,10 @@ def cache_prices(self):
         taxful_total = TaxfulPrice(0, self.currency)
         taxless_total = TaxlessPrice(0, self.currency)
         for line in self.lines.all():
-            taxful_total += line.taxful_price
-            taxless_total += line.taxless_price
-        self.taxful_total_price = taxful_total
-        self.taxless_total_price = taxless_total
+            taxful_total += line.raw_taxful_price
```
tulimaki: IMO this is wrong and order total should be counted from rounded line totals.

tulimaki: Does this change affect on order source also?

Pikkupomo:
Closing this for now before we find a proper fix for this.

PR #890, opened 2016-11-02
Admin Fix Menu (SH-435, SH-433, SH-436, SH-427)

Pikkupomo:
Main menu didn't allow any clicks on mobile size.
The menu was loaded with ajax which caused blinking on load.

In large lists the merchant could not scroll into the last items since they were hidden outside the browser window
Quicklinks menu was way too wide in mobile size, this fix forces it to certain width and alters the styles

* Change the way admin menu is being rendered: show it without ajax
* Fix clicks in mobile size
* Add padding to menu
* Fix quicklinks menu in mobile size

Refs SH-435, SH-433, SH-436, SH-427

### DIFF
--- a/doc/changelog.rst
+++ b/doc/changelog.rst
@@ -23,6 +23,7 @@ Localization
 Admin
 ~~~~~
 
+- Change main menu template and remove ajax loading from main menu.
 - Remove language layer from shop configurations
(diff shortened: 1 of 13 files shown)


tulimaki:
LGTM. Not yet 100% sure whether we should add more tests for this now.

PR #897, opened 2016-11-04
Core: Make contact language required information (SH-439, SH-428)

Pikkupomo:
**Add helpers to identify existing languages**
We are relying on Babel when it comes to translations. It was,
however, possible to use languages that were extinct long time ago.
These helpers will help you identify valid languages for Shuup coding
environment.

**Add `LanguageFormField`**
Also fix issue with extinct languages by using the
newly introduced helpers in Shuup.

** Core: Start actually using `Contact.language`**

**Core: Fix bug with order_source.language**
Source language was never transferred back to Order which caused language being lost when Source was used as a base for Order.

**Core: Make `OrderSource.language` usable**
Old language was pretty much unused and the value was
only applied if the programmer knew that it should be set.

Change the functionality so that if the `OrderSource`
has no language set, the customers
language will be used as a language for `OrderSource` too.

`OrderSource` must have a language set, so if no valid
languages are found, source will use `settings.LANGUAGE_CODE`.

Also add tests to show that this change makes notification emails to
be send in proper language.

Refs SH-439, SH-428


shuup/core/migrations/0011_contact_language.py
```diff
@@ -0,0 +1,20 @@
...
+# Generated by Django 1.9.10 on 2016-11-04 17:37
+from __future__ import unicode_literals
+
+from django.db import migrations
+
+
+class Migration(migrations.Migration):
+
+    dependencies = [
+        ('shuup', '0010_update_managers'),
```
tulimaki: This should be dependent on 11 and this migration 12.

shuup/core/order_creator/_source.py
```diff
@@ -246,6 +248,25 @@ def status(self):
...
 
+    @property
+    def language(self):
+        lang = settings.LANGUAGE_CODE
+        if self._language:
+            lang = self._language
+        elif self.customer:
+            lang = self.customer.language
+
+        if not is_existing_language(lang):
```
tulimaki: Is this if really needed it seems that after the customer doesn't exist the language is `settings.LANGUAGE_CODE`

tulimaki:
LGTM after these few issues. Also changelog could be added since there is new things in this PR.

shuup/core/order_creator/_source.py
Pikkupomo: I was thinking this too, but source.language can be set to something strange thus must be validated. imo.

PR #900, opened 2016-11-04
Core: Do not include not shipped products to shipments (SH-107)

tulimaki:
Skip not shipped products when figuring out unshipped products.

Make default products that is created with `testing.factories.create_product`
shippable since it seems that all tests dealing with shipments until today
assumes that this is the case.


shuup/core/models/_orders.py
```diff
@@ -824,6 +826,13 @@ def get_product_summary(self):
             .values_list("product_id", "quantity"))
         for product_id, quantity in lines:
             products[product_id]['ordered'] += quantity
+
+        from ._products import ShippingMode
```
Pikkupomo: Why is this import here?

shuup/core/models/_orders.py
```diff
@@ -694,10 +694,12 @@ def create_shipment_of_all_products(self, supplier=None):
         :return: Saved, complete Shipment object
         :rtype: shuup.shop.models.Shipment
         """
+        from ._products import ShippingMode
```
Pikkupomo: Why is this import here?

doc/changelog.rst
```diff
@@ -12,6 +12,7 @@ Unrealeased
 Core
 ~~~~
 
+- Do not include not shipped products to shipments
```
Pikkupomo: Changelog should also say that created products are `SHIPPED` by default now

shuup/core/models/_orders.py
tulimaki: It can't be imported at the top of _orders since it causes circular import.

shuup/core/models/_orders.py
tulimaki: It can't be imported at the top of _orders since it causes circular import. Probably shouldn't but couldn't figure the reason fast enough.

PR #909, opened 2016-11-08
Front: Admin tools (SH-365)

Pikkupomo:
Make it possible for merchant to make the decision whether to see all products and categories in front or not

Refs SH-365


shuup/core/models/_contacts.py
```diff
@@ -297,8 +297,10 @@ def save(self, *args, **kwargs):
 
     @property
```
gurch101: we should use `cached_property` here since we're reading from config.

PR #910, opened 2016-11-08
Select first eligible product image as primary if none is selected

finmaster19:
For example now if first form happens to be deleted, default product primary
image is not set. So pick default product primary image from eligible forms. 
Eligible forms are already checked but not used.


finmaster19:
This needs more closer look.

PR #911, opened 2016-11-08
Core: Order Status improvements

Pikkupomo:
Order statuses were used by Shuup in a way where
merchant could not modify them at all. This commit will enable the
modification and adding of the Order Statuses.

This commit also adds `OrderStatusManager` which is a helper class
that ensures that default order statuses are always available. Future
plans with this class is that it handles all the order status changes
with plug-in status change rules.

Shuup now has 4 default order statuses:
* Initial
* Processing
* Completed
* Canceled

These order statuses cannot be modified (only name and public
name can be changed) and they must always be present in shuup.

Breakdown:
* Make `OrderStatus` objects manageable
* Add `OrderStatusManager` to handle and identify default order statuses
* Add `is_active` and `public_name` fields to `OrderStatus`
* Add Admin view for order statuses
* Modify `shuup_init` to always update order statuses
* Add `is_partially_shipped` and `is_partially_paid` methods to order

Refs SH-432


shuup/admin/modules/orders/__init__.py
```diff
@@ -111,6 +113,32 @@ def get_urls(self):
                 "shuup.admin.modules.settings.views.ListSettingsView",
                 name="order.list_settings",
                 permissions=get_default_model_permissions(Order),
+            ),
+
+            # order statuses
+            admin_url(
```
gurch101: any reason you can't use `get_edit_and_list_urls` here?

shuup/admin/modules/orders/views/status.py
```diff
@@ -0,0 +1,73 @@
...
+
+from shuup.admin.breadcrumbs import BreadcrumbedView
+from shuup.admin.utils.picotable import ChoicesFilter, Column, TextFilter
+from shuup.admin.utils.views import CreateOrUpdateView, PicotableListView
+from shuup.core.models import OrderStatus, OrderStatusManager, OrderStatusRole
+from shuup.utils.multilanguage_model_form import MultiLanguageModelForm
+
+
+class OrderStatusForm(MultiLanguageModelForm):
+    identifier = forms.CharField(
```
gurch101: use `InternalIdentifierField`

shuup/admin/modules/orders/views/status.py
```diff
@@ -0,0 +1,73 @@
...
+    template_name = "shuup/admin/orders/status.jinja"
+    context_object_name = "status"
+    parent_name = _("Order Statuses")
+    parent_url = "shuup_admin:order_status.list"
+
+
+class OrderStatusListView(PicotableListView):
+    model = OrderStatus
+    default_columns = [
+        Column("identifier", _("Identifier"), linked=True, filter_config=TextFilter(operator="startswith")),
```
gurch101: I don't think this should be a default column. Ditto with name.

Pikkupomo: I think it should, it's the identifier of the field and in this case it's something that matter and helps merchant to identify field even when (s)he has multiple custom statuses

PR #912, opened 2016-11-09
Add more template documentation

Pikkupomo:
Refs SH-431


gurch101:
Not sure this is needed. Looks like most of this info already exists in `theme.rst`

PR #913, opened 2016-11-09
Admin tour + help text updates (SH-430/SH-445)

gurch101:
* Add admin walkthrough
* Show help text as a popover

Refs SH-430/SH-445


Pikkupomo:
The tour is missing Titles in all of the boxes. IMO they should be present, now the tour boxes seem too "naked".

Pikkupomo:
When going through the menu, it would be nice to have all other menu entries (but the one currently described) somehow blurred.

shuup/admin/static_src/base/js/tour.js
```diff
@@ -0,0 +1,192 @@
...
+        }, {
+            text: [
+                gettext("Find anything, and I mean anything that is inside your store. This includes settings, products, users, and addons.")
+            ],
+            attachTo: "#site-search bottom"
+        }, {
+            text: [
+                gettext("Peep your shop and all the cool changes you’ve made.")
+            ],
+            attachTo: ".shop-btn left"
```
gurch101: yea seems to be an issue in the third party lib. Anything near the edges of the screen is missing arrows.

shuup/admin/static_src/base/js/tour.js
```diff
@@ -0,0 +1,192 @@
+/**
+ * This file is part of Shuup.
+ *
+ * Copyright (c) 2012-2016, Shoop Commerce Ltd. All rights reserved.
+ *
+ * This source code is licensed under the AGPLv3 license found in the
+ * LICENSE file in the root directory of this source tree.
+ */
+(function($) {
```
gurch101: This code is used for all tours. When "re-running" a tour, we run the tour for the page + the chrome.

PR #917, opened 2016-11-11
Send admin user email and last login in telemetry

Pikkupomo:
We need to be able to ensure the instances in
cloud are truly active.

Refs SH-460

### DIFF
--- a/shuup_tests/core/test_telemetry.py
+++ b/shuup_tests/core/test_telemetry.py
@@ -27,9 +27,10 @@ from shuup.core.telemetry import (
 from shuup.testing.factories import (
     create_empty_order, create_order_with_product,
     create_product, create_random_company,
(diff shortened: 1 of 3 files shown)


shuup_tests/core/test_telemetry.py
```diff
@@ -46,13 +47,40 @@ def _clear_telemetry_submission():
...
+    data = json.loads(get_telemetry_data(rf.get("/")))
+    assert data.get("host")
+    assert data.get("admin_user") == admin_user.email
     assert not json.loads(get_telemetry_data(None)).get("host")
 
 
 @pytest.mark.django_db
-def test_optin_optout(rf):
+def test_get_telemetry_data_after_login(rf, admin_user):
+    # create users to ensure correct admin is found
```
tulimaki: There is no Shop for this test.

PR #921, opened 2016-11-14
Fix app_label problem in shuup.testing module

JoaRiski:
Trying to import anything from `shuup.testing.factories` in tests could result in
```
RuntimeError: Model class
shuup.testing.models._behavior_components.ExpensiveSwedenBehaviorComponent doesn't declare an
explicit app_label and isn't in an application in INSTALLED_APPS
```
if done in a Django 1.9 environment.

The changes in this commit seem to fix it, but if not, at least the problem will be localized to a specific import instead of failing on all factories imports


JoaRiski:
@tulimaki / @Pikkupomo

PR #924, opened 2016-11-15
Admin: Misc refund fixes (SH-454/SH-455)

gurch101:
* Show line text for refund line picker
* Fix ability to add more than one refund row at a time
* Allow refunding by arbitrary amount
* Allow refund amount/quantity to be editable

Refs SH-454/SH-455

### DIFF
--- a/doc/changelog.rst
+++ b/doc/changelog.rst
@@ -22,6 +22,7 @@ Localization
 Admin
 ~~~~~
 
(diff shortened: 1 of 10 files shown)


tulimaki:
Can you add more tests around the refund admin view and/or form. Maybe even browser tests. Other that this looks good to me.

PR #925, opened 2016-11-16
Core: Create currency class to use when rounding money amounts (SH-92)

chessbr:
* Add Money.as_rounded() to round money amounts using the Currency class
* Create Admin to create/edit currencies
* Add USD, EUR, BRL, CNY and JPY currencies on shuup_init management command

Refs SH-92


tulimaki:
ok to test

chessbr:
Some telemetry tests failed here too. Is something wrong with the server?

shuup/utils/money.py
```diff
@@ -60,3 +63,30 @@ def unit_matches_with(self, other):
...
+            # the digits info is not cached yet, fetch it
+            if self.currency not in digits_cache:
+                from shuup.core.models._currencies import Currency
+
+                currency = Currency.objects.filter(code=self.currency).first()
+                digits_cache[self.currency] = currency.decimal_places if currency else 0
+
+            return self.new(numbers.bankers_round(self.value, digits_cache[self.currency]))
+        else:
+            return self.new(numbers.bankers_round(self.value, digits))
```
suutari-ai: * Don't add shuup.core.models.Currency stuff to Money class. See below for better way to do this.
* The method should have `rounding` param defaulting to ROUND_HALF_EVEN. (No need to use `bankers_round` btw., just use `Decimal.quantize`)
* Default precision should be 2 digits rather than 0, because it's the most popular money amount precision around the world.

Rather than adding shuup.core stuff to Money, just add accessors for setting the precisions, e.g.
```python
class Money(...):
    default_precision = decimal.Decimal('0.01')
    precisions = digits_to_precisions({
        'USD': 2, 'EUR': 2,  # sensible defaults, add more of these from some nice source like wikipedia...
    })
    @classmethod
    def set_currency_precisions(cls, digits):
        """
        Set default precisions for rounding money amounts.

        :param digits: Map from currency code to number of digits
        :type digits: dict[str,int]
        """
        assert isinstance(digits, dict)
        assert all(isinstance(x, int) for x in digits.values())
        cls.precisions = digits_to_precisions(digits)

    def as_rounded(self, digits=None, rounding=ROUND_HALF_EVEN):
        precision = self.precisions.get(self.currency, self.default_precision)
        return self.new(self.value.quantize(precision, rounding=rounding))


def digits_to_precisions(mapping):
    return {
        key: decimal.Decimal('0.1') ** digits
        for (key, digits) in mapping
    }
```

Then make sure that the precisions are set when shuup.core is loaded, e.g. by setting them in

chessbr: Agreed to not mix core and models in utils and populating the precisions cache on `AppConfig.ready()` also good. What if I also connect a signal to update the precisions on Currency model change?

chessbr: Well, I think signals will not help to spread the change over the server instances.

chessbr: I am proposing another way:

in `utils.money.py`:
```python
def digits_to_precisions(self, mapping):
    return {key: decimal.Decimal('0.1') ** digits for (key, digits) in mapping}

class DefaultPrecisionProvider(object):
    """
    This class provides precision for a given currency.

    Subclass this to implement custom logic and set it
    to the Money class.
    """

    # the most used in the world
    default_precision = decimal.Decimal('0.01')

    # map of precisions for currencies
    currency_precisions = digits_to_precisions({
        'USD': 2,
        'EUR': 2,
        # fill this with more..
    })

    # map os precisions for digits
    digits_precisions = {}

    @classmethod
    def get_precision_for_currency(cls, currency):
        return cls.currency_precisions.get(currency, cls.default_precision)

    @classmethod
    def get_precision_for_digits(cls, digits):
        if digits not in cls.digits_precisions:
            cls.digits_precisions[digits] = (decimal.Decimal('0.1') ** digits)
        return cls.digits_precisions.get(digits)

    @classmethod
    def set_currency_precisions(cls, digits):
        """
        Set default precisions for rounding money amounts.

        :param digits: Map from currency code to number of digits
        :type digits: dict[str,int]
        """
        assert isinstance(digits, dict)
        assert all(isinstance(x, int) for x in digits.values())
        cls.currency_precisions = digits_to_precisions(digits)

class Money(numbers.UnittedDecimal):
    precision_provider = DefaultPrecisionProvider
    # ...

    @classmethod
    def set_precision_provider(cls, precision_provider):
        cls.precision_provider = precision_provider

    def as_rounded(self, digits=None, rounding=ROUND_HALF_EVEN):
        if digits is None:
            precision = self.precision_provider.get_currency_precision(self.currency)
        else:
            precision = self.precision_provider.get_precision_for_digits(digits)

        return self.new(self.value.quantize(precision, rounding=rounding))
```

in `core.models.currencies`:
```python
class CurrencyPrecisionProvider(DefaultPrecisionProvider):
    @classmethod
    def get_precision_for_currency(cls, currency):

        if currency not in cls.currency_precisions:
            currency = Currency.objects.filter(code=currency).first()
            precision = (decimal.Decimal('0.1') ** currency.decimal_places) if currency else self.default_precision
            cls.currency_precisions[currency] = precision

        return cls.currency_precisions.get(currency, cls.default_precision)
```

and finally, in `core.__ini__`:
```python
class ShuupCoreAppConfig(AppConfig):
    # ....
    def ready(self):
        # ....
       Money.set_precision_provider(CurrencyPrecisionProvider)
```

This way we can change the way the precisions are calculated and fetched.

shuup/core/models/_currencies.py
```diff
@@ -0,0 +1,55 @@
...
+                            editable=True,
+                            validators=[MinLengthValidator(3)],
+                            help_text=_("The ISO-4217 code of the currency"))
+
+    decimal_places = models.PositiveSmallIntegerField(verbose_name=_("decimal places"),
+                                                      validators=[MaxValueValidator(10)],
+                                                      default=2)
+
+    translations = TranslatedFields(
+        name=models.CharField(verbose_name=_("name"), max_length=64),
```
tulimaki: You should be able to get the name with using babel the same way we do it now. So there is probably no need for this attribute.

tulimaki:
Missing changelog. Also consider moving the menu entry reordering in it's own commit.

shuup/admin/modules/currencies/views/list.py
```diff
@@ -0,0 +1,36 @@
...
+    default_columns = [
+        Column(
+            "code", _(u"Code"), sort_field="code",
+            filter_config=TextFilter(
+                filter_field="code",
+                placeholder=_("Filter by code"),
+            )
+        ),
+        Column(
+            "name", _("Name"), sort_field="translations__name",
```
tulimaki: Use babel name here instead of name from he model. You can use the `display` parameter.

shuup/utils/money.py
tulimaki: Instead of setting the precision provider. You could also use getter+settings combo like. Very least the app ready should set the precision provider from settings so it's easily changeable.

Pikkupomo: @chessbr One of the requirements is to have merchant to be able to set the precision for currency. So if we have `DefaultPrecisionProvider` it must offer these in way more dynamic manner:
```python
# map of precisions for currencies
    currency_precisions = digits_to_precisions({
        'USD': 2,
        'EUR': 2,
        # fill this with more..
    })
```
 
What would be the use case of changing the `PrecisionProvider`? There are (imo) two ways of setting the precisions:
* Hard coded list of known currency precisions
* Merchant can override it

Like said before, `DefaultPrecisionProvider` should handle both cases. I like the idea of the `Provider` but can you guys think any real world example of why you would like to use other `Provider` instead of the `DefaultPrecisionProvider`? I can't think any now.

shuup/core/models/_currencies.py
Pikkupomo: @tulimaki means this `name=models.CharField(verbose_name="name", max_length=64)`

chessbr: OK, I can crop it. The name will be fetched (when necessary) using babel currencies map.

shuup/utils/money.py
chessbr: @Pikkupomo The `DefaultPrecisionProvider` can be handle both cases, I can move it to core and Money will have, by default, a Dummy provider to return just the precision of 2 digits.. The core App.ready will set the right precision provider.
I used the provider because one can develop some more fast way to cache the values using whatever source he wants, django cache for example. As I said on my considerations, when the merchant change the currency configuration, the provider I developed will not update the cached store on that global variable.

Actually I am almost submitting another version, which will not use class-based methods. This way the Money will instantiate the class if is not instantiated yet and use it.

shuup/admin/modules/shops/forms.py
```diff
@@ -26,9 +26,8 @@ class Meta:
     def __init__(self, **kwargs):
         super(ShopBaseForm, self).__init__(**kwargs)
         self.fields["logo"].widget = MediaChoiceWidget(clearable=True)
-        locale = get_current_babel_locale()
         self.fields["currency"] = forms.ChoiceField(
-            choices=sorted(locale.currencies.items()),
+            choices=sorted([(currency.code, currency) for currency in Currency.objects.all()]),
```
Pikkupomo: maybe use `choices=[(currency.code, currency) for currency in Currency.objects.all().order_by("translations__name")]` or `"code"` but no need to run `sorted` here imo

chessbr:
Guys, you are awesome. You are a good code guardian. You protect this like it was your son - and it is supposed to be. I remember to do the same with my old friends :) Congrats. This is the reason why Shuup is a charm.

shuup/utils/money.py
```diff
@@ -4,9 +4,52 @@
...
+        # just return the precision or the fallback
+        return self.currency_precisions.get(currency, self.default_precision)
+
+    def get_precision_for_digits(self, digits):
+        """
+        Returns the precision for a given number of digits.
+        :param int digits
+        """
+        # just return the precision or the fallback
+        return self.digits_precisions.get(digits, self.default_precision)
```
suutari-ai: Another 40 lines here...

shuup/core/utils/precision_providers.py
```diff
@@ -0,0 +1,65 @@
...
+        """
+        # make sure it is uppercase
+        currency = currency.upper()
+
+        if currency not in self.currency_precisions:
+            cur = Currency.objects.filter(code=currency).first()
+            precision = make_precision(cur.decimal_places) if cur else self.default_precision
+            self.currency_precisions[currency] = precision
+
+        return super(DefaultPrecisionProvider, self).get_precision_for_currency(currency)
```
suutari-ai: Unneeded 50 lines here...

shuup/utils/money.py
```diff
@@ -60,3 +109,44 @@ def unit_matches_with(self, other):
...
+        :param digits int:
+          The number of digits to round
+        :type rounding: str
+
+        :return A new value rounded to the given digits or currency
+        """
+        cls = type(self)
+
+        if not cls.precision_provider:
+            cls.precision_provider = cls.precision_provider_class()
```
suutari-ai: Checking precision provider (and lazy initializing it) on a call to `as_rounded` doesn't feel like a good idea.

suutari-ai:
And btw. @chessbr, you're amazing! :+1: 

Sorry for my stubborn demands on doing this in a specific way. You're providing a much needed feature and doing very good job with it! Don't let my comments trouble you too much. :smile:

chessbr:
relax @suutari-ai  :smile: I like this, it part of the show. imo, this is what makes a good software, discussing all these points of view to make it better. I will make the changes as soon as I get home.

chessbr:
@suutari-ai some considerations about why I did that way:

- from Django documentation, we must avoid the usage of models inside the `AppConfig.ready`. Actually I tested it trying to update the Money class variable from objects from the DB. The first exception raises when running `makemigrations` with `no such table shuup_currency`. Then a lazy way to update the attributes was required. I put the provider to work with the "lazy" (almost) way to update the "cache". Maybe the use of a singleton pattern on the provider instantiation was the best than to explicitly instantiate the provider inside the `as_rounded()`.
- another issue is update the cache when the merchant wants to override the currencies decimal places. yes, the signal will work, but only for the current server instance which it will use to save the model. On a cluster environment, when more nodes, that change will not be replicated through all the servers, it means that we will need to restart all nodes to get the updated values - the best is to use a cache server where the values will be shared - or use the db cache, but it is not (imo) a cache at all.

Let me give my opinion as a I were a user (merchant). I want the system to use all the default values it has, does not matter the currency I will use, maybe I don' even want to create my currency as it must be already created.
If, for some good reason I need to override the currency value, I just want to "enable" an option saying: "I want to override the default currencies" and then configure what I want or event better, just override those I have configured.

Yes, it's preety much what the issue is asking to do - if I understood it well. What I did not understand is why this issues is asking to use `as_rounded()` only when rounding values - it wasn' suppose to reflect a change in a system-wide? Like the products prices, basket totals, etc. This will lead to a huge change, and maybe this is why the issue is only asking to change only when rounding values.

I really want to make this job done, perfectly.

chessbr:
Ok, let' see @suutari-ai, this is a more lightweight approach, but requires to import `babel` inside `core.utils`, which is actually not that bad. Also, I will not preload that values since the values will be populated on-demand.

`utils.money.py`
```python
def make_precision(digits):
    """ Return a precision for a given number of digits """
    return decimal.Decimal('0.1') ** digits


# the most used in the world
DEFAULT_PRECISION = make_precision(2)
# map of precisions for currencies
CURRENCY_PRECISIONS = {}
# map of precisions for digits
DIGITS_PRECISIONS = {}


def get_babel_digits(currency):
    fractions = get_global('currency_fractions')
    try:
        return fractions[currency][0]
    except KeyError:
        return fractions['DEFAULT'][0]


def get_precision(digits=0, currency=None):
    if digits is None:
        if currency not in CURRENCY_PRECISIONS:
            CURRENCY_PRECISIONS[currency] = get_babel_digits(currency)

        return CURRENCY_PRECISIONS[currency]
    else:
        if digits not in DIGITS_PRECISIONS:
            DIGITS_PRECISIONS[digits] = make_precision(digits)

        return DIGITS_PRECISIONS[digits]

class Money(numbers.UnittedDecimal):
    # ...
    precision_provider_func = get_precision

    @classmethod
    def set_precision_provider_function(cls, precision_provider_func):
        cls.precision_provider_func = precision_provider_func

    def as_rounded(self, digits=None, rounding=ROUND_HALF_EVEN):
        precision = type(self).precision_provider_func(digits, self.currency)
        return self.new(self.value.quantize(precision, rounding=rounding))

```

`core.models.currencies.py`
```python
def get_precision(digits=0, currency=None):
    if digits is None:
        if currency not in CURRENCY_PRECISIONS:
            try:
                CURRENCY_PRECISIONS[currency] = make_precision(Currency.objects.get(code=currency).decimal_places)
            except Currency.DoesNotExists:
                CURRENCY_PRECISIONS[currency] = Money.get_precision(currency=currency)

        return CURRENCY_PRECISIONS[currency]
    else:
        Money.get_precision(digits=digits)
```

`core.__init__.py`
```python
def ready(self):
    # ...
    from shuup.core.models import Currency
    from shuup.utils.money import Money
    Money.set_precision_provider_function(Currency.get_precision)
```

Of course, it still not solve the problem of outdated currency decimal places when changed, but I can put that signal to do the job there, for now.

chessbr:
Can someone request a rebuild on Travis, please?

suutari-ai:
@chessbr: Good stuff! And you're right about the caching. If there is many Shuup app instances running connected to the same database, then they'll need to have a shared cache which gets updated if currency precisions are changed.

I simplified your latest version somewhat. It's here: https://github.com/suutari-ai/shuup/commit/d7601c0d764b0a4047708e87361bb8a7790b7634

What do you think?

chessbr:
Yeap, much better. Good one 👍

chessbr:
@suutari-ai what do you usually do in this cases:
1. I merge with your branch and simply push to this branch
2. I merge with your branch and make interactive rebase to squash the changes

I prefer the first option though. Unless we have other option?

suutari-ai:
Thanks a lot @chessbr!

I'll leave the final call to merge this in for @Pikkupomo or @tulimaki.

tulimaki:
LGTM Thank you @chessbr @suutari-ai!

Pikkupomo:
LGTM too, thanks your hard work @chessbr and @suutari-ai. I'm gonna merge this.

PR #937, opened 2016-11-17
Shipment creation view: Improve extendability

JoaRiski:
Extract the shipment creation code to it's own method
This is done so it's possible to override/extend without overriding the entire form_valid method

No refs

### DIFF
--- a/shuup/admin/modules/orders/views/shipment.py
+++ b/shuup/admin/modules/orders/views/shipment.py
@@ -97,6 +97,25 @@ class OrderCreateShipmentView(ModifiableViewMixin, UpdateView):
         add_form_errors_as_messages(self.request, form)
         return super(OrderCreateShipmentView, self).form_invalid(form)
 
+    def create_shipment(self, order, product_quantities, shipment):
+        """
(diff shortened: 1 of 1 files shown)


shuup/admin/modules/orders/views/shipment.py
```diff
@@ -97,6 +97,15 @@ def form_invalid(self, form):
         add_form_errors_as_messages(self.request, form)
         return super(OrderCreateShipmentView, self).form_invalid(form)
 
+    def create_shipment(self, order, product_quantities, shipment):
+        """
+        This function exists so subclasses can implement custom logic without overriding the entire form_valid method
```
Pikkupomo: this should be wrapped to 79 as it's a comment string

tulimaki: You could actually write open the parameters. For example here it seems to be little open what the `product_quantities`.

JoaRiski:
@Pikkupomo Made changes now, but I hope the param comment are allowed over 79 chars (seen this elsewhere in the code)

PR #939, opened 2016-11-18
Simple Supplier: Use shop price properties (SH-477)

gurch101:
Use shop price properties when in single shop mode

Refs SH-477


tulimaki:
👍

PR #940, opened 2016-11-18
Xtheme wizard and other improvements (SH-479)

Pikkupomo:
This bag of commits adds the lightbox to `base.jinja` in admin among few other fixes.

It also adds style APIs to `Theme` as `Theme` had no API to access styles. 

* `Theme`: Add `has_stylesheets`
  * An easy way to check if `Theme` has any stylesheets available.

* `Theme`: Add `has_images`
  * With this method you can check if there are any images set to the `Theme`.

* `Theme`: Add `get_default_style`
  * Returns default style definition. Default style definition can be set with the new variable `default_style_identifier` introduced also in this commit. If `default_style_identifier` is not set, `Theme` will return the first style definition.

The main feature of this PR is to add screenshots into Xtheme wizard phase and also simplify it (in 	ef6152e)

Refs SH-489


shuup/admin/templates/shuup/admin/base.jinja
```diff
@@ -11,6 +11,7 @@
         <link href="{{ static("shuup_admin/img/favicon.ico") }}" rel="icon" type="image/x-icon">
         <link href="{{ static("shuup_admin/css/select2.css") }}" rel="stylesheet" type="text/css">
         <link href="{{ static("shuup_admin/css/base.css") }}" rel="stylesheet" type="text/css">
+        <link href="{{ static("shuup_admin/css/imagelightbox.css") }}" rel="stylesheet" type="text/css">
```
gurch101: if the plan is to add it to every page, we should add a bower install and let it go into vendor.css/vendor.js. Looks like a compatible fork is available here: https://github.com/rejas/imagelightbox

shuup/admin/templates/shuup/admin/base.jinja
```diff
@@ -51,7 +52,13 @@
         {% block post_content %}{% endblock %}
         <script src="{{ static("shuup_admin/js/vendor.js") }}"></script>
         <script src="{{ static("shuup_admin/js/base.js") }}"></script>
+        <script src="{{ static("shuup_admin/js/imagelightbox.js") }}"></script>
         {% block extra_js %}{% endblock %}
+        <script>
```
gurch101: I know this script is tiny but we should try to keep base files clean. This should go into its own js file - see base/js/tooltip.js or base/js/popover.js for similar tiny scripts :)

shuup/xtheme/_theme.py
```diff
@@ -268,6 +273,61 @@ def get_global_plugin_choices(self, empty_label=None):
         choices.sort()
         return choices
 
+    def has_stylesheets(self):
```
gurch101: can this simply be `return bool(self.stylesheets)`?

gurch101:
I tested locally. Seems to work fine :)

Tomi mentioned pictures should be zoomable so that should be added.

Pikkupomo:
I think he means "by zoomable" that they can be opened in bigger view and are inspectable. But I will check this from him

PR #941, opened 2016-11-21
Taxes: Fix default tax bugs (SH-341)

tulimaki:
If postal code was a wildcard it was never matched to anything (except a '*'). When postal code was changed to empty it was never changed in database level and caused non matching rules.


tulimaki:
This commit is picked from https://github.com/shuup/shuup/pull/886 and is already reviewed.

PR #959, opened 2017-01-04
Notify: Create notify scripts through templates (SH-501)

chessbr:
- Created provide category which contains list of script templates
- Create notification scripts for front events (OrderReceived, PaymentCreated, RefundCreated, ShipmentCreated, ShipmentDeleted and RegistrationReceived)
- Created stock alert limit notification through template
- Create tests for scripts creation through templates

Refs SH-501

notification1
notification2


tulimaki:
add to the whitelist

tulimaki:
ok to test

tulimaki:
Here we should have also quick add for shipment created. Also it would be nice if the merchant could create translated messages straight away (only parler default should be required). Also editing should contain some instructions/usability (its still the same old). One needed thing would be to avoid the multiple changes. Maybe the condition could be left out since it's super confusing right now. If you could make a default condition that is always true. The script has still the general is active checkbox.

tulimaki:
LGTM!

PR #961, opened 2017-01-05
Variation products ordering (SH-505)

strycore:
This is the proposed implementation for product variation ordering. It currently lacks some tests, which I'll add soon.

Please note that commit b9a10a1 is unrelated to the ticket, it might be worth cherry-picking if the PR isn't merged.


shuup/admin/modules/products/static_src/product-variation-variable-editor.js
```diff
@@ -57,10 +57,15 @@ window.VariationVariableEditor = (function(m, _) {
         ];
     }
 
+    function sortingHandle() {
+      return m("i.fa.fa-bars.sorting-handle.pull-left", "");
```
tulimaki: See the indentation in the file which is 4 spaces instead of 2. This is based on our project wide decision.

tulimaki:
Also as you can see from our previous commits on this project please add the refs to the commit messages. Below is from our guide how to write a proper commmit message for Shuup.

Commit messages, like code and comments, must be in English.

The messages must be formatted according to general Git format, i.e.

> Title
> (blank line)
> Content
> (blank line)
> Refs-and-such (Here this would be Refs SH-505)

### Title

* Descriptive, imperative English briefly describing the change made.
* 50ish characters is considered brief enough, but it's not a hard limit. However, keep it brief.

### Content

* Descriptive English wrapped at 72 characters.
* Markdown markup may be used to a sane degree; image embeds aren't what we consider sane here,
  but star-emphasis, bulleted lists, quotes, etc. are.
* The content should describe:
  * What is changed
  * Why it is changed
  * How it is changed
* It's totally alright and even recommended that other commits related to this commit are
  referenced here; for instance "augments the ID change made in f3c03c0e" would be cool.
* Also, for ad-hoc changes from customers (not that that should ever happen), it's also a great
  idea to record customer text as > quotes.

### Refs-and-such

Ticket/issue/HTTP/other possible references.

* For JIRA,
  > Refs DERP-1337
* For HTTP,
  > Refs http://basecamp.com/derp

### Example commit message

    VIP Stores: Hide e-billing address in basket

    Round two of "Hide The Billing Address!".

    Also hide the e-billing address (OVT and operator) from sight in the
    basket if the checkbox for billing address hiding is checked.

    Augments 491292557bc78e5
    Refs PROJECT-16719

tulimaki:
Also generate the new translated strings (in case any new ones). This can be done with `python -m shuup_workbench shuup_makemessages -l en` at the app folder.

tulimaki:
And third small changelog would be nice at doc/changelog.rst.

shuup/utils/models.py
```diff
@@ -49,3 +49,13 @@ def build_or_query(over_fields, term, operator=""):
     def add_term(query_q, field):
         return (query_q | models.Q(**{("%s%s" % (field, operator)): term}))
     return reduce(add_term, over_fields, models.Q())
+
+
+class SortableMixin(models.Model):
+    """ Utility mixin to add manual ordering to models"""
+
+    position = models.SmallIntegerField(default=0, db_index=True)
```
Pikkupomo: Fields that handle ordering in Shuup are normally called `ordering`. I would like to see consistency through all models :)

tulimaki:
Also do not autogenerate the migration name. It's always better if the migration name already tells something from the actual migration.

strycore:
I have made the requested changes and rebased the whole PR in a single commit with an improved commit message.

tulimaki:
Yeah thanks! There seems to be issue on browser test, but pretty sure it's not related to your code. Will have look at it. Will have a final look after that, but previously it already looked good so I doubt that there will anymore changes in your part needed.

PR #970, opened 2017-01-09
Reports: create new reports (SH-504)

chessbr:
- Create product total sales report
- Create new costumers report
- Total sales show customer average sales
- Create customer sales
- Create taxes report
- Create shipping report
- Create refunds reports
- Create coupons usage report
- Improve Select2Multiple search by custom fields 

Refs SH-504

product_sales

new_customers

total_sales

customer_sales

taxes

shipping

refunded_sales

coupons_usage

### DIFF
--- a/shuup/default_reports/reports/customer_sales.py
+++ b/shuup/default_reports/reports/customer_sales.py
@@ -1,0 +1,55 @@
+# -*- coding: utf-8 -*-
+# This file is part of Shuup.
+#
+# Copyright (c) 2012-2017, Shoop Commerce Ltd. All rights reserved.
+#
+# This source code is licensed under the OSL-3.0 license found in the
+# LICENSE file in the root directory of this source tree.
+from django.db.models import Avg, Count, Sum
+from django.utils.translation import ugettext_lazy as _
+
+from shuup.core.models import Contact
+from shuup.default_reports.forms import CustomerSalesReportForm
+from shuup.default_reports.mixins import OrderReportMixin
+from shuup.reports.report import ShuupReportBase
+
+
+class CustomerSalesReport(OrderReportMixin, ShuupReportBase):
+    identifier = "customer_sales_report"
+    title = _("Customer Sales")
+    form_class = CustomerSalesReportForm
+
+    filename_template = "customer-sales-report-%(time)s"
+    schema = [
+        {"key": "customer", "title": _("Customer")},
+        {"key": "order_count", "title": _("Orders")},
+        {"key": "average_sales", "title": _("Average Sales")},
+        {"key": "taxless_total", "title": _("Taxless Total")},
+        {"key": "taxful_total", "title": _("Taxful Total")},
+    ]
+
+    def get_objects(self):
+        return Contact.objects.filter(
+            customer_orders__in=super(CustomerSalesReport, self).get_objects()
+        ).annotate(
+            order_count=Count("customer_orders", distinct=True),
+            average_sales=Avg("customer_orders__taxful_total_price_value"),
+            taxless_total=Sum("customer_orders__taxless_total_price_value"),
+            taxful_total=Sum("customer_orders__taxful_total_price_value")
+        ).filter(order_count__gt=0).order_by("-%s" % self.options["order_by"]).values(
+            "name", "order_count", "average_sales", "taxless_total", "taxful_total"
+        )
+
+    def get_data(self):
+        data = []
+
+        for contact in self.get_objects():
+            data.append({
+                "customer": contact["name"],
+                "order_count": contact["order_count"],
+                "average_sales": self.shop.create_price(contact["average_sales"]).as_rounded().value,
+                "taxless_total": self.shop.create_price(contact["taxless_total"]).as_rounded().value,
+                "taxful_total": self.shop.create_price(contact["taxful_total"]).as_rounded().value,
+            })
+
+        return self.get_return_data(data)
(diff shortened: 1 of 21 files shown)


shuup/default_reports/reports/customer_sales.py
```diff
@@ -0,0 +1,59 @@
...
+        {"key": "order_count", "title": _("Orders")},
+        {"key": "average_sales", "title": _("Average Sales")},
+        {"key": "taxless_total", "title": _("Taxless Total")},
+        {"key": "taxful_total", "title": _("Taxful Total")},
+    ]
+
+    def get_objects(self):
+        orders = super(CustomerSalesReport, self).get_objects()
+
+        contacts = Contact.objects.values("pk", "name").order_by().filter(
```
gurch101: blank `order_by`. Also rather than the call to `sorted` below couldn't you use `order_by` after the `annotate`?

shuup/default_reports/reports/product_total_sales.py
```diff
@@ -0,0 +1,68 @@
...
+
+    def get_objects(self):
+        order_line_qs = OrderLine.objects.products().filter(order__in=super(ProductSalesReport, self).get_objects())
+        return order_line_qs.select_related("product").prefetch_related("taxes").order_by("product__id")
+
+    def get_data(self):
+        data = []
+
+        # group products by id - que queryset must be ordered by id to make this work
+        for key, groups in itertools.groupby(self.get_objects(), lambda pl: pl.product_id):
```
gurch101: can't this chunk of logic just be something along the lines of:
```
qs = OrderLine.objects.products().filter(order__in=super(ProductSalesReport, self).get_objects())
qs.values("product").annotate(quantity=Sum("quantity"), taxful_total=Sum("taxful_total_price_value"), taxless_total=Sum("taxless_total_price_value")).order_by(self.options.get("order_by"))
```

shuup/default_reports/reports/refunds.py
```diff
@@ -0,0 +1,51 @@
...
+
+        for order in orders:
+            if order.get_total_unrefunded_amount() > zero:
+                partially_refunded_orders += 1
+            else:
+                fully_refunded_orders += 1
+
+            total_refunded += order.get_total_refunded_amount()
+
+        data = [
```
gurch101: not sure of the value of differentiating between partial and full refunds. IMO, should just query and annotate `OrderLine.refunds()`

shuup/default_reports/reports/shipping.py
```diff
@@ -0,0 +1,80 @@
...
+            "quantity"
+        ).order_by("order__shipping_method__carrier_id", "order__shipping_method_id")
+
+    def get_data(self):
+        data = []
+
+        # just return a tuple (carrier_ID, shipping_method_ID)
+        get_group_func = (lambda ol: (ol.order.shipping_method.carrier_id, ol.order.shipping_method_id))
+
+        for key, group in itertools.groupby(self.get_objects(), get_group_func):
```
gurch101: This query should be possible with a combination of `values` and `annotate`

shuup/default_reports/reports/taxes.py
```diff
@@ -0,0 +1,70 @@
...
+            filters &= Q(tax__in=tax)
+        if tax_class:
+            filters &= Q(order_line__product__tax_class__in=tax_class)
+
+        return order_line_taxes.filter(filters).values(
+            "tax", "tax__rate"
+        ).annotate(
+            total_charged=Sum("amount_value"),
+            order_count=Count("order_line__order", distinct=True)
+        ).order_by()
```
gurch101: blank `order_by`. Should be able to pass `total_charged` here, no?

shuup/default_reports/reports/total_sales.py
```diff
@@ -23,15 +23,27 @@ class TotalSales(OrderReportMixin, ShuupReportBase):
...
     ]
 
     def get_data(self):
         orders = self.get_objects()
-        total_sales_value = orders.aggregate(total=Sum("taxful_total_price_value"))["total"] or 0
+        aggregation = orders.aggregate(customers=Count("customer", distinct=True),
+                                       total=Sum("taxful_total_price_value"))
+        customer_avg_sale = orders.values("customer").annotate(
+            average=Avg("taxful_total_price_value")
+        ).order_by().aggregate(avg=Avg("average"))["avg"] or 0
```
gurch101: Are you sure we want to take the average of averages? Shouldn't we *just* be aggregating here?

shuup/default_reports/reports/product_total_sales.py
chessbr: Unfortunately, no. `OrderLine` does not have its values cached as `Order` has, it means, the attribute `OrderLine.taxful_total_price_value` does not exist.

Fron `_orders.py`:
```
def cache_prices(self):
        taxful_total = TaxfulPrice(0, self.currency)
        taxless_total = TaxlessPrice(0, self.currency)
        for line in self.lines.all():
            taxful_total += line.taxful_price
            taxless_total += line.taxless_price
        self.taxful_total_price = taxful_total
        self.taxless_total_price = taxless_total
```

shuup/default_reports/reports/shipping.py
chessbr: But the problem is that I can't just Sum `price * quantity` since there more logic behind the `order_line.taxful_price` method, like tax calculation (which comes from other foreign key).
Will let this as it is.

PR #971, opened 2017-01-09
Update to the latest version of gulp

strycore:
This might be needed for compatibility reasons with Node 7 (The project still have issues with NPM 4)


chessbr:
Maybe you should also update other apps like `shuup/xtheme` and `shuup/themes/classic_gray` which are using the same version?

strycore:
After updating both gulp and bower, it is now possible to build resources using node 7 and npm 4.

Issue #974, opened 2017-01-10
Swiss franc not passing validation


l0rn:
Hello!

I wanted to remark, that the swiss currency CHF won't pass model validation, complaining it's not an ISO-4217 Code - though it is. Only problem is that babel has no symbol for that. The symbol for Swiss francs actually is CHF. So checking for ISO-4217 is not accurate if you are only checking for symbols like here:
https://github.com/shuup/shuup/blob/master/shuup/core/models/_currencies.py#L43

PR #975, opened 2017-01-10
Fix wrong ISO-4217 validation

l0rn:
should solve #974 , though i don't know possible implications

### DIFF
--- a/shuup/core/models/_currencies.py
+++ b/shuup/core/models/_currencies.py
@@ -40,7 +40,7 @@ class Currency(models.Model):
         super(Currency, self).clean()
 
         # make sure the code is a valid ISO-4217 currency
-        if self.code not in babel.Locale("en").currency_symbols:
+        if self.code not in babel.Locale("en").currencies:
             raise ValidationError(_('Enter a valid ISO-4217 currency code'))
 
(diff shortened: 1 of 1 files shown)


chessbr:
You're right! Looks good, thanks!

chessbr:
Before merging this, I would ask you to amend the commit message to follow our standard. Something like:
```
Fix wrong ISO-4217 validation

Currencies should be checked against currencies map instead currencies symbols map.
```

l0rn:
Gonna do that.... Still i'm not sure if some functionality relies on the currency symbols. If so that of course will remain a problem if you try to use CHF for which babel won't provide symbols

chessbr:
I tested it here:
```
>>> from shuup.utils.money import Money
>>> from shuup.utils.i18n import format_money
>>> money = Money(23.32, 'CHF')
>>> print(format_money(money))
CHF 23.32
>>> money = Money(23.32, 'USD')
>>> print(format_money(money))
$ 23.32
>>> money = Money(23.32, 'DOESNTEXIST')
>>> print(format_money(money))
DOESNTEXIST 23.32
```
Babel will use the currency code to format the amount. So it should not be a problem when the currency symbol does not exist. :)

PR #985, opened 2017-01-13
API: Improve Order create API (SH-519)

strycore:
The REST API for Orders requires less mandatory fields. Orders can now
be created without a customer, without a shipping or billing address
and without a shipping or billing method.

Refs: SH-519


tulimaki:
LGTM. But let's not merge this to master. We will have API improvements branch for all the API enhancements incoming. When all that is QA tested it will be merged to master for release 1.2.0 I suppose.

PR #990, opened 2017-01-17
Basket REST API (SH-520)

strycore:
The feature still needs some refactoring but all the unit tests pass so I'm submitting this for review so I can get some input and opinions before this gets merged.


tulimaki:
**Don't merge this either!** This will be eventually merged into the `api_improvements` branch. But feel free to review though.

shuup/front/api/basket.py
```diff
@@ -0,0 +1,315 @@
...
+        basket._load()
+        return basket
+
+    @list_route(methods=['post'])
+    def new(self, request, *args, **kwargs):
+        """
+        Create a brand new basket object
+        """
+
+        # TODO: Create a custom storage to use in APIs, probably
```
chessbr: is this really working? if so, remove this comment.

shuup/front/basket/objects.py
```diff
@@ -85,12 +86,26 @@ def can_change_quantity(self):
...
-    def __init__(self, request, basket_name="basket"):
+    def __init__(self, request, key=None):
         super(BaseBasket, self).__init__(request.shop)
-        self.basket_name = basket_name
         self.request = request
-        if request:
-            self.ip_address = request.META.get("REMOTE_ADDR")
+        self.ip_address = request.META.get("REMOTE_ADDR")
+
+        # FIXME Somehow, not all requests provide a session attribute during
```
chessbr: It does not sounds good to store this at the session because IMO it does not change the fact that the user still have to put then `sessionid=xxxxxx` in the header of each request. we must avoid that, specially for mobile devices. the end platform will have to take care of the JWT token and also the `sessionid` (does not makes sense for me). basket should be a item like Product and you can get access to it through the url `/product/{ID}/`.

shuup_tests/api/test_basket_api.py
```diff
@@ -0,0 +1,207 @@
+# -*- coding: utf-8 -*-
```
chessbr: We also have to test here request using, probably the `requests` module. APIClient is stateful, it will retain cookies and thus session. Make requests without authentication to prove it is possible to create a basket and add/remove items without authentication.

shuup/front/basket/objects.py
strycore: It does also work without using a session, the API test suite is using only the URLs to access the basket.

shuup_tests/api/test_basket_api.py
strycore: Yes, that's true. Currently the only proof I have of this is that the test suite was passing before I took care of the session details :D

shuup/front/settings.py
```diff
@@ -10,7 +10,7 @@
 #: This is the easiest way to customize the order creation process
 #: without having to override a single URL or touch the ``shuup.front`` code.
 SHUUP_BASKET_ORDER_CREATOR_SPEC = (
-    "shuup.front.basket.order_creator:BasketOrderCreator")
+    "shuup.core.basket.order_creator:BasketOrderCreator")
```
tulimaki: You should move these under core too.

shuup/front/migrations/0004_move_basket_model.py
```diff
@@ -0,0 +1,21 @@
...
+
+    dependencies = [
+        ('shuup_front', '0003_basket_key_unique'),
+    ]
+
+    operations = [
+        migrations.AlterField(
+            model_name='storedbasket',
+            name='key',
+            field=models.CharField(db_index=True, default=shuup.core.models._stored_basket.generate_key, max_length=32, unique=True),
```
tulimaki: Not how sure you managed to generate this under front, but the stored basket should have migration in core.

PR #991, opened 2017-01-17
Make menu width and font size conistent (SH-523)

strycore:


tulimaki:
Don't merge this either. This will be eventually merged into the `api_improvements` branch. But feel free to review though.

Issue #986, opened 2017-01-15
Sales per Month on the dashboard is bugged on Chrome.


moenad:
Here's a screenshot:

<img width="954" alt="screen shot 2017-01-15 at 3 12 06 am" src="https://cloud.githubusercontent.com/assets/4080121/21959458/f7fbda84-dad0-11e6-9b64-cf30c34fa067.png">

I'm on Mac OS using the latest Chrome version. It works fine on Safari.

As a side note, I really liked your platform guys :) Well done.

PR #992, opened 2017-01-18
Dashboard chart fix (SH-524)

strycore:


chessbr:
If I am not wrong, the initial requirement for this chart was to show data only for the current year.

strycore:
Yes, that's how it was initially coded, after discussing this with @tulimaki we figured it made more sense to display the past 12 months instead future months if 12 months were to be shown at all time.

chessbr:
Could not figure out why this fixes this: https://github.com/shuup/shuup/issues/986
It's because of a missing semicolon?

strycore:
No, the semicolon and formatting stuff was only to get rid of warnings displayed by my editor, the actual fix is to display 12 months of data consistently.

chessbr:
Need to fix errors so we can merge this.

chessbr:
You can also squash commits since it is a single and simply functionality.

PR #994, opened 2017-01-18
Fix browser tests (SH-526)

strycore:
Not an actual PR, I just want to run tests on Travis

### DIFF
--- a/shuup_tests/browser/admin/test_picotable.py
+++ b/shuup_tests/browser/admin/test_picotable.py
@@ -75,7 +75,7 @@ def test_list_views(browser, admin_user, live_server, settings, visit_type):
     _visit_list_view(browser, live_server, visit_type)
     if list_view_settings[visit_type].get("test_pagination", False):
         _test_pagination(browser)
-    # _set_settings(browser, visit_type)
+    _set_settings(browser, visit_type)
 
 
 def _visit_list_view(browser, live_server, list_view_name):
@@ -151,7 +151,7 @@ def _set_settings(browser, setting_type):
     # select settings
     for idx, (index_key, text) in enumerate(addable_fields):
         expected_index = default_column_count + 1 + idx
-        assert browser.is_text_present(text)
+        wait_until_condition(browser, lambda x: x.is_text_present(text))
 
         browser.find_by_xpath("//ul[@id='source-sortable']/li[%d]/button" % index_key).first.click()
         wait_until_appeared_xpath(browser, "//ul[@id='target-sortable']/li[%d]/button" % expected_index)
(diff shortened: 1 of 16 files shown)


shuup_tests/browser/admin/test_picotable.py
```diff
@@ -75,7 +75,7 @@ def test_list_views(browser, admin_user, live_server, settings, visit_type):
     _visit_list_view(browser, live_server, visit_type)
     if list_view_settings[visit_type].get("test_pagination", False):
         _test_pagination(browser)
-    _set_settings(browser, visit_type)
+    # _set_settings(browser, visit_type)
```
tulimaki: Why is this method disabled?

shuup_tests/browser/notify/test_script_template.py
```diff
@@ -53,33 +54,42 @@ def test_generic_script_template(browser, admin_user, live_server, settings, scr
 
     url = reverse("shuup_admin:notify.script.list")
     browser.visit("%s%s" % (live_server, url))
-    wait_until_condition(browser, lambda x: x.is_element_present_by_css("div.btn-toolbar a.btn.btn-info"))
+    wait_until_appeared(browser, "div.btn-toolbar a.btn.btn-info")
+    # wait_until_condition(browser, lambda x: x.is_element_present_by_css("div.btn-toolbar a.btn.btn-info"))
```
tulimaki: Remove this code if not needed.

shuup_tests/browser/notify/test_script_template.py
```diff
@@ -53,33 +54,42 @@ def test_generic_script_template(browser, admin_user, live_server, settings, scr
...
             $('#lang-en .summernote-editor').summernote('editor.insertText', 'NEW CONTENT');
         });
     """)
+    time.sleep(5)
     browser.find_by_id("id_en-subject").fill("custom subject!")
     browser.find_by_css("form button.btn.btn-lg.btn-primary").first.click()
 
-    wait_until_condition(browser, lambda b: b.url.endswith(reverse("shuup_admin:notify.script.list")))
+    time.sleep(5)
+    assert browser.is_text_not_present("This field is required.")
```
tulimaki: Could this be some sort of wait until condition. There also might be better way to check if the error happened because of required fields.

shuup_tests/browser/notify/test_script_template.py
```diff
@@ -53,33 +54,42 @@ def test_generic_script_template(browser, admin_user, live_server, settings, scr
...
-    wait_until_condition(browser, lambda x: x.is_element_present_by_css(button_id))
+    wait_until_appeared(browser, button_id, timeout=20)
     browser.find_by_css(button_id).first.click()
 
     config_url = reverse("shuup_admin:notify.script-template-config", kwargs={"id": identifier})
     wait_until_condition(browser, lambda b: b.url.endswith(config_url), timeout=15)
     wait_until_condition(browser, lambda b: b.is_text_present("Configure the Script Template"))
 
     # click to create the script
+    time.sleep(5)
```
tulimaki: Instead of randomly sleeping here maybe we should have some util for executing scripts. That method could also do the same thing for all the scripts.

PR #999, opened 2017-01-24
Add 'engines' directive to package.json files

strycore:
This won't be enforced unless the user has the 'engine-strict' option enabled but it should at least output a warning.


tulimaki:
How about notify?

PR #1000, opened 2017-01-25
Return the appropriate error code in the HTTP response

adamlwgriffiths:
Update the default error handler to return the appropriate HTTP status code.
The status_code variable is passed to the jinja template for rendering the error, but the HTTP status is always left as the default, which is 200.
This fix is applied to both the admin and front error handlers.
The tests have been updated to no longer check for `status_code == 200`, and now check for the appropriate status value.

I have not been able to get the tests running locally.
It would be appreciated if someone could verify the tests pass.

### DIFF
--- a/doc/changelog.rst
+++ b/doc/changelog.rst
@@ -33,6 +33,9 @@ Addons
 Front
 ~~~~~
 
+- Fix default error handler always returning 200 OK as an HTTP status code.
+  Now returns the appropriate status code.
+
 Xtheme
 ~~~~~~
 
--- a/shuup/admin/error_handlers.py
+++ b/shuup/admin/error_handlers.py
@@ -36,4 +36,4 @@ class AdminPageErrorHandler(ErrorPageHandler):
 
     @classmethod
     def handle_error(cls, request, error_status):
-        return render(request, "shuup/admin/errors/{}.jinja".format(error_status))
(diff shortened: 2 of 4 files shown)


tulimaki:
Here the front tests fails because of this line https://github.com/shuup/shuup/blob/master/shuup_tests/core/test_error_handling.py#L74 and you can see that those handlers return normal http response with status 200. Same goes with the admin tests for error handling.

Also the `FrontPageErrorHandler.handle_error` doesn't seem to be called on those broken tests at all.

I think your change looks a logical though. So if you can fix the tests/prove that change is correct I think this could be merged.

adamlwgriffiths:
Ah, I see what you mean.

I've updated the handlers defined in the tests to return the appropriate status code which should make those tests 'more correct' and also pass.

I've also updated the default handler tests further down in each function to check the status code of the response, not just the content of the response, which should help verify the code change is working.

Again, I cannot run the tests in my environment, if you could verify if this has resolved the issue / is sufficient.

adamlwgriffiths:
Is the CLA an issue in getting this merged?
If so, I can do the pull request from a personal repo instead and sign it as an independent developer.

tulimaki:
Yeah the CLA is currently needed. So if you could either sign it or change the account for this PR.

adamlwgriffiths:
Moved to #1019

PR #1001, opened 2017-01-26
Add longitude and latitude to Address models

strycore:


chessbr:
I think we should have just one migration file and put that `blank=True`. Please squash them.

chessbr:
I think we should have just one migration file and put that `blank=True`. Please squash them.

tulimaki:
And while at it squash the commits too. After the change I think this is good to go.

strycore:
Merged and squashed everything!

PR #1002, opened 2017-01-26
Optimize markdown conversion

finmaster19:
Significantly reduces the memory usage of Markdown parser by calling md.reset() in every iteration when using the same instance of Markdown. This also provides huge speed boost when run against database which has existing rows.

### DIFF
--- a/shuup/core/migrations/0020_md_to_html.py
+++ b/shuup/core/migrations/0020_md_to_html.py
@@ -6,32 +6,30 @@ from django.db import migrations
 from markdown import Markdown
 
 
-def update_field(model, field):
+def update_field(md, model, field):
+    for id, value in model.objects.exclude(**{field: ""}).values_list("id", field):
+        md.reset()
+        try:
+            value = md.convert(value)
+        except:
+            # not valid markdown - leave it as-is
+            continue
+        model.objects.filter(id=id).update(**{field: value})
+
+
+def md_to_html(apps, schema_editor):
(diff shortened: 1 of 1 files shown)


RauliL:
The biggest optimization in this commit is not having to load every instance of model into memory because
1. Model instances with empty field values are excluded.
2. Only database ID and value of the field is pulled from the database.

shuup/core/migrations/0020_md_to_html.py
```diff
@@ -6,32 +6,30 @@
...
-def md_to_html(apps, schema_editor):
     ProductTranslation = apps.get_model("shuup", "ProductTranslation")
     CategoryTranslation = apps.get_model("shuup", "CategoryTranslation")
     PaymentMethodTranslation = apps.get_model("shuup", "PaymentMethodTranslation")
     ShippingMethodTranslation = apps.get_model("shuup", "ShippingMethodTranslation")
-    update_field(ProductTranslation, "description")
-    update_field(CategoryTranslation, "description")
-    update_field(PaymentMethodTranslation, "description")
-    update_field(ShippingMethodTranslation, "description")
+    update_field(md, ProductTranslation, "description")
```
chessbr: this instance could'nt be at the module level?

RauliL: What improvements it would provide?

tulimaki:
@finmaster19 @RauliL @suutari-ai Thank you so much!

PR #1003, opened 2017-01-27
Add payment method and shipping method information to order detail template

finmaster19:
This information will help when processing order or just checking order information.


finmaster19:
Hmm, I'm not sure if this is useful for Shuup project. Feel free to discard if you think so.

tulimaki:
Well usually the shipping method and payment method name is visible through order line. But then again I can't see how this would hurt anybody. If somebody thinks this information alongside with the payment and shipping status is useful then there must be some others too.

PR #1004, opened 2017-01-28
Increase max_length of Shop.identifier to 96 chars

strycore:


tulimaki:
Will do some extra addition to this migration so closing this for now.

PR #1006, opened 2017-01-29
Improve and simplify Front Products API

chessbr:


tulimaki:
LGTM. The browser test shall not block more since this is not going to master.

PR #1009, opened 2017-01-30
Introduce provided order information

finmaster19:
This provide category is used to add extra info rows into Order detail page. Provide objects should inherit from `shuup.admin.modules.orders.utils.OrderInformation` class. Classes inheriting `OrderInformation` are turned into info rows by using the title variable as row title and information property as row data. Includes test for checking that info rows can be added to the Order detail page.

### DIFF
--- a/doc/changelog.rst
+++ b/doc/changelog.rst
@@ -22,6 +22,8 @@ Localization
 Admin
 ~~~~~
 
+- Add new provide category called `admin_order_information` which can be used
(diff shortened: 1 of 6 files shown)


tulimaki:
LGTM. Add this also to "Provide objects should inherit from shuup.admin.modules.orders.utils.OrderInformation class." to the provides.rst. And while you are at it squash the commits and add short changelog, thanks!.

PR #1010, opened 2017-01-30
Add nearby filter for shops

chessbr:
Added nearby filter for shops
Add database indices for product, order and shop dates


shuup/core/api/category.py
```diff
@@ -38,6 +43,8 @@ class CategoryFilter(FilterSet):
     shop = django_filters.ModelChoiceFilter(name="shops",
                                             queryset=Shop.objects.all(),
                                             lookup_expr="exact")
+    slug = django_filters.CharFilter(name="translations__slug",
```
gurch101: don't think we should be filtering on slug...can't we filter on identifier instead? or maybe just grab all root categories?

shuup/front/api/products.py
```diff
@@ -125,7 +131,10 @@ def get_sales_unit(self, shop_product):
 
     def get_is_orderable(self, shop_product):
         customer = getattr(self.context["request"], "contact", AnonymousContact())
```
gurch101: Where is `request.contact` defined? We should probably use `get_person_contact(request.user)` here.

shuup/core/api/category.py
chessbr: I didn't filter by `identifier` because user can't see it on the Shuup Admin. I can add a field to filter by root category.

shuup/front/api/products.py
chessbr: good, didn't know it exists ^^

PR #1011, opened 2017-01-30
Use optional DEFAULT_SHOP_ID setting instead of first shop

strycore:
This introduce a minor change on how the default shop is queried. The default behavior stays the same but can be overridden with a DEFAULT_SHOP_ID setting.  

Some parts of the code still use Shop.objects.first() such as the wizard and sample modules which is probably the desired behavior.

### DIFF
--- a/shuup/admin/modules/products/__init__.py
+++ b/shuup/admin/modules/products/__init__.py
@@ -167,7 +167,7 @@ class ProductModule(AdminModule):
 
     def get_model_url(self, object, kind):
         if isinstance(object, Product):
-            shop = Shop.objects.first()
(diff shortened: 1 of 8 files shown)


gurch101:
Can you please also update line 42 of `admin/views/dashboard.py` to do get the default shop as well. Thanks!

strycore:
@gurch101 it seems that the shop is set from the request here:  https://github.com/shuup/shuup/blob/master/shuup/admin/views/dashboard.py#L42 and shuup/front/middleware.py sets this attribute using get_default()

tulimaki:
No need for this. The multivendor admin improvements should cover this.

PR #1014, opened 2017-01-31
Add password reset APIs

gurch101:
* Add API to send password reset email
* Add API to reset password with valid token

### DIFF
--- a/shuup/front/apps/auth/api.py
+++ b/shuup/front/apps/auth/api.py
@@ -1,0 +1,74 @@
+# This file is part of Shuup.
+#
+# Copyright (c) 2012-2017, Shoop Commerce Ltd. All rights reserved.
+#
+# This source code is licensed under the OSL-3.0 license found in the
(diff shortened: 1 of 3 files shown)


shuup/front/apps/auth/api.py
```diff
@@ -0,0 +1,70 @@
...
+        else:
+            return Response(form.errors, status=status.HTTP_400_BAD_REQUEST)
+
+
+class SetPasswordViewSet(GenericViewSet):
+    def create(self, request):
+        serializer = SetPasswordSerializer(data=request.data)
+
+        if serializer.is_valid():
+            data = serializer.validated_data
```
chessbr: You can always make everything inside the Serializer: http://www.django-rest-framework.org/api-guide/serializers/#overriding-save-directly
then call: 
```
if serializer.is_valid():
    serializer.save()
```

gurch101: :+1: Done

PR #1015, opened 2017-02-02
Send the request object to the Section provides

chessbr:
- This will break backwards compatibility but should be necessary for multi-shop feature since some objects could not know the current show without the request.

### DIFF
--- a/doc/changelog.rst
+++ b/doc/changelog.rst
@@ -12,6 +12,7 @@ Unrealeased
 Core
 ~~~~
 
+- Section receives the request object
 - API: added nearby filter for products
 - API: added nearby filter for shops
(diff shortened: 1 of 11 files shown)


tulimaki:
@strycore can you pick this up for the multivendor branch. And I think we can make the request optional and when there isn't one then assume it's not multishop situation and pick the first shop etc...

@strycore When you have picked up this commit, feel free to close this PR.

chessbr:
@tulimaki even making request optional will break those who have this implemented right? So another way is to rename this methods and then invoke the old ones without the request?

strycore:
PR integrated in the multivendor branch

PR #1017, opened 2017-02-02
setup.py: Fix/update Trove classifiers

suutari-ai:
This PR should allow uploading Shuup to PyPI with Twine.

Also updated some info in the classifiers, e.g. development status.


tulimaki:
Thank you so much @suutari-ai! Will wait for merge a little and check if there is something we can do about that license change in `setup.py`. At the latest before we start with the next release which should happen pretty soon (either 1.1.1. or 1.2.0, depending on how the new features are getting release ready).

PR #1019, opened 2017-02-05
Return the appropriate error code in the HTTP response

adamlwgriffiths:
(This is a re-pull of #1000 under a personal repo and personal CLA)

Update the default error handler to return the appropriate HTTP status code.
The status_code variable is passed to the jinja template for rendering the error, but the HTTP status is always left as the default, which is 200.
This fix is applied to both the admin and front error handlers.
The tests have been updated to no longer check for status_code == 200, and now check for the appropriate status value.

I have not been able to get the tests running locally.
It would be appreciated if someone could verify the tests pass.

### DIFF
--- a/doc/changelog.rst
+++ b/doc/changelog.rst
@@ -33,6 +33,9 @@ Addons
 Front
 ~~~~~
 
+- Fix default error handler always returning 200 OK as an HTTP status code.
+  Now returns the appropriate status code.
+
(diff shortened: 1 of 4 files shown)


adamlwgriffiths:
Just to be 100% clear, I've signed this under my own account. The CLA applies to me personally and not any business or affiliations.

PR #1020, opened 2017-02-08
Front API: Fix for the product variations

tulimaki:
* Make sure that all variation information available
at the response.
* The child product data is stil serialized with normal
product serializer

### DIFF
--- a/shuup/front/api/products.py
+++ b/shuup/front/api/products.py
@@ -11,6 +11,7 @@ from collections import defaultdict
 import six
 from django.db.models import Q
 from django.db.models.expressions import RawSQL
+from django.utils.encoding import force_text
 from django.utils.translation import ugettext_lazy as _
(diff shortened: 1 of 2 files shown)


shuup/front/api/products.py
```diff
@@ -215,13 +216,29 @@ class Meta(NormalProductSerializer.Meta):
...
+                    variation_parent_id=shop_product.product_id).values_list("pk", "sku"):
+                combinations.append({
+                    "result_product_pk": product_pk,
+                    "sku_part": sku,
+                    "hash": None,
+                    "variable_to_value": {}
+                })
+
+        for combination in combinations:
+            child = ShopProduct.objects.filter(product__pk=combination["result_product_pk"]).first()
```
chessbr: I would also put the shop ID on this queryset

tulimaki: True

chessbr:
LGTM

PR #1021, opened 2017-02-08
Adapt the admin to allow multiple vendors

strycore:
* Add middleware to ensure the session has shop set
* Store current shop in session
* Add shop changer to admin navigation
* Make all places using some weird `Shop.objects.first` things
  use the current shop in session
* Filter products, categories and orders to use the current shop


tulimaki:
Good for the dev branch.

PR #1022, opened 2017-02-09
Add password reset API

gurch101:


shuup/front/apps/auth/api.py
```diff
@@ -48,6 +48,25 @@ def save(self):
         user.save()
 
 
+class SetAuthenticatedUserPasswordSerializer(serializers.Serializer):
+    password = serializers.CharField()
+    new_password1 = serializers.CharField()
+    new_password2 = serializers.CharField()
```
chessbr: This `new_password2` is not being used here.

gurch101: Added validation on this field

chessbr:
LGTM

PR #1023, opened 2017-02-09
Front order API

gurch101:
* Add API that allows users to retrieve their own orders.
* Allow products from a pre-existing order to be added to the basket.

The way errors are handled for APIs is a mess right now. We need to eventually come up with a standard approach that applies across all APIs.

### DIFF
--- a/shuup/core/api/basket.py
+++ b/shuup/core/api/basket.py
@@ -32,8 +32,8 @@ from shuup.core.fields import (
     FORMATTED_DECIMAL_FIELD_DECIMAL_PLACES, FORMATTED_DECIMAL_FIELD_MAX_DIGITS
 )
 from shuup.core.models import (
-    get_company_contact, get_person_contact, MutableAddress, OrderLineType,
-    OrderStatus, Product, Shop, ShopProduct
+    get_company_contact, get_person_contact, MutableAddress, Order,
+    OrderLineType, OrderStatus, Product, Shop, ShopProduct
 )
 from shuup.utils.importing import cached_load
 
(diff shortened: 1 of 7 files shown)


shuup/core/api/basket.py
```diff
@@ -297,42 +296,71 @@ def _handle_cmd(self, request, command, kwargs):
...
+                "basket": self.request._request.basket,
                 "shop_id": serializer.data.get("shop") or serializer.validated_data["shop"].pk,
                 "product_id": serializer.data.get("product") or serializer.validated_data["product"].pk,
                 "quantity": serializer.validated_data.get("quantity", 1),
                 "supplier_id": serializer.validated_data.get("supplier")
             }
-            # we call `add` directly, assuming the user will handle variations
-            # as he can know all product variations easily through product API
+            self._handle_cmd(self.request._request, "add", cmd_kwargs)
+            self.request.basket.save()
```
chessbr: that also works, but this access the property directly: `self.request._request.basket`

shuup/core/api/basket.py
```diff
@@ -297,42 +296,71 @@ def _handle_cmd(self, request, command, kwargs):
...
+
+    @detail_route(methods=['post'])
+    def add(self, request, *args, **kwargs):
```
chessbr: put the `@schema_serializer_class(ProductAddBasketSerializer)` here. it is important when generating the documentation, otherwise the docs will show the wrong serializer.

shuup/core/api/basket.py
```diff
@@ -297,42 +296,71 @@ def _handle_cmd(self, request, command, kwargs):
...
+            customer=get_person_contact(request.user),
+            pk=request.data.get("order")
+        ).first()
+        if not order:
+            return Response({"error": "invalid order"}, status=status.HTTP_404_NOT_FOUND)
+        for line in order.lines.products():
             try:
-                self._handle_cmd(request, "add", cmd_kwargs)
-                request.basket.save()
+                self.add_product({"product": line.product_id, "shop": order.shop_id, "quantity": line.quantity})
```
chessbr: what happens if user has already something in the basket? should we clean it first?

shuup/front/api/orders.py
```diff
@@ -0,0 +1,104 @@
...
+from shuup.core.models import get_person_contact, Order, OrderLine, Shop
+
+
+class ShopSerializer(serializers.ModelSerializer):
+    logo = serializers.SerializerMethodField()
+    contact_address = AddressSerializer()
+
+    class Meta:
+        model = Shop
+        fields = ("id", "name", "logo", "options", "contact_address")
```
chessbr: `options` here should be serialized as JSON, like: https://github.com/shuup/shuup/blob/api-improvements/shuup/core/api/shop.py#L31:

```python
options = serializers.JSONField(binary=False, required=False)
```

shuup/front/api/orders.py
```diff
@@ -0,0 +1,104 @@
...
+
+class OrderLineSerializer(serializers.ModelSerializer):
+    image = serializers.SerializerMethodField()
+    taxful_price = serializers.SerializerMethodField()
+
+    class Meta:
+        model = OrderLine
+        fields = "__all__"
+
+    def get_image(self, line):
```
chessbr: just a comment: we need to put this on a Field as other common tasks we do.. we are repeating ourselves 😄 
I will put this on my TODO list to be done on the API refactor.

chessbr:
Need to fix Travis errors

shuup/core/api/basket.py
gurch101: Done. Cleared first + added test for this case.

chessbr:
Awesome!

PR #1024, opened 2017-02-09
API: add set address method on basket endpoint

chessbr:
- Add method to set shipping address on basket
- Saving basket when adding coupon codes
- Add decorator with the serializer on `update_quantity` method to generate docs correctly


tulimaki:
LGTM

PR #1025, opened 2017-02-09
Check that product children are available for the current shop

strycore:
This prevents an error when checking orderability when some child products are not available in a shop


tulimaki:
Why 1 as a quantity. I know it used to be like it, but shouldn't that be `ShopProduct.minimum_purchase_quantity`?

PR #1027, opened 2017-02-10
API: Allow current contact update/fetch

gurch101:


chessbr:
Check if the current user is anonymous - on this case return 404.

shuup/front/apps/registration/api.py
```diff
@@ -11,13 +11,20 @@
 from django.contrib.auth import get_user_model
```
chessbr: Don't like this ViewSet here: `shuup/front/apps/registration/api.py`
Now if do more then registration

shuup/front/apps/registration/api.py
```diff
@@ -69,6 +120,15 @@ def get_view_name(self):
     def get_help_text(cls):
         return _("Users can register to the storefront.")
 
+    def get_object(self):
+        return get_person_contact(self.request.user)
```
chessbr: with anonymous user this should return 404

PR #1030, opened 2017-02-10
Update test for PermissionGroupForm

strycore:


shuup/admin/modules/permission_groups/views/edit.py
```diff
@@ -94,13 +97,19 @@ def clean_members(self):
     def save(self):
         obj = super(PermissionGroupForm, self).save()
         permissions = set()
+        print(self.cleaned_data)
```
tulimaki: print

PR #1035, opened 2017-02-16
Don't cache service prices in checkout templates

finmaster19:
Caching service prices may cause faulty prices when using
Waiving cost behaviour component. If user deletes some products
after passing the Shipping & Payment phase, next time Shipping & Payment
phase will show the cached price which could be the waived price. Though
prices are correct when moving to confirmation view but this has caused
confusion among users.


tulimaki:
@finmaster19 Thank You!

PR #1036, opened 2017-02-16
Create endpoint to return simple products

chessbr:
The original products endpoint was renamed to shop_products


tulimaki:
👍  for tests

PR #1038, opened 2017-02-19
Enable more functionality to addons to rely on

Pikkupomo:
With this PR Shuup will be more extendable than ever before. 

The changes here were required to enable addons to actually add more features into the Shuup front and Shuup admin.

It is now possible for addon to: 
* offer menu items better than before
* offer front product order form
* Extend to payments
* Extend to orderlines
* etc.

Refs SH-543


shuup/admin/menu.py
```diff
@@ -132,6 +144,9 @@
     }
 ]
 
+for menu_updater in get_provide_objects("admin_main_menu_updater"):
```
tulimaki: Can you do this in the `get_menu_entry_categories`?

doc/ref/provides.rst
```diff
@@ -86,6 +86,9 @@ Core
     Additional information rows for Order detail page. Provide objects should inherit
     from `~shuup.admin.modules.orders.utils.OrderInformation` class.
 
+``admin_main_menu_updater``
+    Allows updating the Admin Main Menu with new elements.
```
tulimaki: Here add more information about the required menu updater class.

shuup/front/basket/__init__.py
```diff
@@ -27,15 +27,15 @@ def get_basket_command_dispatcher(request):
...
 
 
-def get_basket(request):
+def get_basket(request, basket_name="basket", basket_class=None):
     """
     :type request: django.http.request.HttpRequest
     :rtype: shuup.front.basket.objects.BaseBasket
     """
-    if not hasattr(request, "basket"):
+    if basket_name == "basket" and hasattr(request, "basket"):
```
tulimaki: So if the basket has name and is passed as a parameter new basket is created every time?

shuup/core/models/_products.py
```diff
@@ -132,6 +133,17 @@ def __str__(self):
 
 
 class ProductQuerySet(TranslatableQuerySet):
+
+    def _get_visible_modes(self):
+        # TODO: What's the point of this?
```
tulimaki: ?

shuup/core/models/_products.py
```diff
@@ -568,7 +571,13 @@ def soft_delete(self, user=None):
             # Bypassing local `save()` on purpose.
             super(Product, self).save(update_fields=("deleted",))
 
+    # TODO (2.0): deprecate this
     def verify_mode(self):
+        warnings.warn(
```
tulimaki: Can you add more insights why this needs to be deprecated?

tulimaki:
In general. Would appreciate better commit messages, like why the changes is done. For example "Move classes to form utils" this is pretty weird since it looks like only new code is added.

`Front: Make order form extendable` and `Enable named baskets and paraller checkout processes` probably could contain some simple tests to prove those work like planned.

shuup/front/basket/__init__.py
suutari-ai: IMHO this shouldn't do any caching for the basket (to request) anyway. It's the front middleware which is supposed to set `request.basket`. But for backward compatibility it's probably best to not change the existing behavior for the default basket.

shuup/core/models/_products.py
Pikkupomo: I noticed, that when moving this to separate function the visible modes actually are "every mode ever". That's why I added the comment.

shuup/core/models/_products.py
Pikkupomo: First, it's wrongly named it should be `set_mode` or similar.

Real world usage
As I was making my addon, I realized that my addon cannot set the `ProductMode` for `Product` as it was always overridden with this magical method. So, this method is not extendable. I created a hacked solution for this in my addon, but it's far from ideal solution imo.

Proposals for 2.0 (at least):
* This method should be renamed
* This method should set the `ProductMode` based on business logic (I actually had a solution for this, but didn't want to add too much in this PR). The solution is to have a set of `ProductMode` setters which can be provided through `provides`. Pretty simple, but effective.
* `ProductMode` should not be `Enum` anymore, but something that could be extended from Addons otherwise we would have to figure out all the possible product modes in the world :)

shuup/front/basket/__init__.py
suutari-ai: My point being: The change that @Pikkupomo proposes here is OK in my opinion.

shuup/core/models/_products.py
suutari-ai: How about making Product polymorphic and replace the whole concept of ProductMode with product class, i.e. the subclass determines what kind of product it is?  Though maybe that's more like Shuup 3.0 stuff then... :)

shuup/core/models/_payments.py
```diff
@@ -33,8 +31,20 @@ class Payment(MoneyPropped, models.Model):
     description = models.CharField(max_length=256, blank=True, verbose_name=_('description'))
 
     class Meta:
+        abstract = True
+
+
+# TODO (2.0): Rename this to OrderPayment
+class Payment(AbsctractPayment):
```
suutari-ai: typo: Absctract -> Abstract

shuup/core/models/_shipments.py
```diff
@@ -156,21 +155,26 @@ def set_received(self, purchase_prices=None, created_by=None):
...
+
+# TODO (2.0): Rename this to ShipmentProduct
 @python_2_unicode_compatible
-class ShipmentProduct(models.Model):
-    shipment = models.ForeignKey(
-        Shipment, related_name='products', on_delete=models.PROTECT, verbose_name=_("shipment")
-    )
-    product = models.ForeignKey(
-        "Product", related_name='shipments', on_delete=models.CASCADE, verbose_name=_("product")
-    )
```
suutari-ai: I don't understand why product needs to be taken away from the abstract base class. (Shipment is separated clearly, because if you subclass AbstractShipment, then you probably want to have subclass of AbstractShipmentProduct for its products, but it would still have products. Right?)

shuup/core/models/_shipments.py
```diff
@@ -58,11 +59,8 @@ def all_except_deleted(self):
         return self.exclude(status=ShipmentStatus.DELETED)
 
 
-class Shipment(models.Model):
-    order = models.ForeignKey(
-        "Order", blank=True, null=True, related_name='shipments', on_delete=models.PROTECT, verbose_name=_("order"))
-    supplier = models.ForeignKey(
-        "Supplier", related_name='shipments', on_delete=models.PROTECT, verbose_name=_("supplier"))
+# TODO (2.0): Rename this to Shipment
+class AbstractShipment(ShuupModel):
```
suutari-ai: Hmm... Is this AbstractShipment really needed?

I know, I know... I was there when we planned this, but I didn't notice then that the `order` field is actually *nullable*. So maybe the addon, which wants to create shipments which are not linked to order, can do so by setting `order=None` and linking the shipment to its own models with external one-to-one field (or FK/M2M).

Isn't it the supplier which updates the stocks? So probably every shipment should have a supplier. Or what do you think?

I'm worried that if an addon creates some FoobarShipments that are derived from AbstractShipment, then nothing will process them and product stock counts won't ever be updated.

shuup/admin/modules/service_providers/views/_edit.py
```diff
@@ -104,27 +102,11 @@ def get_toolbar(self):
...
-            fi for fi in self.values() if isinstance(obj, fi.model))
-
-    def get_by_choice_value(self, choice_value):
-        return self.get(choice_value)
-
-    def get_type_choices(self):
-        return [(x.choice_value, x.choice_text) for x in self.values()]
+# TODO (2.0): Remove
+class _FormInfoMap(FormInfoMap):
+    pass
```
suutari-ai: Don't add these underscored classes here. No need to have them here, since they're private: Nobody shouldn't be using them anyway.

shuup/admin/modules/service_providers/views/_edit.py
```diff
@@ -104,27 +102,11 @@ def get_toolbar(self):
...
-class _FormInfo(object):
-    def __init__(self, form_class):
-        self.form_class = form_class
-        self.model = form_class._meta.model
-        modelmeta = self.model._meta
-        self.choice_value = modelmeta.app_label + '.' + modelmeta.model_name
-        self.choice_text = modelmeta.verbose_name.capitalize()
+# TODO (2.0): Remove
+class _FormInfo(FormInfo):
+    pass
```
suutari-ai: Don't add these underscored classes here. No need to have them here, since they're private: Nobody shouldn't be using them anyway.

shuup/core/models/_products.py
```diff
@@ -568,7 +571,13 @@ def soft_delete(self, user=None):
...
+            "Product:verify_mode will be deprecated. Set the product mode explicitly.",
+            RemovedInFutureShuupWarning
+        )
```
suutari-ai: If you're deprecating something, it shouldn't be called in this code base anymore.

I.e. rename it to e.g. `_set_mode` and then make the deprecated `verify_mode` call `_set_mode` and other places in this class should then call `_set_mode`. If `verify_mode` is called from outside (IIRC admin calls it at least), then those places should be updated to not call it, but set the product mode explicitly (as you state in you deprecation warning).

To make sure you found all places, try raising an exception after the deprecation warning and run the tests.

shuup/core/models/_products.py
```diff
@@ -694,15 +703,15 @@ def make_package(self, package_def):
...
+            if child_product.is_container():
                 raise ImpossibleProductModeException(_("Packages can't be nested"), code="multilevel")
             if quantity <= 0:
                 raise ImpossibleProductModeException(_("Quantity %s is invalid") % quantity, code="quantity")
             ProductPackageLink.objects.create(parent=self, child=child_product, quantity=quantity)
         self.verify_mode()
 
     def get_package_child_to_quantity_map(self):
-        if self.is_package_parent():
+        if self.is_container():
```
suutari-ai: Your commits are not properly ordered, since I remember seeing usages of `is_container` before it's being added.

shuup/core/models/_products.py
```diff
@@ -132,6 +133,17 @@ def __str__(self):
...
+
+    def _get_visible_modes(self):
+        # TODO: What's the point of this?
+        return [
+                ProductMode.NORMAL,
+                ProductMode.PACKAGE_PARENT,
+                ProductMode.SUBSCRIPTION,
+                ProductMode.SIMPLE_VARIATION_PARENT,
+                ProductMode.VARIABLE_VARIATION_PARENT
+        ]
```
suutari-ai: So this is all but `VARIATION_CHILD`. Maybe listing the opposite would be easier to understand?

shuup/core/models/_products.py
suutari-ai: Remove the todo comment. Point of this is to exclude variation children from being listed in product listings. :)

shuup/core/models/_products.py
```diff
@@ -143,10 +155,7 @@ def _visible(self, shop, customer, language=None):
                         shop_products__shop=shop,
                         shop_products__visibility=ShopProductVisibility.NOT_VISIBLE
                     )).filter(
-                mode__in=(
-                    ProductMode.NORMAL, ProductMode.PACKAGE_PARENT,
-                    ProductMode.SIMPLE_VARIATION_PARENT, ProductMode.VARIABLE_VARIATION_PARENT
-                )
+                mode__in=self._get_visible_modes()
```
suutari-ai: This commit should be squashed with the one which add SUBSCRIPTION product mode, since this adds the SUBSCRIPTION to this list.

shuup/front/templates/shuup/front/product/detail.jinja
```diff
@@ -38,9 +38,9 @@
                 {{ product_macros.render_product_basic_information(product) }}
             </div>
             {% placeholder "product_extra_1" %}{% endplaceholder %}
-            {{ product_macros.render_product_order_section(product, variation_variables, variation_children, unorderability_reason, orderable_variation_children) }}
+            {{ order_form }}
             <hr>
-            {% placeholder "product_extra_2" %}{% endplaceholder %}
+            {% placeholder "product_extra_2" global %}{% endplaceholder %}
```
suutari-ai: Why is this placeholder changed to global? (I see no note about it in the commit message.)

shuup/front/templates/shuup/front/product/detail_order_section.jinja
```diff
@@ -1,3 +1,5 @@
 {%- from "shuup/front/macros/product.jinja" import render_product_order_section with context -%}
 
+{% if not product.is_subscription_parent() %}
 {{ render_product_order_section(product, variation_variables, variation_children, unorderability_reason, orderable_variation_children) }}
+{% endif %}
```
suutari-ai: I don't think this `if` is needed, since `render_product_order_section` should work fine even though there is no ProductOrderForm which handles subscription products. It should then just render nothing.

shuup/front/basket/command_dispatcher.py
```diff
@@ -73,8 +74,9 @@ def handle(self, command, kwargs=None):
         except (Problem, ValidationError) as exc:
             if not self.ajax:
                 raise
+            msg = exc.message if hasattr(exc, "message") else exc
```
suutari-ai: Exception.message is deprecated, proper way to get the message is by casting to string:
```python
msg = '{}'.format(exc)
```
(Note: should have `from __future__ import unicode_literals` too)

shuup/front/basket/command_dispatcher.py
```diff
@@ -73,8 +74,9 @@ def handle(self, command, kwargs=None):
...
             response = {
-                "error": force_text(exc, errors="ignore"),
+                "error": force_text(msg, errors="ignore"),
```
suutari-ai: Hmm.. but this was already casting the `exc` to string. What was wrong with it?

shuup_tests/front/test_methods_phase.py
```diff
@@ -39,13 +44,33 @@ def test_method_phase_basic(rf, get_method, data, method_id):
...
+    basket = get_basket(request)
+
+    # add product to bakset
+    supplier = get_default_supplier()
+    product = create_product(printable_gibberish(), shop=shop, supplier=supplier, default_price=50)
+    basket.add_product(supplier=supplier, shop=shop, product=product, quantity=2)
+    assert basket.product_count, "basket has products"
+
+    basket.save()
+    request = apply_request_middleware(request, user=admin_user, person=person, customer=person, basket=basket)
```
suutari-ai: Your commit message:
> Methods phase now requires basket

This is a backward incompatible change, though quite edge case. So, is this OK? Can some code (other than our tests) rely on methods phase working without a basket?

Btw. you're also adding products to the basket. Does the methods phase need products in the basket too?

shuup_tests/admin/test_order_mass_actions.py
```diff
@@ -147,7 +147,7 @@ def test_mass_edit_orders4(rf, admin_user):
     assert response.status_code == 200
 
     if weasyprint:
-        assert response['Content-Disposition'] == 'attachment; filename=order_%s_delivery.pdf' % order.pk
+        assert response['Content-Disposition'] == 'attachment; filename=shipment_%s_delivery.pdf' % order.pk
```
suutari-ai: The `.travis.yml` should probably have `pip install weasyprint` to verify this...

shuup/core/models/_shipments.py
Pikkupomo: product was required information, it might not always be available when doing shipments. maybe `AbstractShipmentProduct` could be called `AbstractShipmentRelation` or something that doesn't indicate a product.

shuup_tests/front/test_methods_phase.py
Pikkupomo: I think it is sort of expected that basket is available in request when rendering methods phase. The whole checkout process should not be available (and it's not) when `request.basket` is unset. The `methods phase now requires a basket` simply implies that the basket is now used in methods phase as it was not before.

shuup/front/basket/command_dispatcher.py
suutari-ai: As discussed in Slack. It's OK to use `exc.message` in this case. So this change is OK.

shuup/front/basket/command_dispatcher.py
suutari-ai: Checked that this will make a difference with ValidationErrors, if the validation error is simple one-field only error (i.e. not a list of errors or dict of errors). So I guess this is better now. No need for changes to the PR.

shuup_tests/front/test_methods_phase.py
suutari-ai: OK, this is fine then.

shuup_tests/admin/test_order_mass_actions.py
suutari-ai: Or better, weasyprint should be in test deps listed in tox.ini.

shuup/core/models/_shipments.py
```diff
@@ -58,11 +59,14 @@ def all_except_deleted(self):
         return self.exclude(status=ShipmentStatus.DELETED)
 
 
-class Shipment(models.Model):
+# TODO (2.0): Rename this to Shipment
+class Shipment(ShuupModel):
```
suutari-ai: Remove the TODO, since the name of the class is Shipment already. :)

shuup/core/models/_shipments.py
```diff
@@ -76,11 +80,12 @@ class Shipment(models.Model):
     objects = ShipmentManager()
 
     class Meta:
+        abstract = True
         verbose_name = _('shipment')
         verbose_name_plural = _('shipments')
 
     def __init__(self, *args, **kwargs):
-        super(Shipment, self).__init__(*args, **kwargs)
+        super(AbstractShipment, self).__init__(*args, **kwargs)
```
suutari-ai: shouldn't be changed

shuup/core/models/_products.py
suutari-ai: Now this seems to be OK

PR #1039, opened 2017-02-20
Make excluded categories optional in CategoryProductsBasketCondition

finmaster19:
Don't require excluded categories in CategoryProductsBasketCondition. Because this change affects only in validation I believe its safe to add this to the migration also.

Or is there some kind of reason why excluded categories is required? People thinks that this is weird to require them.


finmaster19:
I mean when creating a new Basket Campaign and adding a rule "Category products in basket".

PR #1040, opened 2017-02-20
Remove position fixed from top panels on mobile devices.

sekoskel:
Top panel takes 50% of screen space on a phone making viewing anything in shuup admin quite challenging.
Make top panel not fixed for smaller mobile devices.


tulimaki:
@sekoskel Thanks!

PR #1042, opened 2017-02-22
Try to make queries somewhat faster

chessbr:
- Removed distance location from shop_product

- Added shop filter on shop_product

### DIFF
--- a/shuup/front/api/products.py
+++ b/shuup/front/api/products.py
@@ -24,6 +24,7 @@ from shuup.core.models import (
     ShopProduct, ShopProductVisibility
 )
 from shuup.core.pricing._context import PricingContext
+from shuup.core.utils import context_cache
 from shuup.front.utils.product_statistics import get_best_selling_product_info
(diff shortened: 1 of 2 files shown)


shuup/front/api/products.py
```diff
@@ -284,13 +316,23 @@ class Meta(CompleteShopProductSerializer.Meta):
...
+
+        shops_qs = queryset.filter(
             visibility__in=(ShopProductVisibility.LISTED, ShopProductVisibility.ALWAYS_VISIBLE),
             product__deleted=False
-        ).distinct()
+        )
+
+        # filter by shop
+        if shop:
+            shops_qs = shops_qs.filter(shop__id=shop)
```
tulimaki: here instead do `.filter(shop_id=shop)` the `__` causes join if I have understood correctly.

chessbr: I will change this, but the join will be done anyway.

PR #1043, opened 2017-02-23
Introduce new provide category "product_context_extra"

finmaster19:
This provide category can be used to add extra context data to the front product views. Provide objects should inherit from `shuup.front.utils.product.ProductContextExtra` object.
Add test to make sure its working.

This should make it very easy for addons to extend product context data.


tulimaki:
LGTM. Rebase please.

finmaster19:
Rebased

PR #1046, opened 2017-02-24
Front menu extender

Pikkupomo:
It was not possible for addons to provide objects into front menus.

This PR will add a support for Addons to extend main menus in Shuup Front.


chessbr:
LGTM

PR #1049, opened 2017-02-26
Products admin fixes

strycore:
This fixes some issues with shops in the Product admin module, it also increases the page load speed of the product detail view.


tulimaki:
Please run isort.

PR #1051, opened 2017-02-27
Front: Reorder old order

Pikkupomo:
It's now possible to add old order to basket

### DIFF
--- a/shuup/front/apps/personal_order_history/templates/shuup/personal_order_history/order_detail.jinja
+++ b/shuup/front/apps/personal_order_history/templates/shuup/personal_order_history/order_detail.jinja
@@ -1,4 +1,5 @@
 {% extends "shuup/front/dashboard/dashboard.jinja" %}
+{% from "shuup/personal_order_history/macros/buttons.jinja" import render_action_buttons with context %}
 {% from "shuup/personal_order_history/macros/order_detail.jinja" import basic_info, shipping_address, billing_address, status, order_contents with context %}
 
 {% block title %}{% trans identifier = order.identifier %}Order {{ identifier }}{% endtrans %}{% endblock %}
@@ -15,7 +16,8 @@
 {% endblock %}
 
 {% block dashboard_content %}
-    <div class="row">
+    <h2 class="page-header">{% trans identifier = order.identifier %}Order {{ identifier }}{% endtrans %}{{ render_action_buttons(order) }}</h2>
+    <div class="row order-dashboard-container">
         <div class="col-sm-12">
             {{ basic_info() }}
         </div>
--- a/shuup/front/apps/personal_order_history/views.py
+++ b/shuup/front/apps/personal_order_history/views.py
@@ -5,13 +5,16 @@
 # This source code is licensed under the OSL-3.0 license found in the
 # LICENSE file in the root directory of this source tree.
 import django.views.generic
+from django.core.urlresolvers import reverse
+from django.http import HttpResponseRedirect
+from django.views.generic import View
 
-import shuup.core.models
+from shuup.core.models import Order, Supplier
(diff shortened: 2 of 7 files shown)


shuup/front/apps/personal_order_history/views.py
```diff
@@ -26,3 +29,26 @@ class OrderListView(DashboardViewMixin, OrderViewMixin, django.views.generic.Lis
...
+        for line in order.lines.products():
+            request.basket.add_product(
+                supplier=supplier,
+                shop=request.shop,
+                product=line.product,
+                quantity=line.quantity
+            )
+
+        return HttpResponseRedirect(reverse("shuup:basket"))
+    # def dispatch(self, request, *args, **kwargs):
```
tulimaki: This block of code is probably not needed.

tulimaki:
New translation strings and changelog item missing.

shuup/front/apps/personal_order_history/templates/shuup/personal_order_history/order_detail.jinja
```diff
@@ -15,7 +15,8 @@
 {% endblock %}
 
 {% block dashboard_content %}
-    <div class="row">
+    <h1 class="page-header">Order #{{ order.id }}{{ render_action_buttons(order) }}</h1>
+    <div class="row order-dashboard-container">
```
suutari-ai: The added text should be translatable and use the order.identifier. There's already an example at line 4 on how to do that:
```
{% trans identifier = order.identifier %}Order {{ identifier }}{% endtrans %}
```
Is the h1 element really needed here, since there already is a h1 title rendered by the content_title macro (line 6)?

Would it look bad if those actions were placed there without the title? (Maybe to some other place?)

shuup/front/apps/personal_order_history/views.py
```diff
@@ -26,3 +29,26 @@ class OrderListView(DashboardViewMixin, OrderViewMixin, django.views.generic.Lis
...
+    #
+    #     return super(ReorderView, self).dispatch(request, *args, **kwargs)
```
suutari-ai: Remove this commented code.

shuup/front/apps/personal_order_history/templates/shuup/personal_order_history/macros/order_detail.jinja
```diff
@@ -111,3 +111,11 @@
...
     {%- endfor -%}
 {%- endmacro -%}
+
+{% macro render_action_buttons(order) %}
+    <div class="pull-right inline">
+        <a class="btn btn-success" href="{{ url("shuup:reorder-order", pk=order.pk) }}">
+            <i class="fa fa-plus"></i> {% trans %}Add all products to cart{% endtrans %}
+        </a>
+    </div>
+{% endmacro %}
```
suutari-ai: Maybe this macro should be in its own file to allow themes overriding just it. Also if theme has overridden this file (macros/order_detail.jinja) for any Shuup version before this PR, it won't have this new macro. It's possible that the calling side (personal_order_history/order_detail.jinja) is not overridden though and then this missing macro would be a problem. Therefore it would be safer to use a separate file.

shuup/front/apps/personal_order_history/templates/shuup/personal_order_history/order_detail.jinja
Pikkupomo: h1 is required because the header won't work (buttons won't align) properly without that. The solution is also in line with other views, we can unify this later imo.

suutari-ai: OK

suutari-ai:
Thanks @Pikkupomo!

PR #1052, opened 2017-02-28
Fix product csv export mass action

finmaster19:
ids are now ShopProduct ids and not regular product ids because of [323c274](https://github.com/shuup/shuup/commit/323c274). Adds some custom logic which will be used to strip column id so that it can be used with getattr function. Some columns are ShopProduct fields and some Product fields.


finmaster19:
This feels little dirty way of getting those values with `getattr` and stripping the first part of some columns ids. So if someone has better idea, it would be nice :)

PR #1053, opened 2017-03-01
Admin: Fix boolean value presentation errors

Pikkupomo:
Since we are using `django.forms.fields.NullBooleanField` in admin
we should return either None, True or False.

While the `Unknown` option (`None`) will never end up in the database
when product attribute is being saved in admin, it is possible to
create this value through API and that causes admin to break.

This commit allows `None` value reading from the numeric_value in
boolean situtiations.

### DIFF
--- a/shuup/core/models/_attributes.py
+++ b/shuup/core/models/_attributes.py
@@ -227,7 +227,17 @@ class AppliedAttribute(TranslatableModel):
 
     def _get_value(self):
         if self.attribute.type == AttributeType.BOOLEAN:
-            return bool(int(self.numeric_value))
+            """
+            Return Boolean value or None
+
+            Since we are using ``django.forms.fields.NullBooleanField`` in admin
+            we should return either None, True or False.
+
+            While the `Unknown` option (`None`) will never end up in the database
+            when product attribute is being saved in admin, it is possible to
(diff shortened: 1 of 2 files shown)


shuup/core/models/_attributes.py
```diff
@@ -255,22 +265,31 @@ def _get_value(self):
...
-            value = int(value)
-
-        if self.attribute.type == AttributeType.BOOLEAN:
-            value = int(bool(value))
+            if self.attribute.type == AttributeType.BOOLEAN:
+                value = int(bool(value))
+            elif self.attribute.type == AttributeType.INTEGER:
+                value = int(value)
+            else:
+                value = parse_decimal_string(new_value or 0)
```
suutari-ai: I don't see the point of these changes.

shuup/core/models/_attributes.py
```diff
@@ -255,22 +265,31 @@ def _get_value(self):
...
 
         self.numeric_value = value
         self.datetime_value = None
-        self.untranslated_string_value = str(self.numeric_value)
+        self.untranslated_string_value = str(value)
```
suutari-ai: This also is unneeded change.

suutari-ai:
Thanks, Janne!

PR #1058, opened 2017-03-06
Improve variation orderability check performance

gurch101:
Before, the orderability of all variations was checked needlessly.
Now, we only ensure at least one child product is orderable.

### DIFF
--- a/shuup/core/models/_product_shops.py
+++ b/shuup/core/models/_product_shops.py
@@ -278,33 +278,36 @@ class ShopProduct(MoneyPropped, models.Model):
             )
 
         if self.product.mode == ProductMode.SIMPLE_VARIATION_PARENT:
-            sellable = 0
+            sellable = False
             for child_product in self.product.variation_children.all():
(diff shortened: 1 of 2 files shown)


shuup/core/models/_product_shops.py
```diff
@@ -278,22 +278,21 @@ def get_orderability_errors(  # noqa (C901)
             )
 
         if self.product.mode == ProductMode.SIMPLE_VARIATION_PARENT:
-            sellable = 0
+            sellable = False
             for child_product in self.product.variation_children.all():
                 if child_product.get_shop_instance(self.shop).is_orderable(
                         supplier=supplier,
                         customer=customer,
                         quantity=1,
```
tulimaki: While at it can you fix this? I think this should be `child_shop_product.minimum_purchase_quantity` instead of 1.

tulimaki: Same applies to `VARIABLE_VARIATION_PARENT`.

PR #1059, opened 2017-03-06
Improve shop product list API performance

gurch101:
Add more prefetching and serialize and fetch
variation children in bulk.

### DIFF
--- a/shuup/front/api/products.py
+++ b/shuup/front/api/products.py
@@ -9,6 +9,7 @@
 from collections import defaultdict
 
 import six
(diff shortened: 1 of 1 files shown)


tulimaki:
Fix this `ERROR: /home/travis/build/shuup/shuup/shuup/admin/modules/services/forms.py Imports are incorrectly sorted.`. I think the other test fails is not part of this commit

tulimaki:
LGTM. @strycore will take a look into those failing admin tests.

PR #1062, opened 2017-03-08
Adapt .gitignore for local development scripts

strycore:
This PR would allow me to finally have a clean git tree :)

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -38,5 +38,8 @@ pip-log.txt
 sdist/
 target/
 var/
+local/
 node_modules
 .DS_Store
+Makefile
+manage.py


.gitignore
```diff
@@ -38,5 +38,8 @@ pip-log.txt
 sdist/
 target/
 var/
+local/
 node_modules
 .DS_Store
+Makefile
+manage.py
```
suutari-ai: Just curious. What are these used for?

Btw. you can use `.git/info/exclude` for your personal ignores.

strycore: manage.py is actually a bash script with ``python -m shuup_workbench $@`` and the Makefile is a symlink to this file https://github.com/shuup/shuup-tools/blob/master/config/Makefile

I didn't know about the .git/info/exclude trick, that's very useful, thanks!

strycore: Closing this PR, the .git/info/exclude file is good enough.

PR #1063, opened 2017-03-08
API updates

gurch101:


shuup/core/api/front_users.py
```diff
@@ -0,0 +1,153 @@
...
+        if not data.get(user_model.USERNAME_FIELD) and not data.get("email"):
+            raise serializers.ValidationError("username and/or email is required")
+        if user_model.USERNAME_FIELD not in data and user_model.USERNAME_FIELD != "email":
+            email = data.get("email")
+            if email:
+                data[user_model.USERNAME_FIELD] = email
+
+        pwd = data.pop("password")
+        if user_model.objects.filter(**data).exists():
+            raise serializers.ValidationError("User already exists.")
```
tulimaki: We probably don't want to give this information to the caller.

shuup/core/api/front_products.py
```diff
@@ -0,0 +1,607 @@
...
+    def get_price(self, shop_product):
+        key, val = context_cache.get_cached_value(identifier="shop_product_price",
+                                                  item=shop_product,
+                                                  context={},
+                                                  allow_cache=True)
+        if val is not None:
+            return val
+
+        context = self._get_pricing_context(self.context["request"], shop_product.shop)
+        price = shop_product.product.get_price(context).value
```
tulimaki: Discounted price right?

tulimaki: maybe the `.value` here is not correct. We should also return it since the `base_price` should be in the product `PriceInfo`.

tulimaki: Also make sure the caller have some sort of currency information available.

tulimaki:
https://github.com/shuup/shuup/pull/1078 will replace this PR with latest rebased code. We still want to keep this as a backup since not 100% if the rebase was success.

PR #1068, opened 2017-03-20
Improve product unit handling

suutari-ai:
Currently product has only one sales unit which is used in Admin and
Front.  Make it possible to have a separate display unit for a product
to allow having product quantities being rendered in different unit than
the values stored in the database.  This is documented in the added
units.rst file, which I recommended checking before going to details of
this PR.

There is many related fixes and improvements too.  Won't go to details
here.  The commit messages should contain the reasons for each change.

### DIFF
--- a/doc/api/shuup.core.utils.rst
+++ b/doc/api/shuup.core.utils.rst
@@ -36,6 +36,14 @@ shuup.core.utils.forms module
     :undoc-members:
     :show-inheritance:
 
+shuup.core.utils.line_unit_mixin module
+---------------------------------------
+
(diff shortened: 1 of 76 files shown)


chessbr:
LGTM. Maybe @tulimaki should take a look too.
It's a nice feature. I can see other usability for this. Shops which just want to display to price/kg without changing the sales unit, like a 2kg and 5kg rice package. the shop sells several brands and packages sizes of rice and customer wants to see which is the cheapest (price/kg) or even the better cost-benefit. Of course, there will be needed some template customization and the comparison addon to enable this, but at least the core is done. :+1:

PR #1074, opened 2017-03-23
Fix decimal parsing in ProductPriceView._get_quantity

suutari-ai:
Commit c7dedea of PR #1068 didn't have proper value validation in the
added _get_quantity method.  Fix this.

Add a new decimal parser for this too, since the parse_decimal_string
silently accepts invalid values.


chessbr:
LGTM

PR #1078, opened 2017-03-24
API updates

tulimaki:


shuup/core/basket/storage.py
```diff
@@ -0,0 +1,170 @@
...
+    def save(self, basket, data):
+        """
+        :type basket: shuup.core.basket.objects.BaseBasket
+        """
+        stored_basket = self._load_stored_basket(basket)
+        stored_basket.data = data
+        stored_basket.taxless_total_price = basket.taxless_total_price_or_none
+        stored_basket.taxful_total_price = basket.taxful_total_price_or_none
+        stored_basket.product_count = basket.smart_product_count
+        stored_basket.product_count = basket.product_count
```
gurch101: you override `smart_product_count` here

PR #1079, opened 2017-03-25
API updates

tulimaki:

Replaces https://github.com/shuup/shuup/pull/1063


shuup/admin/modules/orders/json_order_creator.py
```diff
@@ -157,7 +162,7 @@ def _process_lines(self, source, state):
             try:
                 self._add_json_line_to_source(source, sline)
             except Exception as exc:  # pragma: no cover
-                self.add_error(exc)
+                self.add_error(ValidationError(str(exc), code="line_error"))
```
suutari-ai: Validation error should be (a) unicode (on py2) and (b) translated (user visible) message.

shuup/core/api/front_products.py
```diff
@@ -0,0 +1,607 @@
...
+        elif image.file:
+            return self.context["request"].build_absolute_uri(image.file.url)
+
+    def get_attributes(self, shop_product):
+        return AttributeSerializer(shop_product.product.attributes, many=True).data
+
+    def get_price(self, shop_product):
+        key, val = context_cache.get_cached_value(identifier="shop_product_price",
+                                                  item=shop_product,
+                                                  context={},
```
tulimaki: Fix context here.

shuup/core/models/_shops.py
```diff
@@ -43,7 +43,7 @@ class Shop(ChangeProtected, TranslatableShuupModel):
     change_protect_message = _("The following fields cannot be changed since there are existing orders for this shop")
```
tulimaki: Should shop status, longitude, latitude has also `db_index`?

setup.cfg
```diff
@@ -17,7 +17,7 @@ combine_as_imports=false
 indent=4
 known_first_party=shuup
 known_standard_library=token,tokenize,enum,importlib
-known_third_party=django,six
+known_third_party=django,six,rest_framework,django_filters
```
tulimaki: Is this really needed?

tulimaki:
As long as we don't want to squash this. I think we don't.

Fix following commit messages:
`Basket fixes`
`It's possible to create basket for certain customer`
`Add basket changes`

PR #1082, opened 2017-03-27
Company registration

sekoskel:
Continue Janne's work on company registration feature.
Enable companies to register in Shuup. Company is inactive by default, can be activated by admin.

### DIFF
--- a/shuup/admin/modules/settings/consts.py
+++ b/shuup/admin/modules/settings/consts.py
@@ -9,3 +9,5 @@
 ORDER_REFERENCE_NUMBER_LENGTH_FIELD = "order_reference_number_length"
 ORDER_REFERENCE_NUMBER_PREFIX_FIELD = "order_reference_number_prefix"
 ORDER_REFERENCE_NUMBER_METHOD_FIELD = "order_reference_number_method"
+
+ALLOW_COMPANY_REGISTRATION = "allow_company_registration"
--- a/shuup/core/templatetags/shuup_common.py
+++ b/shuup/core/templatetags/shuup_common.py
@@ -27,6 +27,7 @@ from django.utils.safestring import mark_safe
 from django.utils.timezone import localtime
 from django_jinja import library
 from jinja2.runtime import Undefined
+from jinja2.utils import contextfunction
 
 from shuup.utils.i18n import (
     format_money, format_percent, get_current_babel_locale
@@ -136,3 +137,29 @@ def json(value):
     if isinstance(value, Undefined):
         value = None
     return mark_safe(json_dump(value, cls=ExtendedJSONEncoder))
+
+
+@library.global_function
+@contextfunction
+def get_shop_configuration(context, name, default=None):
+    """
+    Get configuration variable value for the current shop.
+
+    :type context: jinja2.runtime.Context
+    :type name: str
+    :type default: Any
+    """
+    from shuup import configuration
+    return configuration.get(context.request.shop, name, default)
+
+
+@library.global_function
+def get_global_configuration(name, default=None):
+    """
+    Get global configuration variable value.
+
+    :type name: str
+    :type default: Any
+    """
+    from shuup import configuration
+    return configuration.get(None, name, default)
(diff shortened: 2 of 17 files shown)


shuup/admin/modules/settings/consts.py
```diff
@@ -9,3 +9,5 @@
 ORDER_REFERENCE_NUMBER_LENGTH_FIELD = "order_reference_number_length"
 ORDER_REFERENCE_NUMBER_PREFIX_FIELD = "order_reference_number_prefix"
 ORDER_REFERENCE_NUMBER_METHOD_FIELD = "order_reference_number_method"
+
+SETTING_ALLOW_COMPANY_REGISTRATION = "allow_company_registration"
```
suutari-ai: IMHO this shouldn't have the `SETTING_` prefix.

shuup/xtheme/_theme.py
```diff
@@ -342,6 +342,11 @@ def render_menu_extensions(self, request, location=MenuExtenderLocation.MAIN_MEN
                 items.append(extender.get_rendered_menu_items(request, self))
         return mark_safe("".join(items))
 
+    def is_configuration_enabled(self, request, key, global_setting=True):
+        from shuup import configuration
+        shop = None if global_setting else request.shop
+        return configuration.get(shop, key)
```
suutari-ai: * Why name it is_configuration_enabled, since it just returns the value?

* The `is_configuration_enabled` helper function doesn't feel like it belongs to `xtheme`, since it's not a Xtheme configuration. Better place would be a global template function like `get_shop_configuration` (This can be achieved by adding it to `core/templatetags/shuup_common.py`) or a front template helper `shuup.general.get_shop_configuration`. I think former would be better.

* The `global_setting` argument is not good API. It would be better to add a separate function.

* Also, it shouldn't be needed to pass in the request, since it is available for the template helpers through context. Here's an example:
```python
@django_jinja.library.global_function
@jinja2.contextfunction
def get_shop_configuration(context, name, default=None):
    """
    Get configuration variable value for the current shop.

    :type context: jinja2.runtime.Context
    :type name: str
    :type default: Any
    """
    return configuration.get(context.request.shop, name, default)

@django_jinja.library.global_function
def get_global_configuration(name, default=None):
    """
    Get global configuration variable value.

    :type name: str
    :type default: Any
    """
    return configuration.get(None, name, default)
```

shuup/front/apps/registration/templates/shuup/registration/register.jinja
```diff
@@ -18,6 +18,11 @@
                     {% csrf_token %}
                     {% for field in form.hidden_fields() %}{{ field|safe }}{% endfor %}
                     {% for field in form.visible_fields() %}{{ render_field(field) }}{% endfor %}
+                    {% if xtheme.is_configuration_enabled(request, "allow_company_registration") %}
```
suutari-ai: This "should" look like this (see https://github.com/shuup/shuup/pull/1082/files#r108139152):
```jinja
{% if get_shop_configuration("allow_company_registration") %}
```

shuup/core/templatetags/shuup_common.py
```diff
@@ -136,3 +137,29 @@ def json(value):
...
+@contextfunction
+def get_shop_configuration(context, name, default=None):
+    """
+    Get configuration variable value for the current shop.
+
+    :type context: jinja2.runtime.Context
+    :type name: str
+    :type default: Any
+    """
+    from shuup import configuration
```
suutari-ai: I can't see a reason for this global import.. Can't this be at the top?

shuup/core/templatetags/shuup_common.py
```diff
@@ -136,3 +137,29 @@ def json(value):
...
+
+@library.global_function
+def get_global_configuration(name, default=None):
+    """
+    Get global configuration variable value.
+
+    :type name: str
+    :type default: Any
+    """
+    from shuup import configuration
```
suutari-ai: Same here...

shuup/core/templatetags/shuup_common.py
sekoskel: I guess not, cause tests fail with "ImportError: cannot import name 'ConfigurationItem'" if import is global

suutari-ai:
Good work Janne and Sergei! Thanks!

PR #1084, opened 2017-03-29
Fix ProductCrossSellsPlugin with invalid type

suutari-ai:
Fix following bug: If the "type" paramater is invalid in the config, the
cross sells plugin causes internal server error.

The fix was to simply catch a correct exception type.

The invalid value could end up in the config e.g. after changing the
plugin type from ProductHighlightPlugin to ProductCrossSellsPlugin in
the Xtheme editor.

### DIFF
--- a/doc/changelog.rst
+++ b/doc/changelog.rst
@@ -212,6 +212,7 @@ Front
 Xtheme
 ~~~~~~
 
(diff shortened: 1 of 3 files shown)


sekoskel:
:+1: :shipit:

PR #1086, opened 2017-03-30
front: Fix re-ordering

suutari-ai:
When adding products from an order to basket, don't add child lines or
subscriptions.

Child lines are excluded because otherwise package contents are added
twice.  Subscriptions are excluded because those don't use normal
checkout flow.

### DIFF
--- a/shuup/front/apps/personal_order_history/views.py
+++ b/shuup/front/apps/personal_order_history/views.py
@@ -9,7 +9,7 @@ from django.core.urlresolvers import reverse
 from django.http import HttpResponseRedirect
 from django.views.generic import View
 
-from shuup.core.models import Order, Supplier
+from shuup.core.models import Order, ProductMode, Supplier
 from shuup.front.views.dashboard import DashboardViewMixin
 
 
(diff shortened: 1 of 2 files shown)


shuup/front/apps/personal_order_history/views.py
```diff
@@ -30,6 +30,12 @@ class OrderDetailView(DashboardViewMixin, OrderViewMixin, django.views.generic.D
     template_name = 'shuup/personal_order_history/order_detail.jinja'
     context_object_name = 'order'
 
+    def get_context_data(self, **kwargs):
+        context = super(OrderDetailView, self).get_context_data(**kwargs)
+        reorderable_lines = _get_reorderable_lines(context['order'])
+        context['order_is_reorderable'] = reorderable_lines.exists()
```
sekoskel: what if an order has a normal product and a package product?

suutari-ai: Then the order is reorderable. Only normal products will be reordered though.

sekoskel: nvm my comment, thanks for explaining packaged product thing privately.

PR #1087, opened 2017-03-30
Front: add shop phone and email in notification

chessbr:
Notifications can use that to send emails or send SMS to the shop for any reason


shuup/front/notify_events.py
```diff
@@ -81,12 +83,22 @@ class RefundCreated(Event):
...
         customer_email=order.email,
         customer_phone=order.phone,
+        shop_email=None,
+        shop_phone=None,
         language=order.language
-    ).run()
+    )
+
+    if order.shop.contact_address:
+        params.update(dict(
```
tulimaki: Can you add small test for this new shop phone/email option? Also test could cover the case when the contact address is not there.

PR #1088, opened 2017-03-31
front: Fix thumbnail template tag for SVG files

suutari-ai:
Easy thumbnails doesn't support SVG and the `thumbnailer.get_thumbnail`
call here used to raise an IOError when called with an SVG file.

There is no much sense in thumbnailing SVGs since they are vector
graphics, so fix this bug by returning a direct url to the SVG file
rather than even trying to create thumbnails for them.

This fixes e.g. the maintenance mode page rendering when shop logo is an
SVG file.

### DIFF
--- a/doc/changelog.rst
+++ b/doc/changelog.rst
@@ -202,6 +202,7 @@ Admin
 Front
 ~~~~~
 
+- ``thumbnail`` template tag now returns SVG images as-is instead of crashing
 - Simple CMS and Category views now render metadata based on the description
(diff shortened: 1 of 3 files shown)


chessbr:
Can we have just a little test? There is a method to generate random images you can use and also create a SVG  file should be pretty easy :smile:

suutari-ai:
@chessbr: There you go. Some tests too. Also found another problem while writing the tests (commit dde4902).

PR #1089, opened 2017-03-31
Fix get_shop_configuration templatetag

sekoskel:
`context ` is a jinja's Context class, so context items should be accessed by calling a get method.
make 'Register as a company.' translatable.


suutari-ai:
@sekoskel: Oh, this was partially my fault, since the untested non-working code was from me. Sorry!

Anyway, it's PR author's responsibility to make sure code is tested. I still see no tests for this. Care to add some?

PR #1090, opened 2017-04-03
Fix get_shop_configuration template tag

suutari-ai:
Pick the fix from PR #1089 and add test cases to prove it's now working.


sekoskel:
looks good! :+1:

PR #1092, opened 2017-04-04
Translate company registration etc. to Finnish

suutari-ai:
Make the company registration form more DRY (to reduce number of
to-be-translated messages) and translate it to Finnish.

Add also "shuup.api" to Transifex configuration and translate it to
Finnish.

Translate a couple other found untranslated messages too.


VPAltti:
Finnish translations seemed OK.

PR #1094, opened 2017-04-05
Fix messages in the administration toolbar

suutari-ai:
Use non-shortened term for Admin*, because term "Admin" is hard to
translate, since it can mean "Administrator" or "Administration".  It's
also nice to be clear what's being administrated; for the translators
and for the users.

Change term "Maintenance" to "Maintenance mode" to reuse its
translation.  The former had incoherent translation in Finnish.


VPAltti:
Looks good!

PR #1095, opened 2017-04-06
Tune company registration

suutari-ai:
Tune company registration form:

  * Do not collect address details of the contact person, only the
    company's address matters.

  * Reorganize the registration form so that company details are filled
    in first, then the contact person details and finally the user
    account details.

Inject the JavaScript for handling the regions in the address form in
company registration view too.

Also fix couple person contact related Finnish translations.


shuup/front/apps/registration/templates/shuup/registration/company_register.jinja
```diff
@@ -1,25 +1,9 @@
 {% extends "shuup/front/base.jinja" %}
 {% from "shuup/front/macros/general.jinja" import render_field %}
+{% from "shuup/front/macros/checkout.jinja" import render_address_form %}
```
finmaster19: Why `render_address_form` is imported and in line 57 you create macro with the same name?

suutari-ai: Good catch. I first intended to use the version imported from checkout macros, but later decided to bake my own version, since the imported one wasn't good for this case after all.

I'll remove the import.

suutari-ai:
Updated: Removed the extra import.

PR #1096, opened 2017-04-11
Core: Add misc fields in ShopProduct

chessbr:
Enable to set a custom name, description and short description for a product of a shop.

- API: Fix viewset queryset filter
- API: Add django_filters to enable DRF filters form


doc/changelog.rst
```diff
@@ -12,6 +12,7 @@ Unrealeased
 Core
 ~~~~
 
+- Add name, description and short_description fields to ShoProduct model
```
Pikkupomo: typo `ShoProduct`

shuup/core/basket/objects.py
```diff
@@ -52,7 +52,13 @@ def cache_info(self, request):
         self.gross_weight = product.gross_weight
         self.shipping_mode = product.shipping_mode
         self.sku = product.sku
-        self.text = product.safe_translation_getter("name", any_language=True)
+        self.text = ""
+
+        shop_product = self.shop_product
```
Pikkupomo: useless assign imo

tulimaki:
@gurch101 can you fix these two things for your next commit to the branch.

shuup/core/basket/objects.py
chessbr: if you check the property above, the shop_product has the following code:
```
return self.product.get_shop_instance(self.shop)
```
It is not that expensive, but.. :)

PR #1097, opened 2017-04-11
Importer fixes

Pikkupomo:


tulimaki:
Can you add some extra info to commit messages why these "fixes" is needed and what issues these fix. Also some tests for issues this fixes would be nice.

shuup/importer/importing/importing.py
```diff
@@ -42,7 +44,7 @@ class DataImporter(object):
     def __init__(self, data, shop, language):
         self.shop = shop
         self.data = data
-        self.data_keys = data[0].keys()
+        self.data_keys = [k.replace(u'\ufeff', '') for k in data[0].keys()]
```
suutari-ai: Why don't you just remove the BOM from the beginning of the file while decoding the import file in the first place? It doesn't seem right to check BOMs at this point of import, since this phase should process already decoded data.

shuup/importer/importing/importing.py
```diff
@@ -274,6 +273,18 @@ def _handle_special_row_values(self, mapping, value):
 
     def _handle_row_field(self, field, mapping, orig_value, row_session, target, value):
         value = self._get_field_choices_value(field, value)
+
+        # do some magic
```
suutari-ai: It's magic as long as you don't describe what's happening. Doing magic is in code is bad. Readers should understand what's going on. => Add proper comment, if you think it needs explaining.

shuup/importer/importing/session.py
```diff
@@ -36,8 +36,12 @@ def get_deferred(self, key):
         return self.deferred_attach.get(key, (None, None))[0]
 
     def save(self):
-        if hasattr(self.instance, "shop"):
+        try:
+            self.instance.__class__._meta.get_field('shop')
```
suutari-ai: `__class__` is not needed here.

shuup/importer/transforms.py
```diff
@@ -67,7 +68,7 @@ def process_data(rows):
     if not len(data):
         for y, row in enumerate(rows):
             if y == 0:
-                headers = [x.lower().strip() for x in row]
+                headers = [force_text(x).lower().strip() for x in row]
```
suutari-ai: Who's passing non-unicode data to this function anyway? IMHO it's the callers responsibility to make sure that the passed-in data is already decoded. So rather add a docstring which documents that values in rows should be text, not bytes, and then check the callers that they do proper decoding.

shuup/importer/transforms.py
```diff
@@ -100,13 +101,12 @@ def transform_file(mode, filename, data=None):
     else:
         raise NotImplementedError("Mode %s Not implemented" % mode)
 
-    headers = data[0].keys() if len(data) else []
+    header_keys = data[0].keys() if len(data) else []
+    headers = [k.replace(u'\ufeff', '') for k in header_keys]
```
suutari-ai: BOM-handling doesn't belong here. See the other comment about BOM.

shuup/importer/transforms.py
```diff
@@ -129,6 +129,9 @@ def py3_read_file(data, filename):
         dialect = csv.Sniffer().sniff(f.read(1024))
         f.seek(0)
         for x, row in enumerate(csv.DictReader(f, dialect=dialect)):
-            got_data.update(set(h for (h, d) in six.iteritems(row) if d))
-            data.append(row)
+            got_data.update(set(h.replace(u'\ufeff', '') for (h, d) in six.iteritems(row) if d))
```
suutari-ai: Ditto, see the other BOM comments.

shuup/importer/transforms.py
```diff
@@ -129,6 +129,9 @@ def py3_read_file(data, filename):
...
+            new_row = {}
+            for k, v in row.items():
+                new_row[k.replace(u'\ufeff', '')] = v
```
suutari-ai: Ditto, see the other BOM comments.

shuup_tests/importer/test_product_import.py
```diff
@@ -68,6 +68,8 @@ def test_sample_import_no_match():
     tax_class = get_default_tax_class()
     product_type = get_default_product_type()
     supplier = get_default_supplier()
+    supplier.stock_managed = True
+    supplier.save()
```
suutari-ai: Are you changing this test because it doesn't pass without the change any more? I guess you're just trying to make it test the import with stock management enabled. It would be better to parameterize this (do_stocks = True and do_stocks = False), since that would prove it still works without stock management too.

shuup/core/models/_contacts.py
```diff
@@ -292,10 +292,10 @@ class PersonContact(Contact):
     birth_date = models.DateField(blank=True, null=True, verbose_name=_('birth date'), help_text=_(
         "The birth date of the contact."
     ))
-    first_name = models.CharField(max_length=30, blank=True, verbose_name=_('first name'), help_text=_(
+    first_name = models.CharField(max_length=120, blank=True, verbose_name=_('first name'), help_text=_(
```
chessbr: Yes, this was needed!

shuup/importer/transforms.py
```diff
@@ -125,7 +127,16 @@ def py2_read_file(data, filename):
 def py3_read_file(data, filename):
     got_data = set()
     data = []
-    with open(filename, encoding="utf-8") as f:
+
+    bytes = min(32, os.path.getsize(filename))
+    raw = open(filename, 'rb').read(bytes)
```
suutari-ai: You should also close the file. Also.. it doesn't matter if the file is shorter than the bytes read, so there is no need to do the getsize. I suggest simplying lines 131-137 to:
```python
with open(filename, 'rb') as fp:
    encoding = 'utf-8-sig' if fp.read(2) == codecs.BOM_UTF8 else 'utf-8'
```

shuup/importer/transforms.py
```diff
@@ -125,7 +127,16 @@ def py2_read_file(data, filename):
...
+
+    if raw.startswith(codecs.BOM_UTF8):
+        encoding = 'utf-8-sig'
+    else:
+        encoding = "utf-8"
+
+    with open(filename, encoding=encoding) as f:
```
suutari-ai: How about the py2 reader? Why do we have different py2 and py3 versions anyway?

However.. at least there isn't enough tests, since tests didn't find out that py2 version doesn't strip the BOM.

shuup_tests/importer/test_product_import.py
```diff
@@ -34,6 +35,12 @@ def test_sample_import_all_match(filename):
     product_type = get_default_product_type()
 
     path = os.path.join(os.path.dirname(__file__), "data", "product", filename)
+    if filename == bom_file:
+        import codecs
+        bytes = min(32, os.path.getsize(path))
+        raw = open(path, 'rb').read(bytes)
+        assert raw.startswith(codecs.BOM_UTF8)
```
suutari-ai: This is just testing that the test data has the BOM. You should also assert that the importer strips the BOM.

shuup_tests/importer/test_product_import.py
suutari-ai: Yeah, now it's much better. :+1:

PR #1103, opened 2017-04-20
Add create shipment view succeeded signal and event

finmaster19:
Create a new signal which is dispatched after the shipment_created
has dispatched. Connect notify script to this event so you can be sure
that every function connected to shipment_created signal has already
processed and so the shipment contains all information including data
fetched from 3rd party apis for example.

### DIFF
--- a/shuup/core/models/_orders.py
+++ b/shuup/core/models/_orders.py
@@ -38,7 +38,8 @@ from shuup.core.fields import (
 )
 from shuup.core.pricing import TaxfulPrice, TaxlessPrice
 from shuup.core.signals import (
-    payment_created, refund_created, shipment_created
+    payment_created, refund_created, shipment_created,
+    shipment_created_and_processed
 )
 from shuup.utils.analog import define_log_model, LogEntryKind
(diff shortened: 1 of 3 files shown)


finmaster19:
Here is one implementation which could fix the following problem. `ShipmentCreated` event is also connected to a `shipment_created` signal as some other functions could be, if you create a notify script, for example send email, sometimes notify script is executed before other functions connected to `shipment_created` signal are run. This causes problems with email action for example. Email is sent before other receiver functions are called and so some data could be missing from the email. Using another signal and event which are dispatched after the `shipment_created` signal should fix the problem because then we can be sure that the functions connected to `shipment_created` signal has been called and processed.

tulimaki:
Can we move the current `shipment_created` signal? Based on the comments above it seems to be pretty useless.

suutari-ai:
* A new signal is a good idea, but it should be fired from the same place as the shipment_created signal.
* There is no need for the new event, just connect the existing event to the new signal.
* Then this is clearly a fix and should go to 1.2 branch.

suutari-ai:
Thanks Aleksi!

PR #1104, opened 2017-04-26
Change FreeProductLine quantity field type to QuantityField

finmaster19:
Change FreeProductLine basket line effects quantity field
to QuantityField type because product units can have decimal
values. Add validation to FreeProductLine quantity field
so that quantity must be divisible with the product unit
step.


finmaster19:
Should have made pr to 1.2 branch

PR #1105, opened 2017-04-26
Change FreeProductLine quantity field type to QuantityField

finmaster19:
Change FreeProductLine basket line effects quantity field
to QuantityField type because product units can have decimal
values. Add validation to FreeProductLine quantity field
so that quantity must be divisible with the product unit
step.

### DIFF
--- a/doc/changelog.rst
+++ b/doc/changelog.rst
@@ -73,6 +73,8 @@ Order Printouts
 Campaigns
 ~~~~~~~~~
 
+- Fix handling of non-integer quantity in FreeProductLine
+
 Customer Group Pricing
 ~~~~~~~~~~~~~~~~~~~~~~
 
--- a/shuup/campaigns/admin_module/forms/_basket_effects.py
+++ b/shuup/campaigns/admin_module/forms/_basket_effects.py
@@ -4,6 +4,7 @@
 #
 # This source code is licensed under the OSL-3.0 license found in the
 # LICENSE file in the root directory of this source tree.
+from django.core.exceptions import ValidationError
 from django.utils.translation import ugettext_lazy as _
 
 from shuup.admin.forms.fields import PercentageField
@@ -13,7 +14,7 @@ from shuup.campaigns.models.basket_effects import (
 from shuup.campaigns.models.basket_line_effects import (
     DiscountFromCategoryProducts, DiscountFromProduct, FreeProductLine
 )
-from shuup.core.models import Category
+from shuup.core.models import Category, Product
 
 from ._base import BaseEffectModelForm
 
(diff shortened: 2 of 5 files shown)


shuup/campaigns/admin_module/form_sets.py
```diff
@@ -77,6 +80,27 @@ class BasketLineEffectsFormSet(EffectsFormset):
     def get_queryset(self):
         return self.owner.line_effects.instance_of(self._get_actual_model())
 
+    def clean(self):
+        if any(self.errors):
+            # Don't bother validating the formset unless each form is valid on its own
+            return
+        # Apply cleaning if model is FreeProductLine
```
suutari-ai: shouldn't this rather call super?

shuup/campaigns/admin_module/form_sets.py
```diff
@@ -77,6 +80,27 @@ class BasketLineEffectsFormSet(EffectsFormset):
...
+        if self._get_actual_model() == FreeProductLine:
+            for form in self.forms:
+                products_ids = form.cleaned_data["products"]
+                quantity = form.cleaned_data["quantity"]
+                for product_id in products_ids:
+                    shop_product = ShopProduct.objects.get(product_id=product_id)
```
suutari-ai: Where's shop_id?

shuup/campaigns/admin_module/form_sets.py
```diff
@@ -77,6 +80,27 @@ class BasketLineEffectsFormSet(EffectsFormset):
...
+                    step = shop_product.quantity_step
+                    if quantity % step != 0:
+                        n = (quantity // step)
+                        smaller_qty = max(1, n) * step
+                        larger_qty = (n + 1) * step
+                        form.add_error("quantity",
+                                       _("Quantity doesn't match with product units. "
+                                         "Closest quantities that match are {smaller_qty} and {larger_qty}")
+                                       .format(smaller_qty=smaller_qty, larger_qty=larger_qty))
```
suutari-ai: These should be translated to fi_FI and sv_SE too. You could use the same messages as are used in [purchase multiple check of `ShopProduct.get_orderability_errors`](https://github.com/shuup/shuup/blob/master/shuup/core/models/_product_shops.py#L350) and then just copy the messages, but hmm... maybe it's easier to just refactor the get_orderability_errors method a bit and split the quantity check part into a new method. The you could just do `shop_product.check_quantity_validity(quantity)` here (which would raise the ValidationErrors if the quantity is not OK).

shuup/campaigns/admin_module/form_sets.py
```diff
@@ -77,6 +80,27 @@ class BasketLineEffectsFormSet(EffectsFormset):
     def get_queryset(self):
         return self.owner.line_effects.instance_of(self._get_actual_model())
 
+    def clean(self):
```
suutari-ai: This validation shouldn't be here. It should be in `FreeProductLine.clean`. Don't forget to call super's clean too.

finmaster19:
Made some changes.

finmaster19:
Validation is now done in `FreeProductLineForm` because can't access to m2m relationship from model's clean method because the model is not saved yet. Refactors quantity errors to own method so it can be used else where also.

PR #1106, opened 2017-04-27
Basket API: Add option to set shipping and payment methods

tulimaki:
* Add service serializers for basket
* Add option to set payment and shipping methods through API
* Avoid having lazy gettext objects in the error messages

### DIFF
--- a/shuup/core/api/basket.py
+++ b/shuup/core/api/basket.py
@@ -35,9 +35,12 @@ from shuup.core.fields import (
 )
 from shuup.core.models import (
     Basket, get_company_contact, get_person_contact, MutableAddress, Order,
-    OrderLineType, OrderStatus, Product, Shop, ShopProduct
+    OrderLineType, OrderStatus, PaymentMethod, Product, ShippingMethod, Shop,
+    ShopProduct
 )
 from shuup.utils.importing import cached_load
+
+from .service import PaymentMethodSerializer, ShippingMethodSerializer
 
 
 def get_shop_id(uuid):
@@ -115,8 +118,8 @@ class BasketSerializer(serializers.Serializer):
     unorderable_items = serializers.SerializerMethodField()
     codes = serializers.ListField()
     shipping_address = serializers.SerializerMethodField()
-    shipping_method = serializers.IntegerField()
-    payment_method = serializers.IntegerField()
+    shipping_method = ShippingMethodSerializer()
+    payment_method = PaymentMethodSerializer()
     total_price = serializers.DecimalField(max_digits=FORMATTED_DECIMAL_FIELD_MAX_DIGITS,
                                            decimal_places=FORMATTED_DECIMAL_FIELD_DECIMAL_PLACES)
     total_price = serializers.DecimalField(max_digits=FORMATTED_DECIMAL_FIELD_MAX_DIGITS,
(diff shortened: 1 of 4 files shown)


shuup/core/api/basket.py
```diff
@@ -506,12 +521,35 @@ def set_shipping_address(self, request, *args, **kwargs):
...
+        return self._handle_setting_method(request, ShippingMethod, "shipping_method")
+
+    @detail_route(methods=['post'])
+    def set_payment_method(self, request, *args, **kwargs):
+        return self._handle_setting_method(request, PaymentMethod, "payment_method")
+
+    def _handle_setting_method(self, request, model, attr_field):
+        self.process_request()
+        serializer = MethodIDSerializer(data=request.data)
+        if serializer.is_valid():
```
chessbr: you can invoke `serializer.is_valid(raise_exception=True)` so you don't need to return `return Response(serializers.errors, status=status.HTTP_400_BAD_REQUEST)` explicitly and as a bonus we remove a indentation :)

shuup/core/api/basket.py
```diff
@@ -506,12 +521,35 @@ def set_shipping_address(self, request, *args, **kwargs):
...
+            if request.data.get("id"):
```
chessbr: use the serializer here, you don't need to test this any more.. just use `serializer.validated_data['id']`

shuup/core/api/service.py
```diff
@@ -0,0 +1,42 @@
...
+from rest_framework.serializers import SerializerMethodField
+
+from shuup.core.models import PaymentMethod, ShippingMethod
+
+
+class ServiceSerializerMixin(object):
+    def get_price(self, service):
+        price = 0
+        request = self.context["request"]
+        if request.basket:
```
chessbr: You are assuming that we have the basket on each request, which will be false when not using the Front middleware.
IMO, you should read the `source` from the context and whoever creates the instance of this serializer, needs to pass the `source` to the context to receive a valid price.

```python

source = self.context.get("source")
if source:
    price = service.get_total_cost(source).taxful_price.value
```

then when creating the serializer on a ViewSet
```python
class SomeViewSet(...):
    def get_serializer_context(self):
        context = super(SomeViewSet, self).get_serializer_context()
        context['source'] = self.request.basket
        return context
```

or even creating the serializer manually:
```python
PaymentMethodSerializer(..., context={..., 'source': order})
```

After doing this, you'll need to change the viewsets which uses these serializers

shuup/core/basket/objects.py
```diff
@@ -208,6 +208,38 @@ def billing_address(self, value):
...
+
+        if self.shipping_method_id:
+            return ShippingMethod.objects.get(pk=self.shipping_method_id)
+
+    @shipping_method.setter
+    def shipping_method(self, shipping_method):
+        self.shipping_method_id = (shipping_method.id if shipping_method else None)
+
+        if self.shipping_method_id and hasattr(self, "_data"):
+            self._load()['shipping_method_id'] = self.shipping_method_id
```
chessbr: yeah, this is also at the old code, but this logic do not accept to clearing the shipping method right? this also happens with addresses and payment method. maybe we should accept `None` here?

shuup_tests/api/test_basket_api_set_sevices.py
```diff
@@ -0,0 +1,89 @@
...
+    # Set shipping method for basket
+    shipping_method = factories.get_default_shipping_method()
+    data = {"id": shipping_method.id}
+    response = client.post('/api/shuup/basket/{}-{}/set_shipping_method/'.format(shop.pk, basket.key), data)
+    assert response.status_code == status.HTTP_200_OK
+    response_data = json.loads(response.content.decode("utf-8"))
+    method_data = response_data["shipping_method"]
+    assert method_data["id"] == shipping_method.id
+    assert method_data["translations"]["en"]["name"] == shipping_method.name
+    assert method_data["price"] == 0
```
chessbr: it would be good to have tests which actually have prices

tulimaki:
Refactored setting and getting basket data a little. Now there is ways to unset addresses and methods. Tests should prove that this works. Also refactored the service serializer a little to the direction you suggested.

chessbr:
Awesome work!

PR #1107, opened 2017-04-28
Extend default tax report

finmaster19:
Extend default tax report to include following information:
* Total pretax amount
* total tax amount
* total


tulimaki:
It doesn't say that this fixes something. This implies that the old tests should work as it is? I think you should just create new test for this "tax extend" instead tweaking the old tests.

suutari-ai:
The quantity change was not needed anymore after the fix in commit 33f5987. This is just adding new fields to existing report so there is no need to add a new test.

PR #1108, opened 2017-04-29
Campaigns: Fix issue in the multi shop coupons

tulimaki:
Make sure to include source/order shop in to calculations whether the code can be used or not.

Do not limit coupon code to be unique. In multi shops sharing same code like "XMAS" should be pretty common thing.

Instead limit that shop does not have multiple active basket campaigns for same code. Which would make figuring out which code to use pretty hard.

As a side note. Fix the `seed_source` test to allow shop as a parameter to actually guarantee that the source, shipping and payment method shares the same shop. This method is called all over the places so adding this extra in to it might be useful for future when we add more multi shop tests.

### DIFF
--- a/shuup/campaigns/models/campaigns.py
+++ b/shuup/campaigns/models/campaigns.py
@@ -224,6 +224,15 @@ class BasketCampaign(Campaign):
         return force_text(_("Basket Campaign: %(name)s" % dict(name=self.name)))
 
     def save(self, *args, **kwargs):
+        if self.coupon:
+            code_count_for_shop = BasketCampaign.objects.filter(
+                active=True, shop_id=self.shop.id, coupon__code=self.coupon.code)
+            if not self.id and code_count_for_shop.exists():
+                raise ValidationError(_("Can not have multiple active campaigns with same code."))
+
+            if self.id and code_count_for_shop.exclude(coupon_id=self.coupon.id).exists():
+                raise ValidationError(_("Can not have multiple active campaigns with same code."))
+
         super(BasketCampaign, self).save(*args, **kwargs)
         self.conditions.update(active=self.active)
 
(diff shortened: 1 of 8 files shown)


shuup/campaigns/models/campaigns.py
```diff
@@ -224,6 +224,15 @@ def __str__(self):
         return force_text(_("Basket Campaign: %(name)s" % dict(name=self.name)))
 
     def save(self, *args, **kwargs):
+        if self.coupon:
+            code_count_for_shop = BasketCampaign.objects.filter(
```
chessbr: This checkout should also be done on Coupon save right? As we allow merchants to change the coupon code it is possible to have this situation by just renaming the code :)

Also, the `code_count_for_shop.count() == 1` doesn't seem right.. I would suggest to use of `.exist()` if the queryset returns more than 1, I know, that is nearly impossible, but the `exist()` make more sense for me when reading the code.

shuup/campaigns/models/campaigns.py
```diff
@@ -417,11 +426,6 @@ def has_been_used(self, usage_count=1):
         """ See if code is used the times given """
         return CouponUsage.objects.filter(coupon=self).count() >= usage_count
 
-    def save(self, **kwargs):
-        if Coupon.objects.filter(code__iexact=self.code, active=True).exclude(pk=self.pk).exists():
```
chessbr: You should check here the same logic when saving the code as we merchant can, for some weird reason, change the coupon code OR, just we just disable the coupon code change, once created, do not change anymore.

Also, in the multi-vendor system, several shops sharing the same code and allowing all shops changing the code is some what dangerous. I mean, let's suppose I have a shop and create a coupon code and you also have a shop. You have bad intentions and simply change my coupon options, like usage limit or even the active property. Does that seem right? Maybe we should also have a coupon x shop relation here.

tulimaki: Here the shops is not sharing the actual coupon. They are just sharing the code part of the coupon. The coupon is still onetoone-relation to the basket campaign.

PR #1110, opened 2017-05-02
Fix PersonContact not activated after clicking activate link

finmaster19:
When user clicked activation link in the email, PersonContact
related to the user was not set to active and user still couldn't
login because PersonContact wasn't active altough the auth user was.
Fix this by activating also the PersonContact when user clicks
activate link.

### DIFF
--- a/shuup/front/apps/registration/__init__.py
+++ b/shuup/front/apps/registration/__init__.py
@@ -30,7 +30,7 @@ URL names
 
 import django.conf
 from django.db.models.signals import post_save
-from registration.signals import login_user, user_activated
+from registration.signals import user_activated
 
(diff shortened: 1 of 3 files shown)


tulimaki:
Doesn't seem to be right fix at all. I think you should set the person contact active when the activation link is clicked at email or when ever the user `is_activate` status is toggled. Also you shouldn't remove the tests when `settings.SHUUP_REGISTRATION_REQUIRES_ACTIVATION`.  We have both `ActivationView` and `activation_complete` already customized.

finmaster19:
Yeah was not good idea. Refactored this now so that PersonContact is updated on activation of auth user.

PR #1116, opened 2017-05-11
Handle registration activation email sending via notify scripts

finmaster19:
Disable the default user registration activation email sending.
Extend the RegistrationReceived event so that it can include
activation url. Then this event can be used to send customized
activation email to registered user.

### DIFF
--- a/shuup/front/apps/registration/locale/fi_FI/LC_MESSAGES/django.po
+++ b/shuup/front/apps/registration/locale/fi_FI/LC_MESSAGES/django.po
@@ -21,7 +21,7 @@ msgid ""
 msgstr ""
 "Project-Id-Version: Shuup\n"
 "Report-Msgid-Bugs-To: \n"
-"POT-Creation-Date: 2017-04-06 14:48+0000\n"
+"POT-Creation-Date: 2017-05-15 12:11+0000\n"
 "PO-Revision-Date: 2017-02-10 11:15+0000\n"
 "Last-Translator: Janne Tammi <janne@shuup.com>\n"
 "Language-Team: Finnish (Finland) (http://www.transifex.com/shuup/shuup/"
@@ -44,6 +44,12 @@ msgstr "Asiakas"
 
 msgid "Customer Email"
 msgstr "Asiakkaan sähköpostiosoite"
+
+msgid "Activation URL"
+msgstr "Aktivointi-URL"
+
+msgid "Is User Active"
+msgstr "Onko Käyttäjä Aktiivinen"
 
 msgid "Company Registration Received"
 msgstr "Yrityksen rekisteröityminen vastaanotettu"
--- a/shuup/front/apps/registration/notify_events.py
+++ b/shuup/front/apps/registration/notify_events.py
@@ -5,6 +5,7 @@
 #
 # This source code is licensed under the OSL-3.0 license found in the
 # LICENSE file in the root directory of this source tree.
+from django.core.urlresolvers import reverse
 from django.dispatch import receiver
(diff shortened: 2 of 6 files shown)


shuup/front/apps/registration/notify_events.py
```diff
@@ -13,14 +14,15 @@
...
-from shuup.notify.typology import Email, Model
+from shuup.notify.typology import Email, Model, URL
 
 
 class RegistrationReceived(Event):
     identifier = "registration_received"
     name = _("Registration Received")
     customer = Variable(_("Customer"), type=Model("shuup.Contact"))
     customer_email = Variable(_("Customer Email"), type=Email)
+    activation_url = Variable(_("Activation URL"), type=URL, required=False)
```
suutari-ai: Missing translations. 1.2 should be fully translated to at least English, Finnish and Swedish, so please update the corresponding PO files too.

suutari-ai:
Hmm.. there was a part that disabled e-mail sending of django-registration-redux, but I don't see it anymore. Shouldn't the e-mails of django-registration-redux be disabled?

suutari-ai:
This should have a test that checks that the event is really sent and it contains a correct activation URL.

finmaster19:
@suutari-ai Yes this is missing translations and tests. I opened a pr so I can get feedback about this feature. I was thinking that if we still leave the possibility to send activation email. Now setting `SEND_ACTIVATION_EMAIL` to False disables automatic activation email sending and then you could make your own notify script which handles it or not.

suutari-ai:
But if SEND_ACTIVATION_EMAIL is not set to False and merchant defines a notification script that sends an e-mail from the event, doesn't it then send two e-mails (one from django-registration-redux and one from Shuup's notifications)?

Though it's a bit problematic to just set it to False, because then shops upgrading Shuup from pre-1.2 to 1.2 would lose their activation e-mails. Any ideas how to address that?

One option would be to have a default notification script installed by a data migration, but I'm not sure if that's good, since installing default data is currently handled by shuup_init, not data migrations.

finmaster19:
> But if SEND_ACTIVATION_EMAIL is not set to False and merchant defines a notification script that sends an e-mail from the event, doesn't it then send two e-mails (one from django-registration-redux and one from Shuup's notifications)?

With this approach it would send two different email if SEND_ACTIVATION_EMAIL is True and you create your own notify script which also sends the link. When SEND_ACTIVATION_EMAIL is True, registration-redux sends automatic activation email. Then you don't need to create your own notify script which sends activation email, though you can use this `RegistrationReceived` how you want to create other scripts. But if SEND_ACTIVATION_EMAIL is False, then no automatic activation is send. Then you should at least create one script which uses this event and sends activation link.

suutari-ai:
Yeah, but if the code sends the hard coded activation email anyway, then you can't really customize what's being sent to the customer.

shuup/front/apps/registration/locale/fi_FI/LC_MESSAGES/django.po
```diff
@@ -45,6 +45,9 @@ msgstr "Asiakas"
 msgid "Customer Email"
 msgstr "Asiakkaan sähköpostiosoite"
 
+msgid "Activation URL"
+msgstr "Aktivointi URL"
```
suutari-ai: Aktivointi-URL

finmaster19:
Updated this pr so that you can use `user_is_active` event field as condition when script is run and activation email is sent.

suutari-ai:
Great work, Aleksi! Thanks!

PR #1118, opened 2017-05-12
Basket API: Handle extra_data

Pikkupomo:


shuup/core/api/basket.py
```diff
@@ -137,6 +139,7 @@ class BasketSerializer(serializers.Serializer):
     total_price_of_products = serializers.DecimalField(max_digits=FORMATTED_DECIMAL_FIELD_MAX_DIGITS,
                                                        decimal_places=FORMATTED_DECIMAL_FIELD_DECIMAL_PLACES)
     validation_errors = serializers.SerializerMethodField()
+    extra_data = ExtraDataSerializer()
```
chessbr: maybe here you want to use a [JSONField](http://www.django-rest-framework.org/api-guide/fields/#jsonfield), this way you can put whatever data inside. Otherwise, if some want to customize this, he will need to override the basket viewset, the basket serializer and then the extra data serializer. The serializer is a good one to validate the existence of keys and validate exactly the content set, but as this will be an extra data field, maybe that is not so necessary. Your call :smiley:

shuup/core/api/basket.py
```diff
@@ -543,6 +546,26 @@ def _handle_setting_address(self, request, attr_field):
             data = BasketSerializer(request.basket, context=self.get_serializer_context()).data
             return Response(data, status=status.HTTP_200_OK)
 
+    @schema_serializer_class(ExtraDataSerializer)
+    @detail_route(methods=['post'])
+    def set_extra_data(self, request, *args, **kwargs):
+        self.process_request()
+        serializer = ExtraDataSerializer(data=request.data)
+        serializer.is_valid(raise_exception=True)
+        if serializer.is_valid():
```
chessbr: no need to check here again, you have just checked on the line :554

shuup/core/api/basket.py
```diff
@@ -543,6 +546,26 @@ def _handle_setting_address(self, request, attr_field):
...
+            if not serializer.validated_data:
```
chessbr: so if the dictionary is blank it is wrong? maybe this check should be done in the [serialize.validate()](http://www.django-rest-framework.org/api-guide/serializers/#object-level-validation) method

shuup/core/api/basket.py
```diff
@@ -543,6 +546,26 @@ def _handle_setting_address(self, request, attr_field):
...
+                return Response({"error": "invalid key"}, status=status.HTTP_400_BAD_REQUEST)
+
+            extra_data = {}
+            for k, v in request.data.items():
```
chessbr: you should probably iterate over `serializer.validated_data` since it will parse the values correctly accordingly to the serializer field types

shuup/core/basket/objects.py
```diff
@@ -242,6 +243,20 @@ def payment_method(self, payment_method):
...
+        if self._extra_data:
+            return self._extra_data
+        if hasattr(self, "_data"):
+            return self._load().get('extra_data')
+
+    @extra_data.setter
+    def extra_data(self, value):
+        self._extra_data = value
+        # save also on data
+        if value and hasattr(self, "_data"):
```
chessbr: once set, you can not clear the data anymore? I mean, calling `basket.extra_data = None` or even `basket.extra_data = {}` will have no effect at all?

shuup/core/api/basket.py
```diff
@@ -547,6 +550,21 @@ def _handle_setting_address(self, request, attr_field):
             data = BasketSerializer(request.basket, context=self.get_serializer_context()).data
             return Response(data, status=status.HTTP_200_OK)
 
+    @schema_serializer_class(ExtraDataSerializer)
+    @detail_route(methods=['post'])
```
chessbr: sad, we must make the serializer work here (one day)

shuup/core/basket/objects.py
```diff
@@ -242,6 +243,21 @@ def payment_method(self, payment_method):
...
+            return self._extra_data
+        if hasattr(self, "_data"):
+            return self._load().get('extra_data')
+        return {}
+
+    @extra_data.setter
+    def extra_data(self, value):
+        self._extra_data = value or {}
+        # save also on data
+        if value and hasattr(self, "_data"):
```
chessbr: once set, you can not clear the data anymore? I mean, calling `basket.extra_data = None` or even `basket.extra_data = {}` will have no effect at all?

shuup_tests/api/test_basket_api.py
```diff
@@ -1086,3 +1086,37 @@ def test_basket_reorder_staff_user(settings):
...
+    settings.SHUUP_ENABLE_MULTIPLE_SHOPS = False
+    response = client.post("/api/shuup/basket/new/")
+    assert response.status_code == status.HTTP_201_CREATED
+    basket = Basket.objects.first()
+
+    # test with valid key
+    key = "sales_option_id"
+    value = "testt"
+    data = {key: value}
+    response = client.post('/api/shuup/basket/{}-{}/set_extra_data/'.format(shop.pk, basket.key), data)
```
chessbr: create unit test to clear the `extra_data` passing a clear dict

shuup/core/api/basket.py
```diff
@@ -137,6 +139,7 @@ class BasketSerializer(serializers.Serializer):
     total_price_of_products = serializers.DecimalField(max_digits=FORMATTED_DECIMAL_FIELD_MAX_DIGITS,
                                                        decimal_places=FORMATTED_DECIMAL_FIELD_DECIMAL_PLACES)
     validation_errors = serializers.SerializerMethodField()
+    extra_data = serializers.JSONField(binary=False, required=False)
```
suutari-ai: binary is False by default, no need to specify it.

shuup/core/api/basket.py
```diff
@@ -547,6 +550,21 @@ def _handle_setting_address(self, request, attr_field):
...
+        try:
+            extra_data = {}
+            for k, v in request.data.items():
+                extra_data[k] = v
+            request.basket.extra_data = extra_data
+            request.basket.save()
+            data = BasketSerializer(request.basket, context=self.get_serializer_context()).data
+            return Response(data, status=status.HTTP_200_OK)
+        except:
+            return Response({"error", "Bad Request"}, status=status.HTTP_400_BAD_REQUEST)
```
suutari-ai: Why do you need this separate set_extra_data endpoint? Wouldn't it make more sense to have a single `update` endpoint, which can update the extra_data, billing_address, shipping_address, codes, etc. whichever is set in the PATCH request? E.g. if the client calls `PATCH /api/shuup/basket/BASKETNAME` with argument like `{"extra_data": {"the": "extra data", "to": "set"}}`, it would set just the extra_data and similarly you could set only the billing address by providing just that in the input. That would also allow updating all aspects of the basket in a single request.

PR #1121, opened 2017-05-15
API: Add pagination for best selling products

tulimaki:
Best selling products interface should return same paginated response like when getting list of products. The caller shouldn't need to process the response differently for best selling products.

### DIFF
--- a/shuup/core/api/front_products.py
+++ b/shuup/core/api/front_products.py
@@ -27,7 +27,6 @@ from shuup.core.models import (
 from shuup.core.pricing._context import PricingContext
 from shuup.core.utils import context_cache
 from shuup.core.utils.product_statistics import get_best_selling_product_info
-from shuup.utils.numbers import parse_decimal_string
 
 DISTANCE_PER_DEGREE = 111.045   # 111.045km in a latitude degree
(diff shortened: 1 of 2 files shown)


shuup/core/api/front_products.py
```diff
@@ -445,13 +438,14 @@ def best_selling(self, request):
 
         # take here the top `limit` records, because the filter_queryset below can mess with our work
         product_ids = [
-            d[0] for d in sorted(six.iteritems(combined_variation_products), key=lambda i: i[1], reverse=True)[:limit]
+            d[0] for d in sorted(six.iteritems(combined_variation_products), key=lambda i: i[1], reverse=True)
         ]
 
         shop_products_qs = ShopProduct.objects.filter(product__id__in=product_ids)
         shop_products_qs = self.filter_queryset(shop_products_qs).distinct()
+        shop_products_qs = self.paginate_queryset(shop_products_qs)
```
chessbr: Take a look here: https://github.com/encode/django-rest-framework/blob/master/rest_framework/mixins.py#L39
The `paginate_queryset` can return `None` if no page is needed

PR #1124, opened 2017-05-24
Behavior api

Pikkupomo:


shuup/core/api/service.py
```diff
@@ -10,14 +10,43 @@
...
+        BehaviorComponents are Polymorphic thus it's required to make
+        the `to_representation` trick.
+        """
+        return DynamicBehaviorComponentSerializer(obj, context=self.context).to_representation(obj)
 
 
 class ServiceBaseSerializer(TranslatableModelSerializer):
     price = SerializerMethodField()
     is_available = SerializerMethodField()
+    behavior_components = SerializerMethodField()
```
tulimaki: I don't like this idea since this serializer is used by public API. Also not sure whether there is real need for this.

shuup/core/basket/objects.py
```diff
@@ -178,6 +178,8 @@ def clear_all(self):
         self._data = {}
         self.uncache()
         self.dirty = True
+        self.shipping_method = None
```
tulimaki: Maybe you should set `self.shipping_method_id` to None instead of method. Doing this saves the None to the data.

Pikkupomo: Yes I know, it was kind of my intention to really save that. Dunno if it's okay though

tulimaki:
👍 for the tests.

PR #1125, opened 2017-05-25
CGP: Introduce customer group discounts

chessbr:
Add a discount module to configure discounts by contact group

- CGP: bump cache only if there is shop product (SH-554)
- API: add front shop products price info serializer


shuup/customer_group_pricing/models.py
```diff
@@ -42,3 +42,32 @@ def save(self, *args, **kwargs):
         # check if there is a shop product before bumping the cache
         if self.product.shop_products.filter(shop_id=self.shop.id).exists():
             bump_cache_for_product(self.product, self.shop)
+
+
+class CgpDiscount(MoneyPropped, models.Model):
```
tulimaki: These to models in this file shares almost every field. You probably should get rid of the duplicate code here.

shuup/core/utils/price_display.py
```diff
@@ -74,7 +74,7 @@ def _register(self):
 class PriceDisplayFilter(_ContextFilter):
     def __call__(self, context, item, quantity=1, include_taxes=None, allow_cache=True):
         key, val = context_cache.get_cached_value(
-            identifier=self.cache_identifier, item=item, context=context,
+            identifier=self.cache_identifier, item=item, context=context.get('request', context),
```
Pikkupomo: In which situation context has a request in it? IMO the filter is wrongly called if it has

chessbr: I think it always should have. This is the context from jinja. There is a current code (from master) which gets the request from the context here: https://github.com/shuup/shuup/blob/master/shuup/core/utils/price_display.py#L90
If mine is wrong, then this ^^ is also wrong.

Pikkupomo: I would separate this commit to different PR and let @suutari-ai check it too

tulimaki: Agree. Since it is not part of the initial task no need to merge this as part of this PR.

suutari-ai:
Aren't customer group discounts already part of the campaigns module? What's the point doing the same feature in two modules?

chessbr:
The point here @suutari-ai was mainly performance. The customer group discount will be a lot faster as it does "just a single" query for the product we want to put discount and using catalog campaigns will need well, campaign, rule and effect.

PR #1128, opened 2017-06-02
Core: do not allow adding variation parents in basket

chessbr:
Adding variation parent in the basket shouldn't be possible.


tulimaki:
LGTM.

PR #1145, opened 2017-07-04
Make it possible to override person contact form properties

finmaster19:
Add new setting called SHUUP_PERSON_CONTACT_FIELD_PROPERTIES
which can be used the way as SHUUP_ADDRESS_FIELD_PROPERTIES
setting to override form field properties. You can specify a
dict containing the field name as the key and the value is a
another dict containing the field properties and their new values.

### DIFF
--- a/shuup/front/apps/customer_information/forms.py
+++ b/shuup/front/apps/customer_information/forms.py
@@ -6,6 +6,7 @@
 # This source code is licensed under the OSL-3.0 license found in the
 # LICENSE file in the root directory of this source tree.
 from django import forms
+from django.conf import settings
 from django.core.exceptions import ValidationError
 from django.utils.translation import ugettext_lazy as _
 from enumfields import EnumField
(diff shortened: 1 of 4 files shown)


suutari-ai:
The setting should be in Front too. Also move the change log entry to "Front" section... I'm also hesitating to add any new features to 1.2 branch now, since it's in feature freeze... This is quite minor change though, so might make an exception.

suutari-ai:
Since the setting is used in Front, the setting should be in Front too (shuup/front/settings.py). Also move the change log entry to "Front" section... I'm also hesitating to add any new features to 1.2 branch now, since it's in feature freeze... This is quite minor change though, so might make an exception.

shuup/front/apps/customer_information/forms.py
```diff
@@ -29,6 +30,11 @@ def __init__(self, *args, **kwargs):
             self.fields[field].required = True
         self.initial["language"] = self.instance.language
 
+        field_properties = settings.SHUUP_PERSON_CONTACT_FIELD_PROPERTIES
+        for field, properties in field_properties.items():
+            for prop in properties:
+                setattr(self.fields[field], prop, properties[prop])
+
```
suutari-ai: I think it is OK to add this feature to 1.2, because it's very minor change and the implementation is so simple and straight forward. Also this code clearly does nothing if the setting is not modified from its default value (empty dict), so we should be fairly safe.

finmaster19:
Do you mean I should specify that setting in two different places or just in `front/settings.py`?

suutari-ai:
Just in front/settings.py. Why would you think it should be in Core too?

finmaster19:
I read read your comment fast and just made sure that I understood it correctly

suutari-ai:
Thanks, Aleksi!

PR #1171, opened 2017-12-22
Enable browser tests

tulimaki:
* Update all related packages
* Run admin and front tests in separate builds to prevent
build time to rise
* Do not run notify tests with Travis since not business
critical
* Split few tests to avoid individual tests take too much
time. This caused issues with Travis.

There are some problem spots which get fixed with some extra
waiting. Also it seems to be good practice to initialize
database before initializing tests, otherwise there might
be some random database locks happening.

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -1,5 +1,7 @@
 sudo: false
 language: python
+addons:
+    firefox: "latest"
 cache:
   directories:
     - "$HOME/.cache/pip"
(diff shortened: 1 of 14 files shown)


tulimaki:
Let's keep Travis disabled since still some inconsistency with the results. Locally the tests passes pretty much every single time and these issues with Travis is super hard to reproduce.

PR #1172, opened 2017-12-22
Squash migrations

tulimaki:


tulimaki:
Disabled migrations from one Travis test for you and you might get the idea. Also run the migrations for MySQL for fresh database. I assume no difference in Django version but I have 1.9.13 and mysql 5.7.17. Fails every single time. And I am not the only one having issues with this branch.

suutari-ai:
Travis finds a few errors like this in the tests:
> django.db.utils.IntegrityError: UNIQUE constraint failed: shuup_currency.code

It's caused by invalid tests, since the tests assume that database doesn't have any currencies already, but when the database is created with migrations (instead of "syncdb" which is the nomigration way) it runs also the RunPython commands and those create the initial currencys. This isn't anything new, but was there already before the squashed migrations.

These *unit tests* should be fixed to take the existing currencies into account.

tulimaki:
@suutari-ai Thank you! Will change this PR a little later since now we know the reason for the migration issue. Also will fix these issues too. I think we are going to put that Travis job I changed like it was and then create a new job which is run with migrations.

tulimaki:
I think the main thing we should learn from that coverage complain is https://github.com/shuup/shuup/issues/1174. And we do that separately. Just now there is no time to up the coverage. But we can't keep this hanging these squashed migrations here too long so will merge.

PR #1181, opened 2018-01-08
Admin: Add form field for main products

tulimaki:
Add option for select2 multiselect for only normal and parent products.

### DIFF
--- a/shuup/admin/static_src/base/js/select.js
+++ b/shuup/admin/static_src/base/js/select.js
@@ -7,7 +7,7 @@
  * LICENSE file in the root directory of this source tree.
  */
 
-function activateSelect($select, model, attrs={}) {
+function activateSelect($select, model, searchMode, attrs={}) {
     if(model === undefined) {
         return $select.select2($.extend(true, {
             language: "xx"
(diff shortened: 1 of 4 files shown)


shuup/admin/static_src/base/js/select.js
```diff
@@ -20,7 +20,7 @@ function activateSelect($select, model, attrs={}) {
             url: window.ShuupAdminConfig.browserUrls.select,
             dataType: "json",
             data: function(params) {
-                return {model: model, search: params.term};
+                return {model: model, searchMode: searchMode, search: params.term};
```
Pikkupomo: can this be `{model: model, searchMode, search: params.term}`

shuup/admin/views/select.py
```diff
@@ -72,21 +74,35 @@ def get_data(self, request, *args, **kwargs):
...
+
+            qs = qs.filter(query)
+
+        search_mode = request.GET.get("searchMode")
+        if search_mode and search_mode == "main" and issubclass(cls, Product):
+            qs = qs.filter(mode__in=[
+                ProductMode.SIMPLE_VARIATION_PARENT,
+                ProductMode.VARIABLE_VARIATION_PARENT,
+                ProductMode.NORMAL
+            ])
```
Pikkupomo: Could this be `ProductQuerySet::top_level_products()` or similar, so this could be changed to `qs = qs.top_level_products()`? it would make it more re-usable.

shuup/admin/static_src/base/js/select.js
tulimaki: I think not. At least the eslint complains about it.

shuup/admin/views/select.py
tulimaki: I guess so.

PR #1182, opened 2018-01-08
Enable available payments on order api

Pikkupomo:


shuup/core/models/_orders.py
```diff
@@ -1092,5 +1092,33 @@ def get_customer_name(self):
             if getattr(self, "%s_id" % attr):
                 return getattr(self, attr).name
 
+    def get_available_shipping_methods(self):
+        """
+        Get available shipping methods.
+
+        :rtype: list[ShippingMethod]
+        """
+        from shuup.core.models import ShippingMethod
```
tulimaki: Is this imports here really necessary?

shuup/core/models/_orders.py
```diff
@@ -1092,5 +1092,33 @@ def get_customer_name(self):
...
+            if m.is_available_for(self)
+        ]
+
+    def get_available_payment_methods(self):
+        """
+        Get available payment methods.
+
+        :rtype: list[PaymentMethod]
+        """
+        from shuup.core.models import PaymentMethod
```
tulimaki: Is this imports here really necessary?

shuup/core/models/_orders.py
Pikkupomo: Importing these in "header" causes circular import (using `shuup.core.models`). Direct import might be possible though but it's ugly :)

PR #1191, opened 2018-01-31
API: Improve error message for shop product not found

chessbr:


tulimaki:
Tests and I think we good. @chessbr

PR #1193, opened 2018-02-09
DefaultTax: filter tax rules correctly

chessbr:
When tax rule has a specific customer tax group, do not allow contexts with no customer tax groups to match.

In some situations where a tax is only applied to a specific customer group, it is more than necessary to avoid taxing customers that do not belong to that group or which don't even have a group configured.

So, in cases where we have taxes for companies like B2B and also sell to a regular customer with different prices and also allow anonymous ordering, we don't want to tax the anonymous user the same way we tax companies.

To create a tax rule that applies to any customer group, creating a simple rule with no customer tax group will work as expected.

No refs

### DIFF
--- a/shuup_tests/default_tax/test_default_tax.py
+++ b/shuup_tests/default_tax/test_default_tax.py
@@ -13,7 +13,7 @@ from decimal import Decimal
 import pytest
 from django.test.utils import override_settings
 
-from shuup.core.models import Tax, TaxClass
+from shuup.core.models import CustomerTaxGroup, Tax, TaxClass
 from shuup.core.taxing import get_tax_module, TaxingContext
 from shuup.default_tax.admin_module.views import TaxRuleEditView
 from shuup.default_tax.models import TaxRule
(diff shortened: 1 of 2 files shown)


shuup_tests/default_tax/test_default_tax.py
```diff
@@ -237,6 +237,59 @@ def test_rule_admin(rf, admin_user):
...
+    tax_for_companies = Tax.objects.create(code="companies", rate=0.3, name="Additional tax for companies")
+
+    # create tax group for companies
+    companies_tax_group = CustomerTaxGroup.get_default_company_group()
+
+    # create the tax rule as follows:
+    # - 10% for any kind of customer, no matter what
+    # - 30% only for companies
+    any_tax_rule = TaxRule.objects.create(tax=tax_for_anyone)
+    any_tax_rule.tax_classes.add(tax_class)
```
Pikkupomo: Can you give me a real life example for this? Like why and where anyone would do this? This sounds like a pricing module responsibility for me: "All anonymous customers pay 10% more". 

I have a really hard time to figure out why would any merchant do this and even harder time to figure out why we allow this.

chessbr: To be honest, I don't know any case here in my country for this.
But I have the certainty that me, as not inside any customer tax group, should be taxed the same as a customer tax group. What is the purpose of the `customer tax group` if not to limit?

Pikkupomo:
```So, in cases where we have taxes for companies like B2B and also sell to a regular customer with different prices and also allow anonymous ordering, we don't want to tax the anonymous user the same way we tax companies.``` 

This makes no sense because the merchant has no "authority" over making up "some taxes" for anonymous users, there must be something in the user that tells the tax system which rules should apply, being anonymous is not enough. This has no use case in USA or other countries that have complex taxation.

This might be usable in countries where the taxation is simple (like Finland) and for that, it might be justified. 

Also, this is missing a test that proves that anonymous users were taxed same way as companies.

**Comment by tulimaki** [comment]
@Pikkupomo how this thing used to be.

```
So, in cases where we have taxes for companies like B2B and also sell to a regular customer with different prices and also allow anonymous ordering, we don't want to tax the anonymous user the same way we tax companies.
```

Should would you agree that the companies and regular consumers have usually different tax percentages? I think this applies at least in Finland.

**Comment by Pikkupomo** [comment]
@tulimaki your question has nothing to do with anonymity

1) Same taxes are being paid (in the end)
2) the difference is if we show prices taxless or taxful (Finland). 
3) if we don't want companies to pay taxes in Shuup, we will ensure that they are in company group no tax rules match their group. (tax rules are only for Person Contacts)

**Comment by chessbr** [comment]
> Also, this is missing a test that proves that anonymous users were taxed same way as companies.

It is already there. Anonymous is taxed with `any` and company with `any` AND `companies`

**Comment by tulimaki** [comment]
`the difference is if we show prices taxless or taxful (Finland).` This has nothing to do with the issue I think. The issue is that tax rules designed for specific group like companies shouldn't ably for anonymous contact since we don't how whether it is company or not.

**Comment by tulimaki** [comment]
```

tulimaki:
```

PR #1201, opened 2018-02-20
Sync with 1.4 branch

suutari-ai:
These changes are already released in 1.4 branch as 1.4.1.


shuup/utils/dates.py
```diff
@@ -219,6 +219,20 @@ def local_now(tz=None):
...
+    for value in [start, end]:
+        if not isinstance(value, datetime.date):
+            raise TypeError("Not date or datetime: {!r}".format(value))
+    start_is_datetime = isinstance(start, datetime.datetime)
+    end_is_datetime = isinstance(end, datetime.datetime)
+    if start_is_datetime != end_is_datetime:
+        raise TypeError("Start and end must be same type: {!r} - {!r}"
+                        .format(start, end))
+    # Add +1 day to end if it's a date to make the range inclusive
+    end_delta = datetime.timedelta(days=(1 if not end_is_datetime else 0))
```
chessbr: shoudn't this add the delta when the value is datetime instead?
https://docs.djangoproject.com/en/1.9/ref/models/querysets/#range
From my understanding of the docs, dates are inclusive and datetimes aren't.

suutari-ai: No, it's correct that way. If you have two datetimes, e.g. dt1 = 2005-01-01 08:00:00 and dt2 = 2005-03-31 20:00:00, then the range should be used with those exact endpoints dt1 and dt2, but if they are bare dates, e.g. d1 = 2005-01-01 and d2 = 2005-03-31, then we want the range to include all times within those dates, especially the last day too, which is accomplished by converting them to d1_dt = 2005-01-01 00:00:00 and d2_dt = 2005-03-31 24:00:00 = 2005-04-01 00:00:00.

PR #1207, opened 2018-03-08
API: add MoneyField and encoder for Money instance

chessbr:
- Add MoneyField for serializers that includes the money value and currency
- Add API encoder to Money instance to return only the value of the object

No Refs

### DIFF
--- a/shuup/api/encoders.py
+++ b/shuup/api/encoders.py
@@ -11,12 +11,16 @@ from enum import Enum
 from rest_framework.renderers import JSONRenderer
 from rest_framework.utils import encoders
 
+from shuup.utils.money import Money
+
 LOG = logging.getLogger(__name__)
 
 
 class ExtJSONEncoder(encoders.JSONEncoder):
     def default(self, obj):
         if isinstance(obj, Enum):
+            return obj.value
+        if isinstance(obj, Money):
             return obj.value
         return super(ExtJSONEncoder, self).default(obj)
 
--- a/doc/changelog.rst
+++ b/doc/changelog.rst
@@ -12,6 +12,7 @@ Unrealeased
 Core
 ~~~~
 
+- API: add MoneyField and encoder for Money instances
 - OrderCreator: Dispatch a signal when adding lines to order
 - Enable refunds for order API
 - API: Improved suppliers stock endpoints
--- a/shuup/api/fields.py
+++ b/shuup/api/fields.py
@@ -18,7 +18,8 @@ from django.utils.encoding import force_text
 from rest_framework import serializers
 
 from shuup.core.fields import (
-    FORMATTED_DECIMAL_FIELD_DECIMAL_PLACES, FORMATTED_DECIMAL_FIELD_MAX_DIGITS
+    FORMATTED_DECIMAL_FIELD_DECIMAL_PLACES, FORMATTED_DECIMAL_FIELD_MAX_DIGITS,
+    MONEY_FIELD_DECIMAL_PLACES
 )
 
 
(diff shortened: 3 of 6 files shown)


shuup/api/encoders.py
```diff
@@ -11,13 +11,17 @@
...
+
 LOG = logging.getLogger(__name__)
 
 
 class ExtJSONEncoder(encoders.JSONEncoder):
     def default(self, obj):
         if isinstance(obj, Enum):
             return obj.value
+        if isinstance(obj, Money):
+            return obj.value
```
suutari-ai: The currency should be exported too. Maybe a money instance should be encoded as `{'value': obj.value, 'curreny': obj.currency}` ?

chessbr: good question. this sounds like a special field for me, not a job for the encoder. what do you think?

tulimaki: Yeah I think valid point. And sure we want to keep the currency information.

chessbr: I guess I can create a the field to be used with `Money` objects, it will parse and serialize the object as you said and the encoder can be a fallback to just the float value.
We did this because we were having issues with a particular scenario with builtin `decimal.py`. We have a Python 3.4 compiled without decimal C module, weird though. And rebuilding/updating python at that point was not an option, so we had to "fix" our code.
The error was:
```
ValueError: could not convert string to float: '0E-9 USD'
  File "django/core/handlers/base.py", line 174, in get_response
    response = self.process_exception_by_middleware(e, request)
  File "django/core/handlers/base.py", line 172, in get_response
    response = response.render()
  File "django/template/response.py", line 160, in render
    self.content = self.rendered_content
  File "rest_framework/response.py", line 72, in rendered_content
    ret = renderer.render(self.data, accepted_media_type, context)
  File "rest_framework/renderers.py", line 103, in render
    separators=separators
  File "json/__init__.py", line 237, in dumps
    **kw).encode(obj)
  File "json/encoder.py", line 192, in encode
    chunks = self.iterencode(o, _one_shot=True)
  File "json/encoder.py", line 250, in iterencode
    return _iterencode(o, 0)
  File "shuup/api/encoders.py", line 21, in default
    return super(ExtJSONEncoder, self).default(obj)
  File "rest_framework/utils/encoders.py", line 47, in default
    return float(obj)
  File "shuup/utils/_unitted_decimal.py", line 176, in __float__
    return super(UnittedDecimal, self).__float__(**kwargs)
  File "python3.4/decimal.py", line 1614, in __float__
    return float(s)
```

See that the code executing python code instead of the C Module. See the required version of the lib here: https://github.com/python/cpython/blob/3.4/Lib/decimal.py#L149 The server has 2.4.0. So this was breaking there.
I couldn't create a good unit test for this scenario, basically if I remove this import here https://github.com/python/cpython/blob/3.4/Lib/decimal.py#L6400 I can make tests crash :smile: So if you have a good idea, I appreciate. Now you have context @suutari-ai , I will edit the commit and PR info to include this too.

suutari-ai: Being able to do `float(Money(10, 'USD'))` is not a very good feature anyway. IMHO it could even be better if that wouldn't work and there was just the explicit way to get the value out of the Money object, i.e. `Money(10, 'USD').value`.

If there is serializers that expect to serialize the money object as simple float, they're simply wrong and should be fixed to either serialize them as

suutari-ai:
Hmm.. the new MoneyField serializer is not used anywhere, so it seems unneeded still. Should it rather be added when there is a real use case for it inside Shuup API? Or maybe some of the existing Money serializers should be updated to use it?

I'm against the another change of adding a special case for Money to `ExtJSONEncoder`, because explicit is better than implicit and that change would make money amounts exported as bare numbers implicitly. Also wouldn't that need some kind of decoder part too to work for importing money amounts too?

shuup/api/encoders.py
chessbr: yes, indeed. well, maybe `Money.__float__` should raise exception then.
I put the `MoneyField` in the PR already, it deals with the object `{"currency: "USD", "value": 10}`. To deal with Money values, I guess a `ReadOnlyField(source="order.taxful_total_price.value")` is enough.

tulimaki:
We did local fix for this. I think it was our mistake from begin with when we started to serializing Moneys without correct serializer. I guess this is on hold since Shuup doesn't have any use cases for this kind of serializer.

PR #1213, opened 2018-03-21
API: return product taxful and taxless base price

chessbr:
No Refs


tulimaki:
LGTM, I guess it is good to have this serialized for product.

PR #1216, opened 2018-03-30
Report: change date filter field to DateTime

chessbr:
Make it possible to filter objects also by time.

This is useful for filtering orders in a time range inside a single day, for example.


tulimaki:
LGTM

PR #1229, opened 2018-04-20
Front: Improve front for multivendor and customization

tulimaki:


shuup/front/utils/user.py
```diff
@@ -0,0 +1,20 @@
...
+# This file is part of Shuup.
+#
+# Copyright (c) 2012-2018, Shoop Commerce Ltd. All rights reserved.
+#
+# This source code is licensed under the OSL-3.0 license found in the
+# LICENSE file in the root directory of this source tree.
+from django.contrib.auth.models import AnonymousUser
+
+
+def is_admin_user(request):
```
chessbr: maybe we could create some `AdminUserProvider` that checks whether the current request user is an admin and open that through provides.

tulimaki: Let's do the admin user provider some other time.

PR #1234, opened 2018-04-26
Provides: add setting to blacklist undesired

chessbr:


tulimaki:
I think this is a good idea. I definitely can see situation when the app provides something we need, but also some unwanted extra features. This way we should also avoid all kind of mad visibility/should show issues.

PR #1235, opened 2018-04-27
Fix issue with category js

Pikkupomo:
if ReloadProducts was called from product page, it caused an unnecessary js error

### DIFF
--- a/shuup/front/static_src/js/category.js
+++ b/shuup/front/static_src/js/category.js
@@ -69,6 +69,9 @@ function getFilterString(state) {
 function reloadProducts(filterString) {
     const $cont = $("#ajax_content");
     const $prods = $(".products-wrap");
+    if (!$prods.length) {
+        return;
+    }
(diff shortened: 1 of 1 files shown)


shuup/front/static_src/js/category.js
```diff
@@ -69,6 +69,9 @@ function getFilterString(state) {
 function reloadProducts(filterString) {
     const $cont = $("#ajax_content");
     const $prods = $(".products-wrap");
+    if (!$prods.length) {
```
chessbr: Are you sure this is the issue? The length is tested above and then `$cont.offset().top` is used.

tulimaki: Yeah add some more details about the issue. How this works if I add so many filters that there is no products at all? Then how the reload products is supposed to work when I start disabling those filters?

Pikkupomo:
I'll close this now, it might have been caused by other issue.

PR #1236, opened 2018-05-01
Simple CMS: add shop in Page

chessbr:
- Add shop field in Page model
- Change admin to consider the current shop
- Fix unit tests to consider the shop in pages
- Update translation strings


shuup/simple_cms/models.py
```diff
@@ -37,9 +38,13 @@ def visible(self, dt=None):
         qs = self.filter(q)
         return qs
 
+    def for_shop(self, shop):
```
tulimaki: Since we don't offer shared pages should we just pass the shop to visible queryset filter as well. Now I need to know to separately (basically always) also call this for shop queryset filter.

chessbr:
Fixed!

tulimaki:
LGTM

PR #1240, opened 2018-05-04
Hide empty product list filters

Pikkupomo:
This commit solves the problem where category filters were shown on front even if there were no options available.

### DIFF
--- a/shuup/front/utils/sorts_and_filters.py
+++ b/shuup/front/utils/sorts_and_filters.py
@@ -12,6 +12,10 @@ import six
 from django import forms
 from django.conf import settings
 from django.db.models import Q
+from django.forms import (
+    ChoiceField, ModelChoiceField, ModelMultipleChoiceField,
+    MultipleChoiceField
+)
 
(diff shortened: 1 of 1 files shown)


shuup/front/utils/sorts_and_filters.py
```diff
@@ -177,8 +181,16 @@ def __init__(self, request, shop, category, *args, **kwargs):
...
         for extend_obj in _get_active_modifiers(shop, category):
             for field_key, field in extend_obj.get_fields(request, category) or []:
+                is_choice_field = isinstance(field, (
+                    ModelMultipleChoiceField, ModelChoiceField, ChoiceField, MultipleChoiceField
+                ))
+                has_choices = (is_choice_field and len(field.choices))
+
                 if field_key not in self.fields:
-                    self.fields[field_key] = field
+                    if is_choice_field and has_choices:
```
tulimaki: Isn't this just:
```
if not has_choices:
    continue
```

PR #1241, opened 2018-05-09
Admin: Media browser for multiple shops

tulimaki:
* Shop is assigned for file or folder while it is uploaded to Shuup.
* In case any shop is not assigned for the file or folder Shuup assumes the object is shared.
* Disable edit and remove for all shared files or folders.
* Hide Django admin URLs from workbench


shuup/admin/modules/media/views.py
```diff
@@ -64,18 +69,69 @@ def _filer_folder_to_json_dict(folder, children=None):
...
+
+
+def _is_folder_shared(folder):
+    if not settings.SHUUP_ENABLE_MULTIPLE_SHOPS:
+        return False
+
+    media_folder = MediaFolder.objects.filter(folder=folder).first()
+    if not media_folder:
+        return True
+    return bool(media_folder.shops.count() > 1)
```
chessbr: hum.. the folder is shared only when `count() > 1`? what if `shops.count() == 0` ?
also, should shop3 has access to the file which is set for shop1 and shop2? at least this will be a valid test in this case

shuup/admin/modules/media/views.py
```diff
@@ -64,18 +69,69 @@ def _filer_folder_to_json_dict(folder, children=None):
...
+
+
+def _get_folder_query_filter(shop):
+    return Q(Q(media_folder__isnull=True) | Q(media_folder__shops=shop))
```
chessbr: add `media_folder__shops__isnull=True` too ?

shuup/admin/modules/media/views.py
```diff
@@ -64,18 +69,69 @@ def _filer_folder_to_json_dict(folder, children=None):
...
+
+
+def _is_file_shared(file):
+    if not settings.SHUUP_ENABLE_MULTIPLE_SHOPS:
+        return False
+
+    media_file = MediaFile.objects.filter(file=file).first()
+    if not media_file:
+        return True
+    return bool(media_file.shops.count() > 1)
```
chessbr: same above. the commit description says that if the shops is null, it is shared so you should also consider when `count() == 0`

shuup/admin/modules/media/views.py
```diff
@@ -64,18 +69,69 @@ def _filer_folder_to_json_dict(folder, children=None):
...
+
+
+def _get_file_query(shop, folder=None):
+    queryset = File.objects.filter(Q(media_file__isnull=True) | Q(media_file__shops=shop))
```
chessbr: add `media_file__shops__isnull=True` too ?

shuup/admin/modules/media/views.py
```diff
@@ -130,33 +186,39 @@ def post(self, request, *args, **kwargs):
...
+        shop = get_shop(self.request)
+        root_folders = get_cached_trees(Folder._tree_manager.filter(_get_folder_query_filter(shop)))
         return JsonResponse({"rootFolder": _filer_folder_to_json_dict(None, root_folders)})
 
     def handle_post_new_folder(self, data):
+        shop = get_shop(self.request)
         parent_id = int(data.get("parent", 0))
         if parent_id > 0:
-            parent = Folder.objects.get(pk=parent_id)
+            parent = _get_folder_query(shop).get(pk=parent_id)
```
chessbr: this could raise `DoesNotExist`, make sure to handle it properly

shuup/admin/modules/media/views.py
```diff
@@ -130,33 +186,39 @@ def post(self, request, *args, **kwargs):
...
-                files = folder.files
+                folder = _get_folder_query(shop).get(pk=folder_id)
+                subfolders = folder.get_children().filter(_get_folder_query_filter(shop))
+                files = _get_file_query(shop, folder)
             else:
                 folder = None
-                subfolders = Folder.objects.filter(parent=None)
-                files = File.objects.filter(folder=None)
+                subfolders = _get_folder_query(shop).filter(parent=None)
+                files = _get_file_query(shop).filter(folder=None)
```
chessbr: check below the line `"error": "Folder does not exist"`
I guess you could make that a translated string right (do not use `ugettext_lazy`)? :sweat_smile: 
Maybe check around this files if there is any other error string to be translated. I guess they are returned to the users.

shuup/admin/modules/media/views.py
```diff
@@ -179,13 +241,14 @@ def handle_get_folder(self, data):
...
+        shop = get_shop(request)
         try:
             folder_id = int(request.POST.get("folder_id") or request.GET.get("folder_id") or 0)
             path = request.POST.get("path") or request.GET.get("path") or None
             if folder_id != 0:
-                folder = Folder.objects.get(pk=folder_id)
+                folder = _get_folder_query(shop).get(pk=folder_id)
             elif path:
-                folder = get_or_create_folder(path)
+                folder = get_or_create_folder(shop, path)
```
chessbr: change the super generic exception `Exception` below by `DoesNotExist` and translate string

shuup/admin/modules/media/views.py
```diff
@@ -210,36 +275,61 @@ def handle_upload(self):
         })
 
     def handle_post_rename_folder(self, data):
-        folder = Folder.objects.get(pk=data["id"])
+        shop = get_shop(self.request)
+        folder = _get_folder_query(shop).get(pk=data["id"])
```
chessbr: handle `DoesNotExist` ?

shuup/admin/modules/media/views.py
```diff
@@ -210,36 +275,61 @@ def handle_upload(self):
...
+            return JsonResponse({"success": False, "message": message})
+
         folder.name = data["name"]
         folder.save(update_fields=("name",))
         return JsonResponse({"success": True, "message": _("Folder renamed.")})
 
     def handle_post_delete_folder(self, data):
-        folder = Folder.objects.get(pk=data["id"])
+        shop = get_shop(self.request)
+        folder = _get_folder_query(shop).get(pk=data["id"])
```
chessbr: handle `DoesNotExist` ?

shuup/admin/modules/media/views.py
```diff
@@ -210,36 +275,61 @@ def handle_upload(self):
...
+            return JsonResponse({"success": False, "message": message})
+
         new_selected_folder_id = folder.parent_id
         message = delete_folder(folder)
         return JsonResponse({"success": True, "message": message, "newFolderId": new_selected_folder_id})
 
     def handle_post_rename_file(self, data):
-        file = File.objects.get(pk=data["id"])
+        shop = get_shop(self.request)
+        file = _get_file_query(shop).get(pk=data["id"])
```
chessbr: handle `DoesNotExist` ?

shuup/admin/modules/media/views.py
```diff
@@ -210,36 +275,61 @@ def handle_upload(self):
...
+            return JsonResponse({"success": False, "message": message})
+
         file.name = data["name"]
         file.save(update_fields=("name",))
         return JsonResponse({"success": True, "message": _("File renamed.")})
 
     def handle_post_delete_file(self, data):
-        file = File.objects.get(pk=data["id"])
+        shop = get_shop(self.request)
+        file = _get_file_query(shop).get(pk=data["id"])
```
chessbr: handle `DoesNotExist` ?

shuup/admin/modules/media/views.py
```diff
@@ -210,36 +275,61 @@ def handle_upload(self):
...
         try:
             file.delete()
         except IntegrityError as ie:
             raise Problem(str(ie))
         return JsonResponse({"success": True, "message": _("File deleted.")})
 
     def handle_post_move_file(self, data):
-        file = File.objects.get(pk=data["file_id"])
+        shop = get_shop(self.request)
+        file = _get_file_query(shop).get(pk=data["file_id"])
```
chessbr: handle `DoesNotExist` ?

shuup/core/models/_media.py
```diff
@@ -0,0 +1,36 @@
...
+
+@python_2_unicode_compatible
+class MediaFile(models.Model):
+    file = FilerFileField(related_name="media_file", verbose_name=_("file"), on_delete=models.CASCADE)
+    shops = models.ManyToManyField(
+        "shuup.Shop", related_name="+", verbose_name=_('shops'),
+        help_text=_("Select which shops you would like the files to be visible in."))
+
+    def __str__(self):  # pragma: no cover
+        return u'%s' % (self.file)
```
chessbr: you don't need need to add `u""` as you are importing `from __future__ import unicode_literals`

shuup/core/models/_media.py
```diff
@@ -0,0 +1,36 @@
...
+from django.utils.translation import ugettext_lazy as _
+from filer.fields.file import FilerFileField
+from filer.fields.folder import FilerFolderField
+
+
+@python_2_unicode_compatible
+class MediaFile(models.Model):
+    file = FilerFileField(related_name="media_file", verbose_name=_("file"), on_delete=models.CASCADE)
+    shops = models.ManyToManyField(
+        "shuup.Shop", related_name="+", verbose_name=_('shops'),
```
chessbr: I like very much using related_names. When you add `+` that will be hidden ans there is no way of selecting all shop media folders through a related queryset like `ShopProducts.objects.filter(shops__media_files__XXX)`. I know you can't see any usage right now but maybe we shouldn't kill this possibility. What do you think?

shuup/admin/modules/media/views.py
tulimaki: If these untranslated "not found" errors starts to happen then let's add the translation separately. These are catched at post already.

shuup/admin/modules/media/views.py
tulimaki: If these untranslated "not found" errors starts to happen then let's add the translation separately. These are catched at post already.

shuup/admin/modules/media/views.py
tulimaki: If these untranslated "not found" errors starts to happen then let's add the translation separately. These are catched at post already.

shuup/admin/modules/media/views.py
tulimaki: If these untranslated "not found" errors starts to happen then let's add the translation separately. These are catched at post already.

shuup/admin/modules/media/views.py
tulimaki: If these untranslated "not found" errors starts to happen then let's add the translation separately. These are catched at post already.

shuup/admin/modules/media/views.py
tulimaki: If these untranslated "not found" errors starts to happen then let's add the translation separately. These are catched at post already.

shuup/admin/modules/media/views.py
tulimaki: If these untranslated "not found" errors starts to happen then let's add the translation separately. These are catched at post already.

shuup/admin/modules/media/views.py
tulimaki: There is already try catch around these queries.

shuup/admin/modules/media/views.py
tulimaki: I think the "post" already handles this.

chessbr:
This happened when accessing `/admin/media/` directly in browser
```shell
Internal Server Error: /admin/media/
Traceback (most recent call last):
  File "/home/christian/dev/shuup-trees/venv/lib/python3.5/site-packages/django/core/handlers/base.py", line 149, in get_response
    response = self.process_exception_by_middleware(e, request)
  File "/home/christian/dev/shuup-trees/venv/lib/python3.5/site-packages/django/core/handlers/base.py", line 147, in get_response
    response = wrapped_callback(request, *callback_args, **callback_kwargs)
  File "/home/christian/dev/shuup/shuup/admin/utils/urls.py", line 95, in _wrapped_view
    return view_func(request, *args, **kwargs)
  File "/home/christian/dev/shuup-trees/venv/lib/python3.5/site-packages/django/views/generic/base.py", line 68, in view
    return self.dispatch(request, *args, **kwargs)
  File "/home/christian/dev/shuup-trees/venv/lib/python3.5/site-packages/django/views/generic/base.py", line 88, in dispatch
    return handler(request, *args, **kwargs)
  File "/home/christian/dev/shuup/shuup/admin/modules/media/views.py", line 166, in get
    return handler(request.GET)
  File "/home/christian/dev/shuup/shuup/admin/modules/media/views.py", line 194, in handle_get_folders
    root_folders = get_cached_trees(Folder._tree_manager.filter(_get_folder_query_filter(shop)))
  File "/home/christian/dev/shuup-trees/venv/lib/python3.5/site-packages/mptt/utils.py", line 240, in get_cached_trees
    _('Node %s not in depth-first order') % (type(queryset),)
ValueError: Node <class 'mptt.querysets.TreeQuerySet'> not in depth-first order
```

PR #1243, opened 2018-05-09
Admin: add JavaScript Mass Action type

chessbr:
It is possible to invoke JavaScript callbacks through mass actions.

Add new mass action function to return extra action info to be rendered in the Html option element as data attributes


tulimaki:
Looks like we need this. But @chessbr let's wait rest of this task before merging.

PR #1247, opened 2018-05-14
Core: make core basket command return added basket line id

chessbr:
This is useful to track the products added to the basket

Add option to pass additional arguments to add command like `extra`, `force_new_line` and `parent_line`


Pikkupomo:
Looks ok to me!

PR #1250, opened 2018-05-18
Checkout improvements

chessbr:
- Core: move basket validation errors for methods to another function
- Front: add shop option to require payment and shipping methods on checkout
- Front: improve feedback messages in checkout phases
- Update translation strings for Core and Front apps


tulimaki:
LGTM

PR #1253, opened 2018-05-24
Add General Data Protection Regulation (GDPR) support

chessbr:
- Core: add models and admin module to support GDPR

- [X] Create option to enable GDPR in Shuup
- [ ] Create GDPR Consent Documents feature
- [ ] Create Consent Management feature
- [ ] Create Data Management feature

# WIP


tulimaki:
Instead of putting the models in the core itself. Should this be just an app here with own models. Like notify, apps, admin, customer group pricing etc...

chessbr:
@tulimaki maybe. I started like that, but as other apps are going to access that to change their behaviour according to GDPR settings (or whether it is disabled) I put this in Shuup core. For example, now I am adding some property inside Simple CMS page, that would be a lot weird importing cross apps and not from Shuup core. I talked with @Pikkupomo about this today, we decided that GDPR is not an addon, but we didn't decided whether this is separated app.

Pikkupomo:
### Registration and Login

*  should have privacy policy accept instead of terms and conditions 
* should have a link to privacy policy

### User dashboard -> My Data
* download all my data is broken
* left menu disappears

shuup/gdpr/admin_module/toolbar.py
```diff
@@ -0,0 +1,46 @@
...
+from shuup.admin.toolbar import PostActionButton
+
+
+class AnonymizeContactToolbarButton(PostActionButton):
+    def __init__(self, object, **kwargs):
+        kwargs["icon"] = "fa fa-user-times"
+        kwargs["text"] = _("Anonymize")
+        kwargs["confirm"] = _(
+            "This action will replace all contact personal data with random values making it "
+            "impossible to be identified. The account will also be deactivated and the pending order will be canceled. "
```
Pikkupomo: `The account will also be deactivated and the pending order will be canceled` =>  `The account will also be deactivated and any pending order(s) will be canceled`

shuup/gdpr/admin_module/views.py
```diff
@@ -0,0 +1,116 @@
...
+
+
+class GDPRDownloadDataView(BaseContactView):
+    def post(self, request, *args, **kwargs):
+        contact = self.get_object()
+        contact.add_log_entry("User personal data download requested", kind=LogEntryKind.NOTE, user=self.request.user)
+        from shuup.gdpr.utils import get_all_contact_data
+        data = json.dumps(get_all_contact_data(contact))
+        response = HttpResponse(data, content_type="application/json")
+        response["Content-Disposition"] = "attachment; filename=user_data.json"
```
Pikkupomo: maybe have a timestamp on file name as well?

shuup/gdpr/resources.py
```diff
@@ -0,0 +1,53 @@
...
+
+    shop = get_shop(context["request"])
+    gdpr_settings = GDPRSettings.get_for_shop(shop)
+
+    # GDPR not enabled
+    if not gdpr_settings.enabled:
+        return
+
+    # consent already added
+    # TODO: Should we check for changes and ask the consent again here?
```
Pikkupomo: yes, this should be done, Lets do it on second phase

shuup/front/apps/registration/settings.py
```diff
@@ -9,3 +9,6 @@
 #: This corresponds to using the `default` or `simple`
 #: `django-registration` backends.
 SHUUP_REGISTRATION_REQUIRES_ACTIVATION = True
+
+#: The default registration form
+REGISTRATION_FORM = "shuup.front.apps.registration.forms.PersonRegistrationForm"
```
tulimaki: Where is this used? Shouldn't this setting live in registration app? Could the documentation for this setting be better?

shuup/gdpr/anonymizer.py
```diff
@@ -0,0 +1,162 @@
...
+        for order in person.orderer_orders.all():
+            self.anonymize_order(order)
+
+        for basket in person.orderer_baskets.all():
+            self.anonymize_object(basket)
+
+        for basket in person.orderer_core_baskets.all():
+            self.anonymize_object(basket)
+
+        self.anonymize_contact(person.pk)
```
tulimaki: Should we somewhere here also anonymize the user?

chessbr: I let the programmer decide where and how to anonymize the user, it is called in the view when anonymizing the customer

tulimaki: @Pikkupomo is this really correct? If the merchant now anonymize contact he needs to separately know whether there is user linked to the contact and anonymize that too? Isn't user linked to person contact anonymized every single time when the contact is anonymized?

chessbr: Check here @tulimaki  https://github.com/shuup/shuup/pull/1253/files#diff-4dba3d2446f86fab35b2b5abd2e57102R122

shuup/gdpr/admin_module/views.py
```diff
@@ -0,0 +1,129 @@
...
+            # check if there is any company related to the contact
+            # if so, anonymize it if he is the unique member
+            for company in contact.company_memberships.all():
+                if company.members.count() == 1:
+                    anonymizer.anonymize_company(company)
+                    # cancel company orders
+                    for order in Order.objects.incomplete().filter(customer=company):
+                        order.set_canceled()
+
+            if hasattr(contact, "user") and contact.user:
```
tulimaki: Yeah really think this if should be inside the anonymize person. Is this ever called without that. And now it seems we have anonymization logic in multiple places which is not optimal for code readability.

chessbr: ok, I will unity them in a single place

shuup/gdpr/admin_module/views.py
```diff
@@ -0,0 +1,129 @@
...
+            anonymizer = Anonymizer()
+
+            if isinstance(contact, PersonContact):
+                anonymizer.anonymize_person(contact)
+            elif isinstance(contact, CompanyContact):
+                anonymizer.anonymize_company(contact)
+
+            # check if there is any company related to the contact
+            # if so, anonymize it if he is the unique member
+            for company in contact.company_memberships.all():
```
tulimaki: Also this logic IMO should be inside the anonymizer itself.

shuup/gdpr/admin_module/views.py
```diff
@@ -0,0 +1,129 @@
...
+                if company.members.count() == 1:
+                    anonymizer.anonymize_company(company)
+                    # cancel company orders
+                    for order in Order.objects.incomplete().filter(customer=company):
+                        order.set_canceled()
+
+            if hasattr(contact, "user") and contact.user:
+                anonymizer.anonymize_user(contact.user)
+
+            for order in Order.objects.incomplete().filter(customer=contact):
```
tulimaki: Also this logic IMO should be inside the anonymizer itself.

PR #1263, opened 2018-05-31
Front: add custom method to cache MPTT child nodes

chessbr:
Fix already sent to https://github.com/django-mptt/django-mptt/pull/657


tulimaki:
Like discussed inside the team let's go with this until we need something more from that `get_cached_trees`.

PR #1270, opened 2018-06-04
GDPR Tuning

Pikkupomo:
ADMIN: Ensure sane defaults
FRONT: Ensure the link is visible on front


shuup/gdpr/admin_module/views.py
```diff
@@ -40,6 +41,18 @@ class GDPRView(SaveFormPartsMixin, FormPartsViewMixin, CreateOrUpdateView):
     base_form_part_classes = [GDPRBaseFormPart, GDPRCookieCategoryFormPart]
     success_url = reverse_lazy("shuup_admin:gdpr.settings")
 
+    def dispatch(self, request, *args, **kwargs):
```
chessbr: I am not super fan doing this here. Maybe you could move this to `models.py` in `GDPRSettings.get_for_shop()`:
```py
@classmethod
def get_for_shop(cls, shop):
    instance, created cls.objects.get_or_create(shop=shop)
    if created or not instance.cookie_banner_content:
         instance.set_default_content()
    return instance
```

Pikkupomo: yea, I was thinking something like this first but can't remember why I buried it :) Will change

PR #1272, opened 2018-06-05
Admin: Ensure the default value has been selected 

Pikkupomo:
Admin: Ensure the default value has been selected
* In Picotable, if there was no default given, the filter dropdowns had no content. This commit adds the default value there selected by default.

Tasks: Add link to the task comment author

### DIFF
--- a/shuup/tasks/locale/en/LC_MESSAGES/django.po
+++ b/shuup/tasks/locale/en/LC_MESSAGES/django.po
@@ -8,7 +8,7 @@ msgid ""
 msgstr ""
 "Project-Id-Version: PACKAGE VERSION\n"
 "Report-Msgid-Bugs-To: \n"
-"POT-Creation-Date: 2018-05-31 12:47+0000\n"
+"POT-Creation-Date: 2018-06-05 20:21+0000\n"
 "PO-Revision-Date: YEAR-MO-DA HO:MI+ZONE\n"
 "Last-Translator: FULL NAME <EMAIL@ADDRESS>\n"
 "Language-Team: LANGUAGE <LL@li.org>\n"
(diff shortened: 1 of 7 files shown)


shuup/tasks/templates/shuup/admin/tasks/task_edit.jinja
```diff
@@ -44,9 +44,10 @@
                                     </thead>
                                     <tbody>
                                         {% for comment in comments %}
+                                        {% set url = shuup_admin.model_url(comment.author) %}
                                         <tr>
                                             <td class="col-xs-2">{{ comment.created_on|datetime(format="short") }}</td>
-                                            <td class="col-xs-2">{{ comment.author }}</td>
+                                            <td class="col-xs-2"><a href="{{ url }}">{{ comment.author.id }}</a></td>
```
chessbr: is there any purpose to show the customer id instead of the name?

shuup/tasks/locale/en/LC_MESSAGES/django.po
```diff
@@ -17,156 +17,160 @@ msgstr ""
 "Content-Type: text/plain; charset=UTF-8\n"
 "Content-Transfer-Encoding: 8bit\n"
 
+#: admin_module/__init__.py:26 admin_module/__init__.py:53
```
chessbr: why all these file references?

Pikkupomo: wrong makemessages :)

shuup/tasks/templates/shuup/admin/tasks/task_edit.jinja
Pikkupomo: yes, register from front -> go to dashboard -> request data anonymization. 

TL;DR; not all contacts have name :)

I can change it so that if name exists, we show that and if not, we show id (like on the Picotable this PR)

PR #1274, opened 2018-06-06
Front: Update NPM dependencies

tulimaki:


chessbr:
* These files appears as changed.
image

* The Front is not working any more
image
image

Issue #1157, opened 2017-12-22
Nodejs 8 support


tulimaki:
Now the Shuup has few issue with 7+.

tulimaki:
Yeah it seems this got fixed while updating those NPM packages. Tested with Node 7 and 8.

PR #1276, opened 2018-06-07
Update NPM packages for front, admin and xtheme

tulimaki:


chessbr:
Fixes https://github.com/shuup/shuup/issues/1157

PR #1280, opened 2018-06-08
Front: Improve company registration

tulimaki:
* Put all settings around company registration together
* Make all company registration settings per shop
* Enable option for tax number validation
* Add option to order anonymously as a person


shuup/front/admin_module/forms.py
```diff
@@ -0,0 +1,62 @@
...
+                ConfigurationItem.objects.get(shop=self.request.shop, key=key).delete()
+            except ConfigurationItem.DoesNotExist:
+                continue
+
+        for key, value in six.iteritems(form.cleaned_data):
+            if isinstance(value, Enum):
+                value = value.value
+            configuration.set(self.request.shop, key, value)
+
+        return
```
chessbr: why this return so lonely here? should it return something?

shuup/front/apps/registration/templates/shuup/registration/register.jinja
```diff
@@ -18,7 +18,7 @@
                     {% csrf_token %}
                     {% for field in form.hidden_fields() %}{{ field|safe }}{% endfor %}
                     {% for field in form.visible_fields() %}{{ render_field(field) }}{% endfor %}
-                    {% if get_shop_configuration("allow_company_registration") %}
+                    {% if shuup.general.is_company_registration_allowed() %}
```
chessbr: missing the request param

tulimaki: The request is optional for the thing. Just what happened there with customer information was that for some reason the context didn't have the request. Since we pass request for that macro already I thought there might be a reason for all this and added option to pass the request manually from the places where the context is weird.

shuup/front/template_helpers/general.py
```diff
@@ -318,3 +319,9 @@ def get_shop_language_choices(context):
 @contextfunction
 def is_shop_admin(context):
     return is_admin_user(context["request"])
+
+
+@contextfunction
+def is_company_registration_allowed(context, request=None):
+    current_request = request or context["request"]  # From macros it doesn't seem to always pass context correctly
```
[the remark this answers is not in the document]
tulimaki: If there is context without request. But I guess it is ok to fail in these situations...

Issue #1155, opened 2017-12-22
Django 1.11


tulimaki:
Shuup should support Django 1.11.

chessbr:
Done in https://github.com/shuup/shuup/pull/1282

PR #1282, opened 2018-06-11
Add Django 1.11 support

tulimaki:
* [x] Add support for Jinja2 2.10. Moved to #1285
* [x] Decide something for `django-filer 1.3.2 has requirement django-mptt<0.9,>=0.6, but you'll have django-mptt 0.9.0 which is incompatible.` Update: Based on [this](https://github.com/divio/django-filer/commit/302555214e4780a48f254f1571727abe37a30615) it looks like it is safe to allow mptt +0.9 before the django-filer does it's next release.
* [x]  Check what we can do for the test coverage. Now 0.07% decrease sounds like a lot.
* [x] Make sure things still works with Django 1.9.x and 1.8.x
* [x] Enable one Travis job for each Django 1.9.x and 1.8.x
* [x] Look into Django 2.0 support and report #1289 
* [x] Decide whether we enable browser tests with Travis now or later. Update: The browser tests is not that solid. They go through locally pretty well but like in past the Travis adds some extra variables to this. There is at least one push where all tests passed without restarting any jobs. Locally it fails sometimes but even locally it is possible to see this randomness which I guess shouldn't have to be part of this task. Update 2: Yes let's disable for now and let's continue with better time.
* [x] Decide if we should only test old Django with py3. The amount of Travis jobs doesn't look optimal. Maybe we only run all for release and then on post release commit disable some of these extra jobs? Edit: I think we run all of those. With 5 concurrent jobs should be too bad.
* [x] Check what to do to make Codecov happy.

Refs #1155


tulimaki:
@suutari-ai can you take quick look around this before we consider merging this? As a last change I have planned to disable those browser tests for Travis (we try to find better spot to make those more consistent).

.travis.yml
```diff
@@ -19,25 +19,31 @@ matrix:
...
+  - mkdir geckodriver
+  - tar -xzf geckodriver*.tar.gz -C geckodriver
+  - export PATH=$PATH:$PWD/geckodriver
+  - "nvm install 6"
+  - "nvm use 6"
+  - "export DISPLAY=:99.0"
+  - "sh -e /etc/init.d/xvfb start"
+  - "npm install -g selenium-standalone@4.9.1"
+  - "selenium-standalone install"
+  - "selenium-standalone start &"
```
suutari-ai: This should probably moved to a separate script file, like `travis_install_selenium` and the command here could then be just `./travis_install_selenium`.  Running of those should also be done only for the envs that need the selenium stuff, e.g. the listed 4 browser test envs. You can have before_install on per env setting in the Travis build matrix to achieve that:
```yaml
matrix:
  include:
    - ...
    - {python: 3.6, env: TOXENV=py36-django111-nomigrations}
    - python: "3.5"
      env: TOXENV=py35-django19-nomigrations-nocoverage-browser-admin
      before_install: ./travis_install_selenium
    - ...
```

codecov.yaml
```diff
@@ -0,0 +1,2 @@
+ignore:
+  - "shuup/*/management/commands/*"  # ignore management commands
```
suutari-ai: This is cheating. I don't like it. They are either tested or they're not. Making them disappear from the coverage reports just causes confusion.

doc/changelog.rst
```diff
@@ -9,6 +9,16 @@ Unrealeased
...
 
+Core
+~~~~
+
+- Add support for Django 1.11 and for now on Shuup is installed over Django 1.11
+by default. There is still unofficial support for Django 1.8 and 1.9 which
+means that after installing Shuup you can downgrade your Django and Django
+polymorphic versions. We will also still run tests for these old versions
+while adding new features to Shuup. See tox.ini for information about
+downgrades required to run Shuup with old Django.
```
suutari-ai: Shouldn't this be in "General changes" section rather than core? It's concerning all of the Shuup apps, isn't it?

setup.py
```diff
@@ -86,38 +86,38 @@
...
+    'django-enumfields>=0.9,<0.10',
+    'django-filer>=1.3,<2',
+    'django-filter>=1.1,<1.2',
+    'django-jinja>=2.3,<3',
+    'django-mptt>=0.9,<0.10',
+    'django-parler>=1.9,<2',
+    'django-parler-rest>=1.4.2,<2',
+    'django-polymorphic>=0.9.2,<2.1.0',  # For Django 1.8 and 1.9 use version lower 1
+    'django-registration-redux>=1.5,<1.9',
+    'django-timezone-field>=2.1,<3',
```
suutari-ai: Are all these requirement upgrades really needed?

I'd prefer not to change to lowest required version unless they're really a must. It's not Shuup's concern if all those required libraries work together. Shuup should only tell which are the lowest (and highest) versions it needs and then the project installing Shuup should take care that compatible versions are installed. (Usually that is the newest available versions at that point.)

setup.py
```diff
@@ -86,38 +86,38 @@
...
     'djangorestframework-jwt>=1.9,<2',
-    'django_rest_swagger==2.1.0',
+    'django_rest_swagger>=2.1.0,<2.2',
     'factory-boy==2.7.0',
     'fake-factory>=0.5.0,<0.5.4',
-    'Jinja2>=2.8,<2.9',
-    'jsonfield>=1.0,<2',
-    'keyring==10.2',
-    'keyrings.alt==2.0',
+    'Jinja2==2.8.1',
```
suutari-ai: Why is jinja pinned to exact version? Doesn't seem right...

setup.py
```diff
@@ -86,38 +86,38 @@
...
+    'jsonfield>=2.0.2,<3',
+    'keyring==11.0.0',
+    'keyrings.alt==2.4',
```
suutari-ai: Neither these should be pinned to exact version but rather to a minimum version (jsonfield>=1, keyring>=10, keyrings.alt>=2)

setup.py
```diff
@@ -86,38 +86,38 @@
...
     'Markdown>=2.6,<3',
-    'openpyxl==2.3.5',
-    'python-dateutil==2.6.0',
+    'openpyxl==2.5.0',
+    'python-dateutil==2.6.1',
```
suutari-ai: openpyxl>=2.3 and python-dateutil>=2.5

setup.py
```diff
@@ -86,38 +86,38 @@
...
     'pytoml>=0.1.0,<0.2',
-    'pytz>=2015.4',
+    'pytz>=2018.3',
```
suutari-ai: No need to update this one, not a Shuup's concern to force anyone to have new timezone data if they don't want to. (Shuup doesn't break even if the timezone data is outdated.)

setup.py
```diff
@@ -86,38 +86,38 @@
...
     'requests>=2.7,<3',
     'secretstorage==2.3.1',
     'six>=1.9,<2',
     'unicodecsv==0.14.1',
-    'xlrd==1.0.0'
+    'xlrd==1.1.0'
```
suutari-ai: xlrd>=1

shuup/core/models/_base.py
```diff
@@ -91,7 +92,11 @@ def get_inherited_managers(self, attrs):
 class PolymorphicTranslatableShuupModel(six.with_metaclass(
         PolyTransModelBase,
         PolymorphicShuupModel, TranslatableShuupModel)):
-    _default_manager = _PolyTransManager()
+
+    if version.get_docs_version() in ["1.8", "1.9"]:
```
suutari-ai: Rather just `import django`  and do `if django.VERSION < (1, 10)`.

shuup/core/models/_service_base.py
```diff
@@ -454,8 +455,13 @@ def get_delivery_time(self, service, source):
         return None
 
 
-class TranslatableServiceBehaviorComponent(six.with_metaclass(
-        PolyTransModelBase,
-        ServiceBehaviorComponent, TranslatableShuupModel)):
+if version.get_docs_version() in ["1.8", "1.9"]:
+    META_CLASS = six.with_metaclass(PolyTransModelBase, ServiceBehaviorComponent, TranslatableShuupModel)
+else:
+    META_CLASS = six.with_metaclass(PolyTransModelBase, ServiceBehaviorComponent, PolymorphicTranslatableShuupModel)
```
suutari-ai: Rather just `import django`  and do `if django.VERSION < (1, 10)`.

shuup/core/models/_service_base.py
```diff
@@ -454,8 +455,13 @@ def get_delivery_time(self, service, source):
         return None
 
 
-class TranslatableServiceBehaviorComponent(six.with_metaclass(
-        PolyTransModelBase,
-        ServiceBehaviorComponent, TranslatableShuupModel)):
+if version.get_docs_version() in ["1.8", "1.9"]:
```
suutari-ai: Rather just `import django`  and do `if django.VERSION < (1, 10)`.

shuup/utils/migrations.py
```diff
@@ -0,0 +1,19 @@
...
+#
+# This source code is licensed under the OSL-3.0 license found in the
+# LICENSE file in the root directory of this source tree.
+import django.db.models
+from django.utils import version
+
+
+def get_managers_for_migration():
+    docs_version = version.get_docs_version()
+    if docs_version == ["1.8", "1.11"]:
```
suutari-ai: Rather just `import django`  and do `if django.VERSION < (1, 10)`.

shuup_tests/admin/test_service_providers.py
```diff
@@ -106,8 +110,11 @@ def test_invalid_service_provider_type(rf, admin_user):
             "selected": bool(option.get("selected")),
             "value": option["value"],
         })
-    assert options[0]["selected"] is True
-    assert [x["selected"] for x in options[1:]] == [False, False]
+
+    if version.get_docs_version() in ["1.8", "1.9"]:
```
suutari-ai: Rather just `import django`  and do `if django.VERSION < (1, 10)`.

shuup_workbench/settings/utils.py
```diff
@@ -8,3 +9,33 @@ def __contains__(self, item):
 
     def __getitem__(self, item):
         return "notmigrations"
+
+
+def get_disabled_migrations():
+    if version.get_docs_version() in ["1.8", "1.9"]:
```
suutari-ai: Rather just `import django`  and do `if django.VERSION < (1, 10)`.

shuup/admin/forms/widgets.py
```diff
@@ -26,6 +26,16 @@
     Contact, PersonContact, Product, ProductMode, ShopProduct
 )
 
+if version.get_docs_version() in ["1.8", "1.9"]:
```
suutari-ai: Rather just `import django`  and do `if django.VERSION < (1, 10)`.

shuup/campaigns/migrations/0013_hourconfition_fix.py
```diff
@@ -0,0 +1,28 @@
...
+
+    dependencies = [
+        ('campaigns', '0012_basket_campaign_undiscounted'),
+    ]
+
+    operations = [
+        migrations.AlterField(
+            model_name='hourbasketcondition',
+            name='days',
+            field=models.CharField(verbose_name='days', max_length=255, validators=[django.core.validators.RegexValidator(re.compile('^[\\d,]+\\Z' if django.utils.version.get_docs_version() == "1.8" else '^\\d+(?:\\,\\d+)*\\Z', 32), code='invalid', message='Enter only digits separated by commas.')]),
```
suutari-ai: You changed this migration manually, did you?

How about wrapping the code so that it can be read more easily? :)

Also, rather than using the `get_docs_version` just `import django`  and do `if django.VERSION < (1, 10)`.

shuup/campaigns/migrations/0013_hourconfition_fix.py
```diff
@@ -0,0 +1,28 @@
...
+        ),
+        migrations.AlterField(
+            model_name='hourcondition',
+            name='days',
+            field=models.CharField(verbose_name='days', max_length=255, validators=[django.core.validators.RegexValidator(re.compile('^[\\d,]+\\Z' if django.utils.version.get_docs_version() == "1.8" else '^\\d+(?:\\,\\d+)*\\Z', 32), code='invalid', message='Enter only digits separated by commas.')]),
```
suutari-ai: same here

shuup/core/models/_service_base.py
suutari-ai: This is not a meta class you're changing here, but the base class, so the name of this variable is misleading.

I think you should rather do something like this:
```python
import django


_translatable_model = (
    PolymorphicTranslatableShuupModel if django.VERSION >= (1, 10)
    else TranslatableShuupModel)


 class TranslatableServiceBehaviorComponent(six.with_metaclass(
        PolyTransModelBase,
        ServiceBehaviorComponent, _translatable_model)):
```

suutari-ai:
Good work guys!

There's quite a lot to review, but tried to skim it through.  There was some minor stuff with the Travis and requirements, plus something on how to check Django version in code. See my comments above.

tulimaki:
@suutari-ai Thanks a lot. Most of those sounds like fix and should make this PR better.

doc/changelog.rst
suutari-ai: Now OK

travis_install_selenium
```diff
@@ -0,0 +1,12 @@
+
```
suutari-ai: This script needs a shebang (`!/bin/sh`) and must be executable (`chmod +x travis_install_selenium` Though I don't know if it already is, since I'm not sure how GitHub will report file modes...)

setup.py
tulimaki: Yeah shouldn't be needed but can't fix all the issues as part of this task. @suutari-ai it's locked since that is pretty much only version that works with our front (see https://github.com/shuup/shuup/issues/1285).

shuup/core/models/_service_base.py
```diff
@@ -454,8 +455,13 @@ def get_delivery_time(self, service, source):
         return None
 
 
+_translatable_model = (
+    PolymorphicTranslatableShuupModel if django.VERSION > (1, 11)
+    else TranslatableShuupModel)
```
suutari-ai: should probably be `>= (1, 11)`, right?

shuup/core/models/_service_base.py
```diff
@@ -454,8 +455,13 @@ def get_delivery_time(self, service, source):
         return None
 
 
+_translatable_model = (
+    PolymorphicTranslatableShuupModel if django.VERSION > (1, 11)
```
suutari-ai: should probably be `>= (1, 11)`, right?

suutari-ai:
Ship it! :shipit:

tulimaki:
**Do not merge yet. I will be making one release and updating Transifex before this. So by waiting a sec we can avoid Git battles.**

PR #1291, opened 2018-06-18
GDPR improvements + Add Revision

Pikkupomo:


shuup/simple_cms/admin_module/views.py
```diff
@@ -128,6 +113,9 @@ def clean_parent(self):
     def save(self, commit=True):
         if not hasattr(self.instance, "shop") or not self.instance.shop:
             self.instance.shop = get_shop(self.request)
+        if self.cleaned_data.get("page_type") == PageType.REVISIONED:
```
tulimaki: Could this be done in model level instead since likely not something the admin module need to know about?

shuup/gdpr/models.py
```diff
@@ -5,14 +5,16 @@
 #
 # This source code is licensed under the OSL-3.0 license found in the
 # LICENSE file in the root directory of this source tree.
-from __future__ import unicode_literals
+from __future__ import unicode_literals, with_statement
```
tulimaki: What are these `with_statement`?

shuup/gdpr/resources.py
```diff
@@ -15,10 +15,28 @@
...
+
+    view_name = getattr(view_class, "__name__", "")
+    if view_name == "EditorView":
+        return False
+
+    return True
+
+
+def add_gdpr_consent_resources(context, content):
+    # TODO: should we ignore admin urls?
```
tulimaki: Not sure. Maybe remove this in case this is what you already. Commit message could be better.

shuup/gdpr/models.py
chessbr: To enable `with reversion.create_revision():` in py2

shuup/gdpr/resources.py
Pikkupomo: it's a leftover from old code (see line 19), wasn't planning on doing anything regarding this

tulimaki: Ok based on the commit message and line 26 I expected this to actually ignore the admin URLs. Improve the commit message then.

shuup/simple_cms/admin_module/views.py
Pikkupomo: I thought it first, but I didn't want to introduce `def save` into the `Page` Model. Now that I think of it, it might make sense as it dries out the code quite a bit. Thanks

shuup/gdpr/resources.py
Pikkupomo: haha true :)

PR #1301, opened 2018-06-21
GDPR: Try to fix migrations

Pikkupomo:
If migrations were made with data in database, it caused havoc. This Commit should fix the issue.


tulimaki:
Replaced by https://github.com/shuup/shuup/pull/1302

PR #1303, opened 2018-06-21
Fix SimpleCMS children test

chessbr:
A front unit test was breaking subsequent tests by not reverting the timezone set


Pikkupomo:
LGTM

PR #1305, opened 2018-06-21
Core: enable ForeignKey constraint for SQLite database backends

chessbr:
This prevents bugs when developing and prevents errors when in production

### DIFF
--- a/shuup/testing/__init__.py
+++ b/shuup/testing/__init__.py
@@ -8,6 +8,13 @@
 from shuup.apps import AppConfig
 
 
+def activate_sqlite_fk_constraint(sender, connection, **kwargs):
+    """Enable integrity constraint with SQLite."""
+    if connection.vendor == 'sqlite':
+        cursor = connection.cursor()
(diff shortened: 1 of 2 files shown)


Pikkupomo:
Can you rebase this

tulimaki:
Not sure. Is the core ready really right spot for this? Should this be behind debug mode or something? Sounds like a dev thing that is now checked and added too every time app is loaded no matter what.

Or maybe add this for tests only and then rest should be devs responsibility to add if he believes he needs it.

chessbr:
Right @tulimaki I moved that to tests only. I put it on core because of these points:

* It checks whether it is SQLite, so if it is sqlite, you are probably running in dev mode. And even if you are running in a bizarre sqlite production, you want this foreign key constraint to be enabled to prevent objects with wrong values
* It will be enabled for all Shuup apps and for any custom app, by default, the same way other database backends does (like mysql)

tulimaki:
Travis in bad mood. Should be fine.

Issue #1189, opened 2018-01-25
Shuup notify has two migrations with the same ID


finmaster19:
There are 2 migrations with the ID 0002 in the [notify app](https://github.com/shuup/shuup/tree/master/shuup/notify/migrations). 
This may cause problems with some shuup shops? At least initializing
new database still works well..

chessbr:
I guess that is not a bug. The name actually doesn't matter at all, just the migration tree. That name (0002_..) was auto-generated from `makemigrations`, probably because the last migration was `0001_squashed_0003_alter_names.py`, thus, the next is 0002 by checking the prefix. IMO, not a problem. Are you facing issues @finmaster19 ?

finmaster19:
These squashed migrations causes little bit problems in shuup if you haven't applied all migrations before the squashed ones. But as a workaround you can do this:

Comment out all other shuup modules from the INSTALLED_APPS but shuup.core and maybe shuup.admin and then migrate shuup to the latest migration to get the 0001_squashed_0039 applied. After that the other apps can be added back to INSTALLED_APPS and the whole migration should run fine. This might needed to be done for every app one at the time that has problems with squashed migrations.

I noticed that when I had a shuup project which was like a year old, I had to move three newest migrations plus the squashed one away from notify app and then run `python manage.py migrate`. This way I was able to run the `0003_alter_names.py` migration. If I didn't move those three migrations out of the folder and I had only 2 first ones applied, I received `CommandError: Conflicting migrations detected; multiple leaf nodes in the migration graph: (0005_shop_not_null, 0003_alter_names in shuup_notify).
To fix them run 'python manage.py makemigrations --merge'`

finmaster19:
I did take sql dump of my old test shuup db before I started running migrations if someone is interested to try what kind of error messages these squashed migrations might cause if you haven't applied all the migrations before squashed ones.

PR #1316, opened 2018-06-25
Notify: Fix migrations

tulimaki:
Shop for notifications and scripts was added in `0002_shop_scripts`
which was not included in squashed migrations. This caused weird
situation when updating from older projects which didn't yet
have shop migrated.

Refs #1189


tulimaki:
@chessbr you don't have to wait codecov here. But double check that all good with this. It's important we have all migrations working normally before we consider squashing these.

doc/changelog.rst
```diff
@@ -21,6 +21,12 @@ Importer
 
 - Require only shop change permission to execute data imports
 
+Notify
+~~~~~~
+
+- Fix migrations. This would old projects to fake migrations until
```
chessbr: is this message correct? Looks strange, maybe `This requires old projects to fake migrations until notify.0006_shop_not_null` ?

PR #1319, opened 2018-06-26
Front: move function to return available languages to utils

chessbr:
This enable 3rd party modules to use the same util without passing jinja render context

Refs POS-2438

### DIFF
--- a/shuup/front/template_helpers/general.py
+++ b/shuup/front/template_helpers/general.py
@@ -10,7 +10,7 @@ from collections import defaultdict
 import six
 from django.conf import settings
 from django.core.paginator import Paginator
-from django.utils.translation import get_language, get_language_info, ugettext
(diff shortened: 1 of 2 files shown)


shuup/front/template_helpers/general.py
```diff
@@ -307,13 +308,7 @@ def _get_page_range(current_page, num_pages, range_gap=5):
 
 @contextfunction
 def get_shop_language_choices(context):
```
Pikkupomo: The naming is strange as this uses `settings` as a source for languages and not `shop` (`request.shop`). I can see this is not something you added just now, but should we consider changing this? `get_default_language_choices` and when we have mechanics in place for shop languages, add that separately?

chessbr:
Replaced by https://github.com/shuup/shuup/pull/1319

PR #1324, opened 2018-06-27
Simple CMS: Add option to provide admin form parts

tulimaki:


shuup/simple_cms/admin_module/views.py
```diff
@@ -159,6 +181,15 @@ def save_form(self, form):
     def get_queryset(self):
         return super(PageEditView, self).get_queryset().for_shop(get_shop(self.request)).filter(deleted=False)
 
+    def get_toolbar(self):
```
chessbr: This method has the same implementation of its super, hence, it's not needed here. If you want to add provides entry point here this could makes sense, but the logic for that is missing.

shuup/simple_cms/admin_module/views.py
```diff
@@ -159,6 +181,15 @@ def save_form(self, form):
...
+        save_form_id = self.get_save_form_id()
+        if save_form_id:
+            return get_default_edit_toolbar(self, save_form_id)
+
+    @atomic
+    def form_valid(self, form):
```
chessbr: the `save_form` method above is no longer needed

Issue #1323, opened 2018-06-27
Fix Parler integration


chessbr:
The following test fails when model instances are cached:

```py
@pytest.mark.django_db

def test_product_page_languages_cache():
    activate("en")
    shop = get_default_shop()
    variable_name = "Color"
    parent = create_product("test-sku-1", shop=shop)
    variation_variable = ProductVariationVariable.objects.create(product=parent, identifier="color", name=variable_name)
    ProductVariationVariableValue.objects.create(variable=variation_variable, identifier="red", value="Red")
    ProductVariationVariableValue.objects.create(variable=variation_variable, identifier="blue", value="Blue")
    combinations = list(parent.get_all_available_combinations())
    assert len(combinations) == 2
    for combo in combinations:
        assert not combo["result_product_pk"]
        child = create_product("xyz-%s" % combo["sku_part"], shop=shop, supplier=get_default_supplier(), default_price=20)
        child.link_to_parent(parent, combination_hash=combo["hash"])

    combinations = list(parent.get_all_available_combinations())
    assert len(combinations) == 2
    parent.refresh_from_db()
    assert parent.is_variation_parent()
    client = SmartClient()
    product_url = reverse('shuup:product', kwargs=dict(pk=parent.pk, slug=parent.slug))

    # visit the page in EN language
    response, soup = client.response_and_soup(product_url)
    assert get_language() == "en"
    assert "Select {}".format(variable_name) in response.content.decode("utf-8")
    assert soup.find("option", text="Red")
    assert soup.find("option", text="Blue")

    # visit the page in language (FI)
    # the translations should fallback to EN
    client.post(reverse("shuup:set-language"), data={"language": "fi"})
    response, soup = client.response_and_soup(product_url)
    assert get_language() == "fi"
    assert "Select {}".format(variable_name) in response.content.decode("utf-8")
    assert soup.find("option", text="Red")
    assert soup.find("option", text="Blue")

    # clear the context cache
    from shuup.core import cache
    cache.clear()

    # the first request will be in FI language, the product has no translation for that
    # so the fallback should be returned
    for time in range(2):
        response, soup = client.response_and_soup(product_url)
        assert get_language() == "fi"
        assert "Select {}".format(variable_name) in response.content.decode("utf-8")
        assert soup.find("option", text="Red")       # <=== Fails here
        assert soup.find("option", text="Blue")
```

The fix is here: https://github.com/django-parler/django-parler/pull/228
We should just bump Parler after that PR is merged

[...]

PR #1325, opened 2018-06-27
General: cache object ids instead of the model instances

chessbr:
Parler has a bug that fails to load translations after they are fetched from cache

Check the PR that fixes the bug in Parler https://github.com/django-parler/django-parler/pull/228


tulimaki:
Related to https://github.com/shuup/shuup/issues/1323 and likely should be changed back when the parler gets the fixes merged.

Also the front browser tests are cool (see. https://github.com/shuup/shuup/pull/1327)

PR #1326, opened 2018-06-28
Use hash of string in section name in case non latin symbols

soulilya:
When using translations with non latin charset (in my case russian) kebabCase in js return empty string. All sections that not have defined id in jinja templates have the same url like "-section" and become not active. So use hash function to define some name

### DIFF
--- a/shuup/admin/static_src/base/js/side-nav.js
+++ b/shuup/admin/static_src/base/js/side-nav.js
@@ -14,6 +14,18 @@ $(function() {
     function getSectionNavigatorModule($navigateeForm) {
         const $sections = $navigateeForm.find(".content-block");
 
+        const hashCode = function(s) {
+			var h = 0, l = s.length, i = 0;
+			if ( l > 0 )
+			  while (i < l)
+				h = (h << 5) - h + s.charCodeAt(i++) | 0;
+			if ( h < 0 )
+			{
+				h = 0 - h;
+			}
+			return h;
+		  };
+
         const navigationListItems = _.compact(_.map($sections, function(section) {
             const $section = $(section);
             const $blockTitle = $section.find(".block-title");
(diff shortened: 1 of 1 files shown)


soulilya:
You choose. I cant for now say what better solution for project. Becouse of code. In some places, for example, DOM id names pattern as "colon-1","colon-2", "colon-3" ....
In some places id jinja macros is set, but moustly not. Im comfused. 

I can fix that:

1. Fix on view context. In this case i can add id varible to jinja templates that uses original title name of section without translates and without spaces and uppercaces and id in macro doesnt need enymore.
2. Fix macro id. In this case i can write script that finds all macros and insert id from original title name of section without translates and without spaces and uppercaces.
3. Fix name id with js. For readable names may use this lib  https://github.com/andyhu/transliteration  

Code that you provide for prevent numbers < 0; Example: -95152, -45554, -5755757. etc


>Четверг, 28 июня 2018, 19:03 +03:00 от Panu Tulimäki <notifications@github.com>:
>
>@tulimaki requested changes on this pull request.
>I guess  this is the reference for the method.  https://gist.github.com/hyamamoto/fd435505d29ebfa3d9716fd2be8d42f0
>Add this reference to the code too. Fix indentations and add comment why this is required at the end:
>if (h<0) {
>  h = 0 - h;
>}
>
>Also almost sound like sections names should be more readable, but if this makes the admin left side menu better for non latin charset then I think we can do this. Hash is better choice than empty string here anyways.
>—
>You are receiving this because you authored the thread.
>Reply to this email directly,  view it on GitHub , or  mute the thread .

______________________
Разаработка и продвижение сайтов
https://www.facebook.com/sitebuypodolsk/

tulimaki:
Ok. I think we can go with this for now. If there is some better solutions showing up then we double check this decision.

Fix the styles so the Travis passes and I think we can merge this.

tulimaki:
Replaced with https://github.com/shuup/shuup/pull/1332. Fixed the indentations while enabling Russian translations. Thanks @soulilya.

PR #1328, opened 2018-06-28
Importer:  enable importer modules to provide example files

chessbr:
Example files are available for download in admin view.
The example file helps users to understand what should be the expected file structure

Refs POS-2439

image


tulimaki:
🚢 🇮🇹

PR #1335, opened 2018-07-01
Build static resources using Parcel

chessbr:
Make static source build standard. Use [ParcelJS](https://parceljs.org/) to make build simpler.

- [X] Build Notify
- [X] Build Regions
- [X] Build GDPR
- [x] Build Xtheme
- [x] Build Front
- [x] Build Classic Gray Theme
- [x] Check what to do with `generated_resources.txt`
- [x] Fix unit tests

:warning: Admin app is not included because of the coming overhaul


shuup/gdpr/package.json
```diff
@@ -2,31 +2,22 @@
   "name": "shuup.gdpr",
   "version": "1.0.0",
   "description": "Enables General Data Protection Regulation (GDPR) in Shuup",
-  "main": "webpack.config.js",
+  "main": "",
   "scripts": {
-    "build": "npm i && webpack --env.dist",
-    "watch": "webpack --watch --env.dev --colors --progress"
+    "build:js": "./node_modules/.bin/parcel build -d static/ -o shuup-gdpr static_src/gdpr.js",
```
suutari-ai: You shouldn't need to have `./node_modules/.bin/` prefix in package.json scripts, since that should already be in PATH while running the scripts.

shuup/gdpr/package.json
```diff
@@ -2,31 +2,22 @@
...
+    "build:styles": "./node_modules/.bin/parcel build --no-source-maps -d static/ -o shuup-gdpr static_src/gdpr.less",
+    "watch:js": "./node_modules/.bin/parcel watch --no-hmr -d static/ -o shuup-gdpr static_src/gdpr.js",
+    "watch:styles": "./node_modules/.bin/parcel watch --no-hmr --no-source-maps -d static/ -o shuup-gdpr static_src/gdpr.less",
```
suutari-ai: Don't separate the watch commands since you should be able to watch both js and styles simultaneously.

shuup/gdpr/package.json
chessbr: I guess we are going to install parcel globally to build static source, that could be better in the end

chessbr: @suutari-ai I guess you are mistaken. Try to install parcel locally and try running that `npm run parcel`. At least here with npm 6.1 it doesn't work at all.

shuup/gdpr/package.json
chessbr: I will probably try using https://github.com/mysticatea/npm-run-all or https://github.com/kimmobrunfeldt/concurrently

shuup/gdpr/package.json
chessbr: Now I get it what you meant. When the command is inside a npm script it doesn't need the node_modules/.bin prefix. Yes, indeed it works properly. Just running that through the command line directly (not installed globally) doesn't work.

suutari-ai: Exactly.

shuup/gdpr/package.json
suutari-ai: I would also be a good idea to move the parcel commands to a separate scripts, since it's a PITA to have a kilometer long run script in the package.json.

chessbr: Hum, ok I will think in a some sort of script. I just want to make things easy to find. Currently to check how Admin is built, you have to open thousand gulp files. That is not so good, specially for new/noob users. We could have some script that uses Parcel API but I am not happy to create one just to replace this long line. A makefile will bee good as we could make things more organized but that will depend of having it installed so also not a good solution. So either a shell script or a node script. @suutari-ai thoughts?

chessbr: Maybe some variables to reduce duplicated values: http://www.marcusoft.net/2015/08/npm-scripting-configs-and-arguments.html#npm-configuration

shuup_setup_utils/resource_building.py
```diff
@@ -70,7 +71,8 @@ def _find_package_json_dirs(self):
         items = excludes.walk_excl(self.root_directory)
         for (dirpath, dirnames, filenames) in items:
             if 'package.json' in filenames:
-                if 'generated_resources.txt' in filenames:
+                package = json.load(open(os.path.join(dirpath, "package.json")))
```
chessbr: still check for `'generated_resources.txt' in filenames` for backwards compatibility

shuup_tests/browser/front/test_category_view.py
```diff
@@ -286,6 +286,7 @@ def show_sorts_for_the_category_only(browser, category):
 def sort_category_products_test(browser, category):
     # Lowest price first
     click_element(browser, "button[data-id='id_sort']")
+    # WARNING: data-original-index was removed after bootstrap-select 1.6.3
```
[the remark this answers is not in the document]
chessbr: Yes, just warning for the one who bump this bootstrap-select. He will need to fix these unit tests.

doc/changelog.rst
```diff
@@ -9,6 +9,12 @@ Unrealeased
   When releasing next version, the "Unreleased" header will be replaced
   with appropriate version header and this help text will be removed.
 
+General
+~~~~~~~
+
+- Changed the way static resources are built.
+  Parcel is used to build all apps and `generated_resources.txt` is auto-generated.
```
suutari-ai: What's the point of generated_resources.txt if it's autogenerated?

It was supposed to list all of those files which should be excluded when checking if the resource generation cache is up-to-date or not. (I.e. currently the build system generates a hash for the state of the *input* files of the built application by hashing all files recursively, but excluding the files from `generated_resources.txt` (since those are output files not input files).) I can't see how this could work if generated_resources.txt is not there yet.

If you just want to get rid of generated_resources.txt, much better solution would be to improve the resource generation caching mechanisms to be smarter that it would know which files are input and which are output by some other means.

chessbr: Cool, I will think in a better solution for this

PR #1336, opened 2018-07-02
Added custom model resolver to templates

soulilya:


shuup/front/template_helpers/urls.py
```diff
@@ -7,15 +7,24 @@
...
 
 from shuup.core.models import Category, Product
+from shuup.utils.importing import cached_load
 
 
 @contextfunction
 def model_url(context, model, absolute=False):
     uri = None
+
+    model_url_method = cached_load('SHUUP_MODEL_URL_RESOLVER_SPEC')
```
chessbr: Put this setting also in `shuup/front/settings.py` and document it. Move the original implementation to some other place like `shuup/front/utils/urls.py` and set this function as the `SHUUP_MODEL_URL_RESOLVER_SPEC` default value

chessbr:
@soulilya Finally, create some unit test that proves this new setting will work. I guess a client requesting the storefront category page and checking that the url is generated correctly. Add an entry in `docs/changelog.rst` saying the addition of the new setting.

shuup/front/template_helpers/urls.py
soulilya: I create new PR with fixes. Tests passes.

PR #1340, opened 2018-07-03
General: add model url resolver as a setting for Front and Admin

chessbr:
Now it is possible to customize the way model URLs are resolved.

Replaces https://github.com/shuup/shuup/pull/1336

### DIFF
--- a/doc/changelog.rst
+++ b/doc/changelog.rst
@@ -9,6 +9,11 @@ Unrealeased
   When releasing next version, the "Unreleased" header will be replaced
   with appropriate version header and this help text will be removed.
 
+General
+~~~~~~~
+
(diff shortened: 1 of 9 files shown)


chessbr:
Thanks @soulilya for the contribution.

chessbr:
LGTM. What do you think @tulimaki ?

tulimaki:
I guess this is ok. Would be even better if this was done with providers. So basically I would be able to provide model url resolver from my app. And then we loop all provided resolves until the url is resolved for the model.

But like said I think this is still better than the old one. @chessbr

chessbr:
Yeah, that would be really better. @tulimaki I will change this soon then.

PR #1343, opened 2018-07-03
Admin: Make orderlist show everything by default

Pikkupomo:
Orderlist showing only initial status caused a lot of problems with the customers because it seemed that orders 'disappeared' from the list when managing them.


Pikkupomo:
thanks, will look into it!

Issue #1220, opened 2018-04-05
Placeholder contents should honor active object


Pikkupomo:
### The Problem
Consider the following: you have two products in your store and want to have different `placeholder` content on both product pages. This is not possible with Shuup currently as the placeholders are pretty fixed.

### The Solutions?
1. Add a way for `Plugin` to know where it's needed. For example, when adding a `Plugin` to a `placeholder` the user could select which pages this would be shown. 
2. Make `placeholder` more context-aware to ensure it knows where it was added and what it should render.

to be continued...

PR #1345, opened 2018-07-03
Introduce per object placeholders

tulimaki:
* [x] Add option to do per object placeholders.
* [x] Add tests to prove this works
* [x] Add documenation to all new things
* [x] Add changelog
* [x] Document the new provide
* [x] Add option to pretend anonymous, person or company contact as a shop staff member while editing with xtheme.
* [x] Add browser test for the xtheme editing
* [x] Fix issue with deleting last cell
* [x] Add tests around admin tools
* [ ] Move admin tools styles outside the front styles so updating those would require re-releasing all existing themes.
* [ ] Add some help texts around the visibility limits.

Here we decided to add the visibility limitation to placeholders since that should make it easier for merchant to understand and preview. Also the logic behind "what should be shown for one placeholder" is easier.

Basically now the placeholder can have multiple layouts inside and these different layouts can be provided from apps. Then based on current context the right layouts is shown for each placeholder.

Editing the placeholders is pretty much untouched.

Refs #1220

### DIFF
--- a/shuup/core/models/_contacts.py
+++ b/shuup/core/models/_contacts.py
@@ -20,6 +20,10 @@ from timezone_field.fields import TimeZoneField
 from shuup import configuration
 from shuup.core.fields import InternalIdentifierField, LanguageField
 from shuup.core.pricing import PriceDisplayOptions
+from shuup.core.utils.users import (
+    is_user_all_seeing, should_force_anonymous_contact,
+    should_force_person_contact
+)
 from shuup.utils.analog import define_log_model
 
 from ._base import PolymorphicShuupModel, TranslatableShuupModel
@@ -354,10 +358,7 @@ class PersonContact(Contact):
 
     @cached_property
     def is_all_seeing(self):
-        if self.user_id and self.user.is_superuser:
-            all_seeing_key = "is_all_seeing:%d" % self.user_id
-            return configuration.get(None, all_seeing_key, False)
-        return False
+        return is_user_all_seeing(self.user)
 
 
 class AnonymousContact(Contact):
@@ -432,6 +433,10 @@ def get_person_contact(user):
     """
     if not user or user.is_anonymous():
         return AnonymousContact()
+
+    if should_force_anonymous_contact(user):
+        return AnonymousContact()
+
     defaults = {
         'is_active': user.is_active,
         'first_name': getattr(user, 'first_name', ''),
@@ -454,10 +459,31 @@ def get_company_contact(user):
       CompanyContact (or none) of which user's PersonContact is a member
     :rtype: CompanyContact|None
     """
+    if should_force_person_contact(user):
+        return None
+
     contact = get_person_contact(user)
     if not contact:
         return None
     return contact.company_memberships.filter(is_active=True).first()
+
+
+def get_company_contact_for_shop(shop):
+    company = CompanyContact.objects.get_or_create(
+        identifier="shop-contact-%s" % shop.pk, defaults={"name": shop.public_name}
+    )[0]
+    company.shops.add(shop)
+    return company
+
+
+def get_company_contact_for_shop_staff(shop, user):
+    assert user.is_staff
+    if not getattr(user, "is_superuser", False):
+        assert shop.staff_members.filter(id=user.id).exists()
+
+    contact = get_company_contact_for_shop(shop)
+    contact.members.add(get_person_contact(user))
+    return contact
 
 
 CompanyContactLogEntry = define_log_model(CompanyContact)
--- a/shuup/front/views/misc.py
+++ b/shuup/front/views/misc.py
@@ -7,7 +7,11 @@
 # LICENSE file in the root directory of this source tree.
 from django.http import HttpResponseRedirect
 
-from shuup import configuration
+from shuup.core.models import get_company_contact_for_shop_staff
+from shuup.core.utils.users import (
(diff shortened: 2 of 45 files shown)


shuup/xtheme/layout/_base.py
```diff
@@ -215,6 +217,12 @@ def __init__(self, theme, placeholder_name, rows=None):
         if rows:
             self.rows.extend(rows)
 
+    def is_valid_context(self, context):
+        return True
+
+    def get_layout_data_suffix(self, context):
```
chessbr: document here for what this is needed

shuup/xtheme/layout/_contact_group.py
```diff
@@ -0,0 +1,48 @@
...
+from shuup.xtheme.layout.utils import get_customer_from_context
+
+
+class AnonymousContactLayout(Layout):
+    identifier = "xtheme-anonymous-contact-layout"
+    help_text = _("Content in this placeholder is shown for guests.")
+
+    def is_valid_context(self, context):
+        customer = get_customer_from_context(context)
+        return (customer.is_anonymous if customer else False)
```
chessbr: is this test correct? Testing an `AnonymousContact` will always return `False`:
```py
>>> bool(AnonymousContact())
False
```
This way, the context should be valid if customer is `None` of `False` IMO

shuup/xtheme/layout/_contact_group.py
```diff
@@ -0,0 +1,48 @@
...
+class PersonContactLayout(Layout):
+    identifier = "xtheme-person-contact-layout"
+    help_text = _("Content in this placeholder is shown for person contacts.")
+
+    def is_valid_context(self, context):
+        customer = get_customer_from_context(context)
+        return (isinstance(customer, PersonContact) if customer else False)
+
+    def get_layout_data_suffix(self, context):
+        return self.identifier
```
chessbr: maybe a `ContactLayout` should also exist, or even `AuthenticatedUserLayout` that we can show something when there is actually a customer, doesn't matter the type.

shuup/xtheme/layout/_category.py
```diff
@@ -0,0 +1,21 @@
...
+# This source code is licensed under the OSL-3.0 license found in the
+# LICENSE file in the root directory of this source tree.
+from django.utils.translation import ugettext_lazy as _
+
+from shuup.xtheme.layout import Layout
+
+
+class CategoryLayout(Layout):
+    identifier = "xtheme-category-layout"
+    help_text = _("Content in this placeholder is shown for this category only.")
```
chessbr: maybe show the category name here, it could be confusing for user to have 100% what category we are talking about as products can have n categories

tulimaki:
The last 2 steps in todo I am not sure anymore whether those should be done. Even if I move the resources for example the definite theme still have own `base.jinja` and that would still require a release. Also those are just few colors and font size changes.

What comes to the next steps. I think in the future we have "easy" option to group the layouts inside placeholder at front to minimize that the merchant gets confused from so many different layouts. Also the layout could have ordering added while selecting the layout for edit.

shuup/front/views/misc.py
```diff
@@ -7,15 +7,53 @@
...
         return HttpResponseRedirect(return_url)
-    all_seeing_key = "is_all_seeing:%d" % request.user.pk
-    is_all_seeing = not configuration.get(None, all_seeing_key, False)
-    configuration.set(None, all_seeing_key, is_all_seeing)
+
+    toggle_all_seeing_for_user(request.user)
+    return HttpResponseRedirect(return_url)
+
+
+def force_anonymous_contact(request):
```
chessbr: I guess you are repeating yourself with this logic. I guess you could create a decorator for these `is_admin_user` checks

shuup_tests/xtheme/test_resources.py
```diff
@@ -38,7 +38,10 @@ def render(self, context):
         return self.message
 
 
+@pytest.mark.skip()
 def test_resources():
+    # TODO: Give some reason why this test should pass. Why we should
```
chessbr: what's this?

tulimaki: Test is broken since we don't render placeholders if not editable.

tulimaki: I guess point mainly is that should we just remove the test or change it somehow or what?

chessbr: got it. if the user cases doesn't exist, there is no reason for this test then.

shuup/front/views/misc.py
tulimaki: Yeah likely, but is it worth. Likely not. And I guess it is not guaranteed that we want same logic to all of these.

shuup/core/models/_contacts.py
```diff
@@ -454,12 +459,31 @@ def get_company_contact(user):
...
+        return None
+
     contact = get_person_contact(user)
     if not contact:
         return None
     return contact.company_memberships.filter(is_active=True).first()
 
 
+def get_company_contact_for_shop(shop):
+    return CompanyContact.objects.get_or_create(
```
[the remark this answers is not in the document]
tulimaki: Yeah likely useful to add the shop. But I guess it wouldn't hurt if this would have been more behind the scenes user that is not listed. For now I think I add this.

PR #1348, opened 2018-07-04
Xtheme: add option to show only sale items in highlight plugin

chessbr:
- Add parameter to general template tags to filter only products on sale
- Add option to show only sale items in highlight plugin
- Translate Highlight option types

Refs POS-2484

### DIFF
--- a/shuup/front/template_helpers/general.py
+++ b/shuup/front/template_helpers/general.py
@@ -26,8 +26,28 @@ from shuup.utils.mptt import get_cached_trees
 from shuup.utils.translation import cache_translations_for_tree
 
 
-@contextfunction
-def get_listed_products(context, n_products, ordering=None, filter_dict=None, orderable_only=True):
+def _group_list_items(group_list, number):
+    for i in range(0, len(group_list), number):
+        yield tuple(group_list[i: i + number])
+
+
+def _is_sale_item(product, pricing_context):
+    if product.is_variation_parent():
+        for variation in product.variation_children.all():
+            if variation.get_price_info(pricing_context).is_discounted:
+                return True
+
+    elif product.is_package_parent():
+        for package_child in product.get_all_package_children():
+            if package_child.get_price_info(pricing_context).is_discounted:
+                return True
+
+    return product.get_price_info(pricing_context).is_discounted
+
+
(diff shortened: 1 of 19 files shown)


shuup/front/template_helpers/general.py
```diff
@@ -26,8 +26,28 @@
...
+        for variation in product.variation_children.all():
+            if variation.get_price_info(pricing_context).is_discounted:
+                return True
+
+    elif product.is_package_parent():
+        for package_child in product.get_all_package_children():
+            if package_child.get_price_info(pricing_context).is_discounted:
+                return True
+
+    return product.get_price_info(pricing_context).is_discounted
```
Pikkupomo: would it be possible to use `shuup.campaigns.utils.matcher.get_matching_for_product` here instead of all this?

shuup/xtheme/plugins/products.py
```diff
@@ -19,33 +20,57 @@
...
+        sale_items_only = self.config.get("sale_items_only", False)
+
+        if highlight_type == HighlightType.NEWEST.value:
+            products = get_newest_products(context, count, orderable_only, sale_items_only)
+        elif highlight_type == HighlightType.BEST_SELLING.value:
+            products = get_best_selling_products(
+                context,
+                count,
+                orderable_only=orderable_only,
+                sale_items_only=sale_items_only
```
Pikkupomo: no need to name the params here (could keep the same style in all these three. actually, you could do it like this:

```python
if highlight_type == newest:
    f = get_newest_products
elif highlight_type == best_selling:
   f = get_best_selling
else:
    f = get_random_products

products = f(context, count, orderable_only, sale_items_only)
```
or have a mapping for type.

shuup/front/template_helpers/general.py
chessbr: No. Otherwise the discounts module won't be included. This will also add a direct dependence of Shuup Campaigns app here.

shuup/xtheme/plugins/products.py
chessbr: The problem is because not all functions have the same params

Pikkupomo:
LGTM, can be merged

PR #1352, opened 2018-07-05
Pillow 5.2 upgrade fixes

77cc33:
After updating to new Pillow release, there are lot of tests failed. This pull request fixes it.


tulimaki:
@77cc33 Thanks! I have one more commit incoming for this since we only need to limit that pillow version for Python 2.7 users from easy thumbnails.

PR #1353, opened 2018-07-05
Admin: add button to edit selected instances on select inputs

chessbr:
Now one can open edit page for related objects right on the object selector.
image
This makes it handy to open the related object page and make some changes.
Also, more quick add widgets were added in admin.

Refs POS-2486


Pikkupomo:
Does this take permissions into account when rendering edit / add? (it should)

PR #1354, opened 2018-07-05
Remove pillow limitations and give public name for shop on init

tulimaki:


tulimaki:
Yes it is. Good idea!

PR #1360, opened 2018-07-09
Classic Gray: add option to configure shop logo size styles

chessbr:
Add Xtheme option to render extra templates in theme configuration. This enable template injection in the configuration page for more advanced visualization and options.

Refs POS-2476

### DIFF
--- a/shuup/themes/classic_gray/templates/classic_gray/shuup/front/macros/theme/navigation.jinja
+++ b/shuup/themes/classic_gray/templates/classic_gray/shuup/front/macros/theme/navigation.jinja
@@ -1,3 +1,28 @@
+{% macro render_logo() %}
+    {% if request.shop.logo %}
+        {% set height_css = "height: {}px;".format(xtheme.get("shop_logo_height")) if xtheme.get("shop_logo_height") else "" %}
+        {% set width_css = "width: {}px;".format(xtheme.get("shop_logo_width")) if xtheme.get("shop_logo_width") else "" %}
+        {% if xtheme.get("shop_logo_height") and xtheme.get("shop_logo_width") %}
+            {% set cropped_logo = request.shop.logo|thumbnail(size=(xtheme.get("shop_logo_width"), xtheme.get("shop_logo_height"))) %}
+        {% else %}
+            {% set cropped_logo = request.shop.logo|thumbnail(size=(500, 500)) %}
+        {% endif %}
+    {% endif %}
+    <div
+        class="logo {% if not cropped_logo %}no-image{% endif %}"
+        style="{% if cropped_logo %}{{ height_css }}{{ width_css }}{% endif %}"
+    >
+        {% if cropped_logo %}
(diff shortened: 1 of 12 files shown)


Pikkupomo:
The original problem was that if theme provides 500x55 slot for shop logo, merchant could upload anything there and it would look bad / break. 

To solve this problem, we could've added fixed height and just let the merchant set the logo width (crop). But as these fields are here now, let's consider the following:

There is no way for me as a merchant to return to starting point. So, if I change something, I will break the theme indefently. Also, what are these numbers, how should I know what values look good on the theme? 

Furthermore, I would consider the following changes:

1. Theme: should be able to provide default values:
	1.1 default width (example 500), default height (example 55)
	1.2. lock width, lock height
2. Admin Form: Consider providing only width and align and do the crop on background.
3. Admin Form: It should not be possible to break the theme with these selections.
4. Admin Form: By default, all values should be default values if nothing chosen yet.
5. Admin Form: Default values should be shown (from theme, bullet point 1). 
6. Admin Form: Show my current logo dimensions. 
7. Admin Form: If we allow merchant to change the height, provide a range of dimensions
8. Admin Form Bonus: "use default" button after form field
9. Admin Form Bonus: show logo changes realtime (a bit hc though) :)


TL;DR; only width and align should be enough for merchant at this point

chessbr:
I added two buttons:
* Calculate the logo size: it will use some max values and try to fit the image within those values (200x80)
* Reset shop logo values: it will reset the size to the image size and will check the keep aspect ratio.

When the keep aspect ratio is checked, the height field is set to be read only.

image

shuup/themes/classic_gray/templates/classic_gray/shuup/front/macros/theme/navigation.jinja
```diff
@@ -1,3 +1,28 @@
+{% macro render_logo() %}
+    {% if request.shop.logo %}
+        {% set height_css = "height: {}px;".format(xtheme.get("shop_logo_height")) if xtheme.get("shop_logo_height") else "" %}
+        {% set width_css = "width: {}px;".format(xtheme.get("shop_logo_width")) if xtheme.get("shop_logo_width") else "" %}
+        {% if xtheme.get("shop_logo_height") and xtheme.get("shop_logo_width") %}
+            {% set cropped_logo = request.shop.logo|thumbnail(size=(xtheme.get("shop_logo_width"), xtheme.get("shop_logo_height"))) %}
+        {% else %}
+            {% set cropped_logo = request.shop.logo|thumbnail(size=(500, 500)) %}
```
tulimaki: Should this be 500x55 which was the original default.

Pikkupomo:
It's possible to break the theme 

image

logo width 4000px

I guess there should be some sane limitations provided by theme?

Pikkupomo:
Align has next to no effect, consider removing it unless you have plans to widen the logo area

PR #1363, opened 2018-07-10
Front: fix manufacturer modified filter to consider only visible shop products

chessbr:
**Issue:** some manufacturers were available for filtering and when applied, no product was returned. The queryset was not considering invisible shop products that had the manufacturer linked.
Excluding invisible shop products fixes the issue.

The current shop was also added to the manufacturer and shop product queryset to avoid considering other shops.

Refs POS-2497


chessbr:
Coverage seems to be ok

PR #1373, opened 2018-07-13
 Classic Gray: improve the way shop logo is resized for mobile screens

chessbr:
Also add fix in Front middleware to fallback to settings LANGUAGE when no available language is found


tulimaki:
Still approved but fix the tests it seems there is some old theme tests still around. @chessbr

PR #1377, opened 2018-07-13
Core: Fix issues with disappearing services

tulimaki:
There is some cases where the payment or shipping method
can be deleted and basket could contain invalid PKs.
This prevents errors from being raised.

Refs POS-2504


tulimaki:
Replaces https://github.com/shuup/shuup/pull/1371. The fix is in https://github.com/shuup/shuup/pull/1377/files#diff-41c4686acea2149073c66c2057269676R125

Issue #1375, opened 2018-07-13
Tests are running un-cached


Pikkupomo:
Currently, the tests are being run without caches. 

We should run tests using the same caching setup we are using for Shuup in general. The problem now is that we cannot rely on browser tests for example. 

This change was originally merged by me by mistake and I'm sorry about it.

Issue #1347, opened 2018-07-04
Cache is not bumped correctly for Front product template tags


chessbr:
To reproduce, add a highlight plugin that lists only orderable products, load the page and check for the products which are being rendered. Then, make a product unorderable (change visibility). After refreshing the page the product should be gone. Then make the product visible again, it never comes back.

Make sure to test with caches on:
```py
CACHES = {
    'default': {
        'BACKEND': 'django.core.cache.backends.locmem.LocMemCache',
        'LOCATION': 'shuup'
    }
}
```

### Steps to reproduce
- go to front and see a product there
- go to admin, see make it invisible
- go to front: it shouldn't be there (it is gone)
- go to admin and make it visible
- go to front: still NOT there <=== :bug: 

The issue is probably the context cache that is not bumped when products are saved. They are bumped through this post_save signal https://github.com/shuup/shuup/blob/master/shuup/core/__init__.py#L54-L69 but the instances are bumped, not lists.

As you can see here https://github.com/shuup/shuup/blob/master/shuup/front/template_helpers/general.py#L145 the `item=None` when generating the cache key, this means it is not attached to any product or object. Thus, when product is bumped this cache is not in the same namespace.

We can either make sure to bump the namespace (and make sure to bump only this shop namespace) OR disable the cache for those lists.

tulimaki:
@chessbr shouldn't this work now since we reset the caches between tests?

tulimaki:
In progress...

tulimaki:
Fixed!

PR #1380, opened 2018-07-13
Re-enable cache for unit tests

chessbr:
Enabling cache will simulate better the real world scenario when Shuup is deployed.

Refs #1375

### DIFF
--- a/shuup_tests/conftest.py
+++ b/shuup_tests/conftest.py
@@ -12,7 +12,6 @@ from django.core.signals import setting_changed
 from shuup.apps.provides import clear_provides_cache
 from shuup.testing.factories import get_default_shop
 from shuup.utils.importing import clear_load_cache
-from shuup.xtheme import set_current_theme
 from shuup.xtheme.testing import override_current_theme_class
 
 
@@ -53,6 +52,11 @@ def splinter_make_screenshot_on_failure():
 def enable_db_access(db):
     from django.utils.translation import activate
     activate("en")
+
+    # make sure the default cache is also cleared
+    # it is used by third party apps like parler
+    from django.core.cache import cache
+    cache.clear()
 
     from shuup.core import cache
     cache.init_cache()
--- a/shuup_tests/front/test_general_template_helpers.py
+++ b/shuup_tests/front/test_general_template_helpers.py
@@ -6,6 +6,7 @@
 # This source code is licensed under the OSL-3.0 license found in the
 # LICENSE file in the root directory of this source tree.
 import pytest
+from shuup.core import cache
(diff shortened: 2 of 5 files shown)


shuup_tests/conftest.py
```diff
@@ -54,6 +54,9 @@ def enable_db_access(db):
     from django.utils.translation import activate
     activate("en")
 
+    from django.core.cache import cache
```
tulimaki: It think you should clear the shuup cache instead. `from shuup.core import cache`.

shuup_tests/front/test_general_template_helpers.py
```diff
@@ -6,6 +6,7 @@
 # This source code is licensed under the OSL-3.0 license found in the
 # LICENSE file in the root directory of this source tree.
 import pytest
+from django.core.cache import cache
```
tulimaki: Shuup cache instead.

shuup_tests/front/test_general_template_helpers.py
```diff
@@ -52,19 +53,34 @@ def test_get_listed_products_orderable_only():
 
     from shuup.front.template_helpers import general
 
-    assert len(general.get_listed_products(context, n_products, orderable_only=True)) == 0
-    assert len(general.get_listed_products(context, n_products, orderable_only=False)) == 1
+    cache.clear()
```
tulimaki: Summarize a little what is going here. We shouldn't need to clear caches inside one test unless we are just testing something specific with the cache.

chessbr: Ok, I've cleared some too much `cache.clear()`. The point is. As caches are not bumped into a same test, the cache will return invalid objects. If you see a test case, it checks that the method doesn't return nothing, that an order is created and the same function is called again. The cache will take place and return wrong values. I am not trying to test the cache, I am testing that the function will return the expected values. We create orders and add categories directly in code, so caches are not bumped.

We have also this issue that I created when I was changing this file: https://github.com/shuup/shuup/issues/1347
We NEVER bump these caches, this is a bug. We should probably introduce some provide or signal to bump them.

You can also see that I run a `for cache_test in range(2):` in test, so the first loop is to calculate the actual value, and the second to test the cache is properly working.

shuup_tests/conftest.py
chessbr: I clear the Shuup cache above by re-creating it. I have to clear this cache because of 3rd party apps that might be using cache, like Parler. You can see fail in this test log: https://travis-ci.org/shuup/shuup/jobs/403915722

PR #1381, opened 2018-07-13
Classic Gray: improve the way shop logo is resized for mobile screens

tulimaki:
Also add fix in Front middleware to fallback to settings LANGUAGE when no available language is found


tulimaki:
Replaces: https://github.com/shuup/shuup/pull/1373

PR #1389, opened 2018-07-18
Small tuning around Django 1.11 support, admin and xtheme

tulimaki:


Pikkupomo:
LGTM so far

tulimaki:
Fixed! Whoever did the test for [this]( https://github.com/shuup/shuup/pull/1389/files#diff-0aa3152ad390dce83e3b68a275de5563R51) is cool. Tried to remove that `to_python` since it looked like not needed but the tests disagreed instantly. Other fix needed was the shop edit initial staff members which failed on create since m2m field can't be used before initial save.

shuup/admin/modules/suppliers/views/list.py
```diff
@@ -33,6 +35,9 @@ class SupplierListView(PicotableListView):
         Column("module_identifier", _(u"Module"), display="get_module_display", sortable=True)
     ]
 
+    def get_queryset(self):
```
chessbr: I guess you could return all suppliers when superuser, like you did on `edit.py`

tulimaki: This is already the same logic we use with other admin modules. Also I guess it makes it easier for superuser since the filter for shop in picotable is not the easiest task.

PR #1391, opened 2018-07-18
Admin: Add option to filter product with categories

tulimaki:
Well there is certain limitations with our Picotable filters since for example filtering with categories would require to use current shop  to filter the filter options. The name filter here looks pretty handy. When we have the admin design update done maybe we then introduce more options for related filters.

Since pretty much only option to add filters to the columns is define column as default column I removed few to add the categories. Shop shouldn't be needed since the products are already filtered by currently active shop. Also product type shouldn't have too much importance in basic installation.

### DIFF
--- a/shuup/admin/modules/products/views/list.py
+++ b/shuup/admin/modules/products/views/list.py
@@ -33,42 +33,60 @@ class ProductListView(PicotableListView):
     picotable_class = ProductPicotable
 
     default_columns = [
-        Column("primary_image",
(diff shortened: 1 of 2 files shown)


Pikkupomo:
These will now be part of the default columns, which means that they will be on the picotable when going there for the first time. Seems quite a lot of columns visible by default.

tulimaki:
Fixed! Removed 2 to add 2.

PR #1392, opened 2018-07-19
Front: Make admin toolbar better

Pikkupomo:


tulimaki:
Translations shouldn't either be updated directly? Or did you pull these from Transifex. I guess it's ok now, but in the future translations should be updated at Transifex only.

PR #1393, opened 2018-07-19
Xtheme:  make the initially selected language be the active one in form tabs

chessbr:
Make the initial selected language be the current language in multi-language forms to save extra clicks from users.

Improve xtheme to make TextPlugin fallback to empty string instead of `None` and also make the Parler default language be the first one in multi-language forms.

Refs POS-2510

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -52,8 +52,8 @@ matrix:
     - {python: 3.6, env: TOXENV=py36-django19-nomigrations}
     - {python: 2.7, env: TOXENV=py27-django111-nomigrations}
     - {python: 3.5, env: TOXENV=py35-django111-nomigrations}
-    - {python: 3.5, env: TOXENV=py35-django111-nomigrations}
     - {python: 3.6, env: TOXENV=py36-django111}
+    - {python: 3.6, env: TOXENV=py36-django111-nomigrations}
     - python: 3.6
       env: TOXENV=py36-django111-nomigrations-nocoverage-browser-travis-admin
       before_install:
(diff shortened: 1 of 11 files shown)


shuup/admin/template_helpers/shuup_admin.py
```diff
@@ -181,3 +181,9 @@ def get_admin_shop(context):
 @contextfunction
 def is_multishop_enabled(context):
     return settings.SHUUP_ENABLE_MULTIPLE_SHOPS is True
+
+
+@contextfunction
+def get_current_language(context):
```
tulimaki: You should have `{{ LANGUAGE_CODE }}` available at the template.

shuup/xtheme/template_ns.py
```diff
@@ -62,6 +62,11 @@ def get(self, context, name, default=None):
             return theme.get_setting(name, default=default)
         return default
 
+    @contextfunction
+    def get_current_language(self, context):
```
tulimaki: Also here you should have language code in context already.

.travis.yml
```diff
@@ -52,7 +52,6 @@ matrix:
     - {python: 3.6, env: TOXENV=py36-django19-nomigrations}
     - {python: 2.7, env: TOXENV=py27-django111-nomigrations}
     - {python: 3.5, env: TOXENV=py35-django111-nomigrations}
-    - {python: 3.5, env: TOXENV=py35-django111-nomigrations}
```
tulimaki: Likely this should have been py 3.6 without migrations since seems that we have one for Django 1.9 too.

shuup/admin/template_helpers/shuup_admin.py
chessbr: This is not always available in admin. When importing the macro, if one forgot to import `with context`, that won't be available

tulimaki: ok

shuup/xtheme/template_ns.py
tulimaki: I guess this also ok then?

chessbr: I removed this and inserted a `with context` when including this template. Not being a macro helps.

Issue #1376, opened 2018-07-13
Summernote: Add a way to add files from media browser 


Pikkupomo:
The problem with the current Summernote implementation is that it adds all images as inline items and it causes pretty much Shuup instances to explode. I tried to tackle this issue by removing the drag and drop support, but it's still present through the normal image addition.

I would like to see a feature in Shuup where admin user could choose an image from Shuup's own media browser and the image would be added into the content in question as `<img src="mediabrowserimage.jpg" />` instead of `src="data:image/..."`. 

If it's not possible to plug in the media browser in this way, at least the image upload should save the image to media browser and construct a properly linked `<img/>` based on that.

tulimaki:
Related: https://github.com/shuup/shuup/commit/8e7f202cb51a75d2dfafc69fb577c4326542438a

PR #1394, opened 2018-07-20
Admin: make text editor use Shuup media browser when adding pictures

chessbr:
Instead of inserting images in base64 form into the text field, use the current media browser
to handle images, enabling static files serving and caching.

This fixes cases when user adds a large image through drag & drop or even copy/paste.
When the server tried to load the page, it could break and hang forever.

The copy and paste and drag & drop was disabled as they are already implemented in media browser.

Refs POS-2514

# WIP - missing tests


doc/changelog.rst
```diff
@@ -24,6 +24,11 @@ Bug fixes
 - Fix front manufacturer modified filter to consider only visible shop products
 - Make Xtheme TextPlugin fallback to empty string when there is no translation for the current language.
 
+Admin
+~~~~~
+
+- Make Summernote editor use Shuup media browser when adding pictures
```
Pikkupomo: Should this mention that the drag and drop has been disabled as well? and why perhaps?

shuup/xtheme/templates/shuup/xtheme/editor.jinja
```diff
@@ -11,7 +11,7 @@
     <script type="text/javascript" src="{{ url("shuup_admin:js-catalog") }}"></script>
     <script src="{{ static("xtheme/editor.js") }}"></script>
     <link href="https://netdna.bootstrapcdn.com/bootstrap/3.3.5/css/bootstrap.css" rel="stylesheet">
-    <link href="https://cdnjs.cloudflare.com/ajax/libs/summernote/0.8.2/summernote.css" rel="stylesheet">
+    <link href="https://cdnjs.cloudflare.com/ajax/libs/summernote/0.8.10/summernote.css" rel="stylesheet">
```
Pikkupomo: I think we should start serving these as a part of the Shuup installation. I will create a separate issue about this as well, but could you please include and build this CSS file locally instead of the CDN.

chessbr: OK, I guess this could be done after https://github.com/shuup/shuup/pull/1335 gets merged and be part of https://github.com/shuup/shuup/issues/1404 as changing only this won't be super effective now.

tulimaki:
Refs https://github.com/shuup/shuup/issues/1376

PR #1400, opened 2018-07-23
SimpleCMS: make textarea bigger and not resizable

chessbr:
Do not allow resize of the text area as there is full screen mode for that.

Rename Simple CMS app label by Content Pages as it makes more sense for users.

Refs POS-2484

Requires https://github.com/shuup/shuup/pull/1394


Pikkupomo:
There is a bug still with fullscreen.
OS Shuup + Chrome

If you have a static page like this 

image

Open full screen and close it, it will end up like this:
image

It seems this kind of item appears there `<div class="note-toolbar-wrapper panel-default" style="height: 74px;"></div>`

Issue #1163, opened 2017-12-22
Cache product prices


tulimaki:
I believe caching product prices would greatly improve the front performance.

This is also pretty much blocking the product price sorting without custom project specific "hacks".

chessbr:
Yes, I guess we could cache prices but put some flag on the base pricing module to allow this cache. This way we can avoid caching prices for modules that dynamically calculates the price or fetches it from some other source like API etc, then the responsibility is the pricing module to cache it if wanted etc.

PR #1405, opened 2018-07-25
Discounts: New discount module to replace catalog campaigns

tulimaki:
Replacement for catalog campaigns. Catalog campaigns way too complex and adding few campaigns starts to take a toll on performance. There just doesn't seem to be any sane way to improve catalog campaigns either.

This should implement same discount options. Also few extras like discounted price and availability exceptions.

This has worse extendability but in theory the queryset in the module could be extendable by providing query filters. But still some weird discounts can be implemented as separate discount modules which I guess is the way it was supposed to be instead of making the default discount module too complex.

Goal for this discount module is to make product discounts lightning fast without losing the possibilities the catalog campaigns offer. Getting the discounts with one query should take care of the performance part.

For the future we likely want similar thing for basket campaigns. Also we should probably extend coupon codes into this one.

TODO:

* [x] Admin for happy hours
* [x] Admin for archives and delete
* [x] Admin for availability exceptions
* [x] Tests for admin views and forms
* [x] Improve catalog campaign sync. At least happy hours is not yet synced.
* [x] Add data migration for catalog campaign sync (nope, related fields are problematic with data migration so better just offer management command which also makes the sync optional)
* [x] Translations
* [x] Quick add for contact groups and filter groups with shop (waiting https://github.com/shuup/shuup/pull/1409)

Future ides:
* Add basket qty condition. Set product discounts based on the product qty in basket
* Combine basket campaigns into this. In theory we could link effect to the discount model. This effect would be exactly same effect we have already in basket campaigns. Then on product discount we can filter out discounts which have effect set. And for basket campaigns we can still likely loop the matching effects for best discount. As long as we get those effects with one query. Effects could be added to discount through quick add keeping the admin super simple.
* Maybe we should have select2 fields for things like categories, products, contacts and contact groups. Now we only have these select2multiple-fields. On the other hand we don't have quick add widget for these select2multiple-fields.
* Contact quick create

Questions:
* Should we try to simplify this even more and maybe just use shop instead of shops for discounts for example?

Refs https://github.com/shuup/shuup/issues/1163


shuup/product_discounts/models.py
```diff
@@ -0,0 +1,164 @@
...
+                    valid_hours__from_hour__lte=current_local_time,
+                    valid_hours__to_hour__gt=current_local_time,
+                )
+            )
+        )
+
+        if shop:
+            query &= Q(shops=shop)
+
+        return self.filter(query)
```
chessbr: use `distinct()` maybe?

shuup/product_discounts/models.py
```diff
@@ -0,0 +1,164 @@
...
+    end_datetime = models.DateTimeField(
+        null=True, blank=True, verbose_name=_("end date and time"),
+        help_text=_(
+            "The date and time the discount ends. This is only applicable if the discount is marked as active."
+        ))
+
+    # Conditions
+    product = models.ForeignKey(
+        "shuup.Product", related_name="product_discounts", blank=True, null=True,
+        db_index=True, on_delete=models.CASCADE, verbose_name=_("product"))
```
chessbr: FKs already are `db_index=True` by default

shuup/product_discounts/models.py
```diff
@@ -0,0 +1,164 @@
...
+        return self.pk
+
+    class Meta:
+        verbose_name = _('Valid Hours')
+        verbose_name_plural = _('Valid Hours')
+        ordering = ['product_discount', 'weekday', 'from_hour']
+
+    def save(self, **kwargs):
+        if self.to_hour < self.from_hour:
+            raise ValidationError("To hour has to be after from hour", code="valid_hours_error")
```
chessbr: translate error?

shuup/product_discounts/models.py
```diff
@@ -0,0 +1,164 @@
...
+
+    # Conditions
+    product = models.ForeignKey(
+        "shuup.Product", related_name="product_discounts", blank=True, null=True,
+        db_index=True, on_delete=models.CASCADE, verbose_name=_("product"))
+    exclude_selected_category = models.BooleanField(default=False, verbose_name=_("exclude selected category"))
+    category = models.ForeignKey(
+        "shuup.Category", related_name="category_discounts", blank=True, null=True,
+        db_index=True, on_delete=models.CASCADE, verbose_name=_("category"))
+    contact = models.ForeignKey(
```
chessbr: could this be a M2M?

shuup/product_discounts/models.py
```diff
@@ -0,0 +1,164 @@
...
+        "shuup.Contact", related_name="contact_discounts", blank=True, null=True,
+        db_index=True, on_delete=models.CASCADE, verbose_name=_("contacts"))
+    exclude_selected_contact_group = models.BooleanField(
+        default=False, verbose_name=_("exclude selected contact group"))
+    contact_group = models.ForeignKey(
```
chessbr: could this be a M2M?

shuup/core/settings.py
```diff
@@ -69,7 +69,7 @@
 #:
 #: Each discount module may change the price of a product.  See
 #: `shuup.core.pricing.DiscountModule` for details.
-SHUUP_DISCOUNT_MODULES = ["catalog_campaigns", "customer_group_discount"]
+SHUUP_DISCOUNT_MODULES = ["customer_group_discount", "product_discounts"]
```
chessbr: remember to remove the catalog campaings from admin

shuup/product_discounts/module.py
```diff
@@ -0,0 +1,90 @@
...
+        # Apply contact group conditions
+        condition_query &= (
+            Q(contact_group__isnull=True) |
+            (Q(exclude_selected_contact_group=False) & Q(contact_group__id__in=group_ids)) |
+            (Q(exclude_selected_contact_group=True) & ~Q(contact_group__id__in=group_ids))
+        )
+
+        # Get all possible discounts for the current product and context
+        potential_discounts = ProductDiscount.objects.available(shop).filter(
+            condition_query).values_list("discounted_price_value", "discount_amount_value", "discount_percentage")
```
Pikkupomo: As we have many (I mean **many**) rows in DB, maybe consider using `iterator` here as, by default, Django loads all results into memory and starts to iterate them through. With `.iterator()` the memory footprint would be way smaller, I guess :)

More info: https://docs.djangoproject.com/en/2.0/ref/models/querysets/#iterator

shuup/product_discounts/models.py
Pikkupomo: I guess the idea of all this is to prevent joining multiple tables but to get results "instantly" from this one table.

shuup/product_discounts/module.py
```diff
@@ -0,0 +1,90 @@
...
+
+            if discount_percentage:  # Discount percentage per item
+                discounted_prices.append(
+                    max(
+                        price_info.price - price_info.price * discount_percentage,
+                        shop.create_price(0)
+                    )
+                )
+
+        if len(discounted_prices):
```
Pikkupomo: for minor speed up, maybe use `if discounted_prices` here to prevent iteration :)

shuup/product_discounts/module.py
chessbr: Yes we can use it, but MySQL doesn't support this =)

Pikkupomo:
Few findings.

### Admin Form
* I cannot set Discounted price or Discounted amount or Discounted percentage because percentage must always be present. 
	-> Add `required=False` to `DiscountForm:discount_percentage` field.
	-> validate form so that one of these values must be given
* make `name` a required field (I guess it makes sense)
* Overall feeling of the form is that it's pretty complex and fields are not in any particular order. 
    -> name, shops, product it affects, start time and end time, discount fields (price, amount, percentage). After these fields (and after `<hr>`) the rest of the fields which provide finer tuning on visibility etc.

### Admin General
* `DiscountListView:get_queryset` returns only items for current shop, should superuser see all?

### Price Caching
* If availability exception invalidates, the price won't change. 
* In fact prices are cached pretty hard and changes in admin (re-activating discount, changing price, amount, percentage) won't invalidate caches either. 

Overall, this is **fast** and functional. The admin form needs a bit usability love and explanations on how everything work, but other than that, golden stuff.

shuup/discounts/admin/modules/_coupon_codes.py
```diff
@@ -0,0 +1,54 @@
...
+
+    def get_urls(self):
+        delete = admin_url(
+            "^discounts_coupon_codes/(?P<pk>\d+)/delete/$",
+            "shuup.discounts.admin.views.CouponCodeDeleteView",
+            name="discounts_coupon_codes.delete",
+            permissions=get_default_model_permissions(CouponCode)
+        )
+
+        return [delete] + get_edit_and_list_urls(
```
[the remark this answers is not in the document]
tulimaki: I think the current one is more readable.

shuup/discounts/admin/views/_coupon_codes.py
```diff
@@ -0,0 +1,136 @@
...
+
+class CouponCodeEditView(CreateOrUpdateView):
+    model = CouponCode
+    form_class = CouponCodeForm
+    template_name = "shuup/discounts/edit.jinja"
+    context_object_name = "discounts"
+
+    def get_queryset(self):
+        if getattr(self.request.user, "is_superuser", False):
+            return CouponCode.objects.all()
```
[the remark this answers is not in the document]
tulimaki: On purpose. I guess this allows superuser to crate multishop discounts while still only listing the most relevant ones. Also the delete might be a little safer if the shop needs to be also selected.

tulimaki:
Fixed! Add contact group quick add and shop limitation. Improved the discount admin a little (separated the inputs).

Didn't make the name required. We don't need the name so let's keep it optional. It is useful for picotable but that is about it. Small shop doesn't need it.

What is this cache thing? There shouldn't be any caches in front of the price or discount modules.

shuup/discounts/models/_discounts.py
```diff
@@ -0,0 +1,160 @@
...
+        "shuup.Contact", related_name="contact_discounts", blank=True, null=True,
+        on_delete=models.CASCADE, verbose_name=_("contact"),
+        help_text=_("Select contact for this discount."))
+    exclude_selected_contact_group = models.BooleanField(
+        default=False, verbose_name=_("exclude selected contact group"),
+        help_text=_("Exclude contacts in selected contact group from this discount."))
+    contact_group = models.ForeignKey(
+        "shuup.ContactGroup", related_name="contact_group_discounts", blank=True, null=True,
+        on_delete=models.CASCADE, verbose_name=_("contact group"),
+        help_text=_("Select contact for this discount."))
```
chessbr: contact *group*

shuup/discounts/management/commands/sync_catalog_campaigns.py
```diff
@@ -0,0 +1,173 @@
...
+)
+from shuup.core.models import (
+    Category, CompanyContact, ContactGroup, PersonContact, Product
+)
+from shuup.discounts.models import Discount, HappyHour, TimeRange
+
+
+class Command(BaseCommand):
+
+    def handle(self, *args, **options):
```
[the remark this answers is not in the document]
tulimaki: Can't be migrations since related objects are too hard. Migrating this is up to who ever does the update. There is always option to keep using the old catalog campaigns too until those are "consumed".

shuup/discounts/modules.py
```diff
@@ -0,0 +1,91 @@
...
+
+            if discount_amount_value:  # Discount amount value per item
+                discounted_prices.append(
+                    max(
+                        price_info.price - shop.create_price(discount_amount_value) * price_info.quantity,
+                        shop.create_price(0)
+                    )
+                )
+
+            if discount_percentage:  # Discount percentage per item
```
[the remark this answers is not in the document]
tulimaki: Yes but only the best one is applied of course. Couldn't find reason to limit to one effect per discount.

shuup/discounts/utils.py
```diff
@@ -0,0 +1,71 @@
...
+
+        condition_query &= (
+            Q(coupon_code__isnull=True) |
+            coupon_queries
+        )
+    else:
+        condition_query &= Q(coupon_code__isnull=True)
+
+    # Get all possible discounts for the current product and context
+    return Discount.objects.available(shop).filter(
```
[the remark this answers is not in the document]
tulimaki: Not sure if distinct needed, but I can add it.

shuup/discounts/admin/views/_happy_hours.py
```diff
@@ -0,0 +1,198 @@
...
+                self.fields["weekdays"].initial = weekdays
+                self.fields["from_hour"].initial = from_hour
+                self.fields["to_hour"].initial = to_hour
+
+        # Since we touch these views in init we need to reset some
+        # widgets and help texts after setting the initial values.
+        self.fields["from_hour"].widget = TimeInput()
+        self.fields["to_hour"].widget = TimeInput()
+        help_texts = [
+            ("from_hour", _("12pm is considered noon and 12am as midnight.")),
```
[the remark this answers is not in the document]
tulimaki: Both from hour and to hour are included the range.

shuup/discounts/admin/views/_happy_hours.py
```diff
@@ -0,0 +1,198 @@
...
+            ("to_hour", _("12pm is considered noon and 12am as midnight. End time is considered match.")),
```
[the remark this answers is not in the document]
tulimaki: Like the help text says to hour is included in range.

shuup/discounts/management/commands/sync_catalog_campaigns.py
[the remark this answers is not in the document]
chessbr: well, you removed it from admin, so maybe you can put it back and deprecated in some way, like preventing to create new ones?

shuup/discounts/utils.py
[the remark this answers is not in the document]
chessbr: if for some reason, some of the ORs in the query match the same object, it will be returned twice (or even more times), hence the distinct. Unless you are 100% the query won't match the same object more than once.

shuup/core/pricing/_context.py
```diff
@@ -41,18 +41,23 @@ class PricingContext(PricingContextable):
...
         assert shop is not None, "shop is required"
         assert customer is not None, (
             "customer is required (may be AnonymousContact though)")
 
         self.shop = shop
         self.customer = customer
+        self.basket = basket
+        if basket:
+            assert basket.shop == shop
+            assert basket.customer == customer
```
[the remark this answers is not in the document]
tulimaki: Let's keep it unified with the rest of the code in this method. It's kind a lost situation if these asserts gets triggered. We need to evaluate separately what we should do all these asserts or do we do anything. In theory you could open issue about these.

shuup/discounts/admin/mass_actions.py
```diff
@@ -0,0 +1,53 @@
...
+from shuup.admin.utils.picotable import PicotableMassAction
+from shuup.discounts.models import Discount
+
+
+class ArchiveMassAction(PicotableMassAction):
+    label = _("Archive Product Discounts")
+    identifier = "archive_discounts"
+
+    def process(self, request, ids):
+        query = Q(pk__in=ids)
```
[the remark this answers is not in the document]
tulimaki: Did the 3rd option.

shuup/discounts/admin/views/_exceptions.py
```diff
@@ -0,0 +1,140 @@
...
+            if object.pk else None)
+        return get_default_edit_toolbar(self, self.get_save_form_id(), delete_url=delete_url)
+
+    def get_form_kwargs(self):
+        kwargs = super(AvailabilityExceptionEditView, self).get_form_kwargs()
+        kwargs["request"] = self.request
+        return kwargs
+
+
+class AvailabilityExceptionDeleteView(DetailView):
```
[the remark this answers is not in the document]
tulimaki: I guess I wanted the custom messages.

Actual reason: thought there might be a reason we haven't use much of that Django delete view in Shuup admin but I guess the delete view should work. Well it does pretty much the same thing so not changing these.

shuup/discounts/admin/views/_active_list.py
```diff
@@ -0,0 +1,88 @@
...
+        if instance.discount_percentage:
+            effects.append(format_percent(instance.discount_percentage))
+
+        return ','.join(effects)
+
+    def format_end_datetime(self, instance, *args, **kwargs):
+        return get_locally_formatted_datetime(instance.end_datetime) if instance.end_datetime else ""
+
+    def get_queryset(self):
+        return Discount.objects.active(get_shop(self.request))
```
chessbr: The reason why I was not seeing the campaign I've just created is because the `active` here.. the dates were not in the range neither the active checkbox was selected.. does this make sense?

shuup/discounts/templates/shuup/discounts/discount_edit.jinja
```diff
@@ -0,0 +1,26 @@
+{% extends "shuup/admin/base.jinja" %}
+{% from "shuup/admin/macros/general.jinja" import single_section_form with context %}
+
+{% block content %}
+    {% call single_section_form("discounts_form", form) %}
+        {{ bs3.field(form.name) }}
```
chessbr: you are missing the `shops` field here. as superuser, I can't see the field.

shuup/discounts/modules.py
```diff
@@ -0,0 +1,91 @@
...
+    get_active_discount_for_code, get_potential_discounts_for_product
+)
+
+
+class ProductDiscountModule(DiscountModule):
+    identifier = "product_discounts"
+    name = _("Product Discounts")
+
+    def discount_price(self, context, product, price_info):
+        shop = context.shop
```
chessbr: when `shops` is `None` the campaign is not being applied

shuup/discounts/admin/views/_active_list.py
tulimaki: Yeah that is missing shop. And yes it is on purpose that active and not active discounts is shown in separate list to avoid confusion.

shuup/discounts/modules.py
tulimaki: That is exactly what we want atm.

chessbr:
so, I am having issues with happy hours.
1) it looks like the timezone is not being considered.. I set up the happy hours to be from 04:50pm to 04:53pm, it doesnt work, but if I set from  04:50pm to 07:53pm (UTC) it works..
2) happy hours won't work for cached prices I guess, at least the time expires and it is still there (when active) or it doesn't activate when the time comes..
image
The top price is cached, and the bottom isn't

Issue #1403, opened 2018-07-23
Picotable generic filters


Pikkupomo:
Picotable should have generic filters for all known datatypes. Currently, if there is nothing defined for the field, it's not searchable.

Pikkupomo:
Updated description, added saveable defaults

Pikkupomo:
The second part of this task is now on PR (https://github.com/shuup/shuup/pull/1576)

PR #1409, opened 2018-07-26
Add shop to Contact Group and registration shop to contact

Pikkupomo:
Overhaul contact group and contact shops.


shuup/admin/modules/contacts/views/list.py
```diff
@@ -52,7 +53,10 @@ class ContactListView(PicotableListView):
         Column("groups", _("Groups"),
                filter_config=ChoicesFilter(ContactGroup.objects.all(), "groups"),
                display="get_groups_display"),
-        Column("shops", _("Shops"), filter_config=Select2Filter(Shop.objects.all()), display="get_shop_display")
+
+        # TODO: Fix these filters, Shop.objects.all() is wrong...
+        Column("shops", _("Shops"), filter_config=Select2Filter(Shop.objects.all()), display="get_shop_display"),
```
tulimaki: Maybe we can't have this filters at all if we can't limit the shop. You also have always option to add text filter and allow option to filter by shop name. Pretty much the same thing.

shuup/testing/utils.py
```diff
@@ -45,6 +49,18 @@ def apply_request_middleware(request, **attrs):
         activate(current_language)
 
     assert request.shop
+
+    if not attrs.get("skip_session", False):
+        frm = inspect.stack()[1]
+        mod = inspect.getmodule(frm[0])
+        if mod.__name__.startswith("shuup_tests.admin"):
+            # admin uses `get_shop` which relies on this "sessioned" shop id
+            if not request.session:
```
tulimaki: Use `shuup.admin.set_providers.set_shop` here to ensure you are using same logic as the Shuup admin.

shuup/admin/modules/contact_groups/views/forms.py
```diff
@@ -19,23 +20,36 @@
...
+        shop = get_shop(self.request)
+        self.fields["shop"] = forms.ModelChoiceField(
+            queryset=Shop.objects.filter(pk=shop.id),
+            initial=shop,
+            widget=HiddenInput(),
+            label=_('shop'),
+            required=False)
+
+    def clean_shop(self):
+        return get_shop(self.request)
```
tulimaki: What is this? Consider settings `self.shop = get_shop(self.request)` while you pop out the request at init.

shuup/admin/modules/contact_group_price_display/views/forms.py
```diff
@@ -0,0 +1,108 @@
...
+            required=False)
+
+        self.fields['price_display_mode'] = forms.ChoiceField(
+            choices=PRICE_DISPLAY_MODE_CHOICES,
+            label=_("Price display mode"),
+            initial=get_price_display_mode(self.request, self.instance),
+            help_text=_("Set how prices are displayed to contacts in this group."))
+
+    def clean_shop(self):
+        return get_shop(self.request)
```
tulimaki: What is this? See the previous comment on setting `self.shop`.

shuup/admin/modules/contact_group_price_display/views/forms.py
```diff
@@ -0,0 +1,108 @@
...
+    class Meta:
+        model = ContactGroupPriceDisplay
+        fields = ("group", "shop",)
+
+    def __init__(self, *args, **kwargs):
+        self.request = kwargs.pop("request")
+        super(ContactGroupPriceDisplayForm, self).__init__(*args, **kwargs)
+        shop = get_shop(self.request)
+        if self.instance.pk:
+            self.fields["group"].choices = [(self.instance.group.id, self.instance.group.name)]
```
tulimaki: Add quick add filter for groups.

shuup/admin/modules/contact_group_price_display/views/forms.py
```diff
@@ -0,0 +1,108 @@
...
+        return 'none'
+    if contact_group.shop:
+        assert contact_group.shop == shop
+    price_display = contact_group.price_display_options.for_group_and_shop(contact_group, shop)
+    taxes = price_display.show_prices_including_taxes
+    hide = price_display.hide_prices
+    if hide is None and taxes is None:
+        return 'none'
+    elif hide:
+        return 'hide'
```
tulimaki: Should these use the actual choices instead these magical values?

tulimaki: Well looks like old code moved around. Maybe.

shuup/admin/modules/contact_group_price_display/views/list.py
```diff
@@ -0,0 +1,60 @@
...
+
+class ContactGroupPriceDisplayListView(PicotableListView):
+    model = ContactGroupPriceDisplay
+    default_columns = [
+        Column("group", _(u"Group"), display="group"),
+        Column("display_mode", _(u"Display Mode"), display="show_display_mode")
+    ]
+
+    def get_queryset(self):
+        if self.request.user.is_superuser:
```
tulimaki: Should probably use `getattr(self.request.user, "is_superuser", False)` instead.

shuup/front/middleware.py
```diff
@@ -18,6 +18,7 @@
 
 from shuup.core.middleware import ExceptionMiddleware
 from shuup.core.models import Contact, get_company_contact, get_person_contact
+from shuup.core.models._contacts import get_price_display_options
```
tulimaki: You shouldn't import anything from `_contacts`.

shuup/core/models/_contacts.py
```diff
@@ -536,6 +566,82 @@ def get_company_contact_for_shop_staff(shop, user):
     return contact
 
 
+def get_price_display_options(contact, **kwargs):
```
tulimaki: Why did you have to move this? Why we can't ask this through contact anymore?

shuup/admin/modules/contact_group_price_display/views/forms.py
Pikkupomo: yea it's old code moved around, I can make them less magical

shuup/admin/modules/contact_groups/views/forms.py
Pikkupomo: It ensures the shop field value never gets set to something we don't want. I'm not sure how the self.shop would help here as it never gets handled.

shuup/admin/modules/contacts/views/list.py
Pikkupomo: Yea, I think this will be fixed as a part of https://github.com/shuup/shuup/issues/1403

tulimaki: So does this really filter the shops somehow. Surely we can't merge this if all the shops end up in the filter. The #1403 is not on roadmap either and might be something we need to wait first wave of admin updates before doing.

Pikkupomo: This is the same way as it's been done before (as you can see, I just moved it) But I can look into this

PR #1411, opened 2018-07-27
Front:  Add privacy page link to the login dropdown (POS-2523)

maximlion:
Added gdpr privacy page link to the login dropdown.


shuup/gdpr/providers.py
```diff
@@ -44,7 +45,11 @@ def get_fields(self, **kwargs):
         for page in get_active_consent_pages(request.shop):
             key = "accept_{}".format(page.id)
             field = forms.BooleanField(
-                label=_("I have read and accept the {}").format(page.title),
+                label=mark_safe(('''I have read and accept the
```
chessbr: you are note using the ugettext here and use double quotes instead: `"`
```py
ugettext("I have read and accept the <a href='{}' target='_blank' class='login-box'>{}</a>.").format(...)
```
don't use ugettext_lazy as when using mark_safe, just to make sure

shuup/gdpr/providers.py
```diff
@@ -44,7 +45,11 @@ def get_fields(self, **kwargs):
...
+                    <a href='{}' target='_blank' class='login-box'>{}</a>.''').format(
+                    reverse("shuup:cms_page", kwargs=dict(url=page.url)),
+                    page.title
+                )),
```
chessbr: remove the help text below as you are adding the link to the label itself

PR #1422, opened 2018-08-11
Front: Do not continue product reload without ajax div

tulimaki:
Things like for example Wishlist plugin might call this reload on wrong times like frontpage for example.


chessbr:
Flake8 test is failing but it is being fixed here https://github.com/shuup/shuup/pull/1424/

PR #1427, opened 2018-08-14
Front: Fix product card image cropping & make product card bigger

maximlion:
Fixed product card image cropping by replacing 500x500 dimensions
with 420x420, enabled smart cropping in settings
Made product card's detail-wrap bigger so now a product with a long name
(up to three lines) doesn't cause the cards to differ.

Refs POS-2543

image
image
image
image
image
image

### DIFF
--- a/doc/changelog.rst
+++ b/doc/changelog.rst
@@ -17,6 +17,7 @@ Core
 Front
 ~~~~~
 
+- Tune product card and product list styles for better alignment
 - Bump template helpers caches when shop products and manufacturers are saved
 
 General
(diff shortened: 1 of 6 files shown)


maximlion:
[The problem now are the moving prices](url)

chessbr:
@maximlion you can always put images directly in GitHub comments to make life easier =)

maximlion:
Maybe make prices fixed?

chessbr:
@tulimaki can you take a new look now?

tulimaki:
screen shot 2018-08-17 at 9 46 24 am

Still not right. Toggle with grid and list is now broken. What if we copy the whole solution for these from definite theme since I think that works? We don't need separate implementation for this in `shuup.front`.

chessbr:
Replaced by https://github.com/shuup/shuup/pull/1437

PR #1435, opened 2018-08-17
Fix filter fields in order status, simple cms and tasks type admin modules

chessbr:
All of them are translation fields and they need to filter by the related field name

Refs POS-2551


tulimaki:
LGTM.

PR #1437, opened 2018-08-17
Classic gray: improve products card styles add carousel in xtheme plugins

chessbr:
Improve products card styles to make them align nicely and add carousel for xtheme plugins

Refs POS-2543


tulimaki:
Yes this is more like it. 🥇

Issue #1439, opened 2018-08-20
Improve resource cache generation


chessbr:
According to @suutari-ai https://github.com/shuup/shuup/pull/1335#discussion_r209906025 we should get rid of `generated-resources.txt` and create a new cache system.

chessbr:
Fixed in https://github.com/shuup/shuup/pull/1441

PR #1441, opened 2018-08-21
Improve the way the static build cache is handled

chessbr:
- [X] Use Parcel cache system to cache builds in a common location
- [X] Create JS lib to help building Shuup static files
- [X] Remove generated_resources.txt file logic for good
- [X] Removed Concurrently and moved to a own build script file scheme
- [X] Update build_resources command and remove old static file cache logic
- [x] Prune `.cache` from MANIFEST

Refs #1439  

🔥 Admin is not included here as it is under heavy refactor and it won't contain any sort of cache for now


chessbr:
@suutari-ai What do you think? Now using Parcel cache system. I put it by default in home dir so it is almost the same logic as before.

Also, to prevent running `npm install` every time and making build faster, I don't call it on `build` targets. If there is node `node_modules` in app folder OR it is a production build OR user want to force install, then I we call `npm install`.

Things looks much better now. I replaced concurrently by a small `build_resources.js` file that uses a Shuup common js lib to create the Parcel CLI command. The idea is to keep the file super small and all the logic to build in the lib that could be also pushed to npm registry so other apps could use it too. I could use Parcel API but spawning sub processes in our script let us have more control on how things are done.

_misc/sanity_utils.py
```diff
@@ -112,9 +112,6 @@ def find_files(
             _remove_ignored_directories(path, dirs, ignored_dirs, ignored_path_regexps)
             for filename in files:
                 filepath = posixpath.join(path, filename)
-                if filename == "generated_resources.txt":
-                    _process_generated_resources(path, filepath, generated_resources)
-                    continue
```
suutari-ai: The _process_generated_resources function is now probably unused and should be removed too.

shuup/front/build_resources.js
```diff
@@ -0,0 +1,28 @@
...
+ * Copyright (c) 2012-2018, Shuup Inc. All rights reserved.
+ *
+ * This source code is licensed under the OSL-3.0 license found in the
+ * LICENSE file in the root directory of this source tree.
+ */
+const { getParcelBuildCommand, runBuildCommands } = require("shuup-static-build-tools");
+
+runBuildCommands([
+    getParcelBuildCommand({
+        cacheFolder: "front",
```
suutari-ai: It's a directory, not a folder. :D   (At least for consistency with the other argument names, this should be cacheDir)

shuup_setup_utils/commands.py
```diff
@@ -47,14 +47,16 @@ class BuildResourcesCommand(distutils.core.Command):
...
     directory = '.'
     user_options = [
         ('mode=', 'm', "build mode: 'development' (default) or 'production'"),
         ('clean', 'c', "clean intermediate files before building"),
         ('force', 'f', "force rebuild even if cached result exists"),
+        ('no-install', 'n', "do not install npm packages before building"),
         ('directory=', 'd', "directory to build in, or '.' for all (default)"),
     ]
-    boolean_options = ['clean', 'force']
+    boolean_options = ['clean', 'force', 'force-install']
```
suutari-ai: This force-install should probably be `no-install`. Right?

suutari-ai:
I like where this is heading. Very nice work!

There was a couple minor issues which I found in the review though. Please check them out.

chessbr:
@suutari-ai Fixed! Thanks! :heart:

tulimaki:
@suutari-ai Ok so now I need to know to add this 
```

suutari-ai:
I think it's good that not all npm packages in the tree are built by default, i.e. if you have package.json files for some other reason, not all are assumed to be for Shuup built static resources.

There could be a backward compatibility provided though. Previously it detected a directory with `package.json` as a directory with "Shuup built static resources" by (1) existence of `generated_resources.txt` next to it but now (2) the `shuup.static_build` is expected to be true in `package.json`. How about doing detecting so that either (1) or (2) would trigger the build? @chessbr

chessbr:
Yeah, maybe I removed backwards compatibility too early. I thought that if they upgrade to latest Shuup version, they would add that flag to the `packages.json` while bumping the Shuup version and done, things solved. I can add that (1) back @tulimaki .

Explaining about the `build_resources.js` files:
- To use Parcel cache system, I had to put more parameters to the command. The command should point to a _common_ place out of the Shuup folder to be persisted after you delete it, for example. The common place is `~/.cache/shuup/[app_name]`
- As the command was getting super large and the command is similar to all apps inside Shuup and `concurrently` was _failing_ miserably when watching files, I decided to create a lib to build this command that can be used by all Shuup apps and also by 3rd party apps that uses Parcel to build. Parcel has an API for that for more complex situations but to keep it simple I created the lib so we can spawn sub processes as we like, so we can build N commands simultaneously.

We need now to push that lib to NPM registry and we can use that more easily.

And when you say `OK so now I need to know to add this "shuup": {static_build: true}`, this is pretty much what we have before, I honestly didn't find any reference to what `generated_resources.txt` was and that was used to build resources, I had to search for that in code. So, we probably need to add documentation around this then.

suutari-ai:
> I honestly didn't find any reference to what generated_resources.txt was and that was used to build resources, I had to search for that in code.

True. The `generated_resources.txt` was just an ugly hack to make the caching work. It was supposed to be replaced with something better pretty soon, so it was never documented. :smile:

tulimaki:
Ok sounds reasonable. I will highlight this change in release and I think we might be good.

Issue #1455, opened 2018-08-29
Add manufacturer logo


tulimaki:
Required by some new cloud themes and sounds like a common requirement.

PR #1457, opened 2018-08-29
Admin: Add logo to manufacturers

maximlion:
Added logo to manufacturer model, made a new form `ShuupAdminNoTranslation`, which is to be used by admin forms that do not need/use translation

Refs https://github.com/shuup/shuup/issues/1455


tulimaki:
Changelog update would have been cool. But will add it as part of the release. 👍

maximlion:
Yeah didn't want to change it because the manufacturer logo is not used anywhere yet.

shuup/admin/forms/_base.py
```diff
@@ -20,3 +21,12 @@ def __init__(self, **kwargs):
             if issubclass(self.fields[field].widget.__class__, AdminFileWidget):
                 self.fields[field].widget = FileDnDUploaderWidget(
                     upload_path="/default", kind="images", clearable=True)
+
+
+class ShuupAdminFormNoTranslation(ModelForm):
```
chessbr: DRY. This code is the same as the class above, you should create some mixin for this

shuup/admin/modules/manufacturers/__init__.py
```diff
@@ -41,7 +42,7 @@ def get_menu_entries(self, request):
         ]
 
     def get_required_permissions(self):
-        return get_default_model_permissions(Manufacturer)
+        return get_default_model_permissions(Manufacturer) | get_default_model_permissions(File)
```
chessbr: I don't think this is really needed

maximlion: I'll change it in the next PR.

maximlion:
@chessbr  I'll change it in the next PR.

PR #1461, opened 2018-08-31
Xtheme: add feature to add global snippet injection

chessbr:
Add feature to inject snippet of codes in all or a set of themes.

Add widget to enable CoreMirror editor

Refs POS-2564


shuup/xtheme/models.py
```diff
@@ -160,3 +173,22 @@ def update_settings(self, update_values):
...
+class Snippet(models.Model):
+    """
+    Inject snippet code globally filtering by themes if configured
+    """
+    shop = models.ForeignKey("shuup.Shop", related_name="snippets")
+    location = models.CharField(max_length=64, verbose_name=_("location"))
+    snippet_type = models.CharField(max_length=20, verbose_name=_("snippet type"), choices=SnippetTypeChoices)
+    snippet = models.TextField(verbose_name=_("snippet"))
+    # list of theme identifiers that will be have this sniipet injected, if None, it means all themes
+    themes = SeparatedValuesField(verbose_name=_("themes"), blank=True, null=True)
```
tulimaki: Why we need themes here?

shuup/xtheme/resources.py
```diff
@@ -213,3 +250,49 @@ def add_resource(context, location, resource):
...
+
+def inject_global_snippet(context, content):
+    if not valid_view(context):
+        return
+
+    from shuup.xtheme import get_current_theme
+    from shuup.xtheme.models import Snippet, SnippetType
+    shop = get_shop(context["request"])
+
+    for snippet in Snippet.objects.filter(shop=shop):
```
tulimaki: This likely needs to be cached. We can't loop over snippets on every page load.

shuup/xtheme/admin_module/__init__.py
```diff
@@ -96,3 +98,45 @@ def get_notifications(self, request):
                         title=_("Theming"),
                         url="shuup_admin:xtheme.config"
                     )
+
+
+class XthemeSnippetsAdminModule(AdminModule):
```
tulimaki: Make this available from theme configuration (action button over table) and the front toolbar.

shuup/xtheme/models.py
chessbr: you can enable injection just for some themes

shuup/xtheme/resources.py
chessbr: true

shuup/xtheme/models.py
tulimaki: Make sure optional then and add help text too.

chessbr: it is optional, I just added the help text to the form, I will move it to here

Issue #1431, opened 2018-08-17
Adding products while making order in the admin dashboard fails


igordavydsson:
Trying to make a new order in the admin dashboard. When I search for products, trying to add them to this new order, JS adds a new "empty" line, because it cannot handle product's stock count properly at [this line](https://github.com/shuup/shuup/blob/master/shuup/admin/modules/orders/static_src/create/reducers/lines.js#L43).

![peek 2018-08-17 11-57](https://user-images.githubusercontent.com/24206160/44257595-2abf6780-a215-11e8-8b5c-58ea4d4e2556.gif)

Generated `order-creator.js` static file has the following lines:
```javascript
function getFormattedStockCounts(line) {
	    var physicalCount = (0, _numbers2.default)(line.physicalCount);
	    var logicalCount = (0, _numbers2.default)(line.logicalCount);

	    return {
	        physicalCount: physicalCount.toFixed(line.salesDecimals) + " " + line.salesUnit,
	        logicalCount: logicalCount.toFixed(line.salesDecimals) + " " + line.salesUnit
	    };
	}
```

For some reason, both `physicalCount` and `logicalCount` are strings, not integers.
All works fine, once I change them to:
```javascript
function getFormattedStockCounts(line) {
	    var physicalCount = parseInt((0, _numbers2.default)(line.physicalCount));
	    var logicalCount = parseInt((0, _numbers2.default)(line.logicalCount));

	    return {
	        physicalCount: physicalCount.toFixed(line.salesDecimals) + " " + line.salesUnit,
	        logicalCount: logicalCount.toFixed(line.salesDecimals) + " " + line.salesUnit
	    };
	}
```

PR #1463, opened 2018-09-04
Admin: Fix adding stock products to the order

maximlion:
toFixed in ensureNumericValue was returning a string instead of an int or a float,
that's what was breaking the addition of stock products to the order in admin.

Refs #1431


tulimaki:
No need I can add one during the next release.

PR #1479, opened 2018-09-07
Xtheme: fix snippet injection url

chessbr:
Something happened to the correct url in some rebase, probably

No Refs


chessbr:
already fixed :disappointed:

PR #1484, opened 2018-09-11
Xtheme: Add Extra CSS/Style class to cells

### LINKED ISSUE #1475: Add option for extra class in xtheme layout cell
Plan is that after this merchant can define some css at snippet admin.

Tip: https://github.com/shuup/shuup/blob/master/shuup/xtheme/views/forms.py#L15

maximlion:
Added "Cell extra class" or "style_class" to cells. This is to be used
by merchants for custom cell CSS.

Fixes #1475


chessbr:
Replaced by https://github.com/shuup/shuup/pull/1490

PR #1495, opened 2018-09-14
General: add option to build static in synchronous mode

chessbr:
This prevents a large memory usage on deploy servers

No Refs


tulimaki:
Included in https://github.com/shuup/shuup/pull/1494

PR #1527, opened 2018-09-22
Fix some admin issues

tulimaki:
The biggest for sure is the "content misalignment" commit with changes to those common macros. But I actually clicked around a little extra and it really looks like good a move.


shuup/admin/modules/service_providers/views/_wizard.py
```diff
@@ -14,7 +14,7 @@
 
 
 class ServiceProviderTypeForm(forms.Form):
-    providers = forms.CharField(label=_("Provider"), required=False)
+    providers = forms.CharField(label=_("Provider"), required=False, widget=forms.HiddenInput())
```
chessbr: we need to add some margin here:
image

PR #1528, opened 2018-09-22
Product filter

soulilya:
Hallo! There is new module product filter. Support filtering with manufacturers, categories, basic attributes and custom attributes. Filters may me chosen from admin panel. Theme is very basic, tested with classic gray theme. Supports vertical and horizontal layouts.


shuup/front/apps/product_filter/views.py
```diff
@@ -0,0 +1,266 @@
...
+        context['filter_form'] = self.form
+        context['layout'] = 'vertical'
+        del self.form.fields['sort']
+        del self.form.fields['q']
+
+        try:
+            settings = BasicFilterSettingsModel.objects.get(
+                shop=get_shop(self.request),
+            )
+            print(settings.layout)
```
JsseL: ☝️

shuup/front/apps/product_filter/views.py
```diff
@@ -0,0 +1,266 @@
...
+                'initial': self.get_initial(),
+            }
+        )
+
+    def get_initial(self):
+        try:
+            objects = AttributesFilterSettingsModel.objects.filter(
+                shop=get_shop(self.request)
+            )
+        except AttributesFilterSettingsModel.DoesNotExist:
```
tulimaki: Filter never throws does not exist. Same applies to rest of the queryset with only filter which returns just empty list when there is no matches.

shuup/front/apps/product_filter/views.py
```diff
@@ -0,0 +1,266 @@
...
+    priority = 3
+    name = 'attributes_settings'
+    form = AttributesFilterSettingsForm
+
+    def get_form_defs(self):
+        try:
+            instance = AttributesFilterSettingsModel.objects.filter(
+                shop=get_shop(self.request)
+            ).first()
+        except AttributesFilterSettingsModel.DoesNotExist:
```
tulimaki: This shouldn't be raised either. `instance` should be `None` when no attribute filters settings are found.

shuup/front/apps/product_filter/tests.py
```diff
@@ -0,0 +1,3 @@
+from django.test import TestCase
```
tulimaki: This file is not needed, but for sure tests are welcome under `shuup_tests` or  `shuup_tests.front`.

shuup/front/apps/product_filter/forms.py
```diff
@@ -0,0 +1,387 @@
...
+                    ])
+                c += 1
+
+        return attributes
+
+    def get_price(self):
+        price_ranges = []
+        max_price = ShopProduct.objects.all().aggregate(Max('default_price_value'))
+
+        max_value = round(max_price['default_price_value__max'])
```
tulimaki: I would guess that this fails without products. Also use shop filter above to get right products for the current shop.

shuup/front/apps/product_filter/forms.py
```diff
@@ -0,0 +1,387 @@
...
+                            query_values[0], query_values[1]
+                        )}
+                    )
+
+        return query_dict
+
+    def get_filters(self, request, data):
+        query = {}
+        data.pop("q")
+        data.pop("sort")
```
tulimaki: Probably not right place to modify the data. Well with for me this caused key error since the sort was missing from data.

shuup/front/apps/product_filter/views.py
tulimaki: For these run `flake8`, `isort` and check the license headers for new files. You can also find this info from those failed Travis jobs.

soulilya:
Hm... I check with zero.

tulimaki:
Replaced by https://github.com/shuup/shuup/pull/1534

PR #1530, opened 2018-09-24
Admin: fix flex-wrapper and picotable item info

### LINKED ISSUE #1529: Admin UI: Fix action buttons for mobile
When using mobile devices, such as iPhone width 320px of width, there would be almost no space to buttons and title. I guess buttons could be either inside a collapsible button or even wrap to a new line
image from ios

tulimaki:
Fixes #1529


chessbr:
I guess this is the expected alignment:
image

This is what is current result:
image

PR #1537, opened 2018-09-25
WIP: Make the menu retractable & add babel-polyfill

### LINKED ISSUE #1501: Admin UI: Add a way to hide (and show) the menu
This could be done like in twitch but with a difference of us toggling between the "mobile mode" and "desktop mode" (instead of that minified menu twitch does). 

If any questions ping @Pikkupomo.

As a related issue, see https://github.com/shuup/shuup/issues/1499

maximlion:
Menu is now retractable on all devices.
`Array.from` in our javascript was breaking the menu completely for older
browsers, adding babel polyfill fixes that.
Fixes #1501

screenshot from 2018-09-24 20-34-38

screenshot from 2018-09-24 20-34-48

### DIFF
--- a/shuup/admin/static_src/base/js/main-menu.js
+++ b/shuup/admin/static_src/base/js/main-menu.js
@@ -58,15 +58,16 @@ const handleMainMenu = () => {
       mainMenu.classList.remove('open');
     });
   }
-
+  // media query change
+  if ($(window).width() >= 1024) {
+    $("body").toggleClass("menu-open");
+    $('.main-menu').addClass('open');
+  }
   if (toggleBtn) {
     toggleBtn.addEventListener('click', () => {
-      if (mainMenu.classList.contains('open')) {
-        mainMenu.classList.remove('open');
-      } else {
-        mainMenu.classList.add('open');
-        hideMainMenu();
-      }
+      $("body").toggleClass("menu-open");
+      $('.main-menu').toggleClass('open');
+
     });
   }
 }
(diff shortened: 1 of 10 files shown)


shuup/admin/static_src/base/js/main-menu.js
```diff
@@ -58,14 +58,42 @@ const handleMainMenu = () => {
...
+      openAdminMenu();
+    } else {
+      closeAdminMenu();
+    }
+  }
+
+  window.addEventListener("resize", widthChange);
+  if ($(window).width() >= 1024) {
+    $('.main-menu').addClass('open');
+  }
```
Pikkupomo: RE: all above:
`body` can take `class` definitions as well. Which means, you can add (for example) `menu-open` class to body. So, your togglemenu could only do `$("body").toggleClass("menu-open")` instead of all this. Maybe?

Then, in styles define `.menu-open #top-header { left: xxx, width: yyy }` etc.

shuup/admin/static_src/base/scss/shuup/navigation.scss
```diff
@@ -121,7 +121,7 @@
     .nav-menu-button {
         padding: 0;
         margin: 0;
-        display: block;
+        display: block !important;
```
Pikkupomo: Why?

maximlion: Thought it was needed but was obviously wrong. The only thing that was blocking the button from showing was the bootstrap d-lg-none.

PR #1588, opened 2018-10-04
 Admin: fix order creator form

chessbr:
### Admin: fix contacts select2 search and picotable
- Allow superuser selecting any contact without shop filtering
- Show the contact type correctly on contact list view when mobile device

### Admin: fix order creator form
- Make scroll to top only render when current value has really changed, preventing calling thousands of redraws with no
  reason
- Make buttons below scroll to top button container clickable
- Make order creator Back button style outline

Refs POS-2626


chessbr:
Replacled by https://github.com/shuup/shuup/pull/1589

PR #1589, opened 2018-10-05
Admin: fix order creator form

chessbr:
### Admin: fix contacts select2 search and picotable
- Allow superuser selecting any contact without shop filtering
- Show the contact type correctly on contact list view when mobile device

### Admin: fix order creator form
- Make scroll to top only render when current value has really changed, preventing calling thousands of redraws with no
  reason
- Make buttons below scroll to top button container clickable
- Make order creator Back button style outline
- Fix styles for small screens
- Fix bottom bar width when menu is opened and closed on large screen sizes
- Fix order creator by showing shop select correctly

Refs POS-2626/POS-2627/ENT-2580

### DIFF
--- a/shuup/admin/modules/contacts/views/list.py
+++ b/shuup/admin/modules/contacts/views/list.py
@@ -118,7 +118,7 @@ class ContactListView(PicotableListView):
         :type instance: shuup.core.models.Contact
         """
         bits = filter(None, [
-            item.get("type"),
+            self.get_type_display(instance),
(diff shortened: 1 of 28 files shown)


tulimaki:
Allow superuser selecting any contact without shop filtering. The contacts have to be filtered based on the shop selected for order. At the order creator there is shop selection when the staff member has multiple shops available. At least we need to talk about this before mergeable.

Issue #1612, opened 2018-10-16
Fix maintenance view


chessbr:
Check scripts errors when shop is in maintenance mode
![image](https://user-images.githubusercontent.com/8770370/47012107-de5fad80-d119-11e8-8a99-2cbf75efde97.png)

PR #1620, opened 2018-10-20
Front: Add missing logo-to-text image to admin

### LINKED ISSUE #1617: Xtheme admin tools live/maintenance badge doesn't seem to work
I would assume this `badge-success` do some coloring for badge. Same goes for `badge-warning` and for maintenance mode. Make the badge green in live-mode and red on maintenance-mode.

maximlion:
- Front: Add missing logo-to-text image to admin. The image is used by maintenance.jinja hence shuup logo during maintenance was invisible. Regarding #1612, I see no script errors on maintenance page so I assume that was fixed.
- Front: Add color to badge-success
image
Fixes #1617


chessbr:
@maximlion Shouldn't we use new logo?

Issue #1628, opened 2018-10-25
Admin: Minor usability issue


Pikkupomo:
When editing an object, for example, carousel advanced options, you are forced to scroll down. When scrolling, the "save" button disappears on the top. 

Huge quality of life improvement would be to show "save" button fixed always when it's not visible on top anymore. This is how it's usually done and it prevents a lot of scrolling. 

Minor thing, but a huge improvement.

PR #1632, opened 2018-10-28
Admin UI: Move save button to bottom right corner when not visible

maximlion:
Item save button is now moved to the bottom right corner when it's not visible.

image

Refs #1628

### DIFF
--- a/shuup/admin/static_src/base/js/form-utils.js
+++ b/shuup/admin/static_src/base/js/form-utils.js
@@ -69,6 +69,23 @@ window.clearErrors = function ($form) {
     $form.parent().find(".errors").empty();
 };
 
+function inViewport (el) {
+
+    var r, html;
+    if ( !el || 1 !== el.nodeType ) { return false; }
+    html = document.documentElement;
+    r = el.getBoundingClientRect();
+
+    return ( !!r
+      && r.bottom >= 0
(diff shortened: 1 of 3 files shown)


chessbr:
I guess instead of making only the save button fixed, why not making the entire toolbar?
something like this:

image

What do you think @Pikkupomo ?

Pikkupomo:
having the entire toolbar fixed causes the end user to have less screen space in use. having the buttons only fixed bottom right might be better? Not sure, I'm not an UI/UX master :P  @chessbr @maximlion

tulimaki:
Yeah maybe this saving space come in handy with tablet and mobile. @chessbr check one more time if this @maximlion solution can be accepted and if it looks so then let's just merge and move on.

shuup/admin/static_src/base/js/form-utils.js
```diff
@@ -69,6 +69,43 @@ window.clearErrors = function ($form) {
...
+    return ( !!r
+      && r.bottom >= 0
+      && r.right >= 0
+      && r.top <= html.clientHeight
+      && r.left <= html.clientWidth
+    );
+
+}
+
+window.onscroll = function() {
```
chessbr: add this event handler inside the block below to ensure all script were loaded before doing anything

Issue #1631, opened 2018-10-27
Admin order creator broken in multishop installation


tulimaki:
Tips:
`keyPrefix:interpolate("order_creator_shop-%s:",[e.shops[0].id`
`vendor.js:169 Uncaught TypeError: Cannot read property 'id' of undefined`

PR #1633, opened 2018-10-28
Admin: Fix order creation errors

maximlion:
- First error was being caused by bad queries, the deleted query & if statement can already be found in Shop.get_for_user.

- Second error was being caused by purgeAll, a function that was removed in v4 of redux-persist. - https://github.com/rt2zz/redux-persist/issues/373

Refs #1631

### DIFF
--- a/shuup/admin/modules/orders/views/edit.py
+++ b/shuup/admin/modules/orders/views/edit.py
@@ -178,8 +178,6 @@ class OrderEditView(CreateOrUpdateView):
     def get_config(self):
         order = self.object
         shop_queryset = Shop.objects.get_for_user(self.request.user).filter(status=ShopStatus.ENABLED)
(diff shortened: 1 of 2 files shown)


shuup/admin/modules/orders/views/edit.py
```diff
@@ -178,8 +178,6 @@ def get_toolbar(self):
     def get_config(self):
         order = self.object
         shop_queryset = Shop.objects.get_for_user(self.request.user).filter(status=ShopStatus.ENABLED)
-        if getattr(self.request.user, "is_superuser", False):
```
chessbr: why removing this? you SHOULD filter shops by current user, otherwise user will see shops that he is not allowed to

chessbr: oh, I see, `get_for_user` does the trick

PR #1634, opened 2018-11-05
Fix parler cache

### LINKED ISSUE #1323: Fix Parler integration

chessbr:
### Front: cache instances instead of ids in template helpers

IDs were used because of a bug of Django Parler which was fixed in django-parler/django-parler#228
As the fix was not release in PyPi, use the last commit of GitHub repository available.

Fixes #1323
Refs ENT-2729

### Core: improve context cache by preventing new queries

Context cache was creating new user group queries everytime for the same context object.
Cache the groups inside the instance so we can reuse that for the same request

Refs ENT-2729

### General: improve caches and optimize queries
- Use shop product cache in templates
- Prefetch and select related objects in several queries to prevent a lot of db queries

Refs ENT-2729

### DIFF
--- a/shuup/front/template_helpers/general.py
+++ b/shuup/front/template_helpers/general.py
@@ -125,12 +125,13 @@ def get_listed_products(context, n_products, ordering=None, filter_dict=None,
     return products_qs[:n_products]
 
 
-def _can_use_cache(product_ids, shop, customer):
+def _can_use_cache(products, shop, customer):
     """
     Check whether the cached products can be still used
 
     If any of the products is no more orderable refetch the products
     """
+    product_ids = [prod.pk for prod in products]
     for supplier in Supplier.objects.filter(shops__in=[shop]):
         for sp in ShopProduct.objects.filter(product__id__in=product_ids, shop=shop):
             if not sp.is_orderable(supplier, customer=customer, quantity=sp.minimum_purchase_quantity):
@@ -142,7 +143,7 @@ def _can_use_cache(product_ids, shop, customer):
 def get_best_selling_products(context, n_products=12, cutoff_days=30, orderable_only=True, sale_items_only=False):
     request = context["request"]
 
-    key, product_ids = context_cache.get_cached_value(
+    key, products = context_cache.get_cached_value(
         identifier="best_selling_products",
         item=cache_utils.get_best_selling_products_cache_item(request.shop),
         context=request,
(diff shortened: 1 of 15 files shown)


shuup/front/template_helpers/general.py
```diff
@@ -125,37 +125,23 @@ def get_listed_products(context, n_products, ordering=None, filter_dict=None,
     return products_qs[:n_products]
 
 
-def _can_use_cache(product_ids, shop, customer):
```
chessbr: I removed this since we are bumping context cache for these template helpers correctly now

tulimaki:
Issues with admin browser tests shouldn't be part of this pr.

PR #1637, opened 2018-11-06
Cache improvements

chessbr:
### Core: fix context cache when key is customer

Make sure to change the key to customer_groups when creating the cache key

### Front: cache thumbnail images when using template tags

Cache thumbnail images to prevent excessive thumbnails queries between requests

### Front: improve template helper by preventing unnecessary db query

Refs ENT-2729

### DIFF
--- a/shuup/core/utils/context_cache.py
+++ b/shuup/core/utils/context_cache.py
@@ -184,44 +184,69 @@ def get_cache_key_for_context(identifier, item, context, **kwargs):
     return "%s:%s_%s" % (namespace, identifier, hash(frozenset(items.items())))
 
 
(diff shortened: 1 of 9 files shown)


shuup/core/utils/context_cache.py
```diff
@@ -187,19 +187,22 @@ def get_cache_key_for_context(identifier, item, context, **kwargs):
 def _get_items_from_context(context):   # noqa (C901)
     items = {}
     if hasattr(context, "items"):
-        for k, v in six.iteritems(context):
+        for k, value in six.iteritems(context):
```
tulimaki: Might as well make the "k" to "key" too.

PR #1638, opened 2018-11-08
Admin: Make only active shops visible in search

maximlion:
Refs ENT-2750

### DIFF
--- a/shuup/admin/modules/shops/__init__.py
+++ b/shuup/admin/modules/shops/__init__.py
@@ -18,7 +18,7 @@ from shuup.admin.utils.urls import (
     admin_url, derive_model_url, get_edit_and_list_urls, get_model_url
 )
 from shuup.admin.views.home import SimpleHelpBlock
-from shuup.core.models import Shop
+from shuup.core.models import Shop, ShopStatus
 
 
 class ShopModule(AdminModule):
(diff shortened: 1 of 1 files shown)


shuup/admin/modules/shops/__init__.py
```diff
@@ -85,7 +85,7 @@ def get_search_results(self, request, query):
 
         minimum_query_length = 3
         if len(query) >= minimum_query_length:
-            shops = Shop.objects.get_for_user(request.user).filter(translations__name__icontains=query)
+            shops = Shop.objects.get_for_user(request.user).filter(translations__name__icontains=query, status=1)
```
JsseL: `status=1`? Perhaps you have some status constant|enum|definition|whatever somewhere that could be used instead of magical number `1`?

tulimaki:
screen shot 2018-11-09 at 3 41 57 pm
 
Here the diff coverage is the important thing. Ideally it should be over 90% in every single PR that touches files included in coverage. Also try to avoid creating new partials.

shuup/admin/modules/shops/__init__.py
```diff
@@ -85,7 +85,8 @@ def get_search_results(self, request, query):
 
         minimum_query_length = 3
         if len(query) >= minimum_query_length:
-            shops = Shop.objects.get_for_user(request.user).filter(translations__name__icontains=query)
+            shops = Shop.objects.get_for_user(request.user).filter(
+                translations__name__icontains=query, status=ShopStatus.ENABLED)
```
JsseL: "status is enabled".. Even I understand that. 👍

PR #1645, opened 2018-11-13
Xtheme: make content titles a bit more visible (ENT-2757)

Pikkupomo:
Refs ENT-2757


Pikkupomo:
image

tulimaki:
Much better than the previous one. Let's go with this for now.

PR #1647, opened 2018-11-19
Core: Fix product price range sorting

maximlion:
`get_price_info` alone would return 0 for products who have range variations. This adds a check for that which if positive returns the cheapest product variation's price.
Also enabled tests for basic sorting.

### DIFF
--- a/shuup/core/models/_products.py
+++ b/shuup/core/models/_products.py
@@ -494,7 +494,11 @@ class Product(TaxableItem, AttributableMixin, TranslatableModel):
         :type context: shuup.core.pricing.PricingContextable
         :rtype: shuup.core.pricing.Price
         """
-        return self.get_price_info(context, quantity).price
+        # Check if product has variations (children) if yes, return cheapest variation price
+        if self.is_variation_parent():
+            return self.get_cheapest_child_price_info(context, quantity).price
+        else:
+            return self.get_price_info(context, quantity).price
 
     def get_base_price(self, context, quantity=1):
         """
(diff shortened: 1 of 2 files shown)


shuup/core/models/_products.py
```diff
@@ -494,7 +494,11 @@ def get_price(self, context, quantity=1):
         :type context: shuup.core.pricing.PricingContextable
         :rtype: shuup.core.pricing.Price
         """
-        return self.get_price_info(context, quantity).price
+        # Check if product has variations (children) if yes, return cheapest variation price
+        if not self.get_priced_children(context, quantity):
```
chessbr: I guess this is a super expensive check.. IMO check the product mode instead. If it is a parent, then return the cheapest child. Thoughts @tulimaki ?

shuup/core/models/_products.py
```diff
@@ -494,7 +494,11 @@ def get_price(self, context, quantity=1):
...
         :rtype: shuup.core.pricing.Price
         """
-        return self.get_price_info(context, quantity).price
+        # Check if product has variations (children) if yes, return cheapest variation price
+        if self.is_variation_parent():
+            return self.get_cheapest_child_price_info(context, quantity).price
+        else:
+            return self.get_price_info(context, quantity).price
 
     def get_base_price(self, context, quantity=1):
```
tulimaki: How about `get_base_price` and `get_price_info`? Maybe this is something that the `shuup.core.pricing.get_price_info` should deal with. Also I don't like that we do this for front product sort which likely has some bigger issues in it.

tulimaki: Maybe even something pricing module should deal with since that is likely only thing that is responsible how this kind of situation is handled. Maybe in some implementations or projects the cheapest price is not shown but price for some specific variation etc...

Issue #1653, opened 2018-11-26
Add option to remove coupon code from basket


tulimaki:
There isn't option to remove coupon code from basket. Should likely be done with basket command at basket view [this](https://github.com/shuup/shuup/blob/master/shuup/core/basket/commands.py#L177).

PR #1660, opened 2018-12-04
Front: Enable coupon removal in basket

maximlion:
Coupons can now be removed in basket. Added "value" field to basketline API so I could get coupon code in JS. (which I needed to send the AJAX request)

Refs #1653


chessbr:
Replaced by https://github.com/shuup/shuup/pull/1665

PR #1676, opened 2018-12-13
Set default value for CMS & Exclude deleted categories 

### LINKED ISSUE #1661: Admin: categories list view wrong queryset
Currently, all categories except deleted are listed in CategoryListView, although we have Deleted status filter option in the picotable which when selected, never shows anything.

Perhaps we should:
- list all categories and set the filter as Visible by default
- remove the deleted filter option

### LINKED ISSUE #1674: Make CMS page available by default
Now when creating new CMS page from admin it is required to separately remember to set "available from" date. Consider autofilling field for new pages.

maximlion:
- CMS: Set default value to current date for Page's available form field 

Fixes #1674 

- Admin: Exclude deleted categories from category status filter 

Fixes #1661


shuup/simple_cms/models.py
```diff
@@ -64,6 +64,7 @@ def for_shop(self, shop):
 class Page(MPTTModel, TranslatableModel):
     shop = models.ForeignKey("shuup.Shop", verbose_name=_('shop'))
     available_from = models.DateTimeField(
+        default=now,
```
tulimaki: Would this work better? https://docs.djangoproject.com/en/1.11/ref/models/fields/#django.db.models.DateField.auto_now_add

shuup/admin/modules/categories/views/list.py
```diff
@@ -17,6 +17,8 @@
 
 class CategoryListView(PicotableListView):
     model = Category
+    choices = list(CategoryStatus.choices())
+    choices.pop(2)  # Remove DELETED status
```
tulimaki: `choices = [status for status in CategoryStatus if status != CategoryStatus.DELETED]` this should be better.

tulimaki: Or just do simple list based on valid statuses to make things more readable since all choices are actually visible here.

shuup/simple_cms/models.py
maximlion: Not really no, the problem with that is that it automatically sets the "Available from" date whether you assign it a value or not. This becomes an issue when the merchant wants the page to become available in 2 days, auto_now_add will set it to today's date.

maximlion: I guess we could just leave the field empty and set it to today's date if it's sent empty (just like we do it now).

chessbr: but will it possible to clear the field if I want?

maximlion: ### Yes but the value will still be set to default by our backend. Perhaps, we should, set the form's initial value to be today's date, and if it's deleted by the merchant the available from value should be null?

chessbr: that is what I think, at least.

maximlion: @chessbr This works as we talked ^^, If default date is deleted from the field, the available from will be set to null.

PR #1694, opened 2019-01-03
Introduce labels and new fields to shop and service

tulimaki:
* Introduce `shuup.core.models.Label`. Use it to label shops and services. For shop this will come helpful in multishop solutions where you might need to group different shops together (maybe for filtering shops in a marketplace front). For services there might be need to front do some special tricks for certain services. For example you might want to separate shipping methods to pickups and deliveries and then in "thank you" page show slightly different content based on shipping method chosen (behavior components is not suitable for this based on our current enterprise solutions).
* Add description and short description for shops.


shuup/core/models/_labels.py
```diff
@@ -0,0 +1,38 @@
...
+@python_2_unicode_compatible
+class Label(TranslatableShuupModel):
+    identifier = InternalIdentifierField(unique=True, max_length=128, editable=True)
+    created_on = models.DateTimeField(auto_now_add=True, editable=False, verbose_name=_('created on'))
+    modified_on = models.DateTimeField(auto_now=True, editable=False, db_index=True, verbose_name=_('modified on'))
+
+    translations = TranslatedFields(
+        name=models.CharField(
+            max_length=64, verbose_name=_("name"),
+            help_text=_("The shop name. This name is displayed throughout admin.")
```
chessbr: wrong help text

Issue #1682, opened 2018-12-17
Wizard: Silent fail fails if no information given


Pikkupomo:
When firing up a store first time, the wizard is shown. The first page of the wizard "Add an address..." fails silently if no info has been entered:

![image](https://user-images.githubusercontent.com/2863405/50116294-a1d62e00-01fe-11e9-88c2-ee76646be349.png)

Errors not shown:
![image](https://user-images.githubusercontent.com/2863405/50116370-ccc08200-01fe-11e9-9abf-bba10f6a5e71.png)

PR #1695, opened 2019-01-03
Wizard: Fix field errors not showing

maximlion:
image

Refs #1682


shuup/admin/static_src/base/js/form-utils.js
```diff
@@ -56,8 +56,9 @@ window.renderFormErrors = function($form, errors) {
             if(fieldName === "__all__") {
                 $form.parent().find(".errors").append('<div class="alert alert-danger">' + fieldErrors + '</div>');
             } else {
-                let $field = $form.find(":input[name='" + formName + "-" + fieldName + "']").parent(".form-group");
-                $field.append("<span class='help-block error-block'>" + fieldErrors + "</span>").addClass("has-error");
+                var x = document.getElementsByName(formName + "-" + fieldName)[0];
```
chessbr: Check this solution:
```js
window.renderFormErrors = function($form, errors) {
    Object.keys(errors).forEach((formName) => {
        const formErrors = errors[formName];
        Object.keys(formErrors).forEach((fieldName) => {
            const fieldErrors = formErrors[fieldName].join(" ");
            if(fieldName === "__all__") {
                $form.parent().find(".errors").append('<div class="alert alert-danger">' + fieldErrors + '</div>');
            } else {
                const field = $("[name=" + formName + "-" + fieldName + "]");
                field.closest(".form-group").addClass("has-error");
                field.closest(".form-group").append("<span class='help-block error-block'>" + fieldErrors + "</span><br>");
            }
        });
    });
};
```

maximlion: Changed that up a bit to match the styles.

PR #1697, opened 2019-01-08
Simple CMS: fix page links plugin to use reverse url

chessbr:


tulimaki:
Let's see what Travis have to say and then merge. Main point in this change is that not in every project simple-cms pages urls is under "/" directly.

PR #1701, opened 2019-01-09
Update copyright year from license headers

tulimaki:


tulimaki:
@chessbr hold on with merging. There is more changes incoming.

PR #1712, opened 2019-01-19
Front: Improve  company registration logic

maximlion:
The problem was that when going to Company Information, it was throwing a full 404 (not the django one),
The button was being shown despite Shop not allowing company registration, which was causing the above mentioned error.
This ensures button is shown if customer is a company contact and if company registration is allowed.
Django 404 error is thrown instead of HttpResponseNotFound.
No refs

### DIFF
--- a/shuup/admin/modules/products/views/delete.py
+++ b/shuup/admin/modules/products/views/delete.py
@@ -15,18 +15,19 @@ from django.utils.translation import ugettext as _
 from django.views.generic import DetailView
 
 from shuup.admin.utils.urls import get_model_url
-from shuup.core.models import Product
+from shuup.core.models import ShopProduct
 
 
 class ProductDeleteView(DetailView):
-    model = Product
+    model = ShopProduct
     context_object_name = "product"
 
     def get(self, request, *args, **kwargs):
-        return HttpResponseRedirect(get_model_url(self.get_object(), shop=self.request.shop))
+        product = self.get_object().product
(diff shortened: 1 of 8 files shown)


shuup/front/apps/customer_information/dashboard_items.py
```diff
@@ -35,8 +36,9 @@ def show_on_dashboard(self):
         return False  # Don't show this on menu since we already have all the information there
 
     def show_on_menu(self):
-        # Only show this on menu if customer is company
-        return isinstance(self.request.customer, CompanyContact)
+        # Only show this on menu if customer is company and if company registration is allowed
+        is_registration_allowed = allow_company_registration(self.request.shop)
+        return (isinstance(self.request.customer, CompanyContact) and is_registration_allowed)
```
chessbr: I don't get it.. If the contact is a already a company, this shouldn't be an issue. Imagine that the shop has this option enabled and suddenly the merchant disables it, what happens to company contacts that we already created? I guess this is not an issue.

shuup/front/apps/customer_information/dashboard_items.py
```diff
@@ -35,7 +35,7 @@ def show_on_dashboard(self):
         return False  # Don't show this on menu since we already have all the information there
 
     def show_on_menu(self):
-        # Only show this on menu if customer is company
+        # Only show this on menu if customer is company and if company registration is allowed
```
chessbr: fix this comment

PR #1713, opened 2019-01-20
Bump product cache when editing with mass action

### LINKED ISSUE #1710: Admin: product visibility mass action doesn't bump caches
After products are cached in xtheme plugins (like highlight) and products are hided through mass action in product admin, the storefront just doesn't change. Looks like we are missing to bump caches.

maximlion:
`.update()` is a direct SQL statement so it wasn't calling save() at all, hence it didn't trigger `post_save` signals at all. This iterates through the whole query and saves changes separately for each item. (Recommended method by Django docs https://docs.djangoproject.com/en/2.1/topics/db/queries/#updating-multiple-objects-at-once )

Fixes #1710

### DIFF
--- a/shuup/admin/modules/products/mass_actions.py
+++ b/shuup/admin/modules/products/mass_actions.py
@@ -14,6 +14,7 @@ from six import string_types
 
 from shuup.admin.modules.products.views.list import ProductListView
 from shuup.admin.modules.settings.view_settings import ViewSettings
+from shuup.admin.shop_provider import get_shop
 from shuup.admin.utils.picotable import (
     PicotableFileMassAction, PicotableMassAction, PicotableRedirectMassAction
 )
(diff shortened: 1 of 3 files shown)


chessbr:
make sure to make codecov happy too

shuup/admin/modules/products/mass_actions.py
```diff
@@ -28,7 +28,9 @@ def process(self, request, ids):
         query = Q(product__pk__in=ids)
         if isinstance(ids, string_types) and ids == "all":
             query = Q()
-        ShopProduct.objects.filter(query).update(visibility=ShopProductVisibility.ALWAYS_VISIBLE)
+        for shop_product in ShopProduct.objects.filter(query).iterator():
```
chessbr: I guess we have a bug here.. when the merchant selects ALL, it will basically set ALL products of ALL shops :scream: 
this is completely wrong.. fix this by adding the admin shop (from the request) in the basic query and make sure to test this

chessbr:
Coverage still not good, please test the other mass actions you've changed and let's move on

maximlion:
Replaced by https://github.com/shuup/shuup/pull/1715

PR #1715, opened 2019-01-22
Bump product cache when using mass action, limit mass action changes

### LINKED ISSUE #1710: Admin: product visibility mass action doesn't bump caches

maximlion:
.update() is a direct SQL statement so it wasn't calling save() at all, hence it didn't trigger post_save signals at all.
This iterates through the whole query and saves changes separately for each item. (Recommended method by Django docs)

Fixes #1710
Replaces https://github.com/shuup/shuup/pull/1713


chessbr:
Replaced by https://github.com/shuup/shuup/pull/1716

PR #1734, opened 2019-02-07
Admin: improve media upload by adding specific permission

chessbr:
### Add admin url just to upload files
Now it is possible to enable dropzone and allow only uploads and disable file browse. One can remove browse permission and allow only the upload one.

### Admin: fix bug in FormPart subclassing check
`isinstance` was used to check whether a class contains a BaseFormPart inherirance.
This is not right as the correct function is `issubclass` when checking class objects.

Refs MYS-25


shuup/admin/forms/widgets.py
```diff
@@ -114,10 +114,14 @@ def render(self, name, value, attrs=None, renderer=None):
 
 
 class FileDnDUploaderWidget(Widget):
-    def __init__(self, attrs=None, kind=None, upload_path="/", clearable=False):
+    def __init__(self, attrs=None, kind=None, upload_path="/", clearable=False, browsable=True, upload_url=None):
```
tulimaki: Where is this browsable parameter passed?

PR #1762, opened 2019-02-24
Generate missing core migrations

tulimaki:


tulimaki:
@chessbr was not expecting these. Is it also generating these for you?

tulimaki:
False alert. Will re-open if something comes up.

PR #1772, opened 2019-03-05
Core: add managed stock flag in ProductStockStatus

chessbr:
- Remove stock behavior attribute from product model
- Allow disabling stock management for specific products

Refs MYS-52


tulimaki:
@jovan-shuup add some more tests around this as separate PR.

PR #1777, opened 2019-03-07
Stock management: unit tests fixed

jovan-shuup:
Fixed two unit tests.

- `admin/test_contact_price_display .py`fixed because it had a bug
- the several other test were falling due to changes in code
- the fix required to change the code for StockCount get or create invocation for Simple Supplier


chessbr:
Included in by https://github.com/shuup/shuup/pull/1772

PR #1796, opened 2019-03-15
Front: tune variation product orderability check at order detail

tulimaki:


shuup/front/utils/product.py
```diff
@@ -114,7 +117,7 @@ def _unpack_orderable_variation_children_from_cache(cached_value):
     return (orderable_variation_children, orderable)
 
 
-def get_orderable_variation_children(product, request, variation_variables):    # noqa (C901)
+def get_orderable_variation_children(product, request, variation_variables, supplier=None):    # noqa (C901)
```
chessbr: I guess you can pass this supplier in the context cache fetch below to prevent returning wrong values when supplier changes

tulimaki: Thats not safe I think. Also this method is not open to world in reality (only place this is called is there at line 62. Maybe that shouldn't be optional but just in case.

chessbr: I mean, check line 124 below: `key, val = context_cache.get_cached_value(` here you don't pass the supplier, this way a different supplier can make the cache fetch a wrong value (from a different supplier).

tulimaki: You were correct :) Tests now covers this.

shuup_tests/front/test_product_price.py
```diff
@@ -104,3 +198,26 @@ def check(shop_product, input_value, expected_output):
...
+    product.sales_unit = sales_unit
+    product.save()
+
+    view = ProductPriceView()
+    view.request = rf.get('/')
+    assert 'quantity' not in view.request.GET
+
+    def check(shop_product, input_value, expected_output):
+        view.request.GET = dict(view.request.GET, quantity=input_value, unitType="not internal")
+        result = view._get_quantity(shop_product)
```
chessbr: shouldnt you do some check here with the expected value?

shuup_tests/front/test_product_detail_context.py
```diff
@@ -0,0 +1,77 @@
...
+#
+# This source code is licensed under the OSL-3.0 license found in the
+# LICENSE file in the root directory of this source tree.
+from __future__ import unicode_literals
+
+from decimal import Decimal
+
+import pytest
+import six
+import test_app
```
chessbr: what is this `test_app`?

shuup_tests/front/test_product_price.py
```diff
@@ -10,13 +10,14 @@
 from decimal import Decimal
 
 import pytest
+import six
+import test_app
```
chessbr: what is this `test_app`?

shuup_tests/front/test_product_detail_context.py
tulimaki: No idea. I will remove those and clean the imports.

shuup_tests/front/test_product_price.py
tulimaki: perhaps

tulimaki:
Fixd!

PR #1805, opened 2019-03-19
Fix campaigns a little

tulimaki:
Causes issues when generating migrations with different versions of Python. User does not have any role in making this field comma separated so maybe the validator is a little bit too much anyways. While at it fix the weekdays field.


tulimaki:
*Do not merge there is another campaign improvement incoming and I rebase this over it once merged.*

Issue #1783, opened 2019-03-11
Notify: fix broken admin styles


chessbr:
The notify app styles looks broken after the overhaul. Make sure they look good.

While at it, use html text editor on send email action.

chessbr:
Fixed

PR #1806, opened 2019-03-20
Notify: Fix toolbar button style

maximlion:
Notify: Fix toolbar button style & user serializer error

The user serializer bug was happening because a User model was being passed to the template
instead of just model's primary key. This ensures that doesn't happen.
image

Refs #1783, #1807


shuup/notify/admin_module/views/editor.py
```diff
@@ -58,6 +58,8 @@ def script_item_editor(request):
         except ValidationError as verr:
             form.add_error(None, verr)
         else:
+            if "recipient" in form.script_item.data:
+                form.script_item.data['recipient']['constant'] = form.script_item.data['recipient']['constant'].pk
```
jovan-shuup: Maybe would be better to identify how to put `pk` in `form.script_item.data['recipient']['constant']`directly so it comes in this point with proper value.. Although, I failed in trying to identify that point :)

PR #1808, opened 2019-03-20
Classic Gray: Add variation button option

Pikkupomo:
Adding this because I had 15 mins of downtime and we had a discussion about this earlier. 

This is just a suggestion on how this could be done. 

image

image


tulimaki:
Replaced by https://github.com/shuup/shuup/pull/1818

PR #1812, opened 2019-03-21
General fixes

chessbr:
### Admin: show help block that nothing is needed to be configured
When there is no help block to be configured, show a block telling that it is all set.

### Core: improve the available until orderability error message
Make it clear that the product is not available for that current date

### Simple supplier: do not spam spam alert notifications inside a single minute
Prevent too much notifications inside a minute for a given product and supplier

Refs MYS-93


tulimaki:
Replaced by https://github.com/shuup/shuup/pull/1814 (just a rebase). Btw like the branch name 💯

PR #1815, opened 2019-03-22
Admin: improve shipment creation

tulimaki:
Improve shipment creation for situation when same product is ordered by multiple supplier. Shipments are still created per supplier and this just adjusts the form to be more clear what is shipped/unshipped for each supplier.


tulimaki:
Replaced by https://github.com/shuup/shuup/pull/1817. I think this is bad idea all thought maybe some day we need that supplier summary thingy for orders for something. Maybe.

PR #1816, opened 2019-03-22
See cart object in admin (TCP-24)

benjdelt:
Previously, there was an automatic 2 hour buffer between the last update of a cart and it being displayed in Orders > Carts. The idea was to prevent the active carts from being rendered. I created a form part in Shops that allows the user to select a number of hours on inactivity before the cart is displayed in the admin menu. The default is 0 which renders the carts immediately.

### DIFF
--- a/shuup/front/admin_module/carts/views/_list.py
+++ b/shuup/front/admin_module/carts/views/_list.py
@@ -13,6 +13,7 @@ from django.utils.html import escape
 from django.utils.timezone import now
 from django.utils.translation import ugettext_lazy as _
 
+from shuup import configuration
 from shuup.admin.utils.picotable import (
     ChoicesFilter, Column, DateRangeFilter, MultiFieldTextFilter, RangeFilter
 )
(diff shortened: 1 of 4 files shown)


shuup/front/admin_module/carts/views/_list.py
```diff
@@ -58,7 +59,8 @@ def get_queryset(self):
         """
         Ignore potentially active carts, displaying only those not updated for at least 2 hours.
         """
-        cutoff = now() - datetime.timedelta(hours=2)
+        hours = configuration.get(self.request.shop, "cart_update_delay", 0) or 0
```
tulimaki: Instead `self.field_name` define variable `"CART_UPDATE_DELAY_CONF_KEY" = "shuup_front_cart_update_delay"`. Use that variable then here and `form_parts.py`. No need to `or 0` (that first zero is already default)

tulimaki:
Replaced by https://github.com/shuup/shuup/pull/1819

PR #1826, opened 2019-03-27
Front: Variation image changer

Pikkupomo:
When changing variations in front, the images never changed. This PR adds a feature where images will change based on the variation children.


chessbr:
It works, but could this be part of this fix? https://github.com/shuup/shuup/issues/1836

tulimaki:
I think we could tackle that #1836 part of this.

PR #1829, opened 2019-03-28
OrderStatusForm: fix form clean method

jovan-shuup:
Fix form clean method

Added unit tests.

Refs TCP-25


shuup/admin/modules/orders/views/status.py
```diff
@@ -34,9 +34,9 @@ def clean(self):
         if self.instance.pk and OrderStatusManager().is_default(self.instance):
             data = self.cleaned_data
             data["identifier"] = self.instance.identifier
-            data["role"] = self.instance.role
-            data["ordering"] = self.instance.ordering
-            data["is_active"] = self.instance.is_active
+            # data["role"] = self.instance.role
```
tulimaki: Commenting not needed. If these lines are not needed then just remove them.

jovan-shuup: Done

PR #1831, opened 2019-03-28
Admin: introduce extra info for items in picotable

chessbr:
Now it is possible to return extra info for each object in picotable list.

There are some special keys like `class` that is used to change the row class list for each item.

Refs MYS-97


chessbr:
image

PR #1833, opened 2019-03-29
Campaigns: fix product active campaigns form part

tulimaki:
Show active campaigns for currently active shop. Do not show catalog campaigns at all since catalog campaigns are not supported and will be deprecated on Shuup 2.0 (discounts-app) shoudl be used instead.

### DIFF
--- a/shuup/campaigns/admin_module/sections.py
+++ b/shuup/campaigns/admin_module/sections.py
@@ -8,13 +8,15 @@
 from django.utils.translation import ugettext_lazy as _
 
 from shuup.admin.base import Section
-from shuup.campaigns.models import BasketCampaign, CatalogCampaign
+from shuup.admin.shop_provider import get_shop
+from shuup.admin.supplier_provider import get_supplier
+from shuup.campaigns.models import BasketCampaign
 from shuup.core.models import ShopProduct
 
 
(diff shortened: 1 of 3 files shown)


shuup/campaigns/admin_module/sections.py
```diff
@@ -25,14 +27,17 @@ def visible_for_object(cls, product, request=None):
     @classmethod
     def get_context_data(cls, product, request=None):
         ctx = {}
-        shop = request.shop
+        shop = get_shop(request)
         try:
             shop_product = product.get_shop_instance(shop)
+            basket_campaigns = BasketCampaign.get_for_product(shop_product)
+            user = request.user
+            if not (getattr(user, 'is_superuser', False) and shop.staff_members.filter(id=user.pk).first()):
```
chessbr: what is the point here with this check? Is this right? You only check whether the user is staff member if he is a superuser? Does this make sense?

tulimaki: Should be or I guess.

tulimaki:
Fixed!

shuup/campaigns/admin_module/sections.py
chessbr: "I guess" :yum: super confident haha

PR #1846, opened 2019-04-02
Campaigns: add campaign supplier to discount lines

tulimaki:
When lines causes new lines to basket it is important that those discount lines are linked to correct supplier based on campaign.

### DIFF
--- a/shuup/campaigns/models/basket_line_effects.py
+++ b/shuup/campaigns/models/basket_line_effects.py
@@ -23,7 +23,7 @@ class BasketLineEffect(PolymorphicShuupModel):
 
     campaign = models.ForeignKey("BasketCampaign", related_name='line_effects', verbose_name=_("campaign"))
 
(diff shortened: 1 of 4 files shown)


shuup/campaigns/models/basket_line_effects.py
```diff
@@ -103,7 +107,7 @@ class DiscountFromProduct(BasketLineEffect):
     def description(self):
         return _("Select discount amount and products.")
 
-    def get_discount_lines(self, order_source, original_lines):
+    def get_discount_lines(self, order_source, original_lines, supplier):
         product_ids = self.products.values_list("pk", flat=True)
         campaign = self.campaign
         campaign_supplier = campaign.supplier if hasattr(campaign, "supplier") and campaign.supplier else None
```
chessbr: maybe this `campaign_supplier` should be the parameter and then fallback to the campaign supplier. The same with other places.

PR #1855, opened 2019-04-08
Admin: Fix bug in Campaign free product effect

maximlion:
`for error in shop_product.get_quantity_errors(self.cleaned_data["quantity"]):`  was throwing an error because `ignore_minimum`  in `get_quantity_errors` has not been set.

Refs ENT-3006


tulimaki:
I think the battle with tests was worth it. Good stuff @maximlion

Issue #1856, opened 2019-04-09
XSS flaw


asteinhauser:
[shuup.zip](https://github.com/shuup/shuup/files/3057302/shuup.zip)

tulimaki:
Let's continue this at https://github.com/shuup/shuup/pull/1857.

PR #1857, opened 2019-04-09
Fix XSS flaws through product short_description and description.

asteinhauser:
Fixing https://github.com/shuup/shuup/issues/1856

### DIFF
--- a/shuup/front/templates/shuup/front/macros/product.jinja
+++ b/shuup/front/templates/shuup/front/macros/product.jinja
@@ -53,7 +53,7 @@ Inheritance order for product macors:
                     {% endif %}
                 </div>
                 {% if show_description and product.short_description %}
-                    <p class="description">{{ product.short_description|safe }}</p>
+                    <p class="description">{{ product.short_description }}</p>
                 {% endif %}
             </div>
         </a>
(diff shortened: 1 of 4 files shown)


Pikkupomo:
Thank you for your contribution @asteinhauser! 

It is not, however, entirely clear to me how this change prevents XSS as the only ones who can change this value are the users with admin credentials and I think we should trust them and in the end, it's the users responsibility to ensure (s)he doesn't use JavaScript in the descriptions. 

If you have a project where you expose this field to the public, you can override your own templates in your project.

Not going to approve this as is, the description is missing and there is no test proving the XSS flaw is there. So please, provide steps on how to create XSS flaw using the `product.

asteinhauser:
There is no way to exploit it without admin credentials. If you don't consider it a problem, it's up to you. I just needed to report it, so that I can publish this XSS bug as part of evaluating the tool with which I discovered it and claim in my publication that I reported all published flaws to maintainers of vulnerable applications - as part of ethical compliance.

chessbr:
I personally doesn't consider this a XSS flaw as this to happen the administrator credentials should be leaked and if that happens it means all the platform is compromised.
But indeed, we can try to add the bleach to the project to "protect" from values that are used by the user which copied the value from somewhere else, for example.
IMO, this fix is invalid and this is not the proper way of handling this. You can convert this into an Issue with the steps and values to reproduce or fix this PR.

PR #1858, opened 2019-04-12
Fix issue 'Manager isn't available; 'auth.User' has been swapped for …

fmipsylone:
Impossible to swap django User model
 
…'users.User''. Remove all references of 'from django.contrib.auth.models import User' in shuup_tests and 'shuup.core.telemetry'.

Replacing all references to:
`from django.contrib.auth.models import User`

By:
```
from django.contrib.auth import get_user_model
User = get_user_model()
```


tulimaki:
Thank you!

PR #1863, opened 2019-04-14
Simple CMS: allow page owner see invisible page

tulimaki:
Allow page creator option to preview the page on site before publishing it.


shuup/simple_cms/models.py
```diff
@@ -52,6 +52,8 @@ def visible(self, shop, dt=None):
             Q(available_from__lte=dt) &
             (Q(available_to__gte=dt) | Q(available_to__isnull=True))
         )
+        if user and not user.is_anonymous():
+            q |= Q(created_by=user)
```
chessbr: this will make the user see even deleted pages

Issue #1841, opened 2019-04-02
Fix issue with picotable column values


tulimaki:
<img width="1471" alt="Screen Shot 2019-04-01 at 5 42 04 PM" src="https://user-images.githubusercontent.com/4940349/55368419-a02b3000-54a6-11e9-85e1-48b7d4a060cb.png">

Avoid rendering "shuup.CustomerTaxGroup.None". Instead use empty string or "-".

volodymyr-code:
https://github.com/shuup/shuup/pull/1868

in check_different_types func - Manager returns '' value
which we can display to the user

PR #1868, opened 2019-04-16
Fix issue with picotable column values

volodymyr-code:
https://github.com/shuup/shuup/issues/1841
in check_different_types func - Manager returns '' value
which we can display to the user


chessbr:
It might be wise to add a unit test to prove the bug doesn't happen anymore

volodymyr-code:
I can add unit test at this weekends

volodymyr-code:
> I can add unit test at this weekends

done

chessbr:
Thank you @volodymyr-code

PR #1871, opened 2019-04-19
New Dockerfile

apiotrowski312:
Hi guys. I updated your (3 years old) Dockerfile.
I prefer working with Docker more than virtualenv. So I created this Dockerfile for testing shuup_workbench, it follows your guide and creates superuser with admin/admin credentials.

I can create and add PR with docker-compose for easier development if you want.

PS: I hope this time CLA will work correctly :D

### DIFF
--- a/Dockerfile
+++ b/Dockerfile
@@ -1,16 +1,21 @@
-FROM node:0.12.5
-MAINTAINER Aarni Koskela <aarni.koskela@shuup.com>
-EXPOSE 8080
-ADD . /var/www/shuup/working_copy
+FROM node:11.14-slim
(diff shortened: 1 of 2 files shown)


tulimaki:
Seems to work @apiotrowski312. And yes should be improvement to old Dockerfile which might not even work. Have not tested that in a long time.

But hey before we merge this can you tweak those commit messages a little and maybe combine those commits into one. Thanks!

> I can create and add PR with docker-compose for easier development if you want.

Awesome, this would be appreciated.

apiotrowski312:
@tulimaki I changed commit messages as you suggested.

chessbr:
Thank you @apiotrowski312 !

PR #1872, opened 2019-04-19
Front: add new products page

tulimaki:
* /c/ which lists products for all visible categories

### DIFF
--- a/shuup/front/templates/shuup/front/product/category.jinja
+++ b/shuup/front/templates/shuup/front/product/category.jinja
@@ -5,23 +5,28 @@
 {% else %}
     {% extends "shuup/front/base.jinja" %}
     {% block extrameta %}
-    {{ macros.render_metadata(category, category.name, category.description, "category") }}
+    {% if category %}
+        {{ macros.render_metadata(category, category.name, category.description, "category") }}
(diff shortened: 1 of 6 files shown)


shuup/front/templates/shuup/front/product/category.jinja
```diff
@@ -5,23 +5,28 @@
...
-        {% call macros.breadcrumb() %}
-            {{ macros.render_breadcrumb_item("/", _("Home")) }}
-            {% for c in category.get_ancestors() if c.is_visible(customer=request.customer) %}
-                {{ macros.render_breadcrumb_item(shuup.urls.model_url(c), c.name) }}
-            {% endfor %}
-            {{ macros.render_breadcrumb_item(None, category.name, True) }}
-        {% endcall %}
+        {% if category %}
+            {% call macros.breadcrumb() %}
+                {{ macros.render_breadcrumb_item("/", _("Home")) }}
```
chessbr: IMO we should keep at least breadcrumbs to Home / Categories

tulimaki: It is better this way IMO. This was also my first idea, but it did look awkward

PR #1875, opened 2019-04-23
Multiple minor improvements

Pikkupomo:
1) Add `RequestSupplierProvider`. This provides an easy way to test supplier related requests
2) Expose `._data` to the public from `SourceLine`. It was strange that this was hidden after all. Basically, you could use the `to_json` but it's not always what you might want.
3) Ensure data is being carried from `SourceLine` to `OrderLine` properly.

### DIFF
--- a/shuup/core/order_creator/_creator.py
+++ b/shuup/core/order_creator/_creator.py
@@ -72,7 +72,10 @@ class OrderProcessor(object):
         order_line.verified = (not order_line.require_verification)
         order_line.source_line = source_line
         order_line.parent_source_line = source_line.parent_line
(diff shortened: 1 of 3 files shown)


shuupdeploy:
We merge this for beta-release. Tests will follow, but this actually tested manually and the current tests should indicate that this should not break any existing features. AKA LGTM but missing tests but we need to merge to further test this with correct environment.

tulimaki:
We merge this for beta-release. Tests will follow, but this actually tested manually and the current tests should indicate that this should not break any existing features. AKA LGTM but missing tests but we need to merge to further test this with correct environment.

PR #1878, opened 2019-04-25
Front: back to old user when stop impersonating

chessbr:
After user stop impersonating, the old user is the active one again.
This prevents the user to do login again.

Refs MOOS-26


tulimaki:
Not in our interests to satisfy Codecov here. Let's do it. Ping me if you think this is mistake or you are willing to improve the coverage around this. I am happy to help.

PR #1879, opened 2019-04-25
added tests for SourceLine extra_data

kkundanmal:


shuup_tests/core/test_order_creator.py
```diff
@@ -118,6 +118,46 @@ def test_order_creator(rf, admin_user):
...
+        supplier=get_default_supplier(),
+        quantity=1,
+        base_unit_price=source.create_price(10),
+        extra= extra_data
+    )
+   
+
+    creator = OrderCreator()
+    order = creator.create_order(source)
+    test_order_line = order.lines[0]
```
tulimaki: `TypeError: 'RelatedManager' object does not support indexing` and `order.lines.first()`.

shuup_tests/core/test_order_creator.py
```diff
@@ -118,6 +118,46 @@ def test_order_creator(rf, admin_user):
...
+        product=get_default_product(),
+        supplier=get_default_supplier(),
+        quantity=1,
+        base_unit_price=source.create_price(10),
+    )
+   
+
+    creator = OrderCreator()
+    order = creator.create_order(source)
+    test_order_line = order.lines[0]
```
tulimaki: Same as above.

shuup_tests/core/test_order_creator.py
kkundanmal: changed to order.lines.first()

PR #1880, opened 2019-04-27
Admin: Customize Shuup admin panel menu

volodymyr-code:


tulimaki:
- Consider making the 2nd level dragging are a little bit bigger
- Deprecate `MenuItem.subcategories` no need to group menu items based on menu category
- Maybe we could change the module from `shuup.admin.modules.admin_menu` to `shuup.admin.modules.menu`

shuup/admin/base.py
```diff
@@ -7,6 +7,7 @@
 # LICENSE file in the root directory of this source tree.
 from __future__ import unicode_literals
 
+import abc
```
tulimaki: Where is this used?

tulimaki:
```
diff --git a/shuup/core/utils/menu.py b/shuup/core/utils/menu.py
index e75ca29fa..70e98b91d 100644
--- a/shuup/core/utils/menu.py
+++ b/shuup/core/utils/menu.py
@@ -28,6 +28,9 @@ class MainMenuUpdater(object):
         """
         for item in self.menu:
             for child in self.updates.get(item["identifier"], []):
-                if child not in item["children"]:
-                    item["children"].append(child)
+                if child not in item.get("children", []):
+                    if "children" not in item:
+                        item["children"] = [child]
+                    else:
+                        item["children"].append(child)
         return self.menu
```
Maybe something like this is required.

shuup/admin/base.py
volodymyr-code: fixed

shuup_tests/core/test_admin_menu.py
```diff
@@ -0,0 +1,18 @@
...
+# This source code is licensed under the OSL-3.0 license found in the
+# LICENSE file in the root directory of this source tree.
+from __future__ import unicode_literals
+
+from shuup.admin.menu import MAIN_MENU
+from shuup.testing.admin_main_menu_updater import SubscriptionMainMenuUpdater
+
+
+def test_menu_updater():
+    updater = SubscriptionMainMenuUpdater(MAIN_MENU)
```
tulimaki: Use the "overrride_provides" instead. First test the "get menu" without this "menu updater" and then with and make sure those updates actually happened.

volodymyr-code: fixed

maximlion:
Having inactive item under an item makes the parent item unusuable. (Dropdown is active so user is not redirected anywhere) ( happens only when parent item has only one item)
Putting originally parent orders under child orders shows None in Orders dropdown. Parent items shouldn't ever be allowed to be child items. Add (parent) to the name of the initial parent items 
Screenshot from 2019-05-16 22-41-26

volodymyr-code:
> Having inactive item under an item makes the parent item unusuable. (Dropdown is active so user is not redirected anywhere) ( happens only when parent item has only one item)
> Putting originally parent orders under child orders shows None in Orders dropdown. Parent items shouldn't ever be allowed to be child items. Add (parent) to the name of the initial parent items
> Screenshot from 2019-05-16 22-41-26

it is future, made for flexibility: we can realize any structure, for example we can move addons under settings or new plugin like "products multivendor" move under addons. I think will be better if we fix admin menu template and activate parent link for your case.

shuup/admin/static_src/base/js/admin-menu.js
```diff
@@ -0,0 +1,91 @@
...
+}
+
+const sortUpdate = (e) => {
+  const menuEntry = $(e.detail.item);
+  // update parent and entrye menu-entries sorter
+  connectSortable(menuEntry.parent());
+  connectSortable(menuEntry.find("> .menu-entries"));
+}
+
+const connectSortable  = (menuEntries) => {
```
chessbr: double space here before `=`

chessbr:
I removed Orders from the Order category to the root level and this is what is rendered:
image

The Orders menu should list the Carts as a child here.

PR #1883, opened 2019-04-30
Dont allow negative payments

bakaja021:


bakaja021:
Test does not do that. 
Will add those, but it confused me with our story from yesterday. I have tested similar thing and asked you regarding payment creation but when unpaid amount is over 0.

shuup_tests/core/test_order_creator.py
```diff
@@ -104,6 +104,14 @@ def test_order_creator(rf, admin_user):
...
     creator = OrderCreator()
     order = creator.create_order(source)
+    zero = Money(0, order.currency)
+    assert order.can_create_payment()
+    assert order.get_total_unpaid_amount() > zero
+    assert not order.is_paid()
+    order.create_payment(order.get_total_unpaid_amount())
+    assert order.is_paid()
+    assert not order.can_create_payment()
+    assert not order.get_total_unpaid_amount() > zero
```
Pikkupomo: This doesn't test the payment with negative amount.

Issue #1717, opened 2019-01-23
Enable APIs to upload images


chessbr:
Currently there is no way to send manufacturer logo through API
Also check Category and Product API.
Check whether it is possible to set a product primary image through the endpoints

chessbr:
Done

PR #1898, opened 2019-05-08
Admin API: Enable manufacturer logo addition

maximlion:
Setting primary product's image is possible through API by putting the ProductMedia id of the picture. 

Refs #1717, ENT-2773

### DIFF
--- a/shuup/core/api/manufacturer.py
+++ b/shuup/core/api/manufacturer.py
@@ -11,16 +11,42 @@ from django.utils.translation import ugettext_lazy as _
 from rest_framework import serializers
 from rest_framework.viewsets import ModelViewSet
 
+from shuup.api.fields import Base64FileField
 from shuup.api.mixins import (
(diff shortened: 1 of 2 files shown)


shuup/core/api/manufacturer.py
```diff
@@ -11,16 +11,42 @@
...
+        else:
+            self.fields["logo"] = serializers.SerializerMethodField()
+
+    def create(self, validated_data):
+        if "logo_path" in validated_data:
+            validated_data.pop("logo_path")
+        return super(ManufacturerSerializer, self).create(validated_data)
+
+    def validate(self, data):
+        if data.get("logo_path"):
```
chessbr: this check is confusing... it must be something like:

```py
if data.get("logo") and data.get("logo_path"):
   # all valid
elif data.get("logo"):
   raise serializers.ValidationError("Path is required when sending a manufacturer logo.")
```

Issue #1708, opened 2019-01-16
Double check Category API


chessbr:
Check whether creating categories and changing hierarchy through API works. I guess we are missing `Category.objects.rebuild()`

PR #1899, opened 2019-05-08
Improve category API

maximlion:
- API: Include shops in category API
When making a category through API the category wouldn't be connected with any shops because the shops field was excluded. This fixes that.
Refs #1708

- API: Enable image upload for Categories
Refs #1717, ENT-2832, #1708

### DIFF
--- a/shuup/core/api/category.py
+++ b/shuup/core/api/category.py
@@ -14,11 +14,12 @@ from parler_rest.fields import TranslatedFieldsField
 from parler_rest.serializers import TranslatableModelSerializer
 from rest_framework import serializers, viewsets
 
-from shuup.api.fields import EnumField
+from shuup.api.fields import Base64FileField, EnumField
 from shuup.api.mixins import PermissionHelperMixin, ProtectedModelViewSetMixin
 from shuup.core.models import (
     Category, CategoryStatus, CategoryVisibility, Shop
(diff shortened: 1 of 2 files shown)


shuup/core/api/category.py
```diff
@@ -27,14 +28,36 @@ class CategorySerializer(TranslatableModelSerializer):
     visibility = EnumField(CategoryVisibility)
     image = serializers.SerializerMethodField()
 
-    class Meta:
-        model = Category
-        exclude = ("shops",)
+    def __init__(self, *args, **kwargs):
+        super(CategorySerializer, self).__init__(*args, **kwargs)
+        request = self.context.get("request")
+        if request.method in ["POST"]:
```
chessbr: why not just `request.method == "POST"` ?

shuup/core/api/category.py
```diff
@@ -27,14 +28,36 @@ class CategorySerializer(TranslatableModelSerializer):
...
+            self.fields["image"] = Base64FileField(required=False, write_only=True)
+            self.fields["image_path"] = serializers.CharField(required=False, write_only=True)
+        else:
```
chessbr: `elif request.method == "GET"`, this prevents returning when method is `HEAD` or any other crazy string

PR #1915, opened 2019-05-22
Admin: add links to payments for Stripe payments (MYS-126)

tulimaki:
Add link to Stripe dashboard for payments created with shuup_stripe-addon.


tulimaki:
@chessbr good work by blocking this. Not a good idea.

But maybe this reveals some requirements for the payments upgrade. Maybe payments should be linked to some "method"/"service" and maybe the payment processor should provide some logic for these URLs as well as option to implement something for refunds.

PR #1916, opened 2019-05-22
Front: force add prefix to registration and login forms to prevent duplicates

chessbr:
Prevent duplicated form field ids when using quick login along with another registration or login form in the same template

No Refs


chessbr:
Replaced by https://github.com/shuup/shuup/pull/1917

PR #1920, opened 2019-05-22
Fix a problem with file title

Pikkupomo:
File title is not required, so it should fallback to product name

Add file size for convenience


shuup/front/templates/shuup/front/macros/order.jinja
```diff
@@ -19,7 +19,7 @@
         <table class="table">
             {% for attachment in attachments %}
                 <tr>
-                    <td><a href="{{ attachment.url }}" target="_blank">{{ attachment.title }}</a></td>
```
tulimaki: File can be None so check that too.

PR #1923, opened 2019-05-24
Admin: Fix contact address form breaking on region copy

### LINKED ISSUE #1921: "Use billing address as shipping address" fails to copy over Region
When this checkbox is clicked, the Region dropdown fails to copy across.

Applies to Contact Address and likely Checkout.

maximlion:
Contact address form was breaking on copying billing address region to shipping address.
This prevents that.
Edited code from [checkout.jinja](https://github.com/shuup/shuup/blob/be28e6eeee099445842bd02b72e03664c0318451/shuup/front/templates/shuup/front/macros/checkout.jinja)  so that it works with admin.

Fixes #1921

### DIFF
--- a/shuup/admin/templates/shuup/admin/contacts/edit.jinja
+++ b/shuup/admin/templates/shuup/admin/contacts/edit.jinja
@@ -16,34 +16,62 @@
 {% endblock %}
 
 {% block extra_js %}
-<script>
-(function(){
-    var shouldCopyFields = false;
-
-    function copyFieldValue() {
-        if(!shouldCopyFields) {
-            return;
+<script type="text/javascript">
+    $(function(){
+        var shouldCopyFields = false;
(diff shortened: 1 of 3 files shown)


shuup/admin/templates/shuup/admin/contacts/edit.jinja
```diff
@@ -16,34 +16,65 @@
...
-    $('.billing-address :input').on("input change", copyFieldValue);
-}());
-</script>
+   <script type="text/javascript">
+        $(function(){
+            var shouldCopyFields = false;
+            function toggleAndCopyFields() {
+                shouldCopyFields = $("#billing-to-shipping").is(":checked");
+                if (shouldCopyFields) {
+                    console.log("Jucer");
```
tulimaki: Remove console.logs.

Issue #1927, opened 2019-05-29
Allow users to click on variation children product card


chessbr:
When a user clicks on a variation children product card through plugins like Cross Sell plugin, the page redirects to the variation parent product (which is correct) but then the variation clicked is not the selected one, especially when it is a variable variation.

Make sure to make this work with variation buttons and select inputs.

PR #1929, opened 2019-05-29
Front: select variation children in product detail view

chessbr:
Make the simple variation option selected in parent product detail view

Refs #1927


tulimaki:
Requires some unit tests. And complex variations requires attention. I would consider moving that "selected child"-logic to the view context.

chessbr:
Replaced by https://github.com/shuup/shuup/pull/1932

Issue #1930, opened 2019-05-31
Check permissions on Picotable toolbar


tulimaki:
Would make sense not to show these "create"-buttons if the user does not have permission to create order. Applies to other admin modules as well.

<img width="520" alt="Screen Shot 2019-05-31 at 9 59 41 AM" src="https://user-images.githubusercontent.com/4940349/58721884-da239000-838a-11e9-9ab0-26d3753f66a4.png">

chessbr:
Done in https://github.com/shuup/shuup/pull/1939

PR #1939, opened 2019-06-05
Add required_permissions into the URLActionButton

volodymyr-code:
https://github.com/shuup/shuup/issues/1930

### DIFF
--- a/shuup/admin/toolbar.py
+++ b/shuup/admin/toolbar.py
@@ -9,7 +9,12 @@ from __future__ import unicode_literals
 
 import json
 
-from django.core.urlresolvers import NoReverseMatch, reverse
+try:
+    from urllib.parse import urlparse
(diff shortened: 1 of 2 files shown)


shuup/admin/toolbar.py
```diff
@@ -184,6 +189,11 @@ def for_model(cls, model, **kwargs):
             except NoModelUrl:
                 return None
             kwargs["url"] = url
+
+        if 'required_permissions' not in kwargs:
+            url = urlparse(kwargs['url']).path
+            kwargs['required_permissions'] = (resolve(url).url_name,)
```
chessbr: how this solves the issue? the best permission to add this model is:
`"%s.add_%s" % (model._meta.app_label, model._meta.model_name)`.. makes sense?
That way it will be `shuup.add_product` for Product model. This is the way we do for admin modules.

PR #1940, opened 2019-06-05
Save and memorise options selected by the user

chessbr:
### Front: only show marketing permissions check on the first checkout
Save a configuration for each customer inside the options field and make sure it is not asked on the checkout anymore

### GDPR: do not ask to consent documents already consented

Refs MYS-132


tulimaki:
Looks fair. Though Travis does not agree so fix the tests and we merge.

PR #1945, opened 2019-06-11
Front: fix manufacturer, supplier and variation filters

tulimaki:


tulimaki:
Screen Shot 2019-06-10 at 9 11 09 PM
About re-enabling those browser tests. Looks pretty solid locally and do not see reason why Travis can't handle these. One reason that these bugs slipped in was that these were disabled.

chessbr:
Check Travis error so we can merge this

PR #1946, opened 2019-06-11
Core: use force_text in models __str__ method to make sure to return strings

chessbr:
`__str__` should always return strings and when using `safe_translation_getter` it returns a `__proxy__` instead

No Refs

### DIFF
--- a/shuup/core/models/_labels.py
+++ b/shuup/core/models/_labels.py
@@ -8,7 +8,7 @@
 from __future__ import unicode_literals
 
 from django.db import models
-from django.utils.encoding import python_2_unicode_compatible
+from django.utils.encoding import force_text, python_2_unicode_compatible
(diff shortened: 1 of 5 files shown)


shuup/core/models/_labels.py
```diff
@@ -32,4 +32,5 @@ class Meta:
         verbose_name_plural = _('labels')
 
     def __str__(self):
-        return self.safe_translation_getter("name", default=self.identifier)
+        from django.utils.encoding import force_text
```
tulimaki: Any reason why the force text is imported under __str__?

chessbr: just to not add another import in the header which increases the time to load the progress. but yeah, that import is already added somewhere else no so big deal, I can change this

PR #1960, opened 2019-06-19
Improve manifest

chessbr:
:construction: Tests needed


chessbr:
This should be solved when apps get separated repository

PR #1963, opened 2019-06-24
Fix inactive users error

blomqma:
Shuup has a bug of not displaying a correct error message to users that are inactive and are trying to log in either in the front login form or in the admin login form. This is due to the confirm_login_allowed function not being called in the form validation process when inactive user is trying to log in. Thus, the default invalid login error is displayed even for inactive users trying to log in.

This can be fixed by overriding the clean method in both front and admin auth forms and manually calling the confirm_login_allowed function in case the user trying to log in is inactive similar to this:
[https://stackoverflow.com/a/46459998](url)

In the front auth view we also need to check that the user exists to pass the unit test regarding invalid login.

### DIFF
--- a/shuup/admin/forms/_auth.py
+++ b/shuup/admin/forms/_auth.py
@@ -5,7 +5,8 @@
 #
 # This source code is licensed under the OSL-3.0 license found in the
 # LICENSE file in the root directory of this source tree.
-from django.contrib.auth import get_user_model
+from django import forms
+from django.contrib.auth import authenticate, get_user_model
 from django.contrib.auth.forms import AuthenticationForm
 from django.contrib.auth.tokens import default_token_generator
 from django.core.exceptions import MultipleObjectsReturned, ObjectDoesNotExist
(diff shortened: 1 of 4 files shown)


shuup/admin/forms/_auth.py
```diff
@@ -48,6 +49,29 @@ def clean_username(self):
 
         return username
 
+    def clean(self):
+        username = self.cleaned_data.get("username")
+        password = self.cleaned_data.get("password")
+
+        if username is not None and password:
+            self.user_cache = authenticate(
+                self.request, username=username, password=password
```
tulimaki: `authenticate(request=self.request, username=username, password=password)` likely fixes it for Django 1.8 and 1.9. Applies to front too.

shuup/front/apps/auth/forms.py
```diff
@@ -54,6 +54,29 @@ def clean_username(self):
...
+                self.request, username=username, password=password
+            )
+            if self.user_cache is None:
+                try:
+                    user_temp = get_user_model().objects.get(username=username)
+                except ObjectDoesNotExist:
+                    user_temp = None
+
+                if user_temp is not None:
+                    self.confirm_login_allowed(user_temp)
```
tulimaki: This also checks whether the contact belongs to the correct shops so should likely be called pretty much always. Now it seems that `self.user_cache is None` when user is inactive and these custom checks are skipped.

tulimaki:
@blomqma added one commit to this at [here](https://github.com/shuup/shuup/pull/1965). Please check that and let me know if I missed something. Adds support to Django 1.8/1.9, fixes styles as well  some extra tests.

blomqma:
> @blomqma added one commit to this at [here](https://github.com/shuup/shuup/pull/1965). Please check that and let me know if I missed something. Adds support to Django 1.8/1.9, fixes styles as well some extra tests.

@tulimaki 
Looks good! Do you think there should be tests for the admin auth form inactive error too since these tests apply only to the front auth form?

blomqma:
> @blomqma added one commit to this at [here](https://github.com/shuup/shuup/pull/1965). Please check that and let me know if I missed something. Adds support to Django 1.8/1.9, fixes styles as well some extra tests.

@tulimaki I think this looks good now that there are tests for the admin auth form too!

PR #1976, opened 2019-07-18
Admin: improve media browsing

tulimaki:


shuup/admin/static_src/base/js/dropzone.js
```diff
@@ -17,7 +17,7 @@ function activateDropzone($dropzone, attrs={}) {
     const uploadPath = attrs.uploadPath || $data.upload_path;
     const addRemoveLinks = $data.add_remove_links;
     const uploadUrl = $data.upload_url || window.ShuupAdminConfig.browserUrls.media;
-    const browsable = ($data.browsable && $data.browsable !== "False");
+    const browsable = (window.ShuupAdminConfig.browserUrls["media"] && $data.browsable && $data.browsable !== "False");
```
chessbr: IMO, now you can remove this whole browsable option here and use the URL check only

PR #1979, opened 2019-07-24
Front: Include image dimensions in Open Graph meta tags

maximlion:
Also added fallback to use products first picture if the product doesn't have
a primary image set
1200 x 630 is the recommended image size for sharing on Facebook.
Refs HEAL-48

### DIFF
--- a/setup.py
+++ b/setup.py
@@ -40,7 +40,7 @@ VERSION_FILE = os.path.join(TOPDIR, 'shuup', '_version.py')
 #      - Add ".post0.dev" suffix to VERSION variable here
 
 NAME = 'shuup'
-VERSION = '1.9.8b3'
+VERSION = '1.9.8b3.post0.dev'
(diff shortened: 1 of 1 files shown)


shuup/front/templates/shuup/front/macros/general.jinja
```diff
@@ -383,8 +383,12 @@
     {% if text %}
         <meta property="og:description" content="{{ text|safe|striptags|truncate(160)|replace("\n", " ")|replace("\"", "'") }}">
     {% endif %}
-    {% if primary_image %}
-        <meta property="og:image" content="{{ request.build_absolute_uri(primary_image.url) }}">
+    {% set product_image = object.primary_image or object.media.first() %}
```
chessbr: Are you sure this `object.media.first()` will return the first `IMAGE`?
I guess we have better solutions and some other utils like `ShopProduct.public_images`

PR #1992, opened 2019-08-14
Add soft_delete to Supplier

volodymyr-code:


shuup/core/models/_suppliers.py
```diff
@@ -30,7 +30,12 @@ class Labels:
         EXTERNAL = _('external')
 
 
-class SupplierQueryset(TranslatableQuerySet):
+class SupplierManager(TranslatableManager):
+
+    def get_queryset(self):
```
tulimaki: Dos this work if I do `Supplier.objects.enabled()`?

volodymyr-code: added
- SupplierManager tests
- SupplierDeleteView and tests

shuup/admin/modules/suppliers/views/delete.py
```diff
@@ -0,0 +1,38 @@
...
+    pk = None
+
+    def dispatch(self, request, *args, **kwargs):
+        if self.pk is None:
+            self.pk = kwargs.pop("pk")
+        return super(SupplierDeleteView, self).dispatch(request, *args, **kwargs)
+
+    def get_redirect_url(self, *args, **kwargs):
+        if settings.SHUUP_ENABLE_MULTIPLE_SUPPLIERS:
+            return reverse("shuup_admin:shuup_multivendor.vendor.list")
```
tulimaki: We can not have this dependency. You have add separate delete view to multivendor addon which implements this view. Could be something like:

```
class SupplierMultivendorDeleteView(SupplierDeleteView):
    def get_redirect_url(self, *args, **kwargs):
            return reverse("shuup_admin:shuup_multivendor.vendor.list")
```

shuup/admin/modules/suppliers/views/delete.py
```diff
@@ -0,0 +1,38 @@
...
+from django.conf import settings
+from django.contrib import messages
+from django.core.urlresolvers import reverse
+from django.utils.translation import ugettext_lazy as _
+from django.views.generic import RedirectView
+
+from shuup.core.models import Supplier
+
+
+class SupplierDeleteView(RedirectView):
```
tulimaki: Why his is redirect view? Could this be same as in [here](https://github.com/shuup/shuup/blob/master/shuup/admin/modules/products/views/delete.py).

```
class SupplierDeleteView(DetailView):
    model = Supplier
    context_object_name = "supplier"

    def get_success_url(self):
        return reverse("shuup_admin:supplier.list")

    def get_queryset(self, *args, **kwargs):
        suppier = self.get_object()
        shop = get_shop(self.request)
        qs = Supplier.objects.filter(shops=shop)
        admin_supplier = get_supplier(self.request)
        if admin_supplier:
            # If admin has supplier enabled allow only delete self
            # althought not good view to enable for vendor
            qs = qs.filter(id=admin_supplier.pk)
        return qs

        supplier = self.get_object()
        return HttpResponseRedirect(get_model_url(supplier, shop=self.request.shop))

    def post(self, request, *args, **kwargs):
        supplier = self.get_object()
        supplier.soft_delete(user=request.user)
        messages.success(request, _(u"%s has been marked deleted.") % supplier)
        return HttpResponseRedirect(self.get_success_url())



class MultivendorSupplierDeleteView(SupplierDeleteView):
    def get_success_url(self):
        return reverse("shuup_admin:shuup_multivendor.vendor.list")
```

shuup/admin/modules/suppliers/views/edit.py
```diff
@@ -76,6 +78,25 @@ class SupplierEditView(SaveFormPartsMixin, FormPartsViewMixin, CreateOrUpdateVie
     base_form_part_classes = [SupplierBaseFormPart, SupplierContactAddressFormPart]
     form_part_class_provide_key = "admin_supplier_form_part"
 
+    def get_toolbar(self):
```
tulimaki: Here would be more simple way to achieve this. But looks it should work. https://github.com/shuup/shuup/blob/master/shuup/admin/modules/contact_groups/views/edit.py#L33

shuup/core/models/_suppliers.py
```diff
@@ -144,5 +150,11 @@ def update_stock(self, product_id):
     def update_stocks(self, product_ids):
         return self.module.update_stocks(product_ids)
 
+    def soft_delete(self):
+        if not self.deleted:
+            self.deleted = True
+            # Bypassing local `save()` on purpose.
```
tulimaki: What would be the reasoning behind this?

shuup/admin/modules/suppliers/views/delete.py
volodymyr-code: `DetailView` used to render detailed template views, in our case, SupplierDeleteView result is redirect, so my choice was `RedirectView`

shuup/admin/modules/suppliers/views/edit.py
volodymyr-code: that code was copy-paste from HEAL custom delete model, will fix it

shuup/core/models/_suppliers.py
volodymyr-code: > What would be the reasoning behind this?

Sorry, don't understand what you mean.

shuup/admin/modules/suppliers/views/delete.py
tulimaki: This is not huge thing as long as it works. You just need to make sure that current user can only delete suppliers that is linked to current shop and supplier active in admin (see the get_queryset).

shuup/core/models/_suppliers.py
tulimaki: Just curios what this bypassing local save is for? I guess we could call `self.save(update_fields=("deleted",))`. Or even update the attr with queryset update `Supplier.objects.filter(id=self.pk).update(deleted=True)`. This might be just copy paste from `shuup.core.models.Product:soft_delete`, but maybe there is actual reason in product model which does not apply with supplier model.

shuup/simple_cms/models.py
```diff
@@ -54,11 +66,11 @@ def visible(self, shop, dt=None, user=None):
         )
         if user and not user.is_anonymous():
             q |= Q(created_by=user)
-        qs = self.for_shop(shop).filter(q)
+        qs = self.filter(q).for_shop(shop)
         return qs
 
     def for_shop(self, shop):
-        return self.filter(shop=shop)
+        return self.get_queryset().for_shop(shop=shop)
```
tulimaki: Doesn't make sense to add this `get_queryset()` just to replace `filter(deleted=False)`?

shuup/simple_cms/models.py
```diff
@@ -54,11 +66,11 @@ def visible(self, shop, dt=None, user=None):
         )
         if user and not user.is_anonymous():
             q |= Q(created_by=user)
-        qs = self.for_shop(shop).filter(q)
+        qs = self.filter(q).for_shop(shop)
```
tulimaki: Why this needs to be changed?

shuup/simple_cms/models.py
```diff
@@ -144,7 +156,7 @@ class Page(MPTTModel, TranslatableModel):
         "Check this if this page should have a visible title"
     ))
 
-    objects = TreeManager.from_queryset(PageQuerySet)()
+    objects = PageQueryManager()
```
tulimaki: We adding this delete view requires this change?

shuup/simple_cms/models.py
```diff
@@ -178,8 +190,8 @@ def is_visible(self, dt=None):
             dt = now()
 
         return (
-            (self.available_from and self.available_from <= dt) and
-            (self.available_to is None or self.available_to >= dt)
+                (self.available_from and self.available_from <= dt) and
```
tulimaki: Again I doubt this should be changed.

shuup/core/models/_suppliers.py
```diff
@@ -83,9 +88,10 @@ class Supplier(ModuleInterface, TranslatableShuupModel):
     translations = TranslatedFields(
         description=models.TextField(blank=True, verbose_name=_("description"))
     )
+    deleted = models.BooleanField(default=False, verbose_name=_("deleted"))
 
     search_fields = ["name"]
-    objects = SupplierQueryset.as_manager()
+    objects = SupplierManager()
```
tulimaki: Really doesn't sound like changing this benefits us too much. If `filter(deleted=False)` is now added every single queryset how one get the deleted items? I think you could filter out deleted directly in the admin views and then also here in enabled queryset-manager.

volodymyr-code: ok, will revert SupplierQueryset

shuup/simple_cms/models.py
volodymyr-code: reverted

shuup/simple_cms/models.py
volodymyr-code: reverted

shuup/simple_cms/models.py
volodymyr-code: reverted `SupplierQueryset ` and added `not_deleted()` method

PR #1995, opened 2019-08-20
API: Bump Django Rest Framework to latest (>3.9.1)

chessbr:
Bump and fix DRF Swagger

Refs AN-28

:warning: Not backwards compatible


chessbr:
Should be done in https://github.com/shuup/shuup-api and https://github.com/shuup/shuup-rest-api

PR #2007, opened 2019-09-02
Core: Fix issue with order source update and refunding lines without quantity

tulimaki:
- Core: fix issue with order source update from order
- Core: do not try to refund lines without quantity


shuup/core/models/_orders.py
```diff
@@ -897,7 +897,7 @@ def create_full_refund(self, restock_products=False, created_by=None):
             "quantity": line.quantity,
             "amount": line.taxful_price.amount,
             "restock_products": restock_products
-        } for line in self.lines.all() if line.type != OrderLineType.REFUND]
+        } for line in self.lines.filter(quantity__gt=0) if line.type != OrderLineType.REFUND]
```
chessbr: why the line type is not also in the queryset filter? :thinking:

PR #2009, opened 2019-09-03
Add option to disable arbitrary refunds (REAL-39)

tulimaki:


shuup/admin/template_helpers/shuup_admin.py
```diff
@@ -44,6 +44,17 @@ def get_menu_entry_categories(context):
     return menu.get_menu_entry_categories(request=context["request"])
 
 
+def is_menu_category_active(category, target_url):
+    def does_identifier_match(always_active_identifier):
+        return bool("%s" % category.identifier == "%s" % always_active_identifier)
```
volodymyr-code: Probably the bool cast is redundant, simpler way:
```
return "%s" % category.identifier == "%s" % always_active_identifier
```

chessbr: `force_text(category.identifier) == force_text(always_active_identifier)`

shuup/admin/template_helpers/shuup_admin.py
```diff
@@ -44,6 +44,17 @@ def get_menu_entry_categories(context):
...
+
+    identifiers = settings.ALWAYS_ACTIVE_MENU_CATEGORY_IDENTIFIERS
+    if any(identifier for identifier in identifiers if does_identifier_match(identifier)):
+        return True
+
+    return any([entry for entry in category.entries if entry.url == target_url])
```
chessbr: you need to be consistent. you don't use brackets on the `any` above but now you use. pick up one :)

PR #2015, opened 2019-09-13
Admin: add suppliers to product list by default

tulimaki:
When multiple suppliers are enabled add suppliers column to default column. Also ensure that when suppliers is used as column the filtering by name is also set for the column.

### DIFF
--- a/shuup/admin/modules/products/views/list.py
+++ b/shuup/admin/modules/products/views/list.py
@@ -7,6 +7,7 @@
 # LICENSE file in the root directory of this source tree.
 from __future__ import unicode_literals
 
+from django.conf import settings
 from django.contrib.staticfiles.templatetags.staticfiles import static
 from django.db.models import Q
 from django.utils.translation import ugettext_lazy as _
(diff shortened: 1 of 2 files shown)


shuup/admin/modules/products/views/list.py
```diff
@@ -103,9 +105,34 @@ class ProductListView(PicotableListView):
...
+            self.default_columns.append(
+                Column(
+                    "suppliers",
+                    _("Suppliers"),
+                    display="format_suppliers",
+                    ordering=8,
+                    filter_config=get_suppliers_filter(),
+                )
+            )
+        super(ProductListView, self).__init__()
```
chessbr: are we officially dropping py 27?

tulimaki: Not just yet.

chessbr: so this needs fixing

tulimaki: how is this not py27 compatible?

chessbr: :man_facepalming: that is right, Monday...

PR #2020, opened 2019-09-16
Admin: Orders, changed queryset filter

volodymyr-code:
Refs REAL-83

### DIFF
--- a/shuup/admin/modules/orders/views/detail.py
+++ b/shuup/admin/modules/orders/views/detail.py
@@ -14,6 +14,7 @@ from django.utils.translation import ugettext as _
 from django.views.generic import DetailView
 
 from shuup.admin.modules.orders.toolbar import OrderDetailToolbar
+from shuup.admin.shop_provider import get_shop
 from shuup.admin.utils.urls import get_model_url
 from shuup.apps.provides import get_provide_objects
 from shuup.core.models import Order, OrderStatus, OrderStatusRole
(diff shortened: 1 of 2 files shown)


tulimaki:
@volodymyr-code where is this coming from? Maybe we could have separate view for this and allow superusers and staff to only see orders per shop by default (this is the logic for everything else ATM).

And secondly does the shop filter work nicely. What if there is 1000 shops is that select2 or do we have lag for render? Maybe the shop filter should could be based on name instead select if not already?

Basically what I am after is that we always show data per currently active shop. And we currently just missing the admin modules allowing cross shop data. And these likely should be separated.

volodymyr-code:
> @volodymyr-code where is this coming from? Maybe we could have separate view for this and allow superusers and staff to only see orders per shop by default (this is the logic for everything else ATM).
That task from @chessbr that fix required for one of our project
Christian, please discuss

shuup/admin/modules/orders/views/detail.py
```diff
@@ -28,6 +28,12 @@ class OrderDetailView(DetailView):
     def get_toolbar(self):
         return OrderDetailToolbar(self.object)
 
+    def get_queryset(self):
+        qs = super(OrderDetailView, self).get_queryset().exclude(deleted=True)
+        if not self.request.user.is_superuser:
+            qs = qs.filter(shop=self.request.shop)
```
chessbr: never use `request.shop` in admin views, this is wrong. always use `get_shop(request)`

shuup/admin/modules/orders/views/list.py
```diff
@@ -64,7 +64,10 @@ def get_toolbar(self):
         return toolbar
 
     def get_queryset(self):
-        return super(OrderListView, self).get_queryset().exclude(deleted=True).filter(shop=self.request.shop)
+        qs = super(OrderListView, self).get_queryset().exclude(deleted=True)
+        if not self.request.user.is_superuser:
+            qs = qs.filter(shop=self.request.shop)
```
chessbr: use `get_shop`

shuup/admin/modules/orders/views/detail.py
tulimaki: Got it old codes here in diff. Carry on.

PR #2023, opened 2019-09-18
Testing: improve mocker for testing purposes

tulimaki:


tulimaki:
I do not think we add unit tests for testing tools. At least not this time.

PR #2028, opened 2019-09-23
Admin: add option to add messages menu entries

tulimaki:


chessbr:
I will think these menus should be populated through provides

PR #2030, opened 2019-09-30
Admin: allow to use custom file_transformer

volodymyr-code:
- set `custom_file_transformer` to True into your `DataImporter` if you want to use internal `transform_file` classmethod
- fixed slug field issue, it raised `Data too long for column 'slug'`

### DIFF
--- a/shuup/importer/admin_module/import_views.py
+++ b/shuup/importer/admin_module/import_views.py
@@ -52,6 +52,8 @@ class ImportProcessView(TemplateView):
                 mode = "xlsx"
             if filename.endswith("csv"):
                 mode = "csv"
+            if self.importer_cls.custom_file_transformer:
+                return self.importer_cls.transform_file(mode, filename)
             return transform_file(mode, filename)
         except (Exception, RuntimeError) as e:
             messages.error(self.request, e)
--- a/shuup/importer/importing/importing.py
+++ b/shuup/importer/importing/importing.py
@@ -44,6 +44,7 @@ class DataImporter(object):
     meta_class_getter_name = "get_import_meta"
     meta_base_class = ImportMetaBase
     extra_matches = {}
+    custom_file_transformer = False
 
     unique_fields = {}
     unmatched_fields = set()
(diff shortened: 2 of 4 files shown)


shuup/importer/_mapper.py
```diff
@@ -112,8 +113,10 @@ def _create_new_object(self, mapped, value):  # noqa (C901)
         if self.is_translated:
             obj.set_current_language(self.handler.language)
             for field in self.translated_fields:
-                if field.name not in ("master", "id", "language_code", "description"):
+                if field.name not in ("master", "id", "language_code", "description", "slug"):
                     setattr(obj, field.name, value)
+                elif field.name == "slug":
+                    setattr(obj, field.name, slugify(value[:field.max_length]))
```
tulimaki: I guess you want to do this to all fields having "max_length" and not just slugfield. Also not sure if worth re-calling slugify here.

chessbr: I am not super expert with this relation mapper but I agree with @tulimaki . Making the max length check here is good but should happen for all fields that have this set, not only for "slug".
IMO, the slugify here depends on the import class as this is a "custom" behavior. Maybe not all importers want to slugify the value again?
Check https://github.com/shuup/shuup/blob/f68d268c72953e548ce0bae93fdced9793ecdb81/shuup/importer/importing/importing.py#L497
Maybe your importer can set this mapper to handle the relation the way you want instead of putting this custom logic inside Shuup OS?

chessbr: Or add way to customize this behavior in your importer.

volodymyr-code: > Making the max length check here is good but should happen for all fields that have this set, not only for "slug".
I'll change it for all fields
> Maybe not all importers want to slugify the value again?
slugify value at this point sets the first time

tulimaki: I guess the point here is that if have defined this slug-field in my excel like "Hello-World" this will make it always "hello-world" which is not ideal. You can override this in your custom product importer, but probably not good thing to do in this default implementation.

volodymyr-code: removed

Issue #2027, opened 2019-09-22
RegistrationPhase hardcoded to NoActivation


yurkobb:
Although there is a `SHUUP_REGISTRATION_REQUIRES_ACTIVATION` settings which defaults to True, the `RegisterPhase` checkout phase is [hard coded](https://github.com/shuup/shuup/blob/master/shuup/front/checkout/checkout_method.py#L93) to inherit from `RegistrationNoActivationView`. 

This means that if one uses `shuup.front.apps.auth` to handle login and registration outside the checkout process, it _will_ require activation by default, unlike registration during checkout.

In the same time it is unclear for me if it is possible to have two separate email notification templates with and without the activation link.

Ideally, `RegisterPhase` needs to be refactored to respect the `SHUUP_REGISTRATION_REQUIRES_ACTIVATION` setting, so that the registration conditions are the same inside and outside of the checkout process.

PR #2032, opened 2019-10-02
Implement a register-with-activation checkout phase [WIP]

yurkobb:
As noted in #2027, the registration within checkout is currently hard coded to simple registration without requiring activation of the account over email. In this pull request, I'm trying to work towards implementing an optional way to have a registration with mandatory activation during checkout. At this stage it is a work in progress - but I need some advice on how to move forward.

I have implemented an additional `RegisterWithActivationPhase` checkout phase (and a test for it *and* the regular `RegisterPhase`). At this stage the user is registered and the email with the activation link is sent.

It is also highly desirable to implement a continuation path of the checkout process after the user clicks the activation link. I was thinking about appending an optional `?next=` or `?continue=` parameter to the activation url generated in the [send_user_registered_notification()](https://github.com/shuup/shuup/blob/829dfaa4fec7ead2a4992016e5a7458a63233c1c/shuup/front/apps/registration/notify_events.py#L45) function if the registration took place during a checkout. This would probably mean adding an attribute to the `request` object in `RegiterWithActivationPhase` that gets passed to the `user_registered` signal and then to `send_user_registered_notification()`. Does this look like a reasonable way to implement it?

### DIFF
--- a/shuup/front/checkout/checkout_method.py
+++ b/shuup/front/checkout/checkout_method.py
@@ -13,7 +13,7 @@ from django.utils.translation import ugettext_lazy as _
 from enumfields import Enum
 
 from shuup.front.apps.auth.views import LoginView
-from shuup.front.apps.registration.views import RegistrationNoActivationView
+from shuup.front.apps.registration.views import RegistrationNoActivationView, RegistrationView
 from shuup.front.checkout import CheckoutPhaseViewMixin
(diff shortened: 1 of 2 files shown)


yurkobb:
(Non-technical note - just added my committed email address as to my github account to avoid any confusion).

PR #2041, opened 2019-10-15
Admin: init shop product suppliers with only supplier

tulimaki:
Initialize shop product suppliers with the only supplier when multiple suppliers is disabled at core settings.

### DIFF
--- a/shuup_tests/admin/test_product_create.py
+++ b/shuup_tests/admin/test_product_create.py
@@ -1,0 +1,34 @@
+# This file is part of Shuup.
+#
+# Copyright (c) 2012-2019, Shoop Commerce Ltd. All rights reserved.
+#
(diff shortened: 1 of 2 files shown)


shuup_tests/admin/test_product_create.py
```diff
@@ -0,0 +1,35 @@
...
+        tax_class = get_default_tax_class()
+        product_type = get_default_product_type()
+        sales_unit = get_default_sales_unit()
+
+        assert ShopProduct.objects.count() == 0
+        client = SmartClient()
+        client.login(username="admin", password="password")
+        soup = client.soup(reverse("shuup_admin:shop_product.new"))
+        supplier_select = soup.find("select", attrs={"name": "shop%s-suppliers" % shop.id})
+        print(supplier_select)
```
chessbr: print

PR #2043, opened 2019-10-17
Admin: do not allow large images to be inserted into summernote

chessbr:
Images are converted into Base64 and they could lead to slowness.
It is a good feature for vector images but adding too much content to
the HTML will lead to poor performance that is hard to track

Refs HEAL-29

### DIFF
--- a/shuup/admin/static_src/base/js/editor.js
+++ b/shuup/admin/static_src/base/js/editor.js
@@ -68,6 +68,24 @@ function activateEditor($editor, attrs = {}) {
                         }
                     }
                 }
+            },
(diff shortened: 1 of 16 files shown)


shuup/admin/static_src/base/js/editor.js
```diff
@@ -68,6 +68,24 @@ function activateEditor($editor, attrs = {}) {
                         }
                     }
                 }
+            },
+            onImageUpload: function(image) {
```
tulimaki: How about simple CMS?

chessbr:
@tulimaki but it does
image

PR #2044, opened 2019-10-17
Admin: add settings to make select2 fields to load data asynchronously

chessbr:
It is a common request of merchants to have categories options loaded
asynchronously or on page load. To solve this issue we are introducing a
new setting `SHUUP_ADMIN_LOAD_SELECT_OBJECTS_ASYNC` that can be adjusted
per project and adjust fields according to the business model.

The setting is enabled by default and will make categories and suppliers
fields to load objects asynchronously (through AJAX requests). When
disabled, the options will be rendered on page load.

Refs HEAL-35

### DIFF
--- a/shuup/admin/modules/products/forms/base_forms.py
+++ b/shuup/admin/modules/products/forms/base_forms.py
@@ -15,13 +15,14 @@ from django.contrib import messages
 from django.core.exceptions import ValidationError
 from django.forms import BaseModelFormSet
 from django.forms.formsets import DEFAULT_MAX_NUM, DEFAULT_MIN_NUM
-from django.utils.translation import ugettext_lazy as _
 from django.utils.translation import ugettext
+from django.utils.translation import ugettext_lazy as _
 from filer.models import Image
 
 from shuup.admin.forms.fields import (
     Select2ModelField, Select2ModelMultipleField
 )
+from shuup.admin.forms.quick_select import NoModel
 from shuup.admin.forms.widgets import (
     FileDnDUploaderWidget, QuickAddCategoryMultiSelect, QuickAddCategorySelect,
     QuickAddDisplayUnitSelect, QuickAddManufacturerSelect,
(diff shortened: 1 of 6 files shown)


shuup/admin/settings.py
```diff
@@ -67,3 +67,10 @@
 #: menu item. Useful in case there is need to always open some
 #: certain menus.
 SHUUP_ALWAYS_ACTIVE_MENU_CATEGORY_IDENTIFIERS = []
+
+#: Indicates whether all object select fields should load options asynchronously.
+#: When True, fields will load options through AJAX requests instead
+#: of generating them while rendering the page. For enviroments with huge amount of options
+#: in fields, like categories, it is best to have this enabled.
+#:
+SHUUP_ADMIN_LOAD_SELECT_OBJECTS_ASYNC = True
```
tulimaki: Could this be ("categories": True) thing where I can limit this per model or something similar. In this solution do we now loose quick adds. This does not seem right.

shuup/admin/modules/products/forms/base_forms.py
```diff
@@ -202,41 +204,65 @@ def __init__(self, **kwargs):
...
-        self.fields["categories"] = Select2ModelMultipleField(
-            initial=initial_categories,
-            model=Category,
-            widget=QuickAddCategoryMultiSelect(initial=initial_categories),
-            required=False
-        )
+                model=Category,
+                widget=QuickAddCategorySelect(
+                    editable_model="shuup.Category",
+                    initial=(self.instance.primary_category if self.instance.pk else None),
```
tulimaki: This looks like taking iterable.

shuup/admin/modules/products/forms/base_forms.py
```diff
@@ -202,41 +204,65 @@ def __init__(self, **kwargs):
...
+                ),
+                required=False
+            )
+        else:
+            categories_choices = [
+                (cat.pk, cat.name)
+                for cat in Category.objects.all_except_deleted(shop=get_shop(self.request))
+            ]
+            self.fields["primary_category"].widget = QuickAddCategorySelect(
+                initial=(self.instance.primary_category if self.instance.pk else None),
```
tulimaki: Iterable...

shuup/admin/modules/products/forms/base_forms.py
chessbr: Iterable? It was like this before. This is a `Select2ModelField` not a multiple field. :thinking:

PR #2045, opened 2019-10-18
Core: Add orderline behaviour attribute

maximlion:
Added orderline behaviour attribute
INHERIT = 1  # Inherits the quantity of the parent line if parent line exists
SKIP = 2  # Skips quantity change upon parent line quantity change
DELETE = 3  # Removes the line upon parent line quantity change

Refs HEAL-37

### DIFF
--- a/shuup/core/order_creator/_source.py
+++ b/shuup/core/order_creator/_source.py
@@ -14,6 +14,7 @@ from django.contrib.auth.models import AnonymousUser
 from django.core.exceptions import ValidationError
 from django.utils.encoding import force_text
 from django.utils.timezone import now
+from django.utils.translation import ugettext_lazy as _
 from enumfields import Enum
 from six import iteritems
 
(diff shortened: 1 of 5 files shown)


shuup/core/models/_order_lines.py
```diff
@@ -102,6 +113,10 @@ class AbstractOrderLine(MoneyPropped, models.Model, Priceful):
 
     created_on = models.DateTimeField(auto_now_add=True, editable=False, db_index=True, verbose_name=_('created on'))
     modified_on = models.DateTimeField(auto_now=True, editable=False, db_index=True, verbose_name=_('modified on'))
+    line_behaviour = EnumIntegerField(
```
chessbr: This is not needed in OrderLine model, once the order is created, this shouldn't be useful anymore

shuup/core/order_creator/_source.py
```diff
@@ -631,13 +643,16 @@ class SourceLine(TaxableItem, Priceful, LineWithUnit):
     quantity = None  # override property from Priceful
     base_unit_price = None  # override property from Priceful
     discount_amount = None  # override property from Priceful
-
+    on_parent_change_behavior = EnumIntegerField(
```
chessbr: ` on_parent_change_behavior = OrderLineBehavior.INHERIT`

PR #2046, opened 2019-10-18
Admin: fix issues and properly test order related querysets

tulimaki:


shuup/admin/modules/orders/views/addresses.py
```diff
@@ -26,6 +26,10 @@ class OrderAddressEditView(UpdateView):
     context_object_name = "order"
     template_name = "shuup/admin/orders/_address_edit.jinja"
 
+    def get_queryset(self):
+        shop_ids = Shop.objects.get_for_user(self.request.user).values_list("id", flat=True)
```
chessbr: maybe use get_shop here

shuup/admin/modules/orders/views/detail.py
```diff
@@ -30,7 +30,8 @@ def get_toolbar(self):
         return OrderDetailToolbar(self.object)
 
     def get_queryset(self):
-        return super(OrderDetailView, self).get_queryset().exclude(deleted=True).filter(shop=get_shop(self.request))
+        shop_ids = Shop.objects.get_for_user(self.request.user).values_list("id", flat=True)
```
chessbr: get_shop()

shuup/admin/modules/orders/views/detail.py
```diff
@@ -53,6 +54,10 @@ def get_context_data(self, **kwargs):
 class OrderSetStatusView(DetailView):
     model = Order
 
+    def get_queryset(self):
+        shop_ids = Shop.objects.get_for_user(self.request.user).values_list("id", flat=True)
```
chessbr: get_shop()

shuup/admin/modules/orders/views/addresses.py
```diff
@@ -26,6 +26,10 @@ class OrderAddressEditView(UpdateView):
...
+        return Order.objects.exclude(deleted=True).filter(shop_id__in=shop_ids)
```
chessbr: why not just
```py
Order.objects.exclude(deleted=True).filter(shop_in=Shop.objects.get_for_user(get_shop(self.request)))
```

tulimaki: get_for_user does not take shop as param :)

shuup/admin/modules/orders/views/addresses.py
tulimaki: For these I do not see necessary for current shop to be active. I just want to ensure that not wrong people can access these views.

PR #2049, opened 2019-10-21
Admin: add signal that is triggered when form is valid in views

chessbr:
It can help to inject extra validations before saving the view object and raise errors

Refs HEAL-30

### DIFF
--- a/shuup/admin/form_part.py
+++ b/shuup/admin/form_part.py
@@ -7,8 +7,10 @@
 # LICENSE file in the root directory of this source tree.
 from django.http import HttpResponseRedirect
 
-from shuup.admin.signals import object_created, object_saved
+from django.db.transaction import atomic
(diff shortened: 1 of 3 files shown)


shuup/admin/form_part.py
```diff
@@ -79,7 +81,14 @@ class SaveFormPartsMixin(object):
     request = None  # Placate "missing field" errors
     object = None  # --"--
 
+    @atomic()
     def save_form_parts(self, form):
+        # trigger signal for extra form validations
+        try:
+            view_form_valid.send(sender=type(self), view=self, form=form, request=self.request)
```
tulimaki: Are you sure this is correct place for this? Should not this be inside the loop starting there at line 94?

chessbr: Do you think we should call this for every form inside the formgroup? IMO this should happen with the parent form.

PR #2063, opened 2019-10-28
Admin: improve the way active menu items is shown in left side menu

tulimaki:
Previously navigation for example to product edit hide the product menu selection from left side menu which caused some confusion.


shuup/admin/template_helpers/shuup_admin.py
```diff
@@ -46,15 +47,29 @@ def get_menu_entry_categories(context):
     return menu.get_menu_entry_categories(request=context["request"])
 
 
-def is_menu_category_active(category, target_url):
+@lru_cache()
+def get_all_target_urls(context, target_url, breadcrumbs):
```
chessbr: can you pass the request here instead of the entire context? If I am not wrong, I already faced some issues when caching context with lru_cache as some weird content couldn't be serialized by the cache.

PR #2064, opened 2019-10-29
Campaigns: Add a variation parent basket condition

maximlion:
- Campaigns: Add a variation parent basket condition
This condition will match on each child in basket of a chosen product.
image
No refs

- Front: Add safe filter to form field help text 
This makes sure that any help text containing HTML is rendered properly.
No refs


shuup/campaigns/models/basket_conditions.py
```diff
@@ -328,3 +329,37 @@ def description(self):
...
+    name = _("Product Variation Child")
+
+    model = Product
+    product = models.ForeignKey(
+        Product,
+        limit_choices_to={
+            'mode__in': [
+                ProductMode.SIMPLE_VARIATION_PARENT,
+                ProductMode.VARIABLE_VARIATION_PARENT]},
+        null=True)
```
chessbr: this indentation looks so weird. try:
```
product = models.ForeignKey(
    Product,
    limit_choices_to={
        'mode__in': [
            ProductMode.SIMPLE_VARIATION_PARENT,
            ProductMode.VARIABLE_VARIATION_PARENT
        ]
    },
    null=True
)
```

shuup/campaigns/admin_module/forms/_basket_conditions.py
```diff
@@ -65,6 +65,11 @@ class Meta(BaseRuleModelForm.Meta):
         model = BasketMaxTotalAmountCondition
 
 
+class ChildrenProductConditionForm(BaseRuleModelForm):
+    class Meta(BaseRuleModelForm.Meta):
+        model = ChildrenProductCondition
```
chessbr: I guess here you need to filter product options to be from the current shop and also from the current supplier, if that exists

PR #2077, opened 2019-11-13
Core: remove API code and move to separated repository

chessbr:
Moved to https://github.com/shuup/shuup-api and https://github.com/shuup/shuup-rest-api


setup.py
```diff
@@ -99,8 +99,6 @@
     'django-reversion>=2,<3',
     'django-timezone-field>=1.2,<3',
     'djangorestframework>=3.5,<3.6',
```
tulimaki: So what is this needed for anymore?

shuup_workbench/settings/base_settings.py
```diff
@@ -198,32 +196,6 @@
 
 SHUUP_PRICING_MODULE = "customer_group_pricing"
 
-REST_FRAMEWORK = {
```
tulimaki: These shouldn't be removed IMO since we going to install those exact same APIs from the new repositories.

chessbr: you're right

setup.py
chessbr: this is needed for GDPR, unfortunately. but it only use serializers and it shouldn't block bumping DRF

PR #2079, opened 2019-11-14
add indexing to text fields in produdcts, categories and suppliers

kerkkoheiskanen:
This should hopefully make searching quicker

refs SHU-56


kerkkoheiskanen:
Whoops.

PR #2084, opened 2019-11-20
Add ability to display text on top of carousel

mikkovedru:
Quite often simply displaying pictures in a carousel is not enough. 
Pictures are not descriptive and appealing enough to motivate people to 
click on banners. Adding text can often solve this issue.

Supports both "Carousel" and "Banner box" types of carousel. It is well 
visible thanks to shadows around the text. The default size of 
padding/font/etc is chosen to work well with the Shuup (sandbox) default 
values.

ref SHU-45


mikkovedru:
Tests are written and passing!

PR #2086, opened 2019-11-26
Add contact creation to gdpr guest ordering

kerkkoheiskanen:
Create contacts when ordering guest orders with gdpr active.
Also adds testing for the feature, as well as some text to admin

refs SHU-41


shuup/front/checkout/confirm.py
```diff
@@ -97,6 +100,29 @@ def get_context_data(self, **kwargs):
...
+        user, created = User.objects.get_or_create(
+            username=email,
+            defaults={
+                'email': email,
+                'first_name': first_name,
+                'last_name': last_name,
+                'is_active': False,
+            }
+        )
+        user.first_name = first_name
```
tulimaki: Instead of this you could use `update_or_create`. In case your plan is to always update these.

shuup/front/checkout/confirm.py
```diff
@@ -123,9 +149,16 @@ def form_valid(self, form):
     def create_order(self):
         basket = self.basket
         assert basket.shop == self.request.shop
-        basket.orderer = self.request.person
-        basket.customer = self.request.customer
-        basket.creator = self.request.user
+        gdpr_settings = GDPRSettings.get_for_shop(basket.shop)
```
tulimaki: I doubt you can get GDPR settings if the "shuup.gdpr" is not in installed apps.

shuup/front/checkout/confirm.py
```diff
@@ -97,6 +100,29 @@ def get_context_data(self, **kwargs):
         context["product_ids"] = ','.join(self._get_product_ids())
         return context
 
+    def _get_guest_order_data_with_gdpr(self):
+        basket = self.basket
+        first_name = basket._billing_address.name
```
tulimaki: You really can't call `basket.billing_address` here https://github.com/shuup/shuup/blob/master/shuup/core/basket/objects.py#L271? Usually these private attributes should not be called outside the class.

shuup/front/checkout/confirm.py
```diff
@@ -97,6 +100,29 @@ def get_context_data(self, **kwargs):
...
+        last_name = basket._billing_address.name_ext
+        email = basket._billing_address.email
+        basket._billing_address.save()
```
tulimaki: Why exactly you are saving these addresses since there is no changes?

shuup/front/checkout/confirm.py
```diff
@@ -123,9 +149,16 @@ def form_valid(self, form):
     def create_order(self):
         basket = self.basket
         assert basket.shop == self.request.shop
-        basket.orderer = self.request.person
-        basket.customer = self.request.customer
-        basket.creator = self.request.user
+        gdpr_settings = GDPRSettings.get_for_shop(basket.shop)
+        if type(self.request.person) == AnonymousContact and has_installed("shuup.gdpr") and gdpr_settings.enabled:
```
tulimaki: Maybe something like this would look nicer.

1. you do not need type check
2. you should deal gdpr settings if gdpr is installed
```
contact, user = None
if not self.request.person and has_installed("shuup.gdpr"):
    from shuup.gdpr.models import GDPRSettings
    gdpr_settings = GDPRSettings.get_for_shop(basket.shop)
    if gdpr_settings.enabled:
          contact, user = self._get_guest_order_data_with_gdpr()  

basket.orderer = (self.request.person or contact)
basket.customer = (self.request.customer or contact)
basket.creator = (self.request.user or user)
```

shuup/front/checkout/confirm.py
```diff
@@ -10,15 +10,18 @@
...
 from django.utils.translation import ugettext_lazy as _
 from django.views.generic import FormView
 
 from shuup.apps.provides import get_provide_objects
-from shuup.core.models import OrderStatus
+from shuup.core.models import AnonymousContact, OrderStatus, get_person_contact
 from shuup.front.basket import get_basket_order_creator
 from shuup.front.checkout import CheckoutPhaseViewMixin
 from shuup.front.signals import checkout_complete
+from shuup.gdpr.models import GDPRSettings
```
tulimaki: Only import GDPRSettings if installed.

shuup/front/checkout/confirm.py
```diff
@@ -97,6 +100,29 @@ def get_context_data(self, **kwargs):
...
 
+    def _get_guest_order_data_with_gdpr(self):
+        basket = self.basket
+        first_name = basket._billing_address.name
+        last_name = basket._billing_address.name_ext
+        email = basket._billing_address.email
+        basket._billing_address.save()
+        basket._shipping_address.save()
+        user, created = User.objects.get_or_create(
+            username=email,
```
tulimaki: Should we raise here if there is no email since not in all installations email is required for address. Maybe GDPR checkouts as guest should have but likely email being None here is a problem.

shuup/gdpr/receivers.py
```diff
@@ -31,4 +33,8 @@ def create_consents_login_allowed(sender, request, user, *args, **kwargs):
 
 @receiver(checkout_complete)
 def create_consents_checkout_complete(sender, request, user, order, *args, **kwargs):
-    create_user_consent_for_all_documents(request.shop, user)
+    gdpr_settings = GDPRSettings.get_for_shop(request.shop)
+    if not user.is_authenticated() and has_installed("shuup.gdpr") and gdpr_settings.enabled:
```
tulimaki: Well under gdpr app we do not need to check whether it is installed. Likely the enabled check should also be under `create_user_consent_for_all_documents` not sure this check needs to be added either. Maybe instead of this line should just be `reate_user_consent_for_all_documents(request.shop, order.orderer.user)`.

shuup_tests/gdpr/test_user_creation.py
```diff
@@ -0,0 +1,135 @@
+import random
```
tulimaki: Missing license header for this new file.

shuup_tests/gdpr/test_user_creation.py
```diff
@@ -0,0 +1,135 @@
...
+    n_orders_post = Order.objects.count()
+    assert n_orders_post > n_orders_pre, "order was created"
+
+    order = Order.objects.first()
+    expected_ip = "127.0.0.2"
+    assert order.ip_address == expected_ip
+    if create_contact:
+        assert order.orderer == contact
+    elif gdpr_enabled:
+        assert type(order.orderer) == PersonContact
```
tulimaki: You could check the user and customer fields here as well and also check the email/username matches to what we expect.

shuup/front/checkout/confirm.py
```diff
@@ -97,6 +100,29 @@ def get_context_data(self, **kwargs):
...
+            defaults={
+                'email': email,
+                'first_name': first_name,
+                'last_name': last_name,
+                'is_active': False,
+            }
+        )
+        user.first_name = first_name
+        user.last_name = last_name
```
tulimaki: So by guest ordering I can update anyones user information and disable the account if I know their email?

shuup/gdpr/receivers.py
```diff
@@ -31,4 +33,8 @@ def create_consents_login_allowed(sender, request, user, *args, **kwargs):
...
+        create_user_consent_for_all_documents(request.shop, order.orderer.user)
```
tulimaki: Maybe `order.creator` since not sure someone can add consent behalf of someone else since here order can be created as staff as well.

PR #2101, opened 2019-12-21
Admin: Add Order & Order Line reports

maximlion:
Added order & order line reports.
Added limit to all Report queries which is 2000.
image

Refs SHU-55

### DIFF
--- a/shuup/default_reports/forms.py
+++ b/shuup/default_reports/forms.py
@@ -10,7 +10,10 @@ from django import forms
 from django.utils.translation import ugettext_lazy as _
 
 from shuup.admin.forms.fields import Select2MultipleField
-from shuup.core.models import Carrier, Contact, ShippingMethod, Tax, TaxClass
+from shuup.core.models import (
+    Carrier, Contact, OrderLineType, OrderStatus, ShippingMethod, Supplier,
+    Tax, TaxClass
+)
 from shuup.reports.forms import BaseReportForm
 
 
@@ -42,6 +45,45 @@ class OrderReportForm(BaseReportForm):
     def initial_contacts(self, key):
         if self.data and key in self.data:
             return Contact.objects.filter(pk__in=self.data.getlist(key))
+        return []
+
+
+class OrderLineReportForm(BaseReportForm):
+
+    order_line_type = forms.MultipleChoiceField(
+        label=_("Order Line Type"),
+        required=False,
+        initial=[OrderLineType.PRODUCT.value],
+        choices=[
+            (line_type.value, line_type.name.capitalize())
+            for line_type in OrderLineType
+        ],
+    )  # Because value of OrderLineType.PRODUCT is 1
+
+    def __init__(self, *args, **kwargs):
+        super(OrderLineReportForm, self).__init__(*args, **kwargs)
+
+        supplier = Select2MultipleField(label=_("Suppliers"),
+                                        model=Supplier,
+                                        required=False,
+                                        help_text=_("Filter order lines by suppliers."))
+        order_status = forms.ModelMultipleChoiceField(label=_("Order status"),
+                                                      required=False,
+                                                      queryset=OrderStatus.objects.all(),
+                                                      help_text=_("Filter order lines by status of their order."))
+
+        suppliers = self.get_initial_suppliers("supplier")
+
+        if suppliers:
+            supplier.initial = suppliers
+            supplier.widget.choices = [(obj.pk, obj.name) for obj in suppliers]
+
+        self.fields["supplier"] = supplier
+        self.fields["order_status"] = order_status
+
+    def get_initial_suppliers(self, key):
+        if self.data and key in self.data:
+            return Supplier.objects.filter(pk__in=self.data.getlist(key))
         return []
 
 
--- a/shuup/default_reports/mixins.py
+++ b/shuup/default_reports/mixins.py
@@ -7,12 +7,12 @@
 # LICENSE file in the root directory of this source tree.
 from django.db.models import Q
 
-from shuup.core.models import Order
+from shuup.core.models import Order, OrderLine
 from shuup.utils.dates import to_datetime_range
 
 
 class OrderReportMixin(object):
-    def get_objects(self):
+    def get_objects(self, paid=True):
         (start, end) = to_datetime_range(self.start_date, self.end_date)
         queryset = Order.objects.filter(
             shop=self.shop, order_date__range=(start, end))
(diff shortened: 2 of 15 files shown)


shuup/default_reports/forms.py
```diff
@@ -45,6 +48,43 @@ def initial_contacts(self, key):
...
+        suppliers = self.initial_supplier("supplier")
+
+        if suppliers:
+            supplier.initial = suppliers
+            supplier.widget.choices = [(obj.pk, obj.name) for obj in suppliers]
+
+        self.fields["supplier"] = supplier
+        self.fields["order_status"] = order_status
+
+    def initial_supplier(self, key):
```
tulimaki: Naming of this function indicates that one supplier is return, but instead it returns list or queryset. Maybe `get_initial_suppliers`

shuup/default_reports/mixins.py
```diff
@@ -26,5 +26,23 @@ def get_objects(self):
             filters &= Q(orderer__in=orderer)
         if customer:
             filters &= Q(customer__in=customer)
+        if paid:
```
tulimaki: Maybe better like this:
```
queryset = queryset.filter(filters).valid()
if paid:
    queryset = queryset.paid()
return queryset.order_by("order_date)
```

shuup/default_reports/reports/customer_sales.py
```diff
@@ -30,7 +30,7 @@ class CustomerSalesReport(OrderReportMixin, ShuupReportBase):
 
     def get_objects(self):
         return Contact.objects.filter(
-            customer_orders__in=super(CustomerSalesReport, self).get_objects()
+            customer_orders__in=super(CustomerSalesReport, self).get_objects()[:self.report_limit]
```
tulimaki: Could you use `settings.DEFAULT_REPORTS_ITEM_LIMIT` instead `from django.conf import settings`?This way we do not need code changes to up the limit. Applies to all reports.

shuup/default_reports/forms.py
```diff
@@ -45,6 +48,43 @@ def initial_contacts(self, key):
...
+        initial=[1])  # Because value of OrderLineType.PRODUCT is 1
+
+    def __init__(self, *args, **kwargs):
+        super(OrderLineReportForm, self).__init__(*args, **kwargs)
+
+        supplier = Select2MultipleField(label=_("Suppliers"),
+                                        model=Supplier,
+                                        required=False,
+                                        help_text=_("Filter order lines by suppliers."))
+        order_status = Select2MultipleField(label=_("Order status"),
```
chessbr: I guess here you could just use `MultipleModelChoiceField` as I doubt we would have more than 10 different statuses and it is better for the merchant to see the list of status instead of trying to find one they want.

shuup/default_reports/reports/orders.py
```diff
@@ -0,0 +1,84 @@
...
+    def get_data(self):
+        data = []
+        orders = self.get_objects(paid=False)
+
+        for order in orders:
+            data.append({
+                "order_num": order.identifier,
+                "order_date": get_locally_formatted_datetime(order.order_date),
+                "status": order.status,
+                "order_line_quantity": order.lines.count(),
```
chessbr: maybe this could consider only products?

shuup/default_reports/reports/orders.py
```diff
@@ -0,0 +1,84 @@
...
+                "order_line_text": line.text,
+                "order_line_quantity": line.quantity,
+                "taxless_unit_price": line.taxless_base_unit_price,
+                "taxful_unit_price": line.taxful_base_unit_price,
+                "taxful_price": line.taxful_price,
+                "type": line.type.name.capitalize(),
+                "created_on": get_locally_formatted_datetime(line.created_on),
+            })
+        if "order_line_type" in self.options.keys() and len(self.options["order_line_type"]) == 1:
+            self.schema.pop()  # Remove type column if there's only one type
```
chessbr: IMO, you can't do that! This schema is not created in each instance but on the class definition. Once you remove this from the class, it won't be there anymore. Test this by making the code remove the last column and then try to run the report that needs this schema, I bet things will blow up.

shuup/default_reports/reports/orders.py
maximlion: Hm, right now a user can see products only if he selects product type in OrderLineReport.

shuup/default_reports/reports/orders.py
maximlion: You're right! Removed it.

shuup/reports/settings.py
```diff
@@ -0,0 +1,9 @@
+# -*- coding: utf-8 -*-
+# This file is part of Shuup.
+#
+# Copyright (c) 2012-2019, Shoop Commerce Ltd. All rights reserved.
+#
+# This source code is licensed under the OSL-3.0 license found in the
+# LICENSE file in the root directory of this source tree.
+
+DEFAULT_REPORTS_ITEM_LIMIT = 2000
```
chessbr, line 9: add documentation for what this is used for

shuup/default_reports/forms.py
```diff
@@ -45,6 +48,43 @@ def initial_contacts(self, key):
         return []
 
 
+class OrderLineReportForm(BaseReportForm):
+
+    order_line_type = forms.MultipleChoiceField(
```
chessbr, line 53: please format this this way to improve readability:

```py
order_line_type = forms.MultipleChoiceField(
    label=_("Order Line Type"),
    required=False,
    initial=[OrderLineType.PRODUCT.value],
    choices=[
        (line_type.value, line_type.name.capitalize())
        for line_type in OrderLineType
    ],
)
```

shuup/default_reports/reports/customer_sales.py
chessbr, line 33: This limit is being added in the wrong place, please put it after the final filter below:
```py
.filter(order_count__gt=0).order_by("-%s" % self.options["order_by"])[:self.report_limit].values(
    "name", "order_count", "average_sales", "taxless_total", "taxful_total"
)
```

shuup/default_reports/reports/orders.py
chessbr: But this will show the total number of lines no matter what

shuup/default_reports/reports/refunds.py
```diff
@@ -26,7 +26,11 @@ class RefundedSalesReport(OrderReportMixin, ShuupReportBase):
     ]
 
     def get_data(self, **kwargs):
-        orders = super(RefundedSalesReport, self).get_objects().filter(lines__type=OrderLineType.REFUND).distinct()
+        orders = super(
```
chessbr, line 29: oh please, format this:

```py
orders = super(RefundedSalesReport, self).get_objects().filter(
    lines__type=OrderLineType.REFUND
).distinct()[:self.report_limit]
```

PR #2113, opened 2020-01-20
Help text improvements

mikkovedru:
Big update to improve various texts

A lot of changes, touching almost every individual Shuup project's repo.

There are too many changes to try to describe them separately, but the main ones are:
1. Improve `help_text` descriptions. Fix typos, rewrite and change, write new ones.
2. Change some of the model field names (for example rename `available_from` -> `available_since`; `available_to` `available_until` - this is in a separate commit).
3. Improve some of the documentation.
4. Improve some in-code-comments.
5. Improve `settings.py` files. It's now clearer what each option does.
6. Change some of the naming.
7. Change the Error/message system.

The main Shuup repo's Merge Request is here: https://github.com/shuup/shuup/pull/2113 (with a reasoning behind changes).

# Reasoning
Changes 6 and 7 need more clarification. Here I will also talk about the overall writing style changes.

## Overall changes to writing style
Old: 
```
"See Contacts - Permission Groups to change these"
```
New: 
```
"Search for `Permission Groups` to change these."
```

0. All the descriptions should end with a dot. Until now, it has been very inconsistent.
1. Backticks are seen better than the single quotes, and don't interfere that easily with Python and Django.
2. Later on, backticks could be used to visually distinguish between parameters (markdown/github style)
3. The paths to the proper menu can change and be different (like in the above help text; in sandbox it is located under "Shops" and not under "Contacts"). It's best to tell about the final location (which can be searched for), than to tell the path (which can change).

## Change some of the naming.
It has been very confusing to understand what "staff" means. Look at this picture:
image

It's extremely confusing, to have both types to be named `permissions` and have `staff` in both lists. There is no good explanation on what either of the difference is (even some Shuup programmers don't know that these are two entirely different things, let alone know what that difference is). Even if one knows the answers to the above, it's still very difficult to find where you can change the `Permissions` (not to be confused with `Permission Groups`).

To explain, currently we have two types of permission types:
1. "Permissions". This is basically `is_staff` and `superuser` (Django's default fields) accounts. The former gives access to the Admin Panel and the latter gives all the permissions.
2. "Permission Groups". More granular control. Shuup usually comes with `Vendor` and `Staff` groups.

But now it changes for the better!

0. The `help_text` finally explains the difference and tells where to change stuff.
1. I renamed `Permissions` to `Main Permissions`. There have been many other alternatives (Major permissions, Head, Chief, Principal, Main, Central, Fundamental, Basic, Special, Marketplace...), but `Main Permissions` was chosen as the best (after seeking advice from a native speaker).
2. I renamed `Permission Groups` to `Granular Permission Groups`. Again, there were many other alternatives (like `Additional Permission Groups`), but the winner was the most clear/unambiguous (shows that this group of settings allows for a more granular changes to the permissions), while being clearly different from the `Main Permissions`.
3. `staff` (which is Django's default `is_staff` field) was renamed to `Access to Admin Panel`. Even in code this new name was added alongside the internal field name.
4. `admin` (which is Django's default `superuser` field) was renamed to `Superuser (Full rights)`. In code this new name was added alongside the internal field name.

This is how the same user looks after the changes:
image

It's very clear now and doesn't even require me to explain what `is_staff` is and how does it differ from the other Staff group.

## Changes to the Error/message system.
The old messaging system (from Shuup to the user), especially when it comes to error messages was both inconsistent as well as too confusing at times:
- `X has been cancelled` - is this an error or a success message; do I need to do anything about it?
- `File deleted`. Is it "Error" or "Success"?`Sometimes you just don't know.
- `Recipient is a required field when you don't want to send the email to customer.`. Is this an error, information, warning or something else entirely?
- Or `Product is no longer in the wish list`.

Now the system is more consistent. The basic idea, as implemented by the Bootstrap:
image

Every message should start with telling the status of the message:
1. The most important information comes first, only after come the details. This allows to skim messages and quickly see if something needs attention.
2. Consistent for every single message type. Hugely important!
3. Helps color-blind people.

Great many messages were rewritten (or a rule was created to automatically add a status part of the message to the already translated strings) to adhere to a new format:
- "X has been approved successfully" → "Success! X has been approved."
- "Approval for %s has been revoked successfully" → "Success! Approval for %s has been revoked.".
- `Categories successfully saved!` → `Success! Categories were saved.`

The basic rules of messaging are:
1. There are 4 different types of status ("Success", "Warning", "Info", "Error"). They should be written normally, no CAPS.
2. The status message (Error!, Success!, Warning!, Info!) should come first, not as part of the rest of the sentence.
3. Status should be ending with an exclamation mark.
4. Exclamation mark is only for a status. Don't use exclamation mark at the end of other sentences; only dots.
5. Use verb ("were", "was", "did", etc; "X saved" -> "X was saved"). In vast majority of cases it should be in the past ("was added", "were deleted", "something went wrong", "has been"). There can be exceptions, like if there is a JavaScript code, that is being shown while something is being currently done ("Info! Finalizing installations.").
6. There are some rare cases when it makes sense not to put status in the error message, but your case is most probably not that. Please think at least 10 times if you consider not adding a status. Even `Exception`s should have a status!
7. It's important to remember, that all of the above won't help if the message itself was written badly, without thinking about the purpose and the target audience. There are many guides on how to write a good error/success message. Here is one: https://uxplanet.org/how-to-write-good-error-messages-858e4551cd4


tulimaki:
Fixed by https://github.com/shuup/shuup/pull/2184

PR #2131, opened 2020-02-18
Admin: add option to impersonate staff users

tulimaki:
Mostly a superuser feature but allowed to staff as long as not impersonating superuseres. Also add template helper for logout url which stops the staff user impersonation properly.

### DIFF
--- a/shuup/admin/modules/users/views/detail.py
+++ b/shuup/admin/modules/users/views/detail.py
@@ -48,10 +48,19 @@ NEW_USER_EMAIL_CONFIRMATION_TEMPLATE = _("""
 def get_front_url():
     front_url = None
     try:
-        front_url = reverse("shuup:index")
+        front_url = reverse(django_settings.SHUUP_ADMIN_LOGIN_AS_REDIRECT_VIEW)
     except NoReverseMatch:
         front_url = None
     return front_url
+
+
+def get_admin_url():
+    admin_url = None
+    try:
+        admin_url = reverse(django_settings.SHUUP_ADMIN_LOGIN_AS_STAFF_REDIRECT_VIEW)
+    except NoReverseMatch:
+        admin_url = None
+    return admin_url
 
 
 class BaseUserForm(forms.ModelForm):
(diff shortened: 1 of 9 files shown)


shuup/admin/modules/users/views/detail.py
```diff
@@ -48,12 +48,21 @@
 def get_front_url():
     front_url = None
     try:
-        front_url = reverse("shuup:index")
+        front_url = reverse(django_settings.SHUUP_MULTIVENDOR_LOGIN_AS_REDIRECT_VIEW)
```
chessbr: is this `SHUUP_MULTIVENDOR_` prefix correct here in Shuup OS?

shuup/admin/modules/users/views/detail.py
```diff
@@ -208,13 +217,22 @@ def _build_existing_user(self):
...
+            is_current_user_superuser_or_staff and user.is_active and not user.is_superuser
+        )
+
+        if (can_impersonate and get_front_url() and not user.is_staff):
             self.append(PostActionButton(
                 post_url=reverse("shuup_admin:user.login-as", kwargs={"pk": user.pk}),
                 text=_(u"Login as User"),
                 extra_css_class="btn-gray"
             ))
+        elif can_impersonate and get_admin_url():
```
chessbr: I am not sure I got this. Just because there is admin url means the user can impersonate as staff?
Also, do you think we need two buttons? Can we have just one? If the user is staff then redirect to admin, otherwise redirect to front (using those urls in settings)

shuup/admin/modules/users/views/detail.py
```diff
@@ -333,17 +351,19 @@ def dispatch(self, request, *args, **kwargs):
         return super(UserDetailView, self).dispatch(request, *args, **kwargs)
 
 
-def _check_for_login_as_problems(front_url, impersonator_user, user):
-    if not front_url:
+def _check_for_login_as_problems(redirect_url, impersonator_user, user):
+    if not redirect_url:
```
chessbr: the error message below is weird now

shuup/admin/views/impersonate.py
```diff
@@ -0,0 +1,30 @@
...
+    logout(request)
+
+    user = get_user_model().objects.get(pk=impersonator_user_id)
+    for backend in settings.AUTHENTICATION_BACKENDS:
+        if user == load_backend(backend).get_user(user.pk):
+            user.backend = backend
+            break
+
+    login(request, user)
+    return HttpResponseRedirect(reverse("shuup_admin:user.detail", args=[auth_user_id]))
```
chessbr: maybe check whether the user has permissions to `user.detail`? otherwise send it to dashboard? I guess this can break in multivendor world if the staff user impersonates a vendor and there is no user permission

shuup/admin/modules/users/views/detail.py
tulimaki: I think two buttons is better for permissions. Maybe someone should only impersonate end users and impersonating staff users should only be for some main guys.

tulimaki: Admin url logic is same as with front so if for some reason we do not know the redirect url then let's not show te button.

chessbr: got it, makes sense

PR #2141, opened 2020-02-26
Password recovery fixes

tulimaki:


shuup/core/utils/forms.py
```diff
@@ -134,17 +137,35 @@ def process_user(self, user_to_recover):
...
-        email.send()
+
+        notification_sent = False
+        request = getattr(self, "request", None)
+        if has_installed('shuup.notify') and request and request.shop and self.recover_password_confirm_view_url_name:
+            # Send notification event if any Script enabled
+            from shuup.notify.models import Script
+            from shuup.notify.notify_events import PasswordReset
+            if Script.objects.filter(enabled=True, event_identifier=PasswordReset.identifier).exists():
+                notification_sent = True
```
chessbr: maybe you should just set this if the `PasswordReset` ran without exceptions?

```suggestion
   try:
       PasswordReset(**context).run(shop=self.request.shop)
       notification_sent = True
   except:
        notification_sent = False
```

PR #2151, opened 2020-03-05
Fix notification name

wiltso:
The notification name was <class 'object'> and not name of the notification


tulimaki:
Pleaser rebase and this should be only one commit added.

PR #2167, opened 2020-04-14
Change timezone in middleware

wiltso:
Activated the timezone in the middleware. This was done by adding the timezone field to the users contact from, so the user can select there own timezone.
Then I added the timezone field to the supplier model so the supplier can specify their timezone. This is the fallback timezone if the user haven't specified one.
The last fallback timezone if the nether the supplier of user has specified a timezone is the timezone specified in setting.


shuup/front/middleware.py
```diff
@@ -117,7 +118,10 @@ def _set_basket(self, request):
     def _set_timezone(self, request):
         if request.person.timezone:
             timezone.activate(request.person.timezone)
-            # TODO: Fallback to request.shop.timezone (and add such field)
+        elif get_supplier(request) and get_supplier(request).timezone:
```
tulimaki, line 121: Maybe, just in case that get_supplier makes some quick db check and is not cached to request:

```
else:
     supplier = get_supplier(request)
     if supplier and supplier.timezone:
          timezone.timezone.activate(get_supplier(request).timezone)
```

tulimaki:
Please add app defining this "timezone_field" to requirements at setup.py.

PR #2168, opened 2020-04-15
Admin: Fix breadcrumbs double highlight

maximlion:
image

No refs

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -10,6 +10,7 @@ List all changes after the last release here (newer on top). Each change on a se
 
 ### Fixed
 
+- Discounts: create different admin module for archived discounts to fix breadcrumbs
 - Fix product pagination by not overriding the state with undefined values
 
 ### Changed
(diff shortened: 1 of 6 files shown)


maximlion:
> I am not sure what you doing here, why you need to do this?

https://github.com/shuup/shuup/blob/master/shuup/discounts/admin/modules/_discounts.py#L20
https://github.com/shuup/shuup/blob/master/shuup/discounts/admin/modules/_discounts.py#L42
`get_menu_entries` returns two menu entries, problem is that the Module has a `breadcrumbs_menu_entry` that points to one of them so that one entry will always be activated
Now, Modules rarely return two entries but `CampaignAdminModule` does it https://github.com/shuup/shuup/blob/master/shuup/campaigns/admin_module/__init__.py#L22 and it has no `breadcrumbs_menu_entry`
If I just remove breadcrumbs_menu_entry  then discount edit doesn't activate any entry so I added a parent which will be activated upon it

PR #2171, opened 2020-04-22
Update contact list filters so that only customers are shown by default

blomqma:


tulimaki:
More clean version here https://github.com/shuup/shuup/pull/2172

PR #2181, opened 2020-05-05
Admin: improve product media uploads

tulimaki:
- Add error message when upload fails. At media queue complete do not resave product media if the file-count has not changed. This for example prevents media save when the upload itself fails.
- Add option to override dropzone upload path by using data attribute
- Add upload path to browser URLs and use it to fallback on media uploads when the actual media path is not available.

### DIFF
--- a/shuup/admin/static_src/product/edit_media.js
+++ b/shuup/admin/static_src/product/edit_media.js
@@ -82,20 +82,39 @@ $(function() {
         $html.insertBefore($source);
     }
 
+    function getFileIds(kind) {
+        const $fileInputs = $("#product-" + kind + "-section").find(".file-control input");
+        var fileIds = [];
+        for(var i = 0; i < $fileInputs.length; i++){
+            let fileId = parseInt($($fileInputs[i]).val());
+            if(!isNaN(fileId)) {
+                fileIds.push(parseInt($($fileInputs[i]).val()));
+            }
+        }
+        return fileIds
+    }
+
+    function getFileCount(kind) {
+        return $("#product-" + kind + "-section").data("saved_file_count") || 0;
+    }
+
+    function setFileCount(kind, count) {
+        $("#product-" + kind + "-section").data("saved_file_count", count);
+    }
+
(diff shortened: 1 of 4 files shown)


tulimaki:
do not merge

shuup/admin/static_src/product/edit_media.js
```diff
@@ -82,21 +82,40 @@ $(function() {
...
+    function getFileCount(kind) {
+        return $("#product-" + kind + "-section").data("saved_file_count") || 0;
+    }
+
+    function setFileCount(kind, count) {
+        $("#product-" + kind + "-section").data("saved_file_count", count);
+    }
+
+    function onDropzoneQueueComplete(dropzone, kind) {
+        if(location.pathname.indexOf("new") > 0) {
```
chessbr, line 106: I don't think this is a good idea. Can't we have a flag for this coming from the widged instead of relying on the URL?

shuup/admin/static_src/product/edit_media.js
```diff
@@ -82,21 +82,40 @@ $(function() {
...
+            // save product media the traditional way via the save button when creating a new product
+            return;
+        }
+        const productId = $("#product-" + kind + "-section-dropzone").data().product_id;
+        if (!productId) {
+            return;
+        }
+
+        let fileIds = getFileIds(kind);
+        if (getFileCount(kind) == fileIds.length) {  // Skip add media if file count has not changed
```
chessbr, line 116: `===`

shuup/admin/static_src/product/edit_media.js
```diff
@@ -82,21 +82,40 @@ $(function() {
...
         }
+        return fileIds
+    }
+
+    function getFileCount(kind) {
+        return $("#product-" + kind + "-section").data("saved_file_count") || 0;
+    }
+
+    function setFileCount(kind, count) {
+        $("#product-" + kind + "-section").data("saved_file_count", count);
```
chessbr, line 102: not fency, but..

PR #2183, opened 2020-05-15
CMS: Add ordering to Page choices in Page Links Plugin

maximlion:
Refs Trello


maximlion:
> Doesn't work. Sent you a video

Fixed!

PR #2197, opened 2020-06-02
Basket enhancement

chessbr:
- General: add custom get_ip method and use it everywhere
- Core: Use UUID4 while generating order line ids by default
- Front: add feature for checkout phases to spawn extra phases
- Core: add shops and supplier to ServiceProvider and supplier to Service model
- Admin: change service admin to list only providers that the current user can access
- Admin: fixed issue that was importing User model directly
- Campaigns: do not apply campaigns in baskets configured to a supplier
- Core: ignore lines that are not from the given source while calculating taxes
- Front: remove the custom _load() implementation from the basket as it is the same as the core

Refs JDL-18


tulimaki:
Looks good!

PR #2204, opened 2020-06-10
Add data import permissions

wiltso:
- Added so there is a permission for every type of import.
- Added so if the user doing the import dose not have access to media.browse it will not show how to upload products with images.
- Added so if the user doing the import dose not have access to create new categories it will show a list of all the avilebal categories and thellthe importer to use one/more of those categories.
- Added so if there is a supplier given to the data importer it will use that supplier insted of any other supplier.

refs JDL-14


shuup/importer/importing/importing.py
```diff
@@ -57,10 +57,11 @@ class DataImporter(object):
 
     model = None
 
-    def __init__(self, data, shop, language, supplier=None):
+    def __init__(self, data, shop, language, supplier=None, request=None):
```
chessbr, line 60: nope, remove this request from here

shuup/importer/importing/meta.py
```diff
@@ -83,6 +83,27 @@ def get_import_defaults(self):
         """
         return {}
 
+    def change_row_values(self, row, request):
```
chessbr, line 86: rename this to:

```py
def pre_process_row(self, row):
    ....
    return row
```

then write a new method:

```py
def should_skip_row(self, row):
    return False
```

shuup/importer/admin_module/import_views.py
```diff
@@ -63,7 +65,15 @@ def prepare(self):
         if self.data is None:
             return False
 
-        self.importer = self.importer_cls(self.data, get_shop(self.request), self.lang)
+        context = self.importer_cls.get_import_context_from_request(
```
chessbr, line 68: just pass the request, shop and language here

PR #2206, opened 2020-06-12
Admin: add improved product copy

tulimaki:
Copy supplier, categories, media, stock managed value and customer group price

### DIFF
--- a/shuup/admin/modules/products/views/copy.py
+++ b/shuup/admin/modules/products/views/copy.py
@@ -1,0 +1,63 @@
+# -*- coding: utf-8 -*-
+# This file is part of Shuup.
+#
+# Copyright (c) 2012-2020, Shoop Commerce Ltd. All rights reserved.
+#
+# This source code is licensed under the OSL-3.0 license found in the
+# LICENSE file in the root directory of this source tree.
+from django.contrib import messages
+from django.db.transaction import atomic
+from django.http import HttpResponseRedirect
+from django.utils.translation import ugettext_lazy as _
+from django.views.generic import DetailView
+
+from shuup.admin.signals import product_copied
+from shuup.admin.utils.urls import get_model_url
+from shuup.core.models import (
+    Product, ProductAttribute, ProductMedia, ShopProduct
+)
+from shuup.utils.models import get_data_dict
+
+
+class ProductCopyView(DetailView):
+    model = ShopProduct
+    context_object_name = "product"
+
+    @atomic()
+    def get(self, request, *args, **kwargs):
+        shop_product = self.get_object()
+        product = shop_product.product
+        product_data = get_data_dict(product)
+        product_data.update({
+            "sku": "{}-{}".format(product.sku, Product.objects.count())
+        })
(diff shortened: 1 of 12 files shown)


shuup/admin/modules/products/views/copy.py
```diff
@@ -0,0 +1,75 @@
...
+                new_product.primary_image = media
+                new_product.save()
+
+        for price in CgpPrice.objects.filter(product=product, shop=shop_product.shop):
+            CgpPrice.objects.create(
+                product=new_product, shop=shop_product.shop, group=price.group,
+                price_value=price.price_value
+            )
+
+        for supplier in shop_product.suppliers.all():
```
chessbr, line 68: Why the stocks should be copied to this new product? I don't think it makes sense. Also, don't import simple_supplier here as this module can be replaced by any other. If you need to add stocks, use the interface of the supplier module: `get_stock_status` and `adjust_stock`.

shuup/admin/modules/products/views/copy.py
```diff
@@ -0,0 +1,75 @@
...
+        new_product.save()
+
+        new_shop_product = ShopProduct.objects.create(product=new_product, shop=shop_product.shop)
+        new_shop_product.visibility = shop_product.visibility
+        new_shop_product.purchasable = shop_product.purchasable
+        new_shop_product.default_price_value = shop_product.default_price_value
+        new_shop_product.primary_category = shop_product.primary_category
+        new_shop_product.save()
+
+        new_shop_product.suppliers = shop_product.suppliers.all()
```
chessbr, line 49: use `.set()` instead

shuup/admin/modules/products/views/copy.py
```diff
@@ -0,0 +1,75 @@
...
+        new_shop_product.categories = shop_product.categories.all()
```
chessbr, line 50: use `.set()` instead

shuup/admin/modules/products/views/copy.py
```diff
@@ -0,0 +1,75 @@
...
+                price_value=price.price_value
+            )
+
+        for supplier in shop_product.suppliers.all():
+            origin_product_stock_count = StockCount.objects.get_or_create(supplier=supplier, product=product)[0]
+            new_product_stock_count = StockCount.objects.get_or_create(supplier=supplier, product=new_product)[0]
+            new_product_stock_count.stock_managed = origin_product_stock_count.stock_managed
+            new_product_stock_count.save(update_fields=["stock_managed"])
+
+        messages.success(request, _("Product %s copy successfull.") % new_product)
```
chessbr, line 74: I would recommend you to trigger some custom signal here like:
```py
product_copied.send(sender=type(self), copied=product, copy=new_prouct)
```

this way, other addons can do their copy part if they need to (this is the case of the supplier module and others).

shuup/admin/modules/products/views/copy.py
tulimaki: It does not copy the stock but the stock managed value. But yes maybe this should be only done when the simple supplier is actually installed.

PR #2207, opened 2020-06-13
Core: add task runner to support running tasks using 3rd party services like Celery

chessbr:
Refs JDL-19

### DIFF
--- a/shuup/core/tasks.py
+++ b/shuup/core/tasks.py
@@ -1,0 +1,97 @@
+# -*- coding: utf-8 -*-
+# This file is part of Shuup.
+#
+# Copyright (c) 2012-2020, Shoop Commerce Ltd. All rights reserved.
+#
+# This source code is licensed under the OSL-3.0 license found in the
+# LICENSE file in the root directory of this source tree.
+import json
+
+from shuup.utils.importing import cached_load, load
+
+
+class TaskNotSerializableError(Exception):
+    """
+    Raised when the task can't be serialized.
+    """
+    pass
+
+
+class Task:
+    function = ""       # type: str
+    kwargs = None       # type: Optional[Dict[str, Any]]
+
+    def __init__(self, function, **kwargs):
+        """
+        :param function: A string that represents the function specification.
+            It will be locaded dynamically and executed passing the given kwargs.
+            E.g.: `myapp.my_lib.do_domething`
+
+        :param kwargs: Set of parameter to pass to the `function`. The parameters
+            must be JSON serializable to support multiple task runners implementations.
+
+        :type function: str
+        """
+        assert isinstance(function, str)
+
+        try:
+            json.dumps(kwargs)
+        except TypeError:
+            raise TaskNotSerializableError("Task kwargs is not serializable.")
+
+        self.function = function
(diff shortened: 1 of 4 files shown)


shuup/core/tasks.py
```diff
@@ -0,0 +1,97 @@
...
+        :param function: A string that represents the function specification.
+            It will be locaded dynamically and executed passing the given kwargs.
+            E.g.: `myapp.my_lib.do_domething`
+
+        :param kwargs: Set of parameter to pass to the `function`. The parameters
+            must be JSON serializable to support multiple task runners implementations.
+
+        :type function: str
+        """
+        assert isinstance(function, str)
```
ruohola, line 35: Maybe we shouldn't use `assert` statements for other than helping dev/debuggin. Bad to use for validating types, since they can be disabled with a flag or env variable when starting Python.

shuup/core/tasks.py
```diff
@@ -0,0 +1,97 @@
...
+class TaskNotSerializableError(Exception):
+    """
+    Raised when the task can't be serialized.
+    """
+    pass
+
+
+class Task(object):
+    function = None     # type: str
+    kwargs = None       # type: dict
```
ruohola, line 22: Since the values can be `None`, the correct type comments here would be:

```
function = None     # type: Optional[str]
kwargs = None       # type: Optional[dict]
```

Or more preferably the `dict` type would be specified a bit more:

```
kwargs = None  # type: Optional[Dict[str, Any]]
```

shuup/core/tasks.py
```diff
@@ -0,0 +1,97 @@
...
+        try:
+            json.dumps(kwargs)
+        except TypeError:
+            raise TaskNotSerializableError("Task kwargs is not serializable.")
+
+        self.function = function
+        self.kwargs = kwargs
+
+
+class TaskRunner(object):
```
ruohola, line 46: Should probably get used to just leaving out the `(object)` from all new code we write, since it has no meaning in Python 3.

shuup/core/tasks.py
chessbr: That is exactly for that. For devs not doing shit :)

shuup/core/tasks.py
chessbr: Makes sense, we are used to write python2 compatible code 🤷

shuup/core/tasks.py
chessbr: kwargs should be any dict that can be parsed into JSON, so maybe `Any` there is weird.

shuup/core/tasks.py
ruohola: Hopefully those days are over now :D

shuup/core/tasks.py
ruohola: Type hinting JSON is a long standing issue, but at least I always use the same type alias as Guido does https://github.com/python/typing/issues/182#issuecomment-199532520.

shuup/core/tasks.py
chessbr: many years doing this, give me some time :)

PR #2209, opened 2020-06-18
Admin: Improve window.Messages hiding and showing

ruohola:
- Reset the timeout for hiding the messages when a new message is added.
- Immediately clear the already hidden messages a when new one is added.
- Don't hide messages when clicking just random background elements.
- Allow dismissing all of the messages by clicking any one of them anywhere.

Came up when working on DO-138

### DIFF
--- a/shuup/admin/static_src/base/js/messages.js
+++ b/shuup/admin/static_src/base/js/messages.js
@@ -8,13 +8,15 @@
  */
 window.Messages = (function Messages(document) {
     const queue = [];
-    var container = null;
+    let container = null;
+    let timeOutForHide = null;
+    let timeOutForClear = null;
+
     function createContainer() {
         if (!container) {
             container = document.createElement("div");
             container.id = "message-container";
             document.body.appendChild(container);
-            document.addEventListener("click", hideOnClickOut, false);
         }
     }
     function show() {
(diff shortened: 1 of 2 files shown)


shuup/admin/static_src/base/js/messages.js
```diff
@@ -60,13 +54,12 @@ window.Messages = (function Messages(document) {
 
         const dismissIcon = document.createElement("span");
         dismissIcon.className = "dimiss-icon";
```
ruohola, line 56: I tried to fix this typo here to `"dismiss-icon"` and change the corresponding class name in `shuup/admin/static_src/base/scss/shuup/messages.scss`, but it just wouldn't work and the X icon was not showing.

shuup_tests/browser/admin/test_product_detail.py
```diff
@@ -56,8 +54,7 @@ def test_product_detail(browser, admin_user, live_server, settings):
 
     # Make sure that the tabs are clickable in small devices
     browser.driver.set_window_size(480, 960)
-    time.sleep(0.5)     # Otherwise other elements are still in the way, obscuring
```
ruohola, line 59: Worked fine without these sleeps that are just slowing down the pipeline.

shuup_tests/browser/admin/test_product_detail.py
```diff
@@ -56,8 +54,7 @@ def test_product_detail(browser, admin_user, live_server, settings):
...
-    click_element(browser, "#product-images-section", header_height=960)
+    click_element(browser, "#product-images-section", header_height=0)
```
ruohola, line 57: When the browser was scrolled all the way up with `header_height=960` and it tried to click this id that was outside the viewport, of course then the scroll-up arrow is blocking it:

>     selenium.common.exceptions.ElementClickInterceptedException: Message: element click intercepted: Element <div id="product-images-section" class="content-block " style="display: none;">...</div> is not clickable at point (250, 807). Other element would receive the click: <i class="fa fa-chevron-up"></i>

So instead we scroll all the way down and then try to click the tab, since the "product images" tab is around the bottom of the list. This still makes sure that the tab itself is wide enough so that it can be clicked.

PR #2213, opened 2020-07-02
Admin: Add settings for disabling HTML in descriptions

ruohola:
New settings:
`SHUUP_ADMIN_ALLOW_HTML_IN_PRODUCT_DESCRIPTION` (default `True`)
`SHUUP_ADMIN_ALLOW_HTML_IN_VENDOR_DESCRIPTION` (default `True`)
The default values are `True` for allowing full backwards compatibility.
New projects should most probably set these to `False`.

Refs SHU-24

### DIFF
--- a/setup.py
+++ b/setup.py
@@ -82,6 +82,7 @@ utils.add_exclude_patters([
 
 REQUIRES = [
     'Babel==2.5.3',
+    'bleach==3.1.5',
     'Django>=1.8,<2',
     'django-bootstrap3>=6.1,<10',
     'django-countries>=3.3,<5.3',
--- a/shuup/admin/modules/suppliers/forms.py
+++ b/shuup/admin/modules/suppliers/forms.py
@@ -7,6 +7,7 @@
 # LICENSE file in the root directory of this source tree.
 from __future__ import unicode_literals
 
+import bleach
 from django import forms
 from django.conf import settings
 from django.utils.encoding import force_text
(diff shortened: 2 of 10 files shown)


shuup/admin/modules/suppliers/forms.py
```diff
@@ -63,6 +66,11 @@ def clean(self):
         if stock_managed and not module_identifier:
             self.add_error("stock_managed", _("It is not possible to manage inventory when no module is selected."))
 
+        if not settings.SHUUP_ADMIN_ALLOW_HTML_IN_VENDOR_DESCRIPTION:
+            for key, value in cleaned_data.items():
+                if key.startswith("description__"):
+                    cleaned_data[key] = bleach.clean(value, tags=[])
```
ruohola, line 72: Is there a better way to clean translatable fields?

setup.py
```diff
@@ -82,6 +82,7 @@
 
 REQUIRES = [
     'Babel==2.5.3',
+    'bleach==3.1.5',
```
ruohola, line 85: I think this is a pretty solid tool library to have in reqs https://github.com/mozilla/bleach

shuup/admin/modules/suppliers/forms.py
ruohola: IMHO the current way is cleaner than, but there is a extra loop so idk:

```
if not settings.SHUUP_ADMIN_ALLOW_HTML_IN_VENDOR_DESCRIPTION:
    key = self.trans_name_map[self.language_code]["description"]
    cleaned_data[key] = bleach.clean(cleaned_data[key], tags=[])
```

PR #2225, opened 2020-07-16
Prevent crashing when trying to cache an unpicklable value

ruohola:
No refs

### DIFF
--- a/shuup/core/cache/impl.py
+++ b/shuup/core/cache/impl.py
@@ -7,9 +7,11 @@
 # LICENSE file in the root directory of this source tree.
 from __future__ import unicode_literals
 
+import logging
 import random
 import threading
 import time
+from pickle import PicklingError
 
 from django.conf import settings
 from django.core.cache import caches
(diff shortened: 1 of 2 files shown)


shuup/core/cache/impl.py
```diff
@@ -124,7 +129,12 @@ def set(self, key, value, timeout=None, version=None):
             timeout = get_cache_duration(key)
         if version is None:
             version = self.get_version(key)
-        self._cache.set(key, value, timeout=timeout, version=version)
+        try:
+            self._cache.set(key, value, timeout=timeout, version=version)
+        except PicklingError:
+            LOGGER.warning(
```
chessbr, line 135: I think you should use `LOGGER.exception` here. This is a critical issue, the developer basically did shit here passing somethings that can't be stored.

ruohola: Yeah, I think that makes most sense!

ruohola: Fixd.

PR #2234, opened 2020-07-29
Fix admin order edit tool

wiltso:
- Fixed so the admin order edit tool uses the correct id for the supplier query and not product id


shuup/admin/modules/orders/json_order_creator.py
```diff
@@ -110,6 +110,12 @@ def _process_line_quantity_and_price(self, source, sline, sl_kwargs):
 
     def _process_product_line(self, source, sline, sl_kwargs):
         product_info = sline.pop("product", None)
+        supplier_info = sline.pop("supplier", None)
+
+        if not supplier_info:
+            self.add_error(ValidationError(_("Product line dose not have a supplier"), code="no_supplier"))
```
chessbr, line 116: typo

PR #2239, opened 2020-08-03
Admin: add option to arrange menu for superuses, staff and suppliers

tulimaki:
For now it was only possible to arrange menu per user which is not sufficient while the menu needs to be arranged for the whole group of people like shop staff or vendors.

Allow to create menu custom menu for superusers, staff or suppliers, but remain the possibility to still arrange the menu per user.

Add option to translate each menu arranged for these groups since not all vendors/suppliers necessary speak same language.


shuup/admin/menu.py
```diff
@@ -106,11 +113,33 @@ def extend_main_menu(menu):
...
 
 
-def customize_menu(entries, user):  # noqa (C901)
+def customize_menu(entries, request):  # noqa (C901)
     """
     Merge system menu with customized admin menu
     """
-    customized_admin_menu = configuration.get(None, "admin_menu_user_{}".format(user.pk))
+    customized_admin_menu = configuration.get(
+        None,
```
chessbr, line 121: should this be per shop?

tulimaki:
Admin browser are tested locally and are fine. The browser tests just pretty unstable lately with Travis.

PR #2248, opened 2020-08-13
Add warning text t template

wiltso:
- Added warning text to the template about the ignore column in the example files

refs JDL-31


shuup/importer/templates/shuup/importer/admin/import.jinja
```diff
@@ -16,6 +16,11 @@
             <div class="col-md-6 col-md-offset-3">
                 {% if importer and importer.has_example_file() %}
                 <h2>{% trans %}Example files{% endtrans %}</h2>
+                <p>
+                    {% trans %}If there is a 'Ignore' column in the example file you are using.
+                    Make sure to leave the ignore column empty for all your rows,
+                    or else the row will not be imported.{% endtrans %}
```
chessbr, line 22: Make sure to leave the ignore column empty for all your rows, otherwise the rows won't be imported.

PR #2283, opened 2020-10-07
Notify: Show more Order related variables in editor

ruohola:
Add a new `attributes` attribute to `shuup.notify.base.Variable` for
showing examples of which attributes can be accessed in the script templates.

Refs IP-7

The new `attributes` look like this in the script editor:
**(Note that the the javascript for the variable links didn't work even before this PR, but once it gets fixed the attributes were made so that they should work just fine with that)** 

image

### DIFF
--- a/shuup/notify/base.py
+++ b/shuup/notify/base.py
@@ -67,7 +67,22 @@ class BaseMetaclass(type):
 class Variable(object):
     _creation_counter = 0  # For sorting, incremented by `__init__`
 
-    def __init__(self, name, type=Type, required=True, help_text=""):
+    def __init__(self, name, type=Type, required=True, help_text="", attributes=()):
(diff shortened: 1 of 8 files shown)


shuup/notify/base.py
```diff
@@ -78,6 +89,8 @@ def __init__(self, name, type=Type, required=True, help_text=""):
         self.type = type
         self.required = bool(required)
         self.help_text = help_text
+        self.dotted_accessor = dotted_accessor
+        self.cannot_be_bound = True if self.dotted_accessor else cannot_be_bound
```
ruohola, line 93: I made this so, because I was uncertain if the variable binding would work otherwise properly.

tulimaki (approved):

PR #2286, opened 2020-10-12
Add spanish translations

Hispar:


tulimaki:
@Hispar Thanks for the translations. I think we can include this to release tomorrow. Just need to double check what is wrong with that browser test and we are good with this.

tulimaki:
@Hispar did you btw add these translations manually or did you pull these from Transifex when you created the pull requests. Just checking whether I should push these with Transifex since this is our primary source for all Shuup translations.

tulimaki:
I created pull request what I pulled from Transifex https://github.com/shuup/shuup/pull/2291. Does this override your translations or just amend to those. What you think. I do not speak Spanish and would like to avoid overriding any of your work.

Hispar:
I created this PR before starting to translate in transifex. Most translations should be the same or very similar. 
I haven't used transifex before and wasn't sure of how lo put them in the code, It seems there are more resources than modules.

tulimaki:
Yeah in Transifex we have added few modules which are not at the main repository. With transifex you can use transifex client https://docs.transifex.com/client/installing-the-client and then pull the resources with simly `tx pull -l es`.

But if you give you think that pull request is fine let me know. Other than that I will double check before merging that we are not overriding things. Looks like it just updating the syntax when pulling with client.

As as side note: I will machine translate all those untranslated resources for Spanish and will leave those unreviewed. If you want to add more resources you then should have easier task to just review the already translated strings.

tulimaki:
@Hispar ^

Hispar:
@tulimaki thanks for your work, I've reviewed the PR and seems to be fine. I'll probably find some errors when the translation is in use, but I'll be updating in transifex anything that my users find.

I'm going to check the transifex client, to make things better in future PRs

PR #2289, opened 2020-10-14
Admin: Require media.view-all permission to see 'Root' folder

ruohola:
The reason for this change was that previously all of the items in
the Root folder (aka `folder=None`) were visible for all users with
media browsing permission. This could lead to the situation where
a user uploads some files to that folder and doesn't realize that it
will be visible for many other users as well.

Refs AH-30

As expected, a superuser sees the 'Root' folder in `/admin/media/`:

Screenshot_2020-10-14 20 50 00

But an other kind of staff account, with no `"media.view-all"` permission, cannot see it at all:

image

### DIFF
--- a/shuup/admin/modules/media/static_src/media/browser/BrowserView.js
+++ b/shuup/admin/modules/media/static_src/media/browser/BrowserView.js
@@ -100,7 +100,18 @@ export function controller(config={}) {
     };
     ctrl.reloadFolderTree = function() {
         remote.get({"action": "folders"}).then(function(response) {
-            ctrl.rootFolder(response.rootFolder);
+            if (response.rootFolder.name === "Root" && !response.rootFolder.canSeeRoot) {
+                // Hide the root folder if the user cannot access it.
+                // This is only a visual thing to avoid showing the empty folder.
+                // The views handles the actual permission checking and clears the whole file
+                // QueyrSet of the root folder if the user is not allowed to access it.
+                const currentFolder = response.rootFolder.children[0];
+                ctrl.rootFolder(currentFolder || {});
+                ctrl.currentFolderId(currentFolder ? currentFolder.id : null);
+                ctrl.reloadFolderContents();
+            } else {
+                ctrl.rootFolder(response.rootFolder);
+            }
             ctrl._refreshCurrentFolderPath();
         });
     };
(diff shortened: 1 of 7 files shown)


shuup/admin/modules/media/views.py
```diff
@@ -257,7 +256,11 @@ def handle_get_folder(self, data):
             else:
                 folder = None
                 subfolders = _get_folder_query(shop, self.user).filter(parent=None)
```
ruohola, line 259: This should most likely go inside the if-block as well to get the same treatmeant as files.

shuup/admin/modules/media/static_src/media/browser/BrowserView.js
```diff
@@ -100,7 +100,18 @@ export function controller(config={}) {
     };
     ctrl.reloadFolderTree = function() {
         remote.get({"action": "folders"}).then(function(response) {
-            ctrl.rootFolder(response.rootFolder);
+            if (response.rootFolder.name === "Root" && !response.rootFolder.canSeeRoot) {
+                // Hide the root folder if the user cannot access it.
+                // This is only a visual thing to avoid showing the empty folder.
+                // The views handles the actual permission checking and clears the whole file
```
ruohola, line 106: Typo:

```suggestion
                // The view handles the actual permission checking and clears the whole file
```

shuup/admin/modules/media/static_src/media/browser/BrowserView.js
```diff
@@ -100,7 +100,18 @@ export function controller(config={}) {
     };
     ctrl.reloadFolderTree = function() {
         remote.get({"action": "folders"}).then(function(response) {
-            ctrl.rootFolder(response.rootFolder);
+            if (response.rootFolder.name === "Root" && !response.rootFolder.canSeeRoot) {
```
ruohola, line 103: Idk, maybe checking for `response.rootFolder.id === 0 && ...` would be cleaner instead.

shuup/admin/modules/media/static_src/media/browser/BrowserView.js
```diff
@@ -100,7 +100,18 @@ export function controller(config={}) {
...
+                // Hide the root folder if the user cannot access it.
+                // This is only a visual thing to avoid showing the empty folder.
+                // The views handles the actual permission checking and clears the whole file
+                // QueyrSet of the root folder if the user is not allowed to access it.
+                const currentFolder = response.rootFolder.children[0];
+                ctrl.rootFolder(currentFolder || {});
```
ruohola, line 109: This will actually most likely result in any sibling folders of `currentFolder` to be completely inaccessable...

chessbr: IMO, these folders shouldn't even come from the API. Can this expose things we don't want to?

chessbr (approved):

PR #2293, opened 2020-10-21
Fix products highlights plugin not fully populating

ranta:
If `orderable_only=True` and we have products that are not orderable and they are in in the part that doesn't get cut off by `[:n_products]` then the plugin doesn't fully populated.

Refs CR-86


chessbr (changes requested):
We can't load all products from the database into the memory.
This highlight plugin is in roadmap to be refactored.

PR #2295, opened 2020-10-22
Fix taxless price hiding in checkout

eemeliranta:
Refs IP-15.
Originally it was comparing Taxless and Taxful prices which always resulted in True.

image


chessbr (approved):

ruohola (approved):
One more rebase to sort out the conflicts and to make Travis run 👍

PR #2296, opened 2020-10-22
Add direct supplier shop relationship through a model 

chessbr:
Refs VET-18


tulimaki (commented):
Why we still need supplier.shops and also SupplierShop?

chessbr:
> Why we still need supplier.shops and also SupplierShop?

You can easily manage new supplier shops with that field:
```py
supplier.shops.add(shop)
supplier.shops.set([shop1, shop2])
```
It will, behind the scenes, manage the SupplierShop, like it always did, but now, we can access this M2M model directly when we need to.

tulimaki (approved):
Tests are pretty solid. I will add little bit stronger note to changelog as well to address the changes required to querysets. I think mostly .enabled() is used in projects.

PR #2300, opened 2020-11-05
Core: fix basket attribute issue

chessbr:
Fix the implementation that was using the same memory
object for all baskets instances in the same process.

The `_DataValueProperty` was wrongly implemented while saving
the default value inside the `__init__`. The default object was a dictionary
that was always referencing to the same instance for all baskets instances,
and this object was updated by the same process every time.

The `_DataValueProperty` was replaced by an implementation in a way
that the default value is always a new dict object.


tulimaki:
Ok I will merge this. Some issue with Travis builds but all tests passed once. The admin part which is not affected by this is palying some sort of game (stale).

PR #2305, opened 2020-11-11
Include child categories in product category filter

eemeliranta:
Refs EZI-31
Include Products belonging to Child-Categories of selected category filter(s)

eg.
Instead of giving product three categories:
- Category: Dress / Women / Clothes
- Category: Dress / Women
- Category: Dress

Only the deepest one would be needed:
- Category: Dress / Women / Clothes

Currently If product has only "Dress" category set, it would only be shown in "Dress", not in "Women" or "Clothes".

### DIFF
--- a/shuup/front/forms/product_list_modifiers.py
+++ b/shuup/front/forms/product_list_modifiers.py
@@ -372,7 +372,8 @@ class CategoryProductListFilter(SimpleProductListModifier):
 
         categories = [cat.strip() for cat in categories if cat]
         if categories:
-            return Q(shop_products__categories__in=categories)
+            return Q(shop_products__categories__in=Category.objects.get_queryset_descendants(
+                Category.objects.filter(pk__in=categories), include_self=True))
 
     def get_admin_fields(self):
         default_fields = super(CategoryProductListFilter, self).get_admin_fields()
--- a/shuup/front/views/category.py
+++ b/shuup/front/views/category.py
@@ -63,7 +63,7 @@ class CategoryView(DetailView):
         return {
             "shop_products__shop": self.request.shop,
             "variation_parent__isnull": True,
-            "shop_products__categories": self.object,
+            "shop_products__categories__in": self.object.get_descendants(include_self=True),
             "shop_products__suppliers__in": Supplier.objects.enabled(shop=self.request.shop)
         }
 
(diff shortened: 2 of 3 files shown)


shuup/front/forms/product_list_modifiers.py
```diff
@@ -371,6 +371,12 @@ def get_filters(self, request, data):
             categories = list(categories)
 
         categories = [cat.strip() for cat in categories if cat]
+        category_objects = Category.objects.filter(pk__in=categories)
+
+        descendants = [list(dec.get_descendants().values_list("pk", flat=True)) for dec in category_objects]
```
tulimaki, line 376: I would double think the performance here.

shuup/front/views/category.py
```diff
@@ -63,7 +63,7 @@ def get_product_filters(self):
         return {
             "shop_products__shop": self.request.shop,
             "variation_parent__isnull": True,
-            "shop_products__categories": self.object,
+            "shop_products__categories__id__in": self.object.get_descendants().values_list("pk", flat=True),
```
tulimaki, line 66: Does this include the current category or only children?

eemeliranta: You were correct, it included only children. Patched it to include current category too.

shuup/front/views/category.py
```diff
@@ -60,10 +60,12 @@ def get_queryset(self):
         )
 
     def get_product_filters(self):
+        category_ids = [self.object.pk]
+        category_ids.extend(self.object.get_descendants().values_list("pk", flat=True))
```
chessbr, line 64: Just one time could solve?
```py
categories = self.objectget_descendants(include_self=True)
```

shuup/front/views/category.py
```diff
@@ -60,10 +60,12 @@ def get_queryset(self):
...
         return {
             "shop_products__shop": self.request.shop,
             "variation_parent__isnull": True,
-            "shop_products__categories": self.object,
+            "shop_products__categories__id__in": category_ids,
```
chessbr, line 68: and then:
```py
"shop_products__categories__in": categories,
```

shuup/front/forms/product_list_modifiers.py
chessbr: Agreed, I think there is a way of doing this in a single queryset. You need to understand the MPTT model so you can filter properly. Please check this one: https://github.com/django-mptt/django-mptt/blob/master/mptt/managers.py#L191 
Test that you can use this way:
```py
categories = Category.objects. get_queryset_descendants(Category.objects.filter(pk__in=categories), include_self=True)
```

chessbr (changes requested):

chessbr (approved):

PR #2309, opened 2020-11-18
Add front: clean hook for admin fields in product list modifier

wiltso:
- Add so there is a clean hook for the admin fields for product list form modifier
This is so that if a admin fields value is not JSON serializable we can clean the data from that field and then it can be serializable and save as a shop config.

refs https://trello.com/c/dywdlbNj/5-product-configuration

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -8,6 +8,9 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
 
 List all changes after the last release here (newer on top). Each change on a separate bullet point line
 
+### Added
+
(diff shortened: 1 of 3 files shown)


chessbr (changes requested):
I think this doesn't make sense. You can customize your own ConfigurationForm right?

wiltso:
> I think this doesn't make sense. You can customize your own ConfigurationForm right?
Well I would need to customize the view first  and then the form https://github.com/shuup/shuup-multivendor/blob/master/shuup_multivendor/views/vendor.py#L37 instead of just creating the form

PR #2313, opened 2020-11-23
Add font size 16 to SummerNote text editor

eemeliranta:
Other font sizes were there by default.

Refs MYYR-89


ruohola (approved):
Verified that this works.

PR #2318, opened 2020-12-01
Front: Use measurement units from the settings

wiltso:
- Fixed so the mass and measurement units are the same as the settings and not hard coded

Refs JDL-2

### DIFF
--- a/shuup/front/templates/shuup/front/macros/product_information.jinja
+++ b/shuup/front/templates/shuup/front/macros/product_information.jinja
@@ -113,32 +113,25 @@
                             <dt>{% trans %}Available Until{% endtrans %}</dt>
                             <dd>{{ shop_product.available_until|date }}</dd>
                         {% endif %}
+                        {% set mass_unit, length_unit = get_mass_and_length_units() %}
                         {% if product.width or product.height or product.depth %}
                             <dt>{% trans %}Width{% endtrans %}</dt>
-                            <dd>{{ product.width|floatformat(2) }} mm</dd>
+                            <dd>{{ product.width|floatformat(2) }} {{ length_unit }}</dd>
                             <dt>{% trans %}Height{% endtrans %}</dt>
-                            <dd>{{ product.height|floatformat(2) }} mm</dd>
+                            <dd>{{ product.height|floatformat(2) }} {{ length_unit }}</dd>
                             <dt>{% trans %}Depth{% endtrans %}</dt>
-                            <dd>{{ product.depth|floatformat(2) }} mm</dd>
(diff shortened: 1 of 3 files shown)


shuup/front/templatetags/shuup_front.py
```diff
@@ -58,3 +59,11 @@ def replace_field_attrs(field, **attrs):
     for attr, value in attrs.items():
         setattr(field.field, attr, value)
     return field
+
+
+@django_jinja.library.global_function
+def get_mass_and_measurement_unit():
```
ruohola, line 65: Rename this to `get_mass_and_length_units` and add in the docstring the return type:

```
:rtype: Tuple[str, str]
```

shuup/front/templates/shuup/front/macros/product_information.jinja
```diff
@@ -113,32 +113,25 @@
...
+                            <dd>{{ product.depth|floatformat(2) }} {{ length_unit }}</dd>
                         {% endif %}
                         {% if product.net_weight %}
                             <dt>{% trans %}Net Weight{% endtrans %}</dt>
                             <dd>
-                                {% if product.net_weight < 1000 %}
-                                    {{ product.net_weight|floatformat(2) }} g
-                                {% else %}
-                                    {{ (product.net_weight / 1000)|floatformat(2) }} kg
-                                {% endif %}
```
ruohola, line 131: @tulimaki What do you think, do we need to special case this `g`/`kg` displaying for backwards compatibility?

ruohola (changes requested):
^

shuup/front/templates/shuup/front/macros/product_information.jinja
tulimaki: Well how this would work? You are showing product net weight on a template and that net wight have measurement unit defined by settings?

ruohola: That's why I said *special* case, e.g.

```
{% if mass_unit == "g" %}
    {# Special casing the conversion for backwards compatibility. #} 
    {% if product.net_weight < 1000 %}
        {{ product.net_weight|floatformat(2) }} g
    {% else %}
        {{ (product.net_weight / 1000)|floatformat(2) }} kg
    {% endif %}
{% else %}
    {{ product.gross_weight|floatformat(2) }} {{ mass_unit }}
{% endif %}
```

tulimaki: No conversions. Keep it simple.

tulimaki: No tests.

PR #2325, opened 2020-12-11
Core: optimize performance around cache bumps

tulimaki:
- Remove duplicate queries and actions.
- Optimize queries using id's instead objects

### DIFF
--- a/shuup/core/signals.py
+++ b/shuup/core/signals.py
@@ -19,7 +19,7 @@ payment_created = Signal(providing_args=["order", "payment"], use_caching=True)
 get_basket_command_handler = Signal(providing_args=["command"], use_caching=True)
 pre_clean = Signal(providing_args=["instance"], use_caching=True)
 post_clean = Signal(providing_args=["instance"], use_caching=True)
-context_cache_item_bumped = Signal(providing_args=["item"], use_caching=True)
+context_cache_item_bumped = Signal(providing_args=["item", "shop_id"], use_caching=True)
 order_status_changed = Signal(providing_args=["order", "old_status", "new_status"], use_caching=True)
 
 #: Send from supplier module after the stocks updated have
--- a/shuup/front/signal_handlers.py
+++ b/shuup/front/signal_handlers.py
@@ -11,7 +11,7 @@ from filer.models import Image
 
 from shuup.core import cache
 from shuup.core.models import (
-    Manufacturer, Product, ProductCrossSell, ProductMedia, Shop, ShopProduct
+    Manufacturer, ProductCrossSell, ProductMedia, Shop, ShopProduct
 )
 from shuup.core.signals import context_cache_item_bumped  # noqa
 from shuup.core.utils import context_cache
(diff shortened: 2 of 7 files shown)


shuup/core/signals.py
```diff
@@ -19,7 +19,7 @@
 get_basket_command_handler = Signal(providing_args=["command"], use_caching=True)
 pre_clean = Signal(providing_args=["instance"], use_caching=True)
 post_clean = Signal(providing_args=["instance"], use_caching=True)
-context_cache_item_bumped = Signal(providing_args=["item"], use_caching=True)
+context_cache_item_bumped = Signal(providing_args=["shop_id"], use_caching=True)
```
chessbr, line 22: well, this is certainly not backwards compatible, right? I mean, some 3rd party apps can use this to bump caches passing the item, but in the signal handler here https://github.com/shuup/shuup/pull/2325/files#diff-2dfff9a8fb51a9f6302cb45d759c752ab0ffd8583cfeab9fed9b35652f83b4d1R23 you don't check for that key, you probably should do that.

shuup/front/signal_handlers.py
```diff
@@ -20,25 +20,14 @@
...
-        context_cache.bump_cache_for_item(cache_utils.get_newest_products_cache_item(shop))
-        context_cache.bump_cache_for_item(cache_utils.get_products_for_category_cache_item(shop))
-        context_cache.bump_cache_for_item(cache_utils.get_random_products_cache_item(shop))
-        bump_product_queryset_cache()
-
-    elif issubclass(sender, Product):
-        bump_product_queryset_cache()
+def handle_context_cache_item_bumped(sender, shop_id, **kwargs):
+    shop = Shop.objects.filter(pk=shop_id).first()
+    context_cache.bump_cache_for_item(cache_utils.get_listed_products_cache_item(shop))
```
chessbr, line 25: can you change the `cache_utils` to receive either a shop_id or a shop instance, this way you prevent fetching the shop here for nothing

chessbr (changes requested):

shuup/front/utils/cache.py
```diff
@@ -15,32 +15,35 @@
 
 
 def get_listed_products_cache_item(shop):
-    return LISTED_PRODUCTS_CACHE_ITEM_FMT.format(shop_id=shop.pk)
+    shop_id = (shop.id if hasattr(shop, "pk") else shop)
```
chessbr, line 18: You can annotate the `shop` parameter to receive either a Shop instance OR an int, this makes it clear to the dev that we are expecting an instance, or the shop id
```py
def get_listed_products_cache_item(shop: Union[Shop, int]):
    shop_id = (shop.pk if isintance(shop, Shop) else shop))
    return LISTED_PRODUCTS_CACHE_ITEM_FMT.format(shop_id=shop_id)
```

shuup/front/signal_handlers.py
```diff
@@ -20,25 +20,17 @@
...
-        elif not shop:
-            shop = item.shop
-        context_cache.bump_cache_for_item(cache_utils.get_listed_products_cache_item(shop))
-        context_cache.bump_cache_for_item(cache_utils.get_best_selling_products_cache_item(shop))
-        context_cache.bump_cache_for_item(cache_utils.get_newest_products_cache_item(shop))
-        context_cache.bump_cache_for_item(cache_utils.get_products_for_category_cache_item(shop))
-        context_cache.bump_cache_for_item(cache_utils.get_random_products_cache_item(shop))
-        bump_product_queryset_cache()
+def handle_context_cache_item_bumped(sender, **kwargs):
+    shop_id = kwargs.get("shop_id", None)
```
chessbr, line 24: no, the old parameter was `item` which was the shop instance.
you can do like this:
```py
shop_id = kwargs.get("shop_id", None)
if not shop_id and kwargs.get("item") and isinstance(kwargs["item"], Shop):
   shop_id = kwargs["item"].pk
if not shop_id:
    shop_id = Shop.objects.first().pk
```

tulimaki: well I doubt the "item" is shop id here so plan was to just delete the cache for first shop if the shop id is not passed.

chessbr (approved):

PR #2330, opened 2020-12-15
Xtheme rendering speedup

tulimaki:


shuup/front/static_src/js/xtheme_async_plugins.js
```diff
@@ -0,0 +1,58 @@
...
+                            },
+                            1200: { // breakpoint from 992 up
+                                items : 5,
+                                slideBy: 3
+                            }
+                        }
+                    });
+                },
+                error: function(error) {
+                    that.find('.pretrol-carousel').html("")
```
chessbr, line 53: what is this class?

shuup/xtheme/plugins/products.py
```diff
@@ -39,44 +40,46 @@ class ProductHighlightPlugin(TemplatedPlugin):
     template_name = "shuup/xtheme/plugins/highlight_plugin.jinja"
     fields = [
         ("title", TranslatableField(label=_("Title"), required=False, initial="")),
-        ("type", forms.ChoiceField(
-            label=_("Type"),
-            choices=HighlightType.choices(),
-            initial=HighlightType.NEWEST.value
-        )),
+        ("type", forms.ChoiceField(label=_("Type"), choices=[
+            ("newest", "Newest"),
```
chessbr, line 44: why this change? there is no translations as well

tulimaki:
@chessbr Should I just maybe create completely new plugins for these async product highlight. This way we minimize possible breaks with existing customization and template overrides?

shuup/xtheme/plugins/products.py
```diff
@@ -241,5 +250,12 @@ def get_context_data(self, context):
         return {
             "request": request,
             "title": self.get_translated_value("title"),
-            "products": products_qs
+            "products": products_qs,
+            "data_url": reverse(
+                "shuup:xtheme-product-selections-highlight",
+                kwargs=dict(
+                    product_ids=",".join(products),
```
chessbr, line 257: what is this? what is we have a lot of products? the URL has a limit in size and this looks so wrong.
Can't you render this info in DOM using a `data-attribute` ?

shuup/xtheme/rendering.py
```diff
@@ -117,22 +118,36 @@ def render(self):
...
-                    "content": content,
-                    "wrapper_end": "</div>",
-                })
+            cache_key = "shuup_xtheme_placeholder_%s_%s_%s_%s" % (
+                self.view_config.saved_view_config.pk,
+                language,
+                self.placeholder_name,
+                get_layout_data_key(self.placeholder_name, layout, self.context)
+            )
+            layout_content = cache.get(cache_key)
```
chessbr, line 134: have you really tested this? what if we change some property of the plugin, will it be updated? shouldn't we bump the cache when we update/create a SavedViewConfig?

chessbr (changes requested):

shuup/xtheme/plugins/products.py
tulimaki: Well you still need to pass those id's to the view. I make max 10 choices for this plugin. Someone who wants more can implement something for it.

shuup/xtheme/rendering.py
tulimaki: yes that why `saved_view_config.pk` is in cache key. It bumps whenever you do any changes with the editor for the view.

shuup/xtheme/models.py
```diff
@@ -103,6 +103,7 @@ class SavedViewConfig(models.Model):
     shop = models.ForeignKey(on_delete=models.CASCADE, to="shuup.Shop", related_name="saved_views_config", null=True)
     view_name = models.CharField(max_length=64, db_index=True, verbose_name=_("view name"))
     created_on = models.DateTimeField(auto_now_add=True, verbose_name=_("created on"))
+    modified_on = models.DateTimeField(auto_now=True, editable=False, verbose_name=_("modified on"))
```
chessbr, line 106: should this field have a db_index?

shuup/xtheme/urls.py
```diff
@@ -10,9 +10,37 @@
...
+from shuup.xtheme.views.plugins import (
+    get_category_products_highlight, get_product_cross_sell_highlight,
+    get_product_highlight, get_prouduct_selections_highlight
+)
 
 urlpatterns = [
     url(r"^xtheme/editor/$", EditorView.as_view(), name="xtheme_editor"),
     url(r"^xtheme/(?P<view>.+)/*$", extra_view_dispatch, name="xtheme_extra_view"),
     url(r"^xtheme/$", command_dispatch, name="xtheme"),
+    url(
```
chessbr, line 22: I would be nice to put these async plugins into a different app, what do you think? looks like this is polluting the xtheme base app.

shuup/xtheme/__init__.py
```diff
@@ -44,6 +44,10 @@ class XThemeAppConfig(AppConfig):
             "shuup.xtheme.plugins.products:ProductHighlightPlugin",
             "shuup.xtheme.plugins.products:ProductCrossSellsPlugin",
             "shuup.xtheme.plugins.products:ProductSelectionPlugin",
+            "shuup.xtheme.plugins.products_async:ProductsFromCategoryPlugin",
```
chessbr, line 47: wondering if ALL these plugins could live in a different addon?

shuup/xtheme/plugins/products_async.py
```diff
@@ -0,0 +1,290 @@
...
+    BEST_SELLING = "best_selling"
+    RANDOM = "random"
+
+    class Labels:
+        NEWEST = _("Newest")
+        BEST_SELLING = _("Best Selling")
+        RANDOM = _("Random")
+
+
+class ProductHighlightPlugin(TemplatedPlugin):
```
chessbr, line 39: unfortunately, we are repeating ourselves with all this duplicated code. What is we have some sort of async field that is disabled by default that would then change the way we render the template and the context is gethered?

shuup/xtheme/models.py
tulimaki: I don't think since we never filter or sort by this.

PR #2332, opened 2020-12-17
Front: Checkout Phase

jeromeshuup:
- Fix `_vertical_phases.jinja` incorrectly targets all forms

### DIFF
--- a/shuup/front/templates/shuup/front/checkout/_vertical_phases.jinja
+++ b/shuup/front/templates/shuup/front/checkout/_vertical_phases.jinja
@@ -15,7 +15,7 @@
 {% endif %}
 
 {% block content %}
-    <div id="ajax_content">
+    <div id="ajax_content" class="vertical-checkout-phase">
         {% if view and view.phases %}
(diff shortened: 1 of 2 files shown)


shuup/front/templates/shuup/front/checkout/_vertical_phases.jinja
```diff
@@ -89,7 +89,7 @@
             }
         };
         $(function() {
-            $("form").on("submit", function(e) {
+            $("form").on("submit","[id='addresses'], [id='methods'], [id='payment-form'], [id='confirm_form']", function(e) {
```
tulimaki, line 92: Can we not expect here some id:s for the forms.

Why don't you for example add some class here like `vertical-checkout-phase` or similar then update that jquery selector to something like `$('.vertical-checkout-phase').find("form")` that way we trigger this on submit for forms inside that ajax content.

tulimaki (changes requested):

tulimaki (approved):

PR #2342, opened 2021-01-04
Always use cached shop product when possible

chessbr:
This prevents doing a lot of queries while handling the basket


shuup/core/basket/objects.py
```diff
@@ -44,7 +44,7 @@ def shop_product(self):
 
         :rtype: shuup.core.models.ShopProduct
         """
-        return self.product.get_shop_instance(self.shop)
+        return self.product.get_shop_instance(self.shop, allow_cache=True)
```
tulimaki, line 47: It already looked like a cached value?

PR #2351, opened 2021-01-11
Optimize default front calculations around variation products

tulimaki:
- Front: optimize child product orderability checks a bit
- Front: add option to replace product detail context
- Front: optimize rendering images for variation products
- Core: optimize price range calculations

Note: the price range calculation is also huge help in listing products since the loading time wont be affected on how complex are the variation products.

### DIFF
--- a/shuup/core/utils/price_display.py
+++ b/shuup/core/utils/price_display.py
@@ -161,6 +161,27 @@ class TotalPriceDisplayFilter(_ContextFilter):
         return money(total)
 
 
+def get_priced_children_for_price_range(request, product, quantity, supplier):
+    product_queryset = product.variation_children.visible(
+        shop=request.shop, customer=request.customer
+    ).order_by("shop_products__default_price_value", "pk")
+    if supplier:
+        product_queryset = product_queryset.filter(
+            shop_products__suppliers=supplier
+        )
+
+    low = product_queryset.first()
+    high = product_queryset.last()
+
+    if not (low or high):
+        return
+
+    return [
+        (low, _get_priceful(request, low, quantity, supplier)),
+        (high, _get_priceful(request, high, quantity, supplier)),
+    ]
+
+
 class PriceRangeDisplayFilter(_ContextFilter):
     def __call__(self, context, product, quantity=1, allow_cache=True, supplier=None):
         """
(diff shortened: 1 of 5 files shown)


shuup/front/utils/product.py
```diff
@@ -156,23 +157,27 @@ def get_orderable_variation_children(product, request, variation_variables, supp
     orderable_variation_children = OrderedDict()
     orderable = 0
 
-    for combo_data in product.get_all_available_combinations():
+    product_queryset = product.variation_children.visible(
```
chessbr, line 165: won't the `ProductQuerySet._visible` default parameter `invisible_modes=[ProductMode.VARIATION_CHILD]` be a problem here?

shuup/core/utils/price_display.py
```diff
@@ -161,6 +161,27 @@ def __call__(self, context, source, include_taxes=None):
         return money(total)
 
 
+def get_priced_childer_for_price_range(request, product, quantity, supplier):
```
chessbr, line 164: typo

chessbr (changes requested):

shuup/front/utils/product.py
tulimaki: That is only problem when using "listed" and "searchable" querysets since with those we do not want to list or search though children. As you see here https://github.com/shuup/shuup/blob/master/shuup/core/models/_products.py#L209

shuup/core/utils/price_display.py
```diff
@@ -161,6 +161,27 @@ def __call__(self, context, source, include_taxes=None):
         return money(total)
 
 
+def get_priced_children_for_price_range(request, product, quantity, supplier):
+    product_queryset = product.variation_children.visible(
+        shop=request.shop, customer=request.customer
+    ).order_by("shop_products__default_price_value", "pk")
```
tulimaki, line 167: https://stackoverflow.com/questions/51302345/django-last-and-first-return-the-same-object for this since I guess there is a case when default price value is zero for all children. Then this would ensure that the low and high get different products. Mainly this was needed for the test since it figures out the price from sku.

chessbr (approved):

PR #2355, opened 2021-01-13
Introduce subscription options

chessbr:
Refs SH-40

With this features, it is possible to create addons that can provide subscription logic:

image
image
image


shuup/front/templates/shuup/front/macros/order.jinja
```diff
@@ -146,6 +150,27 @@
     </div>
 {% endmacro %}
 
+{% macro render_line_properties(properties) %}
```
tulimaki, line 153: You now have this macro defined twice it seems. Not sure if bad idea since I think basket and order info rendering could be separated. Changing other shouldn't necessary change the other.

tulimaki (approved):
I think the 1st commit has been already added. You want most likely to rebase in top of the master.

PR #2356, opened 2021-01-14
Front: add detail page for carts

tulimaki:
- Modify carts list to show latest cart first
- Remove the delay filter as useless. No need to hide carts
- Hide "finished" carts by default
- Add detail page for carts
- Add option to add log entries linked to stored baskets
  - This is usefull when we want to log things from checkout phase
- Save basket after shipping or billing address is saved. For
  saving the addresses to basket data
- Save basket after shipping or payment method is saved.
- Save basket just before starting the order creation. This ensures we have latest basket there on store


shuup/front/admin_module/carts/views/_detail.py
```diff
@@ -0,0 +1,55 @@
...
+    template_name = 'shuup/front/admin/stored_basket_detail.jinja'
+
+    def get_queryset(self):
+        qs = super(CartDetailView, self).get_queryset()
+        return qs.prefetch_related("products")
+
+    def get_context_data(self, **kwargs):
+        context = super(CartDetailView, self).get_context_data(**kwargs)
+
+        basket = BaseBasket(
```
tulimaki, line 27: Not sure this is necessary. But this makes it easy to render the template. Also some thought here would be to maybe option to pass this for order creator.

shuup/front/models/stored_basket.py
```diff
@@ -27,16 +28,19 @@ class StoredBasket(MoneyPropped, models.Model):
     # A combination of the PK and key is used to retrieve a basket for session situations.
     key = models.CharField(max_length=32, default=generate_key, verbose_name=_('key'))
 
-    shop = models.ForeignKey(Shop, on_delete=models.CASCADE, verbose_name=_('shop'))
+    shop = models.ForeignKey("shuup.Shop", on_delete=models.CASCADE, verbose_name=_('shop'))
+    supplier = models.ForeignKey(
+        "shuup.Supplier", null=True, blank=True, on_delete=models.CASCADE, verbose_name=_('supplier')
```
chessbr, line 33: `SET_NULL` instead

shuup/front/models/stored_basket.py
```diff
@@ -27,16 +28,19 @@ class StoredBasket(MoneyPropped, models.Model):
...
+    )
 
     customer = models.ForeignKey(
-        Contact, blank=True, null=True,
+        "shuup.Contact", blank=True, null=True,
         on_delete=models.CASCADE,
```
chessbr, line 38: shouldn't be SET_NULL as well?

shuup/front/models/stored_basket.py
```diff
@@ -27,16 +28,19 @@ class StoredBasket(MoneyPropped, models.Model):
...
         related_name="customer_baskets",
         verbose_name=_('customer')
     )
     orderer = models.ForeignKey(
-        PersonContact, blank=True, null=True,
+        "shuup.PersonContact", blank=True, null=True,
         on_delete=models.CASCADE,
```
chessbr, line 44: shouldn't be SET_NULL as well?

chessbr (changes requested):

shuup/front/basket/storage.py
```diff
@@ -97,3 +102,16 @@ def _load_stored_basket(self, basket):
...
     def basket_exists(self, key, shop):
         return self.model.objects.filter(key=key, shop=shop).exists()
+
+    def add_log_entry(self, basket, message, extra={}, kind=LogEntryKind.NOTE):
+        try:
+            shop = basket.shop
+            if shop:
+                basket_kwargs = self.get_basket_kwargs(basket)
+                identifier = self.model.get_key_for_logs(basket_kwargs["key"])
+                shop.add_log_entry(kind=kind, identifier=identifier, message=message, extra=extra)
```
chessbr, line 112: what if adding the log to the `basket.data` ?

shuup/front/checkout/addresses.py
```diff
@@ -133,6 +133,12 @@ def form_valid(self, form):
             self.storage["company"] = form.forms["company"].save(commit=False)
         else:
             self.storage["company"] = None
+
+        process_and_save_basket = any([form[form_key].has_changed() for form_key in form.forms])
+        if process_and_save_basket:
+            self.process()
+            self.basket.save()
+            self.basket.storage.add_log_entry(self.basket, _("Saved addresses."))
```
chessbr, line 141: instead of calling the storage, why can't we create a method in BaseBasket to log some event? By default we could just add some line to `basket._data["logs"]` ?

chessbr (approved):

PR #2358, opened 2021-01-19
Xtheme: fix summernote icons by using the original summernote CSS file

chessbr:
Refs FD-143

**Before**
image

**After**
image


tulimaki:
The sortable issue with browser test is fixed here https://github.com/shuup/shuup/pull/2360

PR #2361, opened 2021-01-20
Admin: add apply filters button and display active filters counter badge

tulimaki:
Do not save and refresh Picotable lists on filter change, but wait that the user selects the "Apply filter" option. Also show badge for active filters to indicate that some content is filtered out. Consider "_all" as not filter.


chessbr (approved):
LGTM

PR #2366, opened 2021-01-23
Add Core: verbose_name to shop product so we can translate it

wiltso:
- Added verbose_name to the shop product class so we can translate it

refs https://trello.com/c/5PsvJ1QJ/110-more-translations


tulimaki (changes requested):
Fix migrations issue. Also rebase on top of the master and that "merge master" commit should not be required.

PR #2370, opened 2021-01-25
Add Core: verbose_name to shop product so we can translate it

tulimaki:
- Added verbose_name to the shop product class so we can translate it

refs https://trello.com/c/5PsvJ1QJ/110-more-translations


tulimaki:
Replaces https://github.com/shuup/shuup/pull/2366 with rebase and fixed migration entry.

PR #2373, opened 2021-01-26
Gracefully handle the error when deleting a PROTECTED MediaFolder

ruohola:
Refs SHU-73

### DIFF
--- a/shuup_tests/admin/test_media_module.py
+++ b/shuup_tests/admin/test_media_module.py
@@ -8,8 +8,11 @@ import django
 import json
 import tempfile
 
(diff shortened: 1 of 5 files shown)


tulimaki (approved):

shuup_tests/admin/test_media_module.py
```diff
@@ -8,8 +8,11 @@
 import json
 import tempfile
 
+import mock
```
ruohola, line 11: We could probably just be using `unittest.mock`, but since this was used everywhere else as well.

PR #2375, opened 2021-01-26
Dashboard improvements

tulimaki:


tulimaki:
Replaces https://github.com/shuup/shuup/pull/2374 and just adding that quick browser test fix.

PR #2384, opened 2021-02-04
Disable scrolling on jquery datetime pickers

ranta:
Disables scrolling on jquery date, datetime and time pickers, as the value went down regardless of scroll direction.
Allowing scrolling when cursor is hovering over the field when its not selected may also lead to mistakes.


ruohola (commented):
Wasn't there any way to fix the scrolling direction bug? I admit that it's a super annoying issue, so literally anything helps.

PR #2392, opened 2021-02-19
key error fix

dzungbe:
key error fix


tulimaki:
This was already fixed here https://github.com/shuup/shuup/pull/2389

PR #2398, opened 2021-02-23
Core: bump attribute name to 256 characters

tulimaki:


ruohola (approved):
No one has ever had a problem with too long db fields :D

PR #2399, opened 2021-02-26
Front: update media upload URL

tulimaki:
Avoid mixing up this with static media path when the ending "/" is not added to some checks.


chessbr (commented):
Well, in really it is nginx that doesn't have this URL configured correctly

chessbr (approved):

ruohola:
> Well, in really it is nginx that doesn't have this URL configured correctly

Yep.

PR #2400, opened 2021-02-26
Admin: fix product module not to fail on object does not exists

tulimaki:
Do not throw 500 if the module can not handle the object. Outcome is clear and that that is that URL can not be defined by that module. Log error and move to next module.

### DIFF
--- a/shuup/admin/modules/products/__init__.py
+++ b/shuup/admin/modules/products/__init__.py
@@ -155,12 +155,12 @@ class ProductModule(AdminModule):
 
     def get_model_url(self, object, kind, shop=None):
         if isinstance(object, Product):
-            if not shop:
-                try:
+            try:
+                if not shop:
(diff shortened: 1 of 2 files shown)


shuup/admin/utils/urls.py
```diff
@@ -221,7 +224,11 @@ def get_model_url(object, kind="detail", user=None,
     :rtype: str
     """
     for module in get_modules():
-        url = module.get_model_url(object, kind, shop)
+        try:
+            url = module.get_model_url(object, kind, shop)
+        except Exception as e:
+            LOGGER.exception(sr(e))
```
chessbr, line 230: I think you should target the correct exception here, maybe NoReverseMatch. But please check about the `raise_permission_denied` parameter from this function, maybe you should respect that as well

chessbr (approved):

PR #2402, opened 2021-03-01
Admin: add option to hide font selection and paste plain text in editors

chessbr:
Convert the script functions into an object that can is called in the base template

Also add more blocks to enable more customization of the base admin template

⚠️  This will break summernote in projects that customize `base.jinja` in Admin as they miss the new initialization block script.

### DIFF
--- a/shuup/admin/templates/shuup/admin/base.jinja
+++ b/shuup/admin/templates/shuup/admin/base.jinja
@@ -79,20 +79,36 @@
         {% endblock %}
         {% block post_content %}{% endblock %}
         {% endif %}
-        <script src="{{ shuup_static("shuup_admin/js/vendor.js") }}"></script>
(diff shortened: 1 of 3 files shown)


shuup/admin/templates/shuup/admin/base.jinja
```diff
@@ -79,20 +79,36 @@
...
+        {% block init_scripts %}
         <script>
-            (function(){
-                if(!window.Messages) return;
-                {% for message in messages -%}Messages.enqueue({
-                    tags: {{ message.tags|json }},
-                    text: {{ (message.tags.capitalize() ~ "! " ~ message.message)|json }}
-                });{% endfor %}
-            }());
+            if (window.ShuupAdminEditors) {
```
tulimaki, line 88: Can't this be inside base.js? Since it does not require anything from the jinja template?

tulimaki (approved):

PR #2403, opened 2021-03-02
Core: move refund order line logic to tax module

tulimaki:
Since each refund line you need to take care of the taxes it is more logical if the tax module handles the whole thing. With some 3rd party taxation tool you need an option to handle refund taxes in different way.

### DIFF
--- a/shuup/core/taxing/_module.py
+++ b/shuup/core/taxing/_module.py
@@ -5,12 +5,21 @@
 # This source code is licensed under the OSL-3.0 license found in the
 # LICENSE file in the root directory of this source tree.
 import abc
+from collections import defaultdict
 from itertools import chain
 
 import six
 from django.conf import settings
+from django.db import models, transaction
+from django.utils.translation import ugettext_lazy as _
 
 from shuup.apps.provides import load_module
+from shuup.core.excs import (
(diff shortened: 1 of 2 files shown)


shuup/core/taxing/_module.py
```diff
@@ -162,3 +171,162 @@ def get_taxed_price(self, context, price, tax_class):
         :rtype: shuup.core.taxing.TaxedPrice
         """
         pass
+
+    def _get_tax_class_proportions(self, order):
+        product_lines = order.lines.products()
+
+        zero = order.lines.first().price.new(0)
```
chessbr, line 178: is this really needed?

shuup/core/taxing/_module.py
```diff
@@ -162,3 +171,162 @@ def get_taxed_price(self, context, price, tax_class):
...
+                tax=line_tax.tax,
+                name=_("Refund for %s" % line_tax.name),
+                amount_value=-line_tax.amount,
+                base_amount_value=-line_tax.base_amount,
+                ordering=1
+            )
+        return refund_line
+
+    @transaction.atomic  # noqa (C901) FIXME: simply this
+    def do_refund(self, order, supplier, created_by, refund_data):
```
chessbr, line 226: can we find a better name? :D

chessbr (approved):

PR #2414, opened 2021-03-15
Language aware lru_cache

ranta:
`lru_cache` is shared between everyone and every language, which means incorrect translations are returned if different language is cached than is in use by user

### DIFF
--- a/shuup/utils/i18n.py
+++ b/shuup/utils/i18n.py
@@ -5,6 +5,8 @@
 #
 # This source code is licensed under the OSL-3.0 license found in the
 # LICENSE file in the root directory of this source tree.
+from functools import lru_cache, wraps
+
 import babel
 import babel.numbers
 from babel import UnknownLocaleError
@@ -12,10 +14,25 @@ from babel.dates import format_datetime
 from babel.numbers import format_currency, format_decimal, parse_pattern
 from django.apps import apps
 from django.utils import translation
-from django.utils.lru_cache import lru_cache
 from django.utils.timezone import localtime
 from django.utils.translation import get_language
 from django.views.decorators.cache import cache_page
+
+
+def lang_lru_cache(func):
+    """Language aware least recently used cache decorator"""
+
+    @lru_cache()
+    def cached(*args, __lang=None, **kwargs):
+        return func(*args, **kwargs)
+
+    @wraps(func)
+    def wrapper(*args, **kwargs):
+        return cached(*args, **kwargs, __lang=translation.get_language())
+
+    wrapper.cache_clear = cached.cache_clear
+
+    return wrapper
 
 
 @lru_cache()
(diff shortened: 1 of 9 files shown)


shuup/utils/i18n.py
```diff
@@ -5,19 +5,32 @@
...
 
+def lang_lru_cache(func):
+    """Language Aware Least-recently-used cache decorator"""
+
+    @lru_cache()
+    def cached(*args, __lang=None, **kwargs):
+        return func(*args, **kwargs)
+
+    def wrapper(*args, **kwargs):
+        return cached(*args, **kwargs, __lang=get_current_babel_locale())
```
ruohola, line 30: I think `translation.get_language()` should be more than fine here.

shuup/utils/i18n.py
```diff
@@ -5,19 +5,32 @@
...
 
 
+def lang_lru_cache(func):
+    """Language Aware Least-recently-used cache decorator"""
+
+    @lru_cache()
+    def cached(*args, __lang=None, **kwargs):
+        return func(*args, **kwargs)
+
+    def wrapper(*args, **kwargs):
```
ruohola, line 29: Add `wraps(func)` here, import `wraps` from `functools`: https://docs.python.org/3/library/functools.html#functools.wraps

shuup/utils/i18n.py
```diff
@@ -5,19 +5,32 @@
...
-from django.utils.lru_cache import lru_cache
 from django.utils.timezone import localtime
 from django.utils.translation import get_language
 from django.views.decorators.cache import cache_page
 
 
+def lang_lru_cache(func):
+    """Language Aware Least-recently-used cache decorator"""
+
+    @lru_cache()
```
ruohola, line 25: No need to call it here, `@lru_cache` is a bit cleaner.

shuup/utils/i18n.py
```diff
@@ -5,19 +5,32 @@
...
 from django.apps import apps
 from django.utils import translation
-from django.utils.lru_cache import lru_cache
 from django.utils.timezone import localtime
 from django.utils.translation import get_language
 from django.views.decorators.cache import cache_page
 
 
+def lang_lru_cache(func):
+    """Language Aware Least-recently-used cache decorator"""
```
ruohola, line 23: `"""Language aware least recently used cache decorator."""`

ruohola (commented):
Nicely implemented the `lang_lru_cache` :)

shuup/utils/i18n.py
ranta: `@lru_cache` doesn't support that syntax :(

ruohola (approved):
```
<div>
    <div>
        <div>
            <div>
                <div>
                    <div>
                        <div>
                            <div>
                                <div>
                                    <div>
                                        <div>
                                            <p>LGTM</p>
                                        </div>
                                    </div>
                                </div>
                            </div>
                        </div>
                    </div>
                </div>
            </div>
        </div>
    </div>
</div>
```

chessbr (approved):

PR #2418, opened 2021-03-18
Vendors can copy products (RM-58)

nvthanh11:
Vendors can copy products (RM-58)


shuup/admin/utils/product.py
```diff
@@ -0,0 +1,63 @@
...
+        else:
+            new_shop_product.suppliers.set(shop_product.suppliers.all())
+            
+        new_shop_product.categories.set(shop_product.categories.all())
+
+        for attribute in product.attributes.all():
+            ProductAttribute.objects.create(product=new_product, attribute=attribute.attribute, value=attribute.value)
+
+        for old_media in product.media.all():
+            media = ProductMedia.objects.create(product=new_product, file=old_media.file, kind=old_media.kind)
```
chessbr, line 55: copy all the other fields as well. maybe you can clone this instance this way:
```py
from shuup.utils.models import copy_model_instance
for media in product.media.all():
    media_copy = copy_model_instance(media)
    media_copy.product = new_product
    media_copy.file = old_media.file
    media_copy.save()
```

I don't think `copy_model_instance` will copy FKs, so I think you need to attach them manually. You also need to go through all the translated fields and copy them, just like in Product.

shuup/admin/utils/product.py
```diff
@@ -0,0 +1,63 @@
...
+from shuup.utils.models import get_data_dict
+
+class ProductCloner():
+    def __init__(self, current_shop, current_supplier):
+        self.current_shop = current_shop
+        self.current_supplier = current_supplier
+    
+    @atomic
+    def clone_product(self, shop_product=None):
+        product = shop_product.product
```
chessbr, line 25: maybe use `shuup.utils.models.copy_model_instance`, then update the SKU like you did and add a "Copy" suffix to the product name `new_product.name = _("{name} - Copy").format(name=product.name)`.
We also should go through all the translated fields and copy them (you need to find a way of getting the translated field dynamically, like [here in `_add_translated_columns`](shuup/admin/modules/settings/view_settings.py)).

shuup/admin/utils/product.py
```diff
@@ -0,0 +1,63 @@
...
+            "sku": "{}-{}".format(product.sku, Product.objects.count())
+        })
+        new_product = Product.objects.create(**product_data)
+        new_product.name = product.name
+        new_product.short_description = product.short_description
+        new_product.description = product.description
+        new_product.slug = product.slug
+        new_product.save()
+
+        new_shop_product = ShopProduct.objects.create(product=new_product, shop=self.current_shop)
```
chessbr, line 37: maybe use `shuup.utils.models.copy_model_instance` and update the product instance before saving. You also need to go through all the translated fields and copy them, just like in Product

shuup/simple_supplier/product_copy_signal_handler.py
```diff
@@ -9,12 +9,15 @@
...
 from shuup.simple_supplier.models import StockCount
-
-
+        
 @receiver(product_copied, dispatch_uid="simple_supplier_product_copied")
-def handle_product_copy(sender, shop, copied, copy, **kwargs):
+def handle_product_copy(sender, shop, copied, copy, supplier=None, **kwargs):
     shop_product = copied.get_shop_instance(shop)
-    for supplier in shop_product.suppliers.all():
+    if supplier == None:
```
chessbr, line 16: do you really need to do this? `shop_product.suppliers.all()` will already have only a single supplier if there is a current supplier

shuup/admin/modules/products/views/copy.py
```diff
@@ -6,58 +6,27 @@
...
-        return HttpResponseRedirect(get_model_url(new_shop_product, shop=self.request.shop))
+        current_supplier = None if request.user.is_superuser else get_supplier(request)
+        cloner = cached_load("SHUUP_ADMIN_PRODUCT_CLONER")(request.shop, current_supplier)
+        copied_shop_product = cloner.clone_product(shop_product=shop_product)
+        messages.success(request, _("Product %s copy successfull.") % copied_shop_product.product)
+        if request.user.is_superuser:
+            return HttpResponseRedirect(get_model_url(copied_shop_product, shop=request.shop))
+        else:
+            return HttpResponseRedirect(reverse("shuup_admin:shuup_multivendor.products_edit", kwargs={"pk":copied_shop_product.id}))
+
```
chessbr, line 32: Fix this
image

shuup/admin/modules/products/views/copy.py
```diff
@@ -6,58 +6,27 @@
...
-                new_product.primary_image = media
-                new_product.save()
-
-        product_copied.send(sender=type(self), shop=shop_product.shop, copied=product, copy=new_product)
-        messages.success(request, _("Product %s copy successfull.") % new_product)
-        return HttpResponseRedirect(get_model_url(new_shop_product, shop=self.request.shop))
+        current_supplier = None if request.user.is_superuser else get_supplier(request)
+        cloner = cached_load("SHUUP_ADMIN_PRODUCT_CLONER")(request.shop, current_supplier)
+        copied_shop_product = cloner.clone_product(shop_product=shop_product)
+        messages.success(request, _("Product %s copy successfull.") % copied_shop_product.product)
```
chessbr, line 27: change the message to `"{product_name} was successfully copied".format(product_name=copied_shop_product.product)`

shuup/admin/modules/products/views/copy.py
```diff
@@ -6,58 +6,27 @@
...
+        if request.user.is_superuser:
```
chessbr, line 28: nope, remove this check. Shuup shouldn't know anything about `shuup_multivendor` addon, it should be agnostic. Instead, implement a `get_next_url()` and customize that in `shuup-multivendor`.

```py
def get_next_url(self):
    if request.GET.get("next"):
        return HttpResponseRedirect(request.GET["next"])
    return HttpResponseRedirect(get_model_url(copied_shop_product, shop=request.shop))
```

chessbr (changes requested):

shuup/admin/utils/product.py
chessbr: @nvthanh11 Sorry, here is the correct link: https://github.com/shuup/shuup/blob/master/shuup/admin/modules/settings/view_settings.py#L160

shuup/admin/utils/product.py
```diff
@@ -0,0 +1,76 @@
...
+    def clone_product(self, shop_product=None):
+        # Clone product
+        product = shop_product.product
+        new_product = copy_model_instance(product)
+        new_product.sku = "{}-{}".format(product.sku, Product.objects.count())
+        new_product.name = ("{name} - Copy").format(name=product.name)
+        new_product.save()
+
+        # Change the name so it will automatically create a translation record
+        for trans in product.translations.all()[1:]:
```
chessbr, line 33: why are you using this `[1:]`? Can't you just use `get_or_create` below instead?

shuup/admin/modules/products/views/copy.py
```diff
@@ -6,58 +6,31 @@
...
-        return HttpResponseRedirect(get_model_url(new_shop_product, shop=self.request.shop))
+        current_supplier = None if request.user.is_superuser else get_supplier(request)
+        cloner = cached_load("SHUUP_ADMIN_PRODUCT_CLONER")(request.shop, current_supplier)
+        copied_shop_product = cloner.clone_product(shop_product=shop_product)
+        messages.success(
+            request,
+            _("{product_name} was successfully copied".format(product_name=copied_shop_product.product))
+        )
+        if request.GET.get("next"):
+            next_url = request.GET.get("next").replace("00", str(copied_shop_product.pk))
```
chessbr, line 33: what is this replace?

chessbr (changes requested):

shuup/admin/modules/products/views/copy.py
nvthanh11: It replaces "00" with new copied product pk to redirect to this new product's detail view, I use `reverse `with "00" as default `pk` to pass data to `next`, then in `CopyView` I replace it by new copied product's pk and redirect.

shuup/admin/utils/product.py
nvthanh11: As I check, when we modified new_product's name in line 29, it will create a new translation records and add it to new_product. So I skip the first one

chessbr (approved):

chessbr (approved):

PR #2427, opened 2021-03-19
Add order status history (rm22)

dzungbe:


shuup/core/migrations/0085_auto_20210317_1802.py
```diff
@@ -0,0 +1,31 @@
+# Generated by Django 2.2.18 on 2021-03-17 11:02
```
chessbr, line 1: rename this migration to human-readable name, like `order_status_history`

shuup/core/models/_orders.py
```diff
@@ -141,7 +143,7 @@ def get_default_complete(self):
 
 
 @python_2_unicode_compatible
-class OrderStatus(TranslatableModel):
+class OrderStatus(TranslatableModel, models.Model):
```
chessbr, line 146: why you need to add this model, it is already being used in `TranslatableModel`?

shuup/core/models/_orders.py
```diff
@@ -182,6 +186,35 @@ def save(self, *args, **kwargs):
...
+
+    next_order_status = models.ForeignKey(
+        "OrderStatus", related_name='next_order_status',
+        blank=True, null=True,
+        on_delete=models.PROTECT,
+        verbose_name=_('next order status')),
+
+    created_on = models.DateTimeField(auto_now_add=True, editable=False, db_index=True, verbose_name=_('created on')),
+
+    description = models.TextField(max_length=200, db_index=True, blank=True, null=True, verbose_name=_('description')),
```
chessbr, line 210: no need to have `db_index` here

shuup/core/models/_orders.py
```diff
@@ -539,8 +596,8 @@ def save(self, *args, **kwargs):
 
         order_changed.send(type(self), order=self)
 
-        if self.status != old_status:
-            order_status_changed.send(type(self), order=self, old_status=old_status, new_status=self.status)
+        # if self.status != old_status:
```
chessbr, line 599: remove the line entirely instead of commenting

shuup/core/models/_orders.py
```diff
@@ -1106,5 +1163,47 @@ def get_available_payment_methods(self):
             if m.is_available_for(self)
         ]
 
+    # check whether user can be able to change status of order
+    def validate_order_status_change(self, next_status: OrderStatus):
```
chessbr, line 1167: no, this method is not right, we shouldn't check this roles here

shuup/core/models/_orders.py
```diff
@@ -1106,5 +1163,47 @@ def get_available_payment_methods(self):
...
+        if curr_role in [OrderStatusRole.COMPLETE, OrderStatusRole.CANCELED] \
+                or (curr_role == OrderStatusRole.PROCESSING and next_status == OrderStatusRole.INITIAL):
+            return False
+
+        return True
+
+    def change_status(self, next_status: OrderStatus, user: User = None, description: str = None):
+        # validate next_status is valid or not
+
+        if not self.validate_order_status_change(next_status):
```
chessbr, line 1189: this method should check whether `next_status` is inside `self.status.allowed_next_statuses.all()`, just like that

shuup/core/models/_orders.py
```diff
@@ -1106,5 +1163,47 @@ def get_available_payment_methods(self):
...
+            raise InvalidOrderStatusError("Error! Can not change status to this status")
```
chessbr, line 1190: make sure to have this error translated: `raise InvalidOrderStatusError(_("Can not change the order to this status."))`

shuup/core/models/_orders.py
```diff
@@ -1106,5 +1163,47 @@ def get_available_payment_methods(self):
...
+            # create a new OrderStatusHistory entry
+            order_history = OrderStatusHistory(order=self, previous_order_status=self.status, next_order_status=next_status,
+                                               description=description,
+                                               creator=user
+                                               )
+            order_history.save()
+
+            # end OrderStatusHistory creatation
+            order_status_changed.send(type(self), order=self, old_status=old_status, new_status=self.status)
+        except Exception as changeStatusError:
```
chessbr, line 1205: you should never do such a try/catch like this, too generic, remove this try/catch block

shuup/core/models/_orders.py
```diff
@@ -1106,5 +1163,47 @@ def get_available_payment_methods(self):
...
+
+        if not self.validate_order_status_change(next_status):
+            raise InvalidOrderStatusError("Error! Can not change status to this status")
+
+        try:
+            # update new status of oder
+            self.status = next_status
+
+            # create a new OrderStatusHistory entry
+            order_history = OrderStatusHistory(order=self, previous_order_status=self.status, next_order_status=next_status,
```
chessbr, line 1197: you can format this a little bit better:

```py
order_history = OrderStatusHistory.objects.create(
    order=self, 
    previous_order_status=self.status,
    next_order_status=next_status,
    description=description,
    creator=user
)
```

chessbr (changes requested):

shuup/admin/modules/orders/toolbar.py
```diff
@@ -72,50 +76,59 @@ def _build_order_set_state_button(self):
             )
 
     def _build_set_complete_button(self):
-        self.append(PostActionButton(
```
chessbr, line 75: you can remove this `_build_set_complete_button` completely. This button will be visible in the menu built in `_build_order_set_state_button` method, right?

shuup/admin/modules/orders/toolbar.py
```diff
@@ -72,50 +76,59 @@ def _build_order_set_state_button(self):
...
-            icon="fa fa-trash",
-            disable_reason=(
-                _("Paid, shipped, or canceled orders cannot be canceled")
-                if not self.order.can_set_canceled()
-                else None
-            ),
-            extra_css_class="btn-danger btn-inverse"
-        ))
+        self.append(
+            PostActionButton(
```
chessbr, line 97: same as above, this status will be available in `_build_order_set_state_button`, right?

shuup/core/models/_orders.py
```diff
@@ -1101,10 +1326,39 @@ def get_available_payment_methods(self):
...
+            m
+            for m in PaymentMethod.objects.available(shop=self.shop, products=product_ids)
             if m.is_available_for(self)
         ]
 
+    def change_status(self, next_status: OrderStatus, user: User = None, description: str = None):
+        # validate next_status is valid or not
+
+        if not self.status.allowed_next_statuses.filter(
+            identifier=next_status.identifier
```
chessbr, line 1338: use `pk` instead of identifier

chessbr (changes requested):
Write unit tests to prove this works. Write a test that you call views to change the status of the order from Received to Complete

PR #2434, opened 2021-03-22
Moved `get_final_lines` from `process` to `form_valid`

wiltso:


shuup/front/checkout/methods.py
```diff
@@ -146,6 +145,7 @@ def form_valid(self, form):
 
         if form.has_changed():
             self.process()
+            self.basket.get_final_lines(with_taxes=True)
```
chessbr, line 148: call this after the save

chessbr (approved):

PR #2435, opened 2021-03-22
New code style

chessbr:
Update flake8, isort and add black to format code


ruohola (approved):
LGTM, squash all the formatting commits into one and you could add a `.git-blame-ignore-revs` file in the next commit which contains the hash of the formatting commit with a little explanation. It can then be used with `git config blame.ignoreRevsFile .git-blame-ignore-revs`

https://www.moxio.com/blog/43/ignoring-bulk-change-commits-with-git-blame

ruohola (approved):

PR #2445, opened 2021-03-29
Add option to export customers RM-56

nvthanh11:


chessbr:
Did in https://github.com/shuup/shuup/pull/2442

PR #2449, opened 2021-03-29
Core: allow refunds creation for completed orders

chessbr:
From the business perspective, even after the order was theorically finished,
the customer can always ask for a refund later. It is the shop's refund policy
that dictates whether when refunds can be created or not, not Shuup

Refs FD-196


ranta (approved):
I don't see a reason this shouldn't be allowed 👍

PR #2453, opened 2021-03-31
Admin: Don't require typing characters to show options in Select2 fields

ruohola:
Previously these would require to input 3 characters before showing any
options. It made the choices they contain very undiscoverable.

It's hard to say how much this will affect performance. In theory it
shouldn't hurt it much since these choices will anyways be loaded
asynchronously only after the page has already been rendered.

Before:
image

After:
image

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -11,6 +11,7 @@ List all changes after the last release here (newer on top). Each change on a se
 ### Changed
 
 - Admin: change so TaxClassEditView is a FormPartView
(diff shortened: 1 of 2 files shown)


ranta (approved):
Very welcome change, thank you! 🚀

ruohola:
@chessbr Is always the smartest https://github.com/shuup/shuup/commit/92274ad7e7799cf847684d5d9a29c37887e98a77 after all :)

PR #2455, opened 2021-04-01
Vendors can see hidden product categories in product admin -RM11

nvthanh11:


chessbr:
@nvthanh11 rebase your commits into a single one please

PR #2467, opened 2021-04-16
Attribute related changes

n1ro0:
### Added
- Admin: New type of attribute: CHOICES
- Admin: Selection of multiple choices from attribute for product
- Front: Product filter by attribute choices


shuup/core/migrations/0086_auto_20210415_0847.py
```diff
@@ -0,0 +1,55 @@
+# Generated by Django 2.2.18 on 2021-04-15 08:47
```
chessbr, line 1: Squash all migrations files into a single one and put a name that makes sense

shuup/core/models/_attributes.py
```diff
@@ -119,6 +124,15 @@ class Attribute(TranslatableModel):
...
         ),
     )
+    ordering = models.IntegerField(
+        default=0,
+        help_text=_("The ordering in which your attribute will be displayed."),
+    )
+    max_choices = models.PositiveIntegerField(
+        default=1,
+        verbose_name=_("max choices"),
+        help_text=_("Amount of choices that user can choose from existing options."),
```
chessbr, line 134: add a text saying this will be only used when the type of attribute is choices. maybe we can also create a min_choices which we can use to force users to select a certain number of options?

chessbr (changes requested):
### 1. Add the max_choices field in the base form part
image

You can show/hide it according to the user selection. Add some JS to the template to do it.

### 2. Use a different icon (maybe `fa-list-ol`) for the Choice Options and rename it to Choices
image

### 3. Add some buttons to add and remove forms this way, better user experience. 
image

shuup/utils/fields.py
```diff
@@ -29,3 +29,15 @@ def to_python(self, value):
...
+
+class TypedMultipleChoiceWithLimitField(forms.TypedMultipleChoiceField):
+    def __init__(self, limit=None, **kwargs):
+        self.limit = limit
+        super().__init__(**kwargs)
+
+    def clean(self, value):
+        value = super().clean(value)
+        if self.limit is not None and len(value) > self.limit:
+            raise forms.ValidationError(f"You can't select more than {self.limit} items.")
```
chessbr, line 42: `item(s)`

shuup/front/forms/product_list_modifiers.py
```diff
@@ -576,6 +591,118 @@ def get_admin_fields(self):
...
+
+        return queryset
+
+    def get_admin_fields(self):
+        active, ordering = super(AttributeProductListFilter, self).get_admin_fields()
+        active[1].help_text = _("Allow products to be filtered according to their attributes.")
+
+        attributes = Select2MultipleField(
+            model=Attribute,
+            label=_("Attributes That Can Be Filtered"),
```
chessbr, line 690: `Attributes that can be filtered`

chessbr (changes requested):

shuup/utils/fields.py
```diff
@@ -29,3 +29,20 @@ def to_python(self, value):
...
+    def __init__(self, min_limit=None, max_limit=None, **kwargs):
+        self.min_limit = min_limit
+        self.max_limit = max_limit
+        super().__init__(**kwargs)
+
+    def clean(self, value):
+        value = super().clean(value)
+        if self.min_limit is not None and len(value) < self.min_limit:
+            ending = "" if self.min_limit % 10 == 1 else "s"
+            raise forms.ValidationError(f"You can't select less than {self.min_limit} item{ending}.")
```
chessbr, line 44: maybe use pluralization https://docs.djangoproject.com/en/3.2/topics/i18n/translation/#pluralization

chessbr (approved):

PR #2480, opened 2021-04-27
Fix mass editing validation for ManyToManyField fields.

khchine5-shuup:


shuup/admin/modules/products/views/mass_edit.py
```diff
@@ -54,9 +55,19 @@ def form_valid(self, form):
                 if not v:
                     continue
                 if hasattr(product, k):
-                    setattr(product, k, v)
+                    if isinstance(
+                        product._meta.get_field(k), models.ManyToManyField
+                    ):
+                        setattr(product, k + "_set", v)
```
chessbr, line 61: alright, 2 things:
- Django 2.x doesn't allow us using `product.attributes = [attr1, attr2]` in M2M field for example, you MUST use the `.set()` method instead
- You need to get the real `related_name` instead if relying on `_set`, I am not sure that will always be present.

khchine5-shuup: We use `.set()` instead of of `_set` which will support Django 2/3

khchine5-shuup: Updated it and the tests.

chessbr (approved):

PR #2497, opened 2021-05-19
Admin: show help texts in permission

chessbr:
Add a method in AdminModule that allows returning help texts for permissions

This will allow the developer to define the help texts for the permissions


ranta (approved):
Love it ❤️

PR #2502, opened 2021-05-25
Core: change how price and orderability are cached

chessbr:
Cache product orderability and prices using a list of user groups instead of per contact


ranta (approved):
Good stuff 🤖

PR #2509, opened 2021-06-01
Enable the importer to run asynchronously

chessbr:
Refs RM-4


shuup/core/models/_background_tasks.py
```diff
@@ -0,0 +1,73 @@
...
+
+    class Meta:
+        verbose_name = _("background task")
+        verbose_name_plural = _("background tasks")
+
+
+class BackgroundTaskExecutionStatus(Enum):
+    RUNNING = 0
+    SUCCESS = 1
+    ERROR = 2
```
ranta, line 48: Would tasks always be running, or will they also be waiting in queue for other tasks to finish first? In that case we would need a fourth status also

shuup/core/tasks.py
```diff
@@ -38,22 +69,25 @@ def __init__(self, function, **kwargs):
...
+    def create_task(self, function, stored=False, queue="default", **kwargs) -> Task:
         """
         Create a task to run.
 
         :type function: str
         """
         raise NotImplementedError()
 
-    def run_task(self, task):
+    def run_task(self, task) -> Optional[TaskResult]:
```
ranta, line 90: Nice typing of methods ❤️

shuup/core/models/_background_tasks.py
chessbr: they will start immediately. the TaskRunner is the one that is responsible for this. But I don't think we need a new state for this.

ranta (approved):
This is some pretty good stuff 👍

PR #2510, opened 2021-06-03
Fixed get_supported_product_kinds_for_module()

wiltso:
Fixed
- Core: Only product kind modules that have specified supplier modules will be returned from get_supported_product_kinds_for_module
- Admin: Display the correct error message when adding stock to a product and not only "on text"

refs https://trello.com/c/NiJsBtIW/24-moved-to-sprint-3-original-sprint-2-custom-product-configuration-vendor-to-sell-both-services-and-products

### DIFF
--- a/shuup/core/models/_suppliers.py
+++ b/shuup/core/models/_suppliers.py
@@ -194,7 +194,7 @@ class Supplier(ModuleInterface, TranslatableShuupModel):
     def get_stock_status(self, product_id, *args, **kwargs):
         for module in self.modules:
             stock_status = module.get_stock_status(product_id, *args, **kwargs)
-            if stock_status.handled:
+            if stock_status and stock_status.handled:
                 return stock_status
 
     def get_suppliable_products(self, shop, customer):
--- a/shuup/simple_supplier/admin_module/views.py
+++ b/shuup/simple_supplier/admin_module/views.py
@@ -12,6 +12,7 @@ from django.utils.translation import ugettext, ugettext_lazy as _
 
 from shuup.admin.utils.picotable import ChoicesFilter, Column, TextFilter
 from shuup.admin.utils.views import PicotableListView
+from shuup.core.excs import MissingSupplierModuleException
 from shuup.core.models import Product, Supplier
 from shuup.core.utils import context_cache
 from shuup.simple_supplier.forms import AlertLimitForm, StockAdjustmentForm
@@ -138,6 +139,13 @@ def _process_stock_adjustment(form, request, supplier_id, product_id):
     stock_adjustment = supplier.adjust_stock(
         product_id, delta=data.get("delta"), purchase_price=data.get("purchase_price"), created_by=request.user
     )
+
+    # If the the simple supplier module is disabled for a suppler for some reason.
+    # This will provide a usefull error message to the user that is trying to update the stock.
+    if not stock_adjustment:
+        raise MissingSupplierModuleException(
+            _("Please be in contact with the marketplace owner to get permission to update your product stock")
+        )
     success_message = _get_success_message(
         request, stock_adjustment.supplier, stock_adjustment.product, get_adjustment_success_message(stock_adjustment)
     )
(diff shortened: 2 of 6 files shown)


shuup/core/models/_suppliers.py
```diff
@@ -220,6 +221,10 @@ def adjust_stock(self, product_id, delta, created_by=None, type=None, *args, **k
             if stock:
                 return stock
 
+        raise MissingSupplierModuleException(
```
chessbr, line 224: no, it should just do nothing instead of raising

shuup/core/specs/product_kind.py
```diff
@@ -65,3 +65,4 @@ class DefaultProductKindSpec(ProductKindSpec):
     value = 1
     name = _("Product")
     admin_listing_name = "products"
+    supported_supplier_modules = ["simple_supplier"]
```
chessbr, line 68: wrong, the core shouldn't know ANYTHING about other addons

shuup/core/suppliers/base.py
```diff
@@ -28,7 +28,7 @@ def get_supported_product_kinds_for_module(module_identifier: str) -> Iterable[P
     specs = []
     for product_kind_spec in get_provide_objects("product_kind_specs"):
         supported_modules = product_kind_spec.supported_supplier_modules
-        if not supported_modules or module_identifier in supported_modules:
+        if supported_modules and module_identifier in supported_modules:
```
chessbr, line 31: wrong, if `supported_modules` is blank it should allow any modules.

shuup/simple_supplier/templates/shuup/simple_supplier/admin/macros.jinja
```diff
@@ -18,9 +18,13 @@
                     });
                 },
                 error: function(response) {
+                    var message = gettext("Please be in contact with the marketplace owner to get this action to work");
```
chessbr, line 21: this makes no sense. this is the simple supplier template. this should only be visible when the module is simple supplier

chessbr (changes requested):

shuup/simple_supplier/templates/shuup/simple_supplier/admin/macros.jinja
wiltso: Yea exactly this is only rendered if you are having a product that uses the simple supplier. Also it will on be rendered as the default text when there is an error with adjusting the stock for a product.

shuup/core/suppliers/base.py
```diff
@@ -26,10 +26,15 @@
 @lru_cache()
 def get_supported_product_kinds_for_module(module_identifier: str) -> Iterable[ProductKindSpec]:
     specs = []
+    default_specs = []
```
chessbr, line 29: this also doesn't make any sense

shuup/simple_supplier/templates/shuup/simple_supplier/admin/macros.jinja
chessbr: This error message doesn't tell anything relevant. How would the marketplace owner make this action work?

chessbr (changes requested):

shuup/simple_supplier/admin_module/views.py
```diff
@@ -138,6 +139,10 @@ def _process_stock_adjustment(form, request, supplier_id, product_id):
     stock_adjustment = supplier.adjust_stock(
         product_id, delta=data.get("delta"), purchase_price=data.get("purchase_price"), created_by=request.user
     )
+    if not stock_adjustment:
+        raise MissingSupplierModuleException(
+            _("Please be in contact with the marketplace owner to get permission to update your product stock")
```
chessbr, line 144: makes no sense to raise this here. this view shouldn't even be used if the product doesn't support the simple_supplier module

shuup/core/suppliers/base.py
wiltso: Why not?
First we will return the product specs with and exact match. Whit this I mean a supplier module where the identifier is specified in the `supported_supplier_modules` array in the product spec.
Then if we don't find any exact matches we will return all the product specs that don't have any `supported_supplier_modules`

PR #2521, opened 2021-06-16
Fix MultiLanguageModelForm

wiltso:
- Utils: fix MultiLanguageModelForm so language dependent filed will only be required if the language is required

### DIFF
--- a/shuup/utils/multilanguage_model_form.py
+++ b/shuup/utils/multilanguage_model_form.py
@@ -75,9 +75,9 @@ class MultiLanguageModelForm(TranslatableModelForm):
             for lang in self.languages:
                 language_field = copy.deepcopy(base)
                 language_field_name = "%s__%s" % (f.name, lang)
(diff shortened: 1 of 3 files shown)


shuup/utils/multilanguage_model_form.py
```diff
@@ -75,7 +75,7 @@ def __init__(self, **kwargs):  # noqa (C901)
             for lang in self.languages:
                 language_field = copy.deepcopy(base)
                 language_field_name = "%s__%s" % (f.name, lang)
-                if language_field.required:
+                if language_field.required and lang in self.required_languages:
```
chessbr, line 78: check the line 80, it has the same logic you are just adding. shouldn't you just move it to the line before the `if` ?

chessbr: actually, where is this `required_translated_field_names` even used?

chessbr (approved):

PR #2529, opened 2021-06-30
Add csv report writer

wiltso:
- Reports: add csv report writer

### DIFF
--- a/shuup/reports/writer.py
+++ b/shuup/reports/writer.py
@@ -5,10 +5,10 @@
 #
 # This source code is licensed under the OSL-3.0 license found in the
 # LICENSE file in the root directory of this source tree.
-from __future__ import unicode_literals
-
+import csv
 import six
 from babel.dates import format_datetime
+from datetime import datetime
 from decimal import Decimal
 from django.core.serializers.json import DjangoJSONEncoder
 from django.db import models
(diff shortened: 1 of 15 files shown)


shuup/reports/writer.py
```diff
@@ -122,6 +124,51 @@ def get_filename(self, report):
...
+    writer_type = "csv"
+
+    def __init__(self):
+        super(CSVReportWriter, self).__init__()
+        self.data = []
+
+    def _w(self, content):
+        if content is not None:
+            if isinstance(content, TaxlessPrice) or isinstance(content, TaxfulPrice):
+                content = floatformat(content.amount.value, 2)
```
chessbr, line 139: you can't do this. Each currency has it's own decimal precision. you should:
```py
float(content.amount.as_rounded().value)
```
also fix this in the `ExcelReportWriter`

shuup/reports/writer.py
```diff
@@ -122,6 +124,51 @@ def get_filename(self, report):
...
+
+            if isinstance(content, Decimal):
+                content = floatformat(content, 2)
```
chessbr, line 142: idk, looks wrong to set a fixed number of decimal places here without knowing what it is about. what if:
```py
>>> from decimal import Decimal
>>> value = 32.24738
>>> Decimal(str(value)).normalize().as_tuple().exponent
-5
```
then you can get this exponent and use in the floatformat, use some sort of max() so you prevent using a huge precision, set max as 10?

chessbr (changes requested):

chessbr (approved):

PR #2530, opened 2021-07-01
Front: fix so orders that are canceled can't be payed for

wiltso:
refs https://trello.com/c/9ZvYu2Cy/161-not-allow-users-to-make-a-payment-after-it-has-been-cancelled


shuup/front/views/payment.py
```diff
@@ -46,6 +48,10 @@ def get_context_data(self, **kwargs):
     def dispatch(self, request, *args, **kwargs):
         mode = self.kwargs["mode"]
         order = self.object = self.get_object()
+        if order.is_canceled():
```
chessbr, line 51: put this inside the `mode == "payment"` and create a unit test for this

PR #2534, opened 2021-07-06
GDPR: introduce snippet blocker feature

chessbr:
GDPR cookies can now block an xtheme snippet injection

Refs CS-175

image


ranta (approved):
🚀

PR #2542, opened 2021-07-14
Add product validator and shows a warning for not valid products.

eberonsh:


shuup/admin/modules/products/issues.py
```diff
@@ -0,0 +1,46 @@
...
+#
+# This source code is licensed under the OSL-3.0 license found in the
+# LICENSE file in the root directory of this source tree.
+class ProductValidationIssue:
+    """
+    Class that represents an issue after a validation of a class AdminProductValidation is
+    completed.
+    """
+
+    def __init__(self, code: str, message: str, issue_type: str, is_html: bool = False):
```
chessbr, line 14: let's change this a little:
```py
def __init__(self, message: str, issue_type: str, code: Optional[str] = None, is_html: bool = False):
```

shuup/admin/modules/products/issues.py
```diff
@@ -0,0 +1,46 @@
...
+        :type message: str
+        :param issue_type: The issue's type: can be `info`, `warning` or `error`.
+        :type issue_type: str
+        :param is_html: True if the message contains html code.
+        :type is_html: bool
+        """
+
+        self.code = code
+        self.message = message
+        self.issue_type = issue_type
```
chessbr, line 28: raise a ValueError if issue_type is not one of the options info, warning or error

shuup/admin/modules/products/issues.py
```diff
@@ -0,0 +1,46 @@
...
+        self.is_html = is_html
+
+    def get_issue_type_order(self):
+        if self.issue_type == "error":
+            return 1
+        if self.issue_type == "warning":
+            return 2
+        if self.issue_type == "info":
+            return 3
+        return 1000000
```
chessbr, line 38: shouldn't enter here ever. also rename this method to `get_issue_type_priority()`.

shuup/admin/modules/products/views/edit.py
```diff
@@ -230,9 +231,9 @@ def get_context_data(self, **kwargs):
         context = super(ProductEditView, self).get_context_data(**kwargs)
         orderability_errors = []
 
+        shop = self.request.shop
```
chessbr, line 234: I know you just moved this, but let's fix this, use this instead:
```py
shop = get_shop(self.request)
```

shuup/admin/modules/products/views/edit.py
```diff
@@ -258,4 +259,15 @@ def get_context_data(self, **kwargs):
                     self.object.product, self.request
                 )
 
+        product_validator_provides = sorted(get_provide_objects("admin_product_validator"), key=lambda x: x.order)
```
chessbr, line 262: I think the validators don't need to have an order 🤔

chessbr (changes requested):
Write a unit test that will render at least one validation errors in admin. We don't want to decrease the coverage that much with this new feature.

Also add a new entry into the CHANGELOG.md file with your change, follow the pattern we have there.

shuup/admin/modules/products/views/edit.py
eberonsh: The reason I added an order(priority) is in case there are more ProductValidators and one of them is generating an issue of type error. As all the issues share the same alert notification, sorting by issue type allows to show the alert with the most relevant level.

shuup/admin/modules/products/issues.py
eberonsh: Done

shuup/admin/modules/products/issues.py
eberonsh: Done

shuup/admin/modules/products/issues.py
eberonsh: Done

shuup/admin/modules/products/views/edit.py
eberonsh: Done

PR #2557, opened 2021-07-16
Core: introduce new catalog API

chessbr:
Refs https://trello.com/c/d7V2OrOf/18-priority-1-fix-the-product-pricing-module-logic-for-faster-catalog-fetching


shuup/discounts/models/_discounts.py
```diff
@@ -63,23 +60,19 @@ def available(self, shop=None):
             & (
                 Q(happy_hours__time_ranges__isnull=True)
                 | Q(
+                    # TODO: check whether there is a bug here when a time
+                    # advances a day in a UTC day, e.g: 11pm - 01am
```
heast, line 64: I think this might be okay as long as everything is tz-aware:
```>>> import pytz
>>> from django.utils.timezone import now, localtime
>>> # UTC first, to make it earlier
>>> utc = now()
>>> utc.hour 
16
>>> # Eastern Daylight Time is later, though the hour is a smaller number
>>> edt = localtime(timezone=pytz.timezone("America/New_York"))
>>> edt.hour 
12
>>> utc < edt
True
```

heast, line 64: I think there is a bug, actually, though I didn't think so at first:
```
>>> # UTC first, so as to be the earliest of the two
>>> utc_now = now()
>>> # Now EET, which is ahead of UTC
>>> eet_now = localtime(timezone=pytz.timezone("Europe/Helsinki"))
>>> happy_hour_start = utc + timedelta(hours=3)
>>> # HappyHour discount is still in the future
>>> happy_hour_start < eet_now
False
>>> eet_now += timedelta(hours=3)
>>> # HappyHour discount has started
>>> happy_hour_start < eet_now
True
>>> # But...
>>> happy_hour_start.time() < eet_now.time()
False
>>> happy_hour_start.time()
datetime.time(22, 11, 55, 802365)
>>> eet_now.time()
datetime.time(1, 12, 18, 652294)
```

heast (approved):

PR #2559, opened 2021-07-16
Improve GDPR anonymization so we can have better tool to complete the…

eberonsh:
… anonymization.


shuup/gdpr/tasks.py
```diff
@@ -0,0 +1,32 @@
...
+
+def anonymize(shop_id: int, contact_id: int = None, user_id: int = None):
+
+    shop = Shop.objects.filter(
+        pk=shop_id,
+    ).first()
+
+    contact = None
+    if contact_id:
+        contact = PersonContact.objects.get(pk=contact_id)
```
chessbr, line 22: `get()` won't return None if it doesn't exist, it will raise an exception.
so you can do it like this instead:
```py
contact = PersonContact.objects.filter(pk=contact_id).first()
if not contact:
    contact = CompanyContact.objects.filter(pk=contact_id).first()
```

shuup/gdpr/tasks.py
```diff
@@ -0,0 +1,32 @@
...
+        if not contact:
+            contact = CompanyContact.objects.get(pk=contact_id)
+
+    user = None
+    if user_id:
+        user = get_user_model().objects.get(pk=user_id)
```
chessbr, line 28: same here, use `.filter(pk=user_id).first()` instead

chessbr (changes requested):

chessbr (approved):

chessbr (approved):

PR #2560, opened 2021-07-20
Fix basket quantity for parent package products

wiltso:
- Core: Fix basket quantity for parent package products when caching lines


shuup/core/basket/objects.py
```diff
@@ -525,7 +525,7 @@ def _cache_lines(self):  # noqa (C901)
                                 break
                         if orderable:
                             orderable_lines.append(line)
-                            orderable_counter[product.id] += quantity
+                            orderable_counter[product.id] = quantity
```
chessbr, line 528: what if the package contain the same product more than once?

wiltso: But this is the parent

chessbr (approved):

PR #2564, opened 2021-07-23
Preserve comment during checkout phases

hyptocrypto:


hyptocrypto:
This helps. But if the page refreshes, the comment is cleared.

shuup/front/templates/shuup/front/checkout/confirm.jinja
```diff
@@ -18,6 +18,13 @@
 {% block extrajs %}
     {{ super() }}
     <script>
+        window.addEventListener("unload", function() {
+            window.localStorage.removeItem("confirm-comment");
+        });
+        $("#id_comment").on("change keydown paste input", function(){
+            window.localStorage.setItem("confirm-comment", $("#id_comment").val());
```
chessbr, line 25: can you use the basket.key as pasrt of this storage key?

PR #2570, opened 2021-07-30
Do not let users to set value with decimals while adjusting stock qua…

eberonsh:
[Do not let users to set value with decimals while adjusting stock quantity if the sales unit doesn't permit](https://trello.com/c/fpFHrqIl/62-do-not-let-users-to-set-value-with-decimals-while-adjusting-stock-quantity-if-the-sales-unit-doesnt-permit)


shuup/simple_supplier/forms.py
```diff
@@ -27,6 +27,7 @@ class StockAdjustmentForm(forms.Form):
 
     def __init__(self, *args, **kwargs):
```
chessbr, line 28: change this __init__ to receive an optional `sales_unit` instance

shuup/simple_supplier/forms.py
```diff
@@ -37,7 +38,9 @@ def clean_delta(self):
         delta = self.cleaned_data.get("delta")
         if delta == 0:
             raise ValidationError(_("Only non-zero values can be added to stock."), code="stock_delta_zero")
-
+        # rounds towards zero
+        rounding_margin = 0.4999  # * pow(0.1, self.sales_unit_decimals)
+        delta = round(float(delta) - rounding_margin, self.sales_unit_decimals)
```
chessbr, line 43: you can then use Decimal to do the precision calculation:
```py
precision = Decimal("0.1") ** self.sales_unit.decimals
return Decimal(delta).quantize(precision)
```

chessbr (changes requested):

shuup/simple_supplier/forms.py
```diff
@@ -27,6 +28,7 @@ class StockAdjustmentForm(forms.Form):
 
     def __init__(self, *args, **kwargs):
         super(StockAdjustmentForm, self).__init__(*args, **kwargs)
+        self.sales_unit = kwargs.get("sales_unit")
```
chessbr, line 31: use:
```py
def __init__(self, sales_unit: Optional[SalesUnit] = None, *args, **kwargs):
    self.sales_unit = sales_unit
    super().__init__(*args, **kwargs)
```

shuup/simple_supplier/forms.py
```diff
@@ -23,22 +25,34 @@ class StockAdjustmentForm(forms.Form):
             currency_name=get_currency_name(settings.SHUUP_HOME_CURRENCY),
         )
     )
-    delta = forms.DecimalField(label=_("Quantity"))
+    delta = DecimalPlaceField(label=_("Quantity"))
+    decimal_places = forms.IntegerField(widget=forms.HiddenInput())
 
     def __init__(self, *args, **kwargs):
+        self.decimals = 0
+        if len(args) > 0 and "decimal_places" in args[0]:
```
eberonsh, line 33: Read decimal_places value from POST data.

shuup/simple_supplier/forms.py
```diff
@@ -23,22 +25,34 @@ class StockAdjustmentForm(forms.Form):
...
+            self.decimals = int(args[0]["decimal_places"])
         super(StockAdjustmentForm, self).__init__(*args, **kwargs)
         if not ShuupSettings.get_setting("SHUUP_ENABLE_MULTIPLE_SHOPS"):
             self.fields["purchase_price"].label = format_lazy(
                 _("Purchase price per unit ({currency_name})"),
                 currency_name=get_currency_name(Shop.objects.first().currency),
             )
+        if self.decimals == 0 and "decimal_places" in self.initial:
```
eberonsh, line 41: Read decimals from initial data.

shuup/simple_supplier/forms.py
```diff
@@ -23,22 +25,34 @@ class StockAdjustmentForm(forms.Form):
...
+            self.decimals = self.initial["decimal_places"]
+        self.fields["delta"].decimal_places = self.decimals
```
eberonsh, line 43: Format widget dynamically considering the value passed for decimal places.

shuup/simple_supplier/utils.py
```diff
@@ -147,7 +147,13 @@ def get_stock_adjustment_div(request, supplier, product):
         "product": product,
         "supplier": supplier,
         "delta_step": pow(0.1, product.sales_unit.decimals) if product.sales_unit.decimals else 0,
-        "adjustment_form": StockAdjustmentForm(initial={"purchase_price": purchase_price, "delta": None}),
+        "adjustment_form": StockAdjustmentForm(
+            initial={
+                "purchase_price": purchase_price,
+                "delta": None,
+                "decimal_places": product.sales_unit.decimals,
```
eberonsh, line 154: Passing decimals instead of sales_unit as we need to recover it from the POST data also in a hidden field.

shuup/simple_supplier/forms.py
```diff
@@ -23,22 +25,34 @@ class StockAdjustmentForm(forms.Form):
             currency_name=get_currency_name(settings.SHUUP_HOME_CURRENCY),
         )
     )
-    delta = forms.DecimalField(label=_("Quantity"))
+    delta = DecimalPlaceField(label=_("Quantity"))
+    decimal_places = forms.IntegerField(widget=forms.HiddenInput())
```
eberonsh, line 29: Added a hidden field decimal_places to recover the decimal places value for the validation from the POST data.

CHANGELOG.md
```diff
@@ -16,6 +16,7 @@ List all changes after the last release here (newer on top). Each change on a se
 
 ### Fixed
 
+- Do not let users to set value with decimals while adjusting stock quantity if the sales unit doesn't permit.
```
chessbr, line 19: att this to the Unreleased version

shuup/admin/forms/fields.py
```diff
@@ -37,6 +37,19 @@ def widget_attrs(self, widget):
...
 
+class DecimalPlaceField(DecimalField):
+    def widget_attrs(self, widget):
+        attrs = super().widget_attrs(widget)
+        if self.min_value is not None:
+            attrs["min"] = self.min_value
+        if self.max_value is not None:
+            attrs["max"] = self.max_value
+        if self.decimal_places == 0:
+            attrs["onkeydown"] = "if(event.key==='.'){event.preventDefault();}"
```
chessbr, line 48: this won't work for different locations, for example here in Brazil the decimal separator is `,`. Can't you control this through the input `step='0.01'` property instead?

shuup/simple_supplier/forms.py
chessbr: not good. IMO we should configure the field according to the decimal places passed through __init__. then you can create the field accordingly:

```py
class StockAdjustmentForm(forms.Form):
    delta = forms.DecimalField(label=_("Quantity")) # keep as it is

    def __init__(self, sales_unit: Optional[SalesUnit] = None, *args, **kwargs):
       super().__init__(*args, **kwargs)
       if sales_unit:
       self.fields["delta"].decimal_places = sales_unit.decimal_places
```

could this work? please check https://docs.djangoproject.com/en/2.2/ref/forms/fields/#django.forms.DecimalField.decimal_places

chessbr (changes requested):

shuup/simple_supplier/forms.py
eberonsh: `sales_unit` is available when we render the form but not when the process it after the user posts the response. We could do it if we do a specific query for retrieving the sales_unit of the product.

chessbr (approved):

PR #2571, opened 2021-08-03
Adds font object selector.

eberonsh:
Adds font object selector form Admin Theme form.


chessbr (changes requested):
add the new permission to XthemeFontsAdminModule module as well

chessbr (approved):

PR #2572, opened 2021-08-03
Refactored object selectors, corrected permission names.

eberonsh:


shuup/admin/modules/permission_groups/__init__.py
```diff
@@ -40,7 +40,7 @@ def get_model_url(self, object, kind, shop=None):
         return derive_model_url(PermissionGroup, "shuup_admin:permission_group", object, kind)
 
     def get_extra_permissions(self) -> Iterable[str]:
-        return ["permission_group.object_selector"]
+        return ["group.object_selector"]
```
chessbr, line 43: is this correct?

chessbr (changes requested):
I think you should create an utils that will return the permission for a model this way we can have consistent results across the platform

shuup/admin/modules/permission_groups/__init__.py
eberonsh: Yes, PermissionGroup comes from: 
`from django.contrib.auth.models import Group as PermissionGroup`

chessbr: ah, makes sense

chessbr (approved):

shuup/admin/modules/services/__init__.py
```diff
@@ -55,12 +56,12 @@ def get_model_url(self, object, kind, shop=None):
         return derive_model_url(self.model, self.url_name_prefix, object, kind)
 
     def get_extra_permissions(self) -> Iterable[str]:
-        return ["carrier.object_selector", "paymentmethod.object_selector"]
+        return [get_object_selector_permission_name(Carrier), get_object_selector_permission_name(PaymentMethod)]
```
eberonsh, line 59: It seems I missed the permission for ShippingMethod.

chessbr (approved):

PR #2578, opened 2021-08-05
Move API keys fields to use django-mirage-field module.

eberonsh:


shuup/core/management/commands/migrator.py
```diff
@@ -0,0 +1,113 @@
+from django.apps import apps as installed_apps
+from django.db import connections, router
+from mirage.crypto import Crypto
+from tqdm import tqdm
+
+
+class MirageMigrator:
```
eberonsh, line 7: We need a Migrator class capable of setting a id field as a parameter for handling the child classes of ServiceProvider (which doesn't have an id field). The mirage Migrator class doesn't allow that.

shuup/core/management/commands/shuup_migrate_encrypted_field.py
```diff
@@ -0,0 +1,57 @@
...
+# Copyright (c) 2012-2021, Shuup Commerce Inc. All rights reserved.
+#
+# This source code is licensed under the OSL-3.0 license found in the
+# LICENSE file in the root directory of this source tree.
+from django.core.management.base import BaseCommand
+
+from .migrator import MirageMigrator as Migrator
+
+
+class Command(BaseCommand):
```
eberonsh, line 12: Same here, it handles the parameter idfield and uses MirageMigrator instead of mirage.migration.Migrator.

chessbr (changes requested):
I don't think we need a Migrator. There is no field here to be migrated.
You should do a data migration to migrate the fields to the encrypted field on every addon (like this: https://github.com/luojilab/django-mirage-field#1-migrations), not here.

In Shuup Core, you just need to have a new API:

```py
encrypted_configuration.set(shop: Optional[Shop], key, value)
encrypted_configuration.get(shop: Optional[Shop], key)
```
and your new model.

On every addon that have encrypted data inside a configuration key, you need to create a data migration like this:

```py
shop_config_keys = [
    "my_secret_1",
    "my_api_key",
    "another_api_secret"
]
global_config_keys = [
    "my_secret_1",
    "my_api_key",
    "another_api_secret"
]

def encrypt(apps, schema_editor):
    for key in global_config_keys:
        not_encrypted = configuration.get(None, key)
        encrypted_configuration.set(None, key, not_encrypted)

    for shop in Shop.objects.all():
        for key in shop_config_keys:
            not_encrypted = configuration.get(shop, key)
            if not_encrypted is not None:
                encrypted_configuration.set(shop, key, not_encrypted)
    

def decrypt(apps, schema_editor):
    for key in global_config_keys:
        encrypted = encrypted_configuration.get(None, key)
        configuration.set(None, key, not_encrypted)

    for shop in Shop.objects.all():
        for key in shop_config_keys:
            encrypted = encrypted_configuration.get(shop, key)
            if encrypted is not None:
                configuration.set(shop, key, encrypted)
```

chessbr:
@eberonsh If https://github.com/luojilab/django-mirage-field/pull/17 merge and release soon, we don't need to keep this custom migrator here.

shuup/configuration.py
```diff
@@ -118,6 +121,8 @@ def _get_configuration_from_db(shop):
     configuration = {}
     for conf_item in ConfigurationItem.objects.filter(shop=shop):
         configuration[conf_item.key] = conf_item.value
+    for conf_item in EncryptedConfigurationItem.objects.filter(shop=shop):
+        configuration[conf_item.key] = conf_item.value
```
chessbr, line 125: This will override the configurations with the same key that are not encrypted. Think you can add a prefix for encrypted items: `f"encrypted:{key}"`.
In the get() method also pass the `encrypted=False` flag so we can chose whether to get encrypted items or not.

eberonsh: But the point is that only the encrypted value remains, right? So why do we keep the non encrypted value?

chessbr (approved):

PR #2579, opened 2021-08-05
Core: suppliers must have supplier modules to ship

bill-shuup:
Prevent suppliers with no supply modules from creating shipments.

Reference: https://trello.com/c/YdNmeTJ0/99-fix-bug-that-makes-it-not-possible-to-create-shipments-in-shuup-os

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -8,18 +8,16 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
 
 List all changes after the last release here (newer on top). Each change on a separate bullet point line
 
+### Changed
+
+- BREAKING: Suppliers with no supplier modules cannot create shipments
+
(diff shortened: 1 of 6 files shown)


shuup/core/models/_orders.py
```diff
@@ -863,6 +864,11 @@ def create_shipment(self, product_quantities, supplier=None, shipment=None):
         if not supplier:
             supplier = shipment.supplier
 
+        if not supplier.modules:
```
chessbr, line 867: you already have the exception inside `supplier.ship_products`, no need to have it here.

chessbr (changes requested):

shuup/core/models/_orders.py
bill-shuup: Fixed.

chessbr (approved):

PR #2580, opened 2021-08-06
Use the PhoneNumberField for phone numbers

bill-shuup:
Reference: https://trello.com/c/wxT4FDxV/101-add-support-for-phonefield


chessbr (changes requested):
This field must be used across the platform, in all places in front and admin. I also think this deserves a special field widget to help users to select the country on the input field.

PR #2581, opened 2021-08-09
Move settings to db

eberonsh:


shuup/admin/modules/settings/forms/system.py
```diff
@@ -72,8 +73,40 @@ class OrderSettingsForm(BaseSettingsForm):
...
 
 
 class OrderSettingsFormPart(BaseSettingsFormPart):
     form = OrderSettingsForm
     name = "order_settings"
+
+
+class CoreSettingsForm(BaseSettingsForm):
+    title = _("Core Settings")
+    home_currency = forms.CharField(
```
chessbr, line 90: make this be a MovelChoiceField and the model is Currency

shuup/admin/modules/settings/forms/system.py
```diff
@@ -72,8 +73,40 @@ class OrderSettingsForm(BaseSettingsForm):
...
+        label=_("Home Currency"),
+        help_text=_(
+            "This option defines the currency in that all the monetary values are expressed. "
+            "Enter a valid ISO-4217 currency code."
+        ),
+        required=True,
+    )
+    address_home_country = forms.CharField(
```
chessbr, line 98: this can have the list of countries as options. we use django-countries, so you probably can use it here as well. maybe there is a form field ready for that

shuup/setting_keys.py
```diff
@@ -0,0 +1,11 @@
+# This file is part of Shuup.
+#
+# Copyright (c) 2012-2021, Shuup Commerce Inc. All rights reserved.
+#
+# This source code is licensed under the OSL-3.0 license found in the
+# LICENSE file in the root directory of this source tree.
+
+
+SHUUP_HOME_CURRENCY = "home_currency"
```
chessbr, line 9: move this file to `shuup/core/settings_keys.py`. You will need to split the keys into different apps: core, front, xtheme, etc.

chessbr (changes requested):

shuup/apps/provides.py
```diff
@@ -170,61 +171,64 @@ def override_provides(category, spec_list):
         _provide_specs[category] = old_provides
 
 
-def load_module(setting_name, provide_category):
+def load_module(configuration_name, provide_category):
```
chessbr, line 174: it is not a good idea to touch on this function. I recall we are using this in several places in addons and projects. I would rather create a new  function to load from a configuration or create a new parameter to control this.

shuup/apps/settings.py
```diff
@@ -119,3 +121,23 @@ def reload_apps():
     django.apps.apps.app_configs.clear()
     django.apps.apps.ready = False
     django.setup()
+
+
+def get_configuration_setting(setting_key: str):
+    from shuup import configuration
+
+    modules = [
+        "shuup.core.setting_keys",
```
chessbr, line 130: `shuup/apps` shouldn't know anything abount `shuup/core` or shuup/admin`. Can you create some provide key to load them dynamically?

shuup/campaigns/admin_module/__init__.py
```diff
@@ -117,4 +113,6 @@ def get_permissions_help_texts(self) -> Iterable[str]:
 
 
 def _show_catalog_campaigns_in_admin():
-    return bool("catalog_campaigns" in settings.SHUUP_DISCOUNT_MODULES)
+    from shuup import configuration
+
+    return bool("catalog_campaigns" in configuration.get(None, SHUUP_DISCOUNT_MODULES))
```
chessbr, line 118: fallback to a blank list just in case:

```py
bool("catalog_campaigns" in (configuration.get(None, SHUUP_DISCOUNT_MODULES) or []))
```

shuup/core/migrations/0099_move_settings_to_db.py
```diff
@@ -0,0 +1,83 @@
...
+    SHUUP_ORDER_SOURCE_MODIFIER_MODULES,
+    SHUUP_PRICING_MODULE,
+    SHUUP_REFERENCE_NUMBER_LENGTH,
+    SHUUP_REFERENCE_NUMBER_METHOD,
+    SHUUP_REFERENCE_NUMBER_PREFIX,
+    SHUUP_TAX_MODULE,
+    SHUUP_TELEMETRY_ENABLED,
+    SHUUP_VOLUME_UNIT,
+)
+from shuup.front.setting_keys import SHUUP_FRONT_MAX_UPLOAD_SIZE
```
chessbr, line 35: you can't import shuup.from in shuup.core. they are different apps that shouldn't depent on each other.
front can depend on core but not the inverse. the more independent they are the better.

shuup/core/migrations/0099_move_settings_to_db.py
```diff
@@ -0,0 +1,83 @@
...
+from shuup.reports.setting_keys import DEFAULT_REPORTS_ITEM_LIMIT
```
chessbr, line 36: same here, create one migration for each app

shuup/utils/filer_forms.py
```diff
@@ -0,0 +1,28 @@
...
+from django import forms
+from django.core.validators import FileExtensionValidator
+
+from shuup import configuration
+from shuup.core.setting_keys import SHUUP_ALLOWED_UPLOAD_EXTENSIONS
+from shuup.utils.filer import file_size_validator
+
+
+class UploadFileForm(forms.Form):
+    file = forms.FileField(
```
chessbr, line 19: add translated label for both forms

shuup/core/migrations/0099_move_settings_to_db.py
```diff
@@ -0,0 +1,83 @@
...
+
+
+def move_settings_to_db(apps, schema_editor):
+    configuration.set(None, SHUUP_HOME_CURRENCY, settings.SHUUP_HOME_CURRENCY)
```
chessbr, line 40: for every setting, use a getattr and fallback to the value that was in settings.py:

```py
configuration.set(None, SHUUP_ALLOW_ANONYMOUS_ORDERS, getattr(settings, "SHUUP_ALLOW_ANONYMOUS_ORDERS", True))
```

We need to do this because we need to delete the entries from the settings.py, we shouldn't keep those anymore.

chessbr (changes requested):
Good work, almost there!

Also we need to do the following changes:

- Remove all the settings that we move to DB from the `settings.py`
- We need to change `shuup_init` management command to set the default values when it is ran. I think we can create a signal that we can trigger inside it so all apps can listed to it and ensure their initial settings, like this:

1) Create a signal `shuup_inited` inside `shuup/core/signals.py`.
2) Inside `shuup/core/management/commands/shuup_init.py`, before `set_current_theme` call you trigger it: `shuup_inited.send(sender=type(self))`

3) Move the `set_current_theme` to `shuup/xtheme/signal_handlers.py`:
```py
@receiver(shuup_inited)
def on_shuup_inited(sender, **kwargs):
    set_current_theme("shuup.themes.classic_gray", Shop.objects.first())
```

4) Let's ensure the default settings for core, front and reports in their respective signal_handlers.py:

in core:
```
@receiver(shuup_inited)
def on_shuup_inited(sender, **kwargs):
    configuration.set(None, SHUUP_HOME_CURRENCY, "EUR")
    configuration.set(None, SHUUP_ALLOW_ANONYMOUS_ORDERS, True)
    ... all other settings
```

the same with front and reports. This is to make sure NEW installations have the settings we expect them to have without breaking things.

shuup/admin/modules/settings/forms/system.py
```diff
@@ -37,13 +42,16 @@ def save(self, form):
 
         for key in form.fields.keys():
             try:
+                # delete also from EncryptedConfigurationItem
```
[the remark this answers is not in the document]
eberonsh: When I started the task a decade ago it wasn't still merged. But now I can. :)

shuup/admin/modules/settings/forms/system.py
```diff
@@ -70,10 +78,313 @@ class OrderSettingsForm(BaseSettingsForm):
...
+        required=True,
+    )
+    order_reference_number_length = forms.IntegerField(
+        label=_("Order Reference number length"),
+        help_text=_("The default length of reference numbers generated by certain reference number generators."),
+        required=True,
+    )
+    order_reference_number_prefix = forms.CharField(
+        label=_("Order Reference number prefix"),
+        help_text=_("#: An arbitrary (numeric) default prefix for certain reference number generators."),
```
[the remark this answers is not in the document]
eberonsh: ups, I'll leave evidence behind

shuup/admin/modules/settings/forms/system.py
```diff
@@ -70,10 +78,313 @@ class OrderSettingsForm(BaseSettingsForm):
...
+    home_currency = forms.ModelChoiceField(
+        label=_("Home Currency"),
+        queryset=Currency.objects.all(),
+        help_text=_(
+            "This option defines the currency in that all the monetary values are expressed. "
+            "Enter a valid ISO-4217 currency code."
+        ),
+        required=True,
+    )
+    address_home_country = LazyTypedChoiceField(
```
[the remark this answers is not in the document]
eberonsh: ObjectSelect2ModelField requires a model, ObjectSelect2MultipleField doesn't but we want to select just one country. So I think LazyTypedChoiceField does a good job here.

shuup/admin/modules/settings/forms/system.py
```diff
@@ -70,10 +78,313 @@ class OrderSettingsForm(BaseSettingsForm):
...
+        ),
+        required=False,
+    )
+    allow_anonymous_orders = forms.BooleanField(
+        label=_("Allow Anonymous Orders"),
+        help_text=_("Whether or not anonymous orders (without a `creator` user) are allowed."),
+        required=False,
+    )
+    default_order_label = forms.CharField(
+        label=_("Default Order Label"),
```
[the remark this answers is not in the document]
eberonsh: In every case, the initial value is taken from configuration, so there is no need. We just need to initialize correctly the configuration values.

shuup/admin/modules/settings/forms/system.py
```diff
@@ -70,10 +78,313 @@ class OrderSettingsForm(BaseSettingsForm):
...
+    )
+
+    def __init__(self, *args, **kwargs):
+        super(CoreSettingsForm, self).__init__(*args, **kwargs)
+        self.fields[SHUUP_HOME_CURRENCY].initial = Currency.objects.filter(
+            code=configuration.get(None, SHUUP_HOME_CURRENCY)
+        ).first()
+        if self.data:
+            # The html input of type checkout doesn't send the value if False. Here we force it to uncheck the value.
+            if "enable_attributes" not in self.data:
```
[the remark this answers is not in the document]
eberonsh: Much better

shuup/apps/provides.py
eberonsh: Ok, I'm going to separate them, but I think is safer the way I implemented, as load_module looks first in the configuration and it falls back to settings. Any add-on using load_module will still work. If there is an add-on relying on the settings that we moved, will no longer work. So I must identify all the cases.

shuup/admin/modules/settings/forms/system.py
```diff
@@ -31,25 +39,36 @@ def get_form_defs(self):
...
+            if isinstance(value, models.Model):
+                value = str(value)
+            encrypt = key in form.encrypted_fields
+            configuration.set(None, key, value, encrypted=encrypt)
         return True
 
 
 class BaseSettingsForm(forms.Form):
     title = None
+    encrypted_fields = []
```
eberonsh, line 71: We need to specify in each form which fields are encrypted.

shuup/core/migrations/0099_move_settings_to_db.py
```diff
@@ -0,0 +1,91 @@
...
+    SHUUP_REFERENCE_NUMBER_METHOD,
+    SHUUP_REFERENCE_NUMBER_PREFIX,
+    SHUUP_TAX_MODULE,
+    SHUUP_TELEMETRY_ENABLED,
+    SHUUP_VOLUME_UNIT,
+)
+
+
+def move_settings_to_db(apps, schema_editor):
+    configuration.set(None, SHUUP_HOME_CURRENCY, getattr(settings, "SHUUP_HOME_CURRENCY", "EUR"))
```
eberonsh, line 35: I took the setting values from shuup.core.settings.

PR #2590, opened 2021-08-12
ProductMedia

bill-shuup:
Update verbose_name to be more intuitive to the user.

Reference: https://trello.com/c/7zhtAx7Y/67-card-46-fb-change-ordering-to-order-of-images


shuup/core/migrations/0097_change_productmedia_verbose_text.py
```diff
@@ -0,0 +1,18 @@
+# Generated by Django 2.2.24 on 2021-08-12 16:45
+
+from django.db import migrations, models
+
+
+class Migration(migrations.Migration):
+
+    dependencies = [
+        ('shuup', '0096_auto_20210811_2253'),
```
chessbr, line 9: this dependency doesn't exist

chessbr (approved):

PR #2598, opened 2021-08-17
SimpleSupplier: enhancements for stock management

chessbr:
- Do not show stock management form part if supplier module is not simple supplier
- Fix template a little

Refs RM-15

### DIFF
--- a/shuup/simple_supplier/templates/shuup/simple_supplier/admin/product_form_part.jinja
+++ b/shuup/simple_supplier/templates/shuup/simple_supplier/admin/product_form_part.jinja
@@ -3,38 +3,42 @@
 {% set ss_form = form["simple_supplier"] %}
 
 {%  macro render_product(suppliers, product, product_url, product_qty) %}
-    {% if suppliers %}
-        <div class="row">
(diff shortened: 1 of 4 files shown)


shuup/simple_supplier/templates/shuup/simple_supplier/admin/product_form_part.jinja
```diff
@@ -3,38 +3,42 @@
...
+        {% endfor %}
+    </div>
 {% endmacro %}
 
 {% macro render_products(form) %}
     {% set product_qty = ss_form.products|length %}
     {% for product in form.products %}
         {% set product_url = shuup_admin.model_url(product) %}
-        {{ render_product(form.get_suppliers(product), product, product_url, product_qty) }}
+        {% set suppliers = form.get_suppliers(product) %}
```
khchine5-shuup, line 34: Should we add here `{% set has_first_draw = false  %}` ?

chessbr: no need, jinja won't raise hese as `has_first_draw` will be basically None

PR #2647, opened 2025-03-04
Update setup.py

NeginTmJ:
I faced error when tried to install,


Thrsh001 (approved):

tpickle-py:
i too had this same fix, need this before i can continue with other reviews

